PR #10, opened 2023-03-11
fix

masci:


nodes/speech2text/pyproject.toml
```diff
@@ -31,25 +31,16 @@ dependencies = [
   "numpy",
   "torch",
   "pydantic",
-  "torchaudio"
+  "torchaudio",
+  "openai-whisper @ git+https://github.com/openai/whisper.git"
```
mayankjobanputra, line 35: Just curious, is there a reason for us to always build from main?

PR #26, opened 2023-05-15
add MilvusDocumentStore

masci:
Milvus will be deprecated in Haystack 1.17 and definitely removed in Haystack 1.19.

To keep using Milvus with Haystack we provide it as a standalone package.


stores/milvus-documentstore/milvus.py
```diff
@@ -0,0 +1,675 @@
+from typing import Any, Dict, Generator, List, Optional, Union
```
ZanSara, line 1: I think this file is a lost copy? There's an identical one under `milvus-documentstore/milvus_documentstore`

ZanSara (changes requested):

ZanSara (approved):
I wonder if the version should rather be 1.0.0 (after all, we know this class works), but I'll leave this choice to you

PR #28, opened 2023-06-01
Add `MilvusDocumentStore` usage example

masci:
The same should be added to the docs page


stores/milvus-documentstore/README.md
```diff
@@ -6,6 +6,17 @@
 ```console
 pip install -e "git+https://github.com/deepset-ai/haystack-extras.git#egg=milvus_documentstore&subdirectory=stores/milvus-documentstore"
 ```
+## Usage
+The package will be available in your code under the name `milvus_documentstore`, so for example you can add a document like this:
```
dfokina, line 10:
```suggestion
You can find the package in your code as `milvus_documentstore`. Then, you can add a document like this:
```

PR #33, opened 2023-09-29
Rename workflow wo extension

anakin87:
See https://github.com/deepset-ai/haystack-extras/pull/32#issuecomment-1740723808


ZanSara (approved):
LGTM!

PR #34, opened 2023-10-10
[instructor] Fix osx install

masci:
The pinned version of `pyarrow` can't install on OSX. Removing the pin still works, not sure if we want to test the package in other environments to confirm.

Also exported the embedders in the root package for convenience.


anakin87 (commented):
It's a strange pin, copied from the original Instructor requirements...

After a check, I'm quite sure this dependency is not needed.
In fact, `datasets` already installs it:
https://github.com/huggingface/datasets/blob/7d357eb4b499cd530c3f4e626f2825a50ee6c8aa/setup.py#L114
(and `pyarrow` is never used directly in the `InstructorEmbedding` package)

I would remove/comment the dependency.

anakin87 (approved):
Thanks!

PR #50, opened 2023-11-13
Elastiscsearch BM25Retriever: set `scale_score` default value to False

anakin87:
Related to https://github.com/deepset-ai/haystack/issues/6271

(I also fixed a test that used `text` instead of `content` in Document initialization).

### DIFF
--- a/document_stores/elasticsearch/tests/test_bm25_retriever.py
+++ b/document_stores/elasticsearch/tests/test_bm25_retriever.py
@@ -15,7 +15,7 @@ def test_init_default():
     assert retriever._document_store == mock_store
     assert retriever._filters == {}
     assert retriever._top_k == 10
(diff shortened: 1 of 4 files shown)


document_stores/elasticsearch/tests/test_bm25_retriever.py
```diff
@@ -15,7 +15,7 @@ def test_init_default():
     assert retriever._document_store == mock_store
     assert retriever._filters == {}
     assert retriever._top_k == 10
-    assert retriever._scale_score
+    assert retriever._scale_score is False
```
silvanocerza, line 18:
```suggestion
    assert not retriever._scale_score
```

silvanocerza (approved):

PR #52, opened 2023-11-14
Elasticsearch Document store - embedding retrieval

anakin87:
Part of  https://github.com/deepset-ai/haystack/issues/5329

- Add basic support for Embedding retrieval in Elasticsearch Document Store

### Notes for the reviewer
- `ElasticSearchEmbeddingRetriever` will be added in a subsequent PR.

- I chose to support only [Approximate kNN](https://www.elastic.co/guide/en/elasticsearch/reference/8.11/knn-search.html#knn-methods), which is the suggested approach compared to Exact, brute-force kNN.

- This Document Store is compatible with Elasticsearch>=8.11
Configuring vector fields in older versions required more manual effort on the part of the user (e.g., explicitly specifying the vector size at index creation).

- I haven't implemented scaling scores in the range [0, 1]. See #53 

- I would like to test also other unhappy paths, such as trying to write documents with embeddings of different sizes.
Concerning this point, we should rework **error handling** during indexing (`write_documents` method).

  _Update_: I encountered several erroneous `DuplicateDocumentError` while working on embedding retrieval.
If we do not want to spend time reworking this part at the moment, I would propose printing a warning with the obtained errors. It seems less misleading to me. WDYT?

### DIFF
--- a/document_stores/elasticsearch/src/elasticsearch_haystack/document_store.py
+++ b/document_stores/elasticsearch/src/elasticsearch_haystack/document_store.py
@@ -53,7 +53,7 @@ class ElasticsearchDocumentStore:
         :param hosts: List of hosts running the Elasticsearch client. Defaults to None
         :param index: Name of index in Elasticsearch, if it doesn't exist it will be created. Defaults to "default"
         :param embedding_similarity_function: The similarity function used to compare Documents embeddings.
-            Defaults to "cosine".
+            Defaults to "cosine". This parameter only takes effect if the index does not yet exist and is created.
             To choose the most appropriate function, look for information about your embedding model.
             To understand how document scores are computed, see the Elasticsearch documentation:
             https://www.elastic.co/guide/en/elasticsearch/reference/current/dense-vector.html#dense-vector-params
@@ -100,6 +100,26 @@ class ElasticsearchDocumentStore:
         Returns how many documents are present in the document store.
         """
         return self._client.count(index=self._index)["count"]
+
+    def _search_documents(self, **kwargs) -> List[Document]:
+        """
+        Calls the Elasticsearch client's search method and handles pagination.
+        """
+
+        documents: List[Document] = []
+        from_ = 0
+        # Handle pagination
+        while True:
+            res = self._client.search(
+                index=self._index,
+                from_=from_,
+                **kwargs,
+            )
+            documents.extend(self._deserialize_document(hit) for hit in res["hits"]["hits"])
+            from_ = len(documents)
+            if from_ >= res["hits"]["total"]["value"]:
+                break
+        return documents
 
     def filter_documents(self, filters: Optional[Dict[str, Any]] = None) -> List[Document]:
         """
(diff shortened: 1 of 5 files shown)


document_stores/elasticsearch/src/elasticsearch_haystack/document_store.py
```diff
@@ -45,19 +52,32 @@ def __init__(self, *, hosts: Optional[Hosts] = None, index: str = "default", **k
 
         :param hosts: List of hosts running the Elasticsearch client. Defaults to None
         :param index: Name of index in Elasticsearch, if it doesn't exist it will be created. Defaults to "default"
+        :param embedding_similarity_function: The similarity function used to compare Documents embeddings.
+            Defaults to "cosine".
+            To choose the most appropriate function, look for information about your embedding model.
+            To understand how document scores are computed, see the Elasticsearch documentation:
+            https://www.elastic.co/guide/en/elasticsearch/reference/current/dense-vector.html#dense-vector-params
```
silvanocerza, line 59: We should specify this is used only if the index specified doesn't exist already.

document_stores/elasticsearch/src/elasticsearch_haystack/document_store.py
```diff
@@ -306,3 +327,57 @@ def _bm25_retrieval(
...
+                "query_vector": query_embedding,
+                "k": top_k,
+                "num_candidates": num_candidates,
+            },
+        }
+
+        if filters:
+            body["knn"]["filter"] = _normalize_filters(filters)
+
+        res = self._client.search(index=self._index, **body)
```
silvanocerza, line 378: We need to handle pagination here.

document_stores/elasticsearch/tests/test_bm25_retriever.py
```diff
@@ -27,7 +27,11 @@ def test_to_dict(_mock_elasticsearch_client):
         "type": "ElasticsearchBM25Retriever",
         "init_parameters": {
             "document_store": {
-                "init_parameters": {"hosts": "some fake host", "index": "default"},
+                "init_parameters": {
+                    "hosts": "some fake host",
+                    "index": "default",
+                    "embedding_similarity_function": "cosine",
```
silvanocerza, line 33: Let's update also `test_init_default` and `test_from_dict`.

document_stores/elasticsearch/tests/test_document_store.py
```diff
@@ -43,6 +49,7 @@ def test_to_dict(self, _mock_elasticsearch_client):
             "init_parameters": {
                 "hosts": "some hosts",
                 "index": "default",
+                "embedding_similarity_function": "cosine",
```
silvanocerza, line 52: Let's update also `test_from_dict`.

document_stores/elasticsearch/tests/test_document_store.py
```diff
@@ -231,3 +249,45 @@ def test_lte_filter_table(self, docstore: ElasticsearchDocumentStore, filterable
...
+    def test_embedding_retrieval_query_documents_different_embedding_sizes(self, docstore: ElasticsearchDocumentStore):
+        """
+        Test that the retrieval fails if the query embedding and the documents have different embedding sizes.
+        """
+        docs = [Document(content="Hello world", embedding=[0.1, 0.2, 0.3, 0.4])]
+        docstore.write_documents(docs)
+
+        with pytest.raises(
+            BadRequestError,
+            match="search_phase_execution_exception",
```
silvanocerza, line 291: Let's not match this error but let's check that a custom error is raised when the received embedding have a different size.

anakin87: I tried implementing a custom error handling logic, but the code became too messy.
Also, the Elasticsearch exception is quite informative:
`elasticsearch.BadRequestError: BadRequestError(400, 'search_phase_execution_exception', 'failed to create query: the query vector has a different dimension [2] than the index vectors [4]')`
So, I would rather not handle this error explicitly.

silvanocerza (approved):
Nice 👍

PR #57, opened 2023-11-16
Elasticsearch - refactor `_search_documents`

anakin87:
The method `_search_documents` was created to properly manage calls to the Elasticsearch client's search method and handle pagination. It also avoids code duplication.

**Changes** 
- improve this method, making it consider the `top_k` value if present (to stop iterating)
- introduce it in `_bm25_retrieval`
- add tests for retrieval with pagination, where this method plays a key role

### DIFF
--- a/document_stores/elasticsearch/src/elasticsearch_haystack/document_store.py
+++ b/document_stores/elasticsearch/src/elasticsearch_haystack/document_store.py
@@ -53,7 +53,7 @@ class ElasticsearchDocumentStore:
         :param hosts: List of hosts running the Elasticsearch client. Defaults to None
         :param index: Name of index in Elasticsearch, if it doesn't exist it will be created. Defaults to "default"
         :param embedding_similarity_function: The similarity function used to compare Documents embeddings.
-            Defaults to "cosine".
+            Defaults to "cosine". This parameter only takes effect if the index does not yet exist and is created.
             To choose the most appropriate function, look for information about your embedding model.
             To understand how document scores are computed, see the Elasticsearch documentation:
             https://www.elastic.co/guide/en/elasticsearch/reference/current/dense-vector.html#dense-vector-params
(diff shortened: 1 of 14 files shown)


document_stores/elasticsearch/src/elasticsearch_haystack/document_store.py
```diff
@@ -115,8 +119,12 @@ def _search_documents(self, **kwargs) -> List[Document]:
                 from_=from_,
                 **kwargs,
             )
+
             documents.extend(self._deserialize_document(hit) for hit in res["hits"]["hits"])
-            from_ = len(documents)
+            from_ += len(documents)
```
silvanocerza, line 124: This doesn't seem correct to me. 🤔

document_stores/elasticsearch/src/elasticsearch_haystack/document_store.py
```diff
@@ -115,8 +119,12 @@ def _search_documents(self, **kwargs) -> List[Document]:
...
+
+            if top_k and from_ >= top_k:
```
silvanocerza, line 126: Let's be explicit like we do above.

```suggestion
            if top_k is not None and from_ >= top_k:
```

document_stores/elasticsearch/src/elasticsearch_haystack/document_store.py
```diff
@@ -326,14 +334,17 @@ def _bm25_retrieval(
...
-        return docs
+        if not scale_score:
+            return documents
+
+        documents_w_scaled_scores = []
+        for doc in documents:
+            doc.score = float(1 / (1 + np.exp(-np.asarray(doc.score / BM25_SCALING_FACTOR))))
+            documents_w_scaled_scores.append(doc)
+
+        return documents_w_scaled_scores
```
silvanocerza, line 347: Can't we do this instead? There shouldn't be any need to create a new list. 🤔 

```suggestion
        if scale_score:
            for doc in documents:
                doc.score = float(1 / (1 + np.exp(-np.asarray(doc.score / BM25_SCALING_FACTOR))))

        return documents
```

document_stores/elasticsearch/tests/test_document_store.py
```diff
@@ -1,6 +1,10 @@
 # SPDX-FileCopyrightText: 2023-present Silvano Cerza <silvanocerza@gmail.com>
 #
 # SPDX-License-Identifier: Apache-2.0
+
+# ruff: noqa: S311
```
silvanocerza, line 5: Can we move this ignore to the line that actually uses `random`?

document_stores/elasticsearch/src/elasticsearch_haystack/document_store.py
anakin87: You're right!

document_stores/elasticsearch/tests/test_document_store.py
anakin87: I had some trouble doing it, but now I have managed.

silvanocerza (approved):
Go go! 🚀

PR #59, opened 2023-11-20
[Elasticsearch Document Store] improve error handling in `write_documents`

### LINKED ISSUE #58: [Elasticsearch Document Store] better error handling in `write_documents`
Currently, we assume that all document writing errors in the Document Store are related to duplicate documents.

We should handle this part better so as not to show misleading exceptions.

anakin87:
Fixes #58

We now distinguish the errors into `DuplicateDocumentError` and generic `DocumentStoreError`.

The `DuplicateDocumentError` is raised only at the appropriate Elasticsearch exception and when the duplicate policy is FAIL.

(Added a test that raises `DocumentStoreError`).

### DIFF
--- a/document_stores/elasticsearch/src/elasticsearch_haystack/document_store.py
+++ b/document_stores/elasticsearch/src/elasticsearch_haystack/document_store.py
@@ -12,7 +12,7 @@ from elasticsearch import Elasticsearch, helpers  # type: ignore[import-not-foun
 from haystack.preview import default_from_dict, default_to_dict
 from haystack.preview.dataclasses import Document
 from haystack.preview.document_stores.decorator import document_store
-from haystack.preview.document_stores.errors import DuplicateDocumentError
+from haystack.preview.document_stores.errors import DocumentStoreError, DuplicateDocumentError
 from haystack.preview.document_stores.protocols import DuplicatePolicy
 
 from elasticsearch_haystack.filters import _normalize_filters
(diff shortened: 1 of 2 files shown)


document_stores/elasticsearch/src/elasticsearch_haystack/document_store.py
```diff
@@ -237,16 +240,29 @@ def write_documents(self, documents: List[Document], policy: DuplicatePolicy = D
...
+            other_errors = []
+            for e in errors:
+                if e["create"]["error"]["type"] == "version_conflict_engine_exception":
+                    if policy == DuplicatePolicy.FAIL:
+                        duplicate_errors_ids.append(e["create"]["_id"])
+                    elif policy == DuplicatePolicy.SKIP:
+                        # when the policy is skip, these errors are OK and we should not raise an exception
+                        continue
+                    else:
+                        other_errors.append(e)
```
anakin87, line 255: if this type of error raises with  `DuplicatePolicy.OVERWRITE` is not related to duplicates

document_stores/elasticsearch/src/elasticsearch_haystack/document_store.py
```diff
@@ -237,16 +240,29 @@ def write_documents(self, documents: List[Document], policy: DuplicatePolicy = D
...
+                else:
+                    other_errors.append(e)
```
silvanocerza, line 257: What do you think of this instead? Seems more straightforward to me. 🤔 


```suggestion
                error_type = e["create"]["error"]["type"]
                if policy == DuplicatePolicy.FAIL and error_type == "version_conflict_engine_exception":
                    duplicate_errors_ids.append(e["create"]["_id"])
                elif policy == DuplicatePolicy.SKIP and error_type == "version_conflict_engine_exception":
                    # when the policy is skip, duplication errors are OK and we should not raise an exception
                    continue
                else:
                    other_errors.append(e)
```

anakin87: To me too :smile:

silvanocerza (approved):

PR #75, opened 2023-12-03
Add `cohere_haystack` integration package

masci:
Moved the `CohereGenerator` from Haystack into a new integration package. Embedders will be added to this same integration.

### DIFF
--- a/integrations/cohere/LICENSE.txt
+++ b/integrations/cohere/LICENSE.txt
@@ -1,0 +1,1 @@
+404: Not Found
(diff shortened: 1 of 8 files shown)


integrations/cohere/LICENSE.txt
```diff
@@ -0,0 +1 @@
+404: Not Found
```
silvanocerza, line 1: That's a `curl` pipe that it didn't go according to plans I guess? 😅

masci: WTF

silvanocerza (approved):

PR #80, opened 2023-12-05
[cohere] Add text and document embedders

masci:
Follow up of https://github.com/deepset-ai/haystack/pull/5831

### DIFF
--- a/integrations/cohere/src/cohere_haystack/embedders/document_embedder.py
+++ b/integrations/cohere/src/cohere_haystack/embedders/document_embedder.py
@@ -3,13 +3,12 @@
 # SPDX-License-Identifier: Apache-2.0
 from typing import List, Optional, Dict, Any
 import os
-from tqdm import tqdm
+import asyncio
 
-from haystack.preview import component, Document, default_to_dict, default_from_dict
-from haystack.preview.lazy_imports import LazyImport
+from haystack import component, Document, default_to_dict
+from cohere import Client, AsyncClient, CohereError
 
-with LazyImport(message="Run 'pip install cohere'") as cohere_import:
-    from cohere import Client, AsyncClient, CohereError
+from .utils import get_async_response, get_response
 
 API_BASE_URL = "https://api.cohere.ai/v1/embed"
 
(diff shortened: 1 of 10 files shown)


masci:
@vrunm PR is good to go!

integrations/cohere/src/cohere_haystack/embedders/document_embedder.py
```diff
@@ -0,0 +1,149 @@
...
+    @component.output_types(documents=List[Document], metadata=Dict[str, Any])
+    def run(self, documents: List[Document]):
+        """
+        Embed a list of Documents.
+        The embedding of each Document is stored in the `embedding` field of the Document.
+
+        :param documents: A list of Documents to embed.
+        """
+
+        if not isinstance(documents, list) or not isinstance(documents[0], Document):
```
julian-risch, line 122: Let's change this to 
`if not isinstance(documents, list) or documents and not isinstance(documents[0], Document):`so that empty lists do not raise an error. If an empty list is provided as input, components should also return an empty list as output.

integrations/cohere/src/cohere_haystack/generator.py
```diff
@@ -4,13 +4,11 @@
...
 import sys
-from typing import Any, Callable, Dict, List, Optional
+from typing import Any, Callable, Dict, List, Optional, cast
 
+from cohere import COHERE_API_URL, Client
+from cohere.responses import Generations
 from haystack import DeserializationError, component, default_from_dict, default_to_dict
-from haystack.lazy_imports import LazyImport
-
-with LazyImport(message="Run 'pip install cohere'") as cohere_import:
```
julian-risch, line 12: I see you're removing this because cohere is installed as a dependency of the cohere-haystack package anyway. 👍 Let's clean this up in the other integrations then too otherwise people might use them as an example: https://github.com/search?q=repo%3Adeepset-ai%2Fhaystack-core-integrations%20LazyImport&type=code Could you please create an issue?

integrations/cohere/src/cohere_haystack/embedders/document_embedder.py
julian-risch: Ideally, we'd also have a test case to check that behavior.

julian-risch (changes requested):
Looks very good to me, almost ready to go. I would like to have the CohereDocumentEmbedder also work with empty lists of documents and suggested a change of an if condition and the addition of a test case. Everything else is ready to be merged. 👍

integrations/cohere/src/cohere_haystack/generator.py
masci: Yup, here you go! https://github.com/deepset-ai/haystack-core-integrations/pull/83

PR #86, opened 2023-12-07
fix cohere pypi version badge and add Embedder note

julian-risch:
Fixing a copy paste error: cohere pypi badge was pointing to chroma
Adding `Embedder` to the readme overview table for cohere now that we merged TextEmbedder and DocumentEmbedder in addition to the existing Generator


vblagoje (approved):
LGTM

PR #88, opened 2023-12-07
Add cohere chat generator

sunilkumardash9:
support for cohere chat endpoint


sunilkumardash9:
Hi @vblagoje, I was curious how this import works as I can't seem to import chat_generator to test files with relative import.
https://github.com/deepset-ai/haystack-core-integrations/blob/4076949668030eab0ef2eeada80f518067c3aa69/integrations/cohere/tests/test_document_embedder.py#L10

integrations/cohere/src/cohere_haystack/chat/chat_generator.py
```diff
@@ -0,0 +1,135 @@
...
+        response = self.client.chat(
+            message=message[0], model=self.model_name, stream=self.streaming_callback is not None, **generation_kwargs
+        )
+        if self.streaming_callback:
+            for chunk in response:
+                if chunk.event_type == "text-generation":
+                    stream_chunk = self._build_chunk(chunk)
+                    self.streaming_callback(stream_chunk)
+            chat_message = ChatMessage(content=response.texts, role=None, name=None)
+            chat_message.metadata.update(
```
vblagoje, line 92: Any way we can try to match OpenAI metadata, they have:
```
{
                "model": chunk.model,
                "index": 0,
                "finish_reason": "stop"
                "usage": {},  # here we put token counts for prompt and response
            }
```

vblagoje: Aha I see token counts are available. @anakin87 should we try to match the OpenAI format here so all the chat generators are more or less interchangeable? Ideally we match the OpenAI format  and then provide more Cohere specific metadata, whatever is available, can't hurt

vblagoje (changes requested):
@sunilkumardash9 solid work, thank you. Would you please address my minor comments and don't forget to add release note using `reno` tool. See contributing section, thanks for your great work so far.

integrations/cohere/src/cohere_haystack/chat/chat_generator.py
anakin87: It would be nice to do that!

vblagoje:
Apologies for the delay @sunilkumardash9 I'll take a look on Friday, play with the generator and hopefully, we will integrate it at that time.

sunilkumardash9:
@vblagoje No issues. Do let me know in case I have missed anything.

vblagoje:
Hey @sunilkumardash9 if I understood their chat [API](https://docs.cohere.com/docs/cochat-beta) correctly, I think we need to set the message parameter to be our last message, i.e message[-1], not to concatenate all messages into one and then sending it over. Rather, we need to pass all messages before the last as chat_history parameter. But please correct me if I'm wrong, I'm looking at the API docs; you have played with the model. 

Also please consider playing with an example chat script to prove to yourself that everything is working as expected.

vblagoje:
Also it seems like we have to translate the standard roles that everyone else agreed upon to cohere roles "User", and "Chatbot". So I think you'll need to do a bit of conversion of ChatMessages given to the CohereChatGenerator run method. 

user_message = {"user_name": "User", "text": message}
bot_message = {"user_name": "Chatbot", "text": answer}

sunilkumardash9:
@vblagoje I was wondering the same after Gemini chat, which has a similar chat history implementation, made some changes to adjust default chat roles to Cohere-specific ones. And it seems to be working well.

vblagoje:
Excellent @sunilkumardash9 took a quick look and this is much better. I'll play with it over the holidays and get back to you early next week

vblagoje:
Hey @sunilkumardash9 , I tried your CohereChatGenerator, and it works well. The only issue I faced was in streaming mode. The first response from LLM streams fine, but when I tried continuing the chat conversation, the second request failed with 400 response. Out of curiosity - do you get the same issue?

sunilkumardash9:
@vblagoje It is working fine for me, even with subsequent requests. There might be some issues with the Cohere endpoint.

vblagoje:
@sunilkumardash9, this looks okay to me as the first version. Let's track the progress at https://github.com/deepset-ai/haystack-core-integrations/issues/150 . Would you please kindly update the task complete there so we can push it toward the finish line?

vblagoje (approved):
Thanks for this contribution @sunilkumardash9

sunilkumardash9:
@vblagoje I think most tasks in #150  have already been done while CohereGenerator was added. Would you like me to complete the rest of it?

anakin87:
Sorry sorry...

For consistency, we are changing each mention of `metadata` in the codebase to `meta`: see https://github.com/deepset-ai/haystack/issues/6570

@sunilkumardash9 can you please update this PR to reflect this change?
Otherwise, I will take care...

vblagoje:
Right, I've seen it too and forgot to mention it to @sunilkumardash9 - thanks @anakin87 Please update these references @sunilkumardash9

sunilkumardash9:
Hi @anakin87, I am getting metadata references in chat_message and chat_role in haystack-ai v2.0.0b3.

anakin87:
@sunilkumardash9 Sorry, my fault.
We haven't released the new version yet, so it is OK to merge this PR.
When we release the new version of Haystack (with modified ChatMessages), we will also update your implementation.

integrations/cohere/releasenotes/notes/add_cohere_chat_generator-ef96bbeaf588281c.yaml
```diff
@@ -0,0 +1,4 @@
+---
+preview:
+  - |
+    Adds `CohereChatGenerator` for text and chat generation.
```
anakin87, line 4: We currently don't have release notes for this repository.
Please remove this file.

PR #95, opened 2023-12-12
chore: fix Wheel package for instructor-embedders, uniform tests and CI

masci:
Name of the folder containing the code should be the same as the package declared in the project file. The latest version of hatchling would fail on such a discrepancy, so I renamed `instructor_embedders` to `instructor_embedders_haystack`.

While I was there, I took the chance to uniform some conventions for tests and CI.


anakin87 (approved):
:+1:

PR #99, opened 2023-12-13
Change default 'input_type' for CohereTextEmbedder

bilgeyucel:
Having `input_type = "search_query"` is a better default value for CohereTextEmbedder according to [Cohere Docs](https://docs.cohere.com/reference/embed)


anakin87:
It makes sense!

Let's also change the tests accordingly...

bilgeyucel:
> It makes sense!
> 
> Let's also change the tests accordingly...

on it! 🙌

anakin87 (approved):
LGTM!

PR #107, opened 2023-12-15
Add support for API key auth when using Google Vertex AI

silvanocerza:


silvanocerza:
After some investigation we realized that API key authentication doesn't work. 

I had the impression it was locally as it was picking up my ADC credentials.

API key authentication with `vertexai` installed through `google-cloud-aiplatform` doesn't work as an exception is raised saying it's not supported.

`google-generativeai` doesn't work either when using an API key, I get this error locally: `User location is not supported for the API use`.

I tried running `google-generativeai` in a Colab but for some reason imports that work locally don't work in Colab.


ADC is annoying to use but works, closing this. 👋

PR #110, opened 2023-12-15
Gradient.ai - updating the SDK

michaelfeil:
Our SDK `gradientai >= 1.4.0` supports now `async` API and threaded batching in our API - the embedding model now exposes `.embed` => This PR aims to improve speed of embeddings over large numbers of documents.

## Dependencies:
- bumped  `gradientai` from `1.3.0` to `1.4.0`.  The SDK always had a minium python version of 3.8, except for some initial experiments (e.g. `1.0.0.dev1`). Therefore, I raised the miniumum version to `1.4.0` 
- added `tqdm` as optional. It should be pre-installed, and we are using no advanced options other than `from tqdm import tqdm`, no version pin required. I think this will improve developer experience, especially since the embeddings are now faster on the server side.

@mhaligowski Can you give this a check before I mark as ready?


integrations/gradient/src/gradient_haystack/embedders/gradient_document_embedder.py
```diff
@@ -31,15 +36,17 @@ def __init__(
         self,
         *,
         model_name: str = "bge-large",
-        batch_size: int = 100,
+        batch_size: int = 32768,
         access_token: Optional[str] = None,
         workspace_id: Optional[str] = None,
         host: Optional[str] = None,
+        progress_bar: bool = True,
```
mhaligowski, line 43: I think the convention is to name the variables with a "booleanish" prefix? `show_progress_bar`? `has_progress_bar`? `progress_bar` looks like it's expecting an actual object.

integrations/gradient/pyproject.toml
```diff
@@ -7,7 +7,7 @@ name = "gradient-haystack"
 dynamic = ["version"]
 description = ''
 readme = "README.md"
-requires-python = ">=3.7"
+requires-python = ">=3.8"
```
mhaligowski, line 10: what's the reason for bumping?

mhaligowski:
LGTM, let's make sure that the tests are ran and we're gtg!

integrations/gradient/src/gradient_haystack/embedders/gradient_document_embedder.py
```diff
@@ -31,15 +36,17 @@ def __init__(
         self,
         *,
         model_name: str = "bge-large",
-        batch_size: int = 100,
+        batch_size: int = 32768,
```
mhaligowski, line 39: do you want `32_768`?

integrations/gradient/pyproject.toml
michaelfeil: With `pip install gradientai` using a py37 environment, it installed `1.0.0.dev-something`

michaelfeil:
All tests passing on my end.

masci:
Thanks folks! FYI, in #103 we removed all the `__about__.py` files, to resolve the conflict just delete the one in your branch and it should be good.

michaelfeil:
@masci Alright.

integrations/gradient/src/gradient_haystack/embedders/gradient_document_embedder.py
```diff
@@ -4,9 +4,19 @@
 from gradientai import Gradient
 from haystack import Document, component, default_to_dict
 
+try:
+    from tqdm import tqdm  # type: ignore # noqa
```
michaelfeil, line 8: @masci Requiring some help here - the CI is failing on this line - not sure why.

integrations/gradient/src/gradient_haystack/embedders/gradient_document_embedder.py
```diff
@@ -75,11 +90,17 @@ def _generate_embeddings(self, documents: List[Document], batch_size: int) -> Li
...
         Batches the documents and generates the embeddings.
         """
-        batches = [documents[i : i + batch_size] for i in range(0, len(documents), batch_size)]
+        if self._progress_bar and tqdm_imported:
+            batches = [documents[i : i + batch_size] for i in range(0, len(documents), batch_size)]
+            progress_bar = tqdm
+        else:
+            # no progress bar
+            progress_bar = _alt_progress_bar  # type: ignore
+            batches = [documents]
```
silvanocerza, line 99: Is this correct? It seems strange to me that batching is ignored if `tqdm` is not shown. 🤔

michaelfeil: Yes - our server side batch size is currently 256. 
We run a internal mapping / mirco-bacthing loop of size 256, but have no progress bar. Therefore, there is no upper limit, you could also set batch_size to 100_000

silvanocerza: Ok, good for me if it's good for you. 👍

silvanocerza (approved):
Cool, merging!

michaelfeil:
Awesome, thanks guys!

Issue #112, opened 2023-12-15
Package build broken if integration folder name contains hyphens


masci:
**Describe the bug**

We recently moved (#103) package versioning from hardcoded strings in the `__about__.py` file to git tags using `setuptools_scm` through `hatch`.

Problem is `setuptools_scm` splits the tag using `-` and having hyphens in the string prepending the version in the tag name, for example `integrations/google_vertex-v0.0.1` confuses the plugin.

There isn't an easy fix, so I propose the following workaround:

- [x] Rename all the tags `integrations/google-vertex-vXXX` to `integrations/google_vertex-vXXX`
- [x] Rename all the tags `integrations/instructor-embedders-vXXX` to `integrations/instructor_embedders-vXXX`
- [x] Push the new tags, CI will attempt to rebuild the packages but will fail because the path `google_vertex` doesn't exist
- [x] Rename the folder `integrations/google-vertex` to `integrations/google_vertex` https://github.com/deepset-ai/haystack-core-integrations/pull/114
- [x] Rename the folder `integrations/instructor-embedders` to `integrations/instructor_embedders` https://github.com/deepset-ai/haystack-core-integrations/pull/114
- [ ] Enforce the new naming convention with a job in the CI https://github.com/deepset-ai/haystack-core-integrations/pull/119

The workaround won't affect the name of the package on PyPI, nor the import paths.

**To Reproduce**

Checkout the latest tag for google vertex and call `hatch version`

**Describe your environment (please complete the following information):**
 - Haystack version:
 - Integration version:

PR #114, opened 2023-12-18
chore: replace - with _

masci:
Part of #112


anakin87 (approved):
All good!

.github/labeler.yml
```diff
@@ -16,7 +16,7 @@ integration:elasticsearch:
 
 integration:google-vertex:
```
silvanocerza, line 17: Should we change the labels too? 🤔

masci: we're going to have two naming conventions now:
- hyphens for package names: `google-vertex-haystack`
- underscores for folder names: `integrations/google_vertex/**`

I think it depends on what are you targeting with (for example in this case) a label: the folder in the source layout, or the "integration project"?

silvanocerza: Ah right, makes sense.

silvanocerza (approved):

PR #133, opened 2023-12-20
Pinecone - filters

anakin87:


integrations/pinecone/src/pinecone_haystack/filters.py
```diff
@@ -0,0 +1,193 @@
...
+        # Remove the "meta." prefix if present.
+        # Documents are flattened when using the PineconeDocumentStore
+        # so we don't need to specify the "meta." prefix.
+        # Instead of raising an error we handle it gracefully.
+        field = field[5:]
+
+    # if field == "content":
+    #     field = "meta.content"
+    # if field == "dataframe":
+    #     field = "meta.dataframe"
```
masci, line 60: let's remove these comments

integrations/pinecone/src/pinecone_haystack/filters.py
```diff
@@ -0,0 +1,193 @@
...
+
+    if "operator" not in condition:
+        msg = f"'operator' key missing in {condition}"
+        raise FilterError(msg)
+    if "value" not in condition:
+        msg = f"'value' key missing in {condition}"
+        raise FilterError(msg)
```
masci, line 67: should we move these at the top? In case of error we wouldn't bother to check `meta.`

integrations/pinecone/src/pinecone_haystack/filters.py
```diff
@@ -0,0 +1,193 @@
...
+
+    return {field: {"$eq": value}}
+
+
+def _not_equal(field: str, value: Any) -> Dict[str, Any]:
+    supported_types = (str, int, float, bool)
+    if not isinstance(value, supported_types):
+        msg = (
+            f"Unsupported type for 'inequal' comparison: {type(value)}. "
```
masci, line 92:
```suggestion
            f"Unsupported type for 'not equal' comparison: {type(value)}. "
```

integrations/pinecone/src/pinecone_haystack/filters.py
```diff
@@ -0,0 +1,193 @@
...
+    "!=": _not_equal,
+    ">": _greater_than,
+    ">=": _greater_than_equal,
+    "<": _less_than,
+    "<=": _less_than_equal,
+    "in": _in,
+    "not in": _not_in,
+}
+
+LOGICAL_OPERATORS = {"AND": "$and", "OR": "$or"}
```
masci, line 193: I would put module constants at the top

masci (changes requested):

integrations/pinecone/src/pinecone_haystack/filters.py
anakin87: The fact is that `COMPARISON_OPERATORS` depends on some functions, so it should live after the function definitions.
And I would prefer `LOGICAL_OPERATORS` to be close.

masci (approved):
All good!

Issue #128, opened 2023-12-20
Pinecone Document Store


anakin87:
## Summary and motivation

Briefly explain the request: why do we need this integration? What use cases does it support?

## Detailed design

Explain the design in enough detail for somebody familiar with Haystack to understand, and for somebody familiar with
the implementation to implement. Get into specifics and corner-cases, and include examples of how the feature is used.
Also, if there's any new terminology involved, define it here.

## Checklist

If the request is accepted, ensure the following checklist is complete before closing this issue.

- [x] The code was merged in the `main` branch
- [x] #161
- [x] There is a Github workflow running the tests for the integration nightly and at every PR
- [x] A label named like `integration:<your integration name>` has been added to this repo
- [x] The [labeler.yml](https://github.com/deepset-ai/haystack-core-integrations/blob/main/.github/labeler.yml) file has been updated
- [x] The package has been released on PyPI
- [x] deepset-ai/haystack-integrations#107
- [x] The integration has been listed in the [Inventory section](https://github.com/deepset-ai/haystack-core-integrations#inventory) of this repo README
- [x] There is an example available to demonstrate the feature
- [x] deepset-ai/devrel-board#241

dfokina:
@anakin87 @bilgeyucel docs are ready and published

PR #149, opened 2023-12-27
[Pinecone] Add example notebook

anakin87:
Part of #128 

Add an example that shows how to use the `PineconeDocumentStore`


vblagoje:
Super cool @anakin87 , I'll take a look today and get back to you

vblagoje (approved):
Works like a charm 🚀 I have two optional suggestions:
- Add index parameter explicitly as whatever default index name is so users can easily change it if needed
- Add the "Open with Colab" link so it is even easier to try these notebooks out

anakin87:
Thx @vblagoje! 

I applied your suggestions...

PR #156, opened 2023-12-29
Gemini with Makersuite

TuanaCelik:
This is the first go at implementing Gemini support with Makersuite. Works with:

`gemini-pro` for just text
`gemini-pro-vision` for text + image
`GeminiChatGenerator` for function calling

Needs:
- [ ] name changes
- [ ] docstrings
- [ ] tests

Contains: example.ipynb witn text generation on it's own, with images, and function calling


TuanaCelik:
PS: Adding the function calling support with these packages was a bit fiddly and the documentation for Gemini suggests that this is still experimental and in beta. So up to you whether you want to keep it or not

integrations/gemini-haystack/README.md
```diff
@@ -0,0 +1,21 @@
...
+[![PyPI - Python Version](https://img.shields.io/pypi/pyversions/gemini-haystack.svg)](https://pypi.org/project/gemini-haystack)
+
+-----
+
+**Table of Contents**
+
+- [Installation](#installation)
+- [License](#license)
+
+## Installation
```
annthurium, line 13: do we direct readers to the docs for more information? I've noticed there are often more comprehensive code samples over there (and I agree it would be a pain to maintain them everywhere, especially for haystack core integrations.

silvanocerza:
I added some tests and related workflow, refactored the serialization of the components and also renamed them by prefixing them with `GoogleAI`. We're also going to rename the one in `vertex-haystack` with the `VertexAI` prefix to keep them in line with these components.

Most noticeable change is the rename of the package from `gemini-haystack` to `google-ai-haystack`. Better to be generic in this case, we don't know if Google is going to release other models or something different.

I also renamed the integration folder name to use underscores or the release process wouldn't work.

silvanocerza (approved):

Issue #154, opened 2023-12-28
Chroma DocumentStore throws error at creation if collection exists


TuanaCelik:
**Describe the bug**
When intializing the `ChromaDocumentStore`, if you rerun the cell without changing the collection name, you will get an error saying telling you collection already exists.

Expected behaviour: Instead of the init calling 'create' by default, if collection already exists, it should simply connect to the existing collection. The current state means we cannot reuse a collection lateron.

**To Reproduce**
Any colab that has a cell with `document_store = ChromaDocumentSrore()`
Here is one for example: https://colab.research.google.com/drive/19NzliNb5ZUo1fUbplwnUrZpNBxs9HvoR?usp=sharing

**Describe your environment (please complete the following information):**
 - Integration version: chroma_haystack-0.9.0

PR #157, opened 2023-12-31
fix: Chroma DocumentStore creation for pre-existing collection name

sahusiddharth:
fixes: #154 

This PR addresses a bug in the ChromaDocumentStore initialization process. 

Used the inbuilt chroma function `get_or_create_collection` which checks if the specified collection already exists. If it does, the ChromaDocumentStore now connects to the existing collection instead of attempting to create a new one.


anakin87 (approved):
Hey!

Thanks for fixing this :heart_decoration:

Issue #151, opened 2023-12-27
The default model for `VertexAIImageGenerator` should be `imagegeneration`


bilgeyucel:
**Describe the bug**
According to [Vertex AI documentation](https://cloud.google.com/vertex-ai/docs/generative-ai/learn/models#imagen-models), the model for image generation is `imagegeneration`. Default model right now is `imagetext` which is a model for image captioning

PR #158, opened 2023-12-31
fixed: The default model of VertexAIImagegenerator

sahusiddharth:
fixes: #151 

This PR addresses a discrepancy in the default model used by VertexAIImageGenerator. According to the Vertex AI documentation, the correct default model for image generation is `imagegeneration`. However, the current default model is set to `imagetext`, which is designed for image captioning.

Updated the default model for VertexAIImageGenerator to `imagegeneration` to align with the Vertex AI documentation.


silvanocerza (approved):
Cool, thanks! 🙏

PR #166, opened 2024-01-03
Add VertexAI prefix to GeminiGenerator and GeminiChatGenerator components

silvanocerza:
To differentiate generators using Gemini we add this `VertexAI` prefix as we'd done with the `google-ai-haystack` integration that has the `GoogleAI` prefix for the Gemini generators.


anakin87 (approved):
:+1:

PR #178, opened 2024-01-08
Cohere - change metadata to meta

anakin87:
Follows #6570

Now that haystack-ai has been released,
we should also update the Cohere Chat Generator, which uses the old `metadata` attribute of Chat Messages.


vblagoje:
Nice one, thanks @anakin87 🚢  it

Issue #177, opened 2024-01-08
Llama.cpp Generator


anakin87:
## Summary and motivation

[Llama.cpp](https://github.com/ggerganov/llama.cpp) is an important project in this field and having an integration with it would be great.
@awinml worked on something related.

## Checklist

If the request is accepted, ensure the following checklist is complete before closing this issue.

- [x] The code was merged in the `main` branch
- [x] Docs are published at https://docs.haystack.deepset.ai/
- [x] There is a Github workflow running the tests for the integration nightly and at every PR
- [x] A label named like `integration:<your integration name>` has been added to this repo
- [x] The [labeler.yml](https://github.com/deepset-ai/haystack-core-integrations/blob/main/.github/labeler.yml) file has been updated
- [x] The package has been released on PyPI
- [x] An integration tile has been added to https://github.com/deepset-ai/haystack-integrations https://github.com/deepset-ai/haystack-integrations/pull/118
- [x] The integration has been listed in the [Inventory section](https://github.com/deepset-ai/haystack-core-integrations#inventory) of this repo README
- [x] There is an example available to demonstrate the feature
- [x] #259

anakin87:
@awinml can you also open a PR in https://github.com/deepset-ai/haystack-integrations? 🙏

awinml:
@anakin87 Yes, I will do that!

PR #179, opened 2024-01-08
feat: Add Llama.cpp Generator

awinml:
### Related Issues:

Migrated from: https://github.com/awinml/llama-cpp-haystack/tree/main

fixes parts of #177 

### Proposed Changes

Introduces `LlamaCppGenerator` to load models quantized using llama.cpp (GGUF) for text generation.

- Examples have been added to the `examples` directory and README, including a standalone example and one that shows usage in a RAG pipeline.
- A new label `llama_cpp` has been added to the [labeler.yml](https://github.com/deepset-ai/haystack-core-integrations/blob/main/.github/labeler.yml).
- A Github workflow (`llama_cpp.yml`) has been added to run the tests suite on the CI.

### Design

```python
@component
class LlamaCppGenerator:
    def __init__(
        self,
        model_path: str,
        n_ctx: Optional[int] = 0,
        n_batch: Optional[int] = 512,
    ):
    ...

    def warm_up(self):
        if self.model is None:
            self.model = Llama(**self.model_kwargs)

    @component.output_types(replies=List[str], meta=List[Dict[str, Any]])
    def run(self, prompt: str, generation_kwargs: Optional[Dict[str, Any]] = None):
        ...
        return {"replies": replies, "meta": [output]}
```

- The `warm_up()` method stores the loaded model config in `self.model`. Llama.cpp loads the model into memory during the generation call, so an explicit backend is not required.

- The `model_path`, `n_ctx`, `n_batch` arguments have been exposed for convenience and can be directly passed to the Generator during initialization as keyword arguments.  

- The `model_kwargs` parameter can be used to pass additional arguments when initializing the model. In case of duplication, these kwargs override `model_path`, `n_ctx`, and `n_batch` init parameters.

- The `generation_kwargs` parameter can be used to pass additional generation arguments like `max_tokens`, `temperature`, `top_k`, `top_p`, etc to the model during inference. 

- The `generation_kwargs` can also be passed to the `run` method of the generator directly.

See Llama.cpp's [model documentation](https://llama-cpp-python.readthedocs.io/en/latest/api-reference/#llama_cpp.Llama.__init__) and [`create_completion` documentation](https://llama-cpp-python.readthedocs.io/en/latest/api-reference/#llama_cpp.Llama.create_completion) for more information on the available model arguments and generation parameters.

### Notes for reviewer

The [OpenChat GGUF](https://huggingface.co/TheBloke/openchat-3.5-1210-GGUF/resolve/main/openchat-3.5-1210.Q3_K_S.gguf) is downloaded in the `tests/models` directory and used for the integration tests. The file size of the model is ~3.5 GB.


README.md
```diff
@@ -73,6 +73,7 @@ deepset-haystack
...
+| [ollama-haystack](integrations/ollama/)                                         | Generator           | [![PyPI - Version](https://img.shields.io/pypi/v/ollama-haystack.svg?color=orange)](https://pypi.org/project/ollama-haystack)                            | [![Test / ollama](https://github.com/deepset-ai/haystack-core-integrations/actions/workflows/ollama.yml/badge.svg)](https://github.com/deepset-ai/haystack-core-integrations/actions/workflows/ollama.yml)                                                               |
+| [llama-cpp-haystack](integrations/llama_cpp/)                                   | Generator           | [![PyPI - Version](https://img.shields.io/pypi/v/ollama-haystack.svg?color=orange)](https://pypi.org/project/llama-cpp-haystack)                         | [![Test / llama-cpp](https://github.com/deepset-ai/haystack-core-integrations/actions/workflows/llama_cpp.yml/badge.svg)](https://github.com/deepset-ai/haystack-core-integrations/actions/workflows/llama_cpp.yml)                                                      |
```
masci, line 79: nit: can we put this entry in alphabetical order?

integrations/llama_cpp/pyproject.toml
```diff
@@ -0,0 +1,184 @@
...
+    "integration: marks tests as slow (deselect with '-m \"not integration\"')",
+]
+addopts = [
+    "--import-mode=importlib",
+]
+
+[[tool.mypy.overrides]]
+module = [
+    "haystack.*",
+    "pytest.*"
```
masci, line 182: The CI is failing on mypy, I think you need to add `llama_cpp` here

masci (changes requested):
Thanks for this contribution! I left a couple of comments but it looks good.

README.md
awinml: Fixed, reordered some of the other entries as well.

masci (approved):
🚀

PR #182, opened 2024-01-09
Add healthcheck to Ollama service

anakin87:
Sometimes Ollama tests fail because the Ollama service is not fully ready when the next steps in the workflow are executed.
See for example https://github.com/deepset-ai/haystack-core-integrations/actions/runs/7454708762/job/20282535972

This PR adds a healthcheck that should avoid this problem.


silvanocerza (approved):
Nice workaround. 👍

PR #184, opened 2024-01-09
Add more docstrings for `ElasticsearchDocumentStore` and `ElasticsearchBM25Retriever`

silvanocerza:
Solves deepset-ai/haystack#6601

### DIFF
--- a/integrations/elasticsearch/src/elasticsearch_haystack/bm25_retriever.py
+++ b/integrations/elasticsearch/src/elasticsearch_haystack/bm25_retriever.py
@@ -52,18 +52,13 @@ class ElasticsearchBM25Retriever:
         Initialize ElasticsearchBM25Retriever with an instance ElasticsearchDocumentStore.
 
         :param document_store: An instance of ElasticsearchDocumentStore.
-        :type document_store: ElasticsearchDocumentStore
         :param filters: Filters applied to the retrieved Documents, for more info
                         see `ElasticsearchDocumentStore.filter_documents`, defaults to None
-        :type filters: Optional[Dict[str, Any]], optional
         :param fuzziness: Fuzziness parameter passed to Elasticsearch, defaults to "AUTO".
                           see the official documentation for valid values:
                           https://www.elastic.co/guide/en/elasticsearch/reference/current/common-options.html#fuzziness
-        :type fuzziness: str, optional
         :param top_k: Maximum number of Documents to return, defaults to 10
-        :type top_k: int, optional
         :param scale_score: If `True` scales the Document`s scores between 0 and 1, defaults to False
-        :type scale_score: bool, optional
         """
 
         if not isinstance(document_store, ElasticsearchDocumentStore):
--- a/integrations/elasticsearch/src/elasticsearch_haystack/bm25_retriever.py
+++ b/integrations/elasticsearch/src/elasticsearch_haystack/bm25_retriever.py
@@ -37,6 +37,7 @@ class ElasticsearchBM25Retriever:
     result = retriever.run(query="Who lives in Berlin?")
     for doc in result["documents"]:
         print(doc.text)
+    ```
     """
 
     def __init__(
--- a/integrations/elasticsearch/src/elasticsearch_haystack/bm25_retriever.py
+++ b/integrations/elasticsearch/src/elasticsearch_haystack/bm25_retriever.py
@@ -11,6 +11,34 @@ from elasticsearch_haystack.document_store import ElasticsearchDocumentStore
 
 @component
 class ElasticsearchBM25Retriever:
+    """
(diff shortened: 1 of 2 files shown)


integrations/elasticsearch/src/elasticsearch_haystack/bm25_retriever.py
```diff
@@ -11,6 +11,34 @@
...
+        Document(text="My name is Paul and I live in New York"),
+        Document(text="My name is Silvano and I live in Matera"),
+        Document(text="My name is Usagi Tsukino and I live in Tokyo"),
+    ]
+    document_store.write_documents(documents)
+
+    result = retriever.run(query="Who lives in Berlin?")
+    for doc in result["documents"]:
+        print(doc.text)
+    """
```
shadeMe, line 40: Looks like we're missing a closing ` ``` ` here?

integrations/elasticsearch/src/elasticsearch_haystack/bm25_retriever.py
```diff
@@ -20,6 +48,24 @@ def __init__(
...
+        Initialize ElasticsearchBM25Retriever with an instance ElasticsearchDocumentStore.
+
+        :param document_store: An instance of ElasticsearchDocumentStore.
+        :type document_store: ElasticsearchDocumentStore
+        :param filters: Filters applied to the retrieved Documents, for more info
+                        see `ElasticsearchDocumentStore.filter_documents`, defaults to None
+        :type filters: Optional[Dict[str, Any]], optional
+        :param fuzziness: Fuzziness parameter passed to Elasticsearch, defaults to "AUTO".
+                          see the official documentation for valid values:
+                          https://www.elastic.co/guide/en/elasticsearch/reference/current/common-options.html#fuzziness
```
shadeMe, line 59: Shouldn't we be following the [RST syntax](https://sublime-and-sphinx-guide.readthedocs.io/en/latest/references.html#links-to-external-web-pages)?

integrations/elasticsearch/src/elasticsearch_haystack/bm25_retriever.py
```diff
@@ -20,6 +48,24 @@ def __init__(
         top_k: int = 10,
         scale_score: bool = False,
     ):
+        """
+        Initialize ElasticsearchBM25Retriever with an instance ElasticsearchDocumentStore.
+
+        :param document_store: An instance of ElasticsearchDocumentStore.
+        :type document_store: ElasticsearchDocumentStore
+        :param filters: Filters applied to the retrieved Documents, for more info
+                        see `ElasticsearchDocumentStore.filter_documents`, defaults to None
```
shadeMe, line 56: Sphinx should automatically generate the default value, right? (Same question everywhere else).

shadeMe (changes requested):

integrations/elasticsearch/src/elasticsearch_haystack/bm25_retriever.py
silvanocerza: As far as I know it doesn't, and I can't find anything about it either to be fair. 🤔

integrations/elasticsearch/src/elasticsearch_haystack/bm25_retriever.py
silvanocerza: Good point. ☝️

silvanocerza: Now that I think about it this should be valid RST syntax, the link will be highlighted and threated as such. 🤔

integrations/elasticsearch/src/elasticsearch_haystack/bm25_retriever.py
shadeMe: I know that it has support for it having used it in [previous projects](https://curated-transformers.readthedocs.io/en/latest/decoders.html#curated_transformers.models.TransformerDecoder.forward), but it might be something that needs to be configured/theme-dependent.

If we currently always specify the default value in the docstring, we'll leave the defaults in this PR. But in that case, we should open an issue to rectify this globally.

silvanocerza: As of now we specify it, same thing in the Haystack core repo. I totally agree that we need more standardization on the documentation side. 

Tagging @dfokina to keep her in the loop.

integrations/elasticsearch/src/elasticsearch_haystack/bm25_retriever.py
shadeMe: I'll take your word for it 👍  

That being said, we might want to discuss elsewhere if using hyperlinks (as opposed to bare URLs) improves readability. And on a related note, we should also add a CI workflow that automatically deploys documentation changes to a preview service

shadeMe (approved):

PR #186, opened 2024-01-10
Setup everything to start working on the Weaviate integration

silvanocerza:
This is the first commit for the Weaviate integration. It doesn't implement anything as of now, it just adds all the necessary files to start working on it.

I also added the label `integration:weaviate`.


masci (approved):
🚀

Issue #131, opened 2023-12-20
Weaviate Document Store


silvanocerza:
## Summary and motivation

We want to support Weaviate as a Document Store in Haystack 2.x, much like we did for Haystack 1.x

## Checklist

If the request is accepted, ensure the following checklist is complete before closing this issue.

- [x] The code was merged in the `main` branch
- [x] Docs are published at https://docs.haystack.deepset.ai/
- [x] There is a Github workflow running the tests for the integration nightly and at every PR
- [x] A label named like `integration:<your integration name>` has been added to this repo
- [x] The [labeler.yml](https://github.com/deepset-ai/haystack-core-integrations/blob/main/.github/labeler.yml) file has been updated
- [x] The package has been released on PyPI
- [x] An integration tile has been added to https://github.com/deepset-ai/haystack-integrations
- [x] The integration has been listed in the [Inventory section](https://github.com/deepset-ai/haystack-core-integrations#inventory) of this repo README
- [x] There is an example available to demonstrate the feature
- [x] The feature was announced through social media

silvanocerza:
Integration has been announced on socials, everything is done.

PR #187, opened 2024-01-10
WeaviateDocumentStore initialization and serialization

silvanocerza:
Relates to #131 

This PR implements a minimal initialization of `WeaviateDocumentStore` and the internal Weaviate client.

It also adds proper serialization of the Document Store.

We also fix the `weaviate-client` version to avoid breakages whenever the next upcoming version is released.

### DIFF
--- a/integrations/weaviate/pyproject.toml
+++ b/integrations/weaviate/pyproject.toml
@@ -26,7 +26,7 @@ classifiers = [
 ]
 dependencies = [
   "haystack-ai",
-  "weaviate-client",
+  "weaviate-client==3.*",
 ]
 
 [project.urls]
@@ -46,6 +46,7 @@ git_describe_command = 'git describe --tags --match="integrations/weaviate-v[0-9
 dependencies = [
   "coverage[toml]>=6.5",
   "pytest",
+  "ipython",
 ]
 [tool.hatch.envs.default.scripts]
 test = "pytest {args:tests}"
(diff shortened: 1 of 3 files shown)


integrations/weaviate/src/weaviate_haystack/document_store.py
```diff
@@ -1,2 +1,157 @@
...
+    def filter_documents(self, filters: Optional[Dict[str, Any]] = None) -> List[Document]:  # noqa: ARG002
+        return []
+
+    def write_documents(
+        self, documents: List[Document], policy: DuplicatePolicy = DuplicatePolicy.NONE  # noqa: ARG002
+    ) -> int:
+        return 0
+
+    def delete_documents(self, document_ids: List[str]) -> None:  # noqa: ARG002
+        return
```
silvanocerza, line 157: `ARG002` is ignored for the time being to let CI pass. In future PRs they'll be implement and the `noqa` removed.

integrations/weaviate/pyproject.toml
```diff
@@ -46,6 +46,7 @@ git_describe_command = 'git describe --tags --match="integrations/weaviate-v[0-9
 dependencies = [
   "coverage[toml]>=6.5",
   "pytest",
+  "ipython",
```
ZanSara, line 49: We need ipython as a dependency?

ZanSara (commented):
Only one question, otherwise LGTM

ZanSara (approved):

PR #191, opened 2024-01-11
Ollama: improve test

anakin87:
This test sometimes fails ([an example](https://github.com/deepset-ai/haystack-core-integrations/actions/runs/7482286030/job/20365554468)) because this small model (orca-mini)
refers to Glasgow as the second largest city in UK. :smile:


ZanSara (approved):
:smile:

Issue #208, opened 2024-01-15
opensearch: mount integration under haystack_integrations.*


masci:
## Checklist

- [x] Update the structure, the code, and examples on haystack-core-integrations https://github.com/deepset-ai/haystack-core-integrations/pull/212
- [x] Update the integration tile on https://github.com/deepset-ai/haystack-integrations https://github.com/deepset-ai/haystack-integrations/pull/133
- [x] Update the docs at https://docs.haystack.deepset.ai/
- [x] Update the notebooks on https://github.com/deepset-ai/haystack-cookbook (if needed)
```[tasklist]
### Tasks
- [x] Update artici
```

TuanaCelik:
I've updated the cookbook and ticked it off

Issue #100, opened 2023-12-13
Have a separate folder for each component type under integrations


bilgeyucel:
**Is your feature request related to a problem? Please describe.**
As a user, I'm familiar with the convention at haystack repository and having the same convention here would make it easy for me to navigate.

**Describe the solution you'd like**
Create a `/generators` folder under `cohere_haystack` integration folder. 
The same can apply to `jina_haystack` and `unstructured`. This way, we can be more open to new components coming from the same providers

**Describe alternatives you've considered**
We can leave it as it is.

**Additional context**
N/A

```[tasklist]
### Tasks
- [ ] https://github.com/deepset-ai/haystack-core-integrations/issues/195
- [ ] https://github.com/deepset-ai/haystack-core-integrations/issues/196
- [ ] https://github.com/deepset-ai/haystack-core-integrations/issues/197
- [ ] https://github.com/deepset-ai/haystack-core-integrations/issues/198
- [ ] https://github.com/deepset-ai/haystack-core-integrations/issues/199
- [ ] https://github.com/deepset-ai/haystack-core-integrations/issues/200
- [ ] https://github.com/deepset-ai/haystack-core-integrations/issues/201
- [ ] https://github.com/deepset-ai/haystack-core-integrations/issues/202
- [ ] https://github.com/deepset-ai/haystack-core-integrations/issues/203
- [ ] https://github.com/deepset-ai/haystack-core-integrations/issues/210
- [ ] https://github.com/deepset-ai/haystack-core-integrations/issues/209
- [ ] https://github.com/deepset-ai/haystack-core-integrations/issues/208
- [ ] https://github.com/deepset-ai/haystack-core-integrations/issues/207
- [ ] https://github.com/deepset-ai/haystack-core-integrations/issues/206
- [ ] https://github.com/deepset-ai/haystack-core-integrations/issues/205
- [ ] https://github.com/deepset-ai/haystack-core-integrations/issues/204
- [ ] https://github.com/deepset-ai/haystack-core-integrations/issues/213
```

masci:
We're going to uniform the import path of all the integrations under the namespace `haystack_integrations.*`.

For example, this means that a document store `FooDocumentStore` shipped with the integration `baz-haystack` will be imported like this:

```py
from haystack_integrations.document_stores.baz import FooDocumentStore
```

Similarly, a generator `AwesomeGenerator` shipped with the integration `foo-haystack` will be imported like this:

```py
[...]

PR #212, opened 2024-01-15
move package under haystack_integrations/*

masci:
Part of #208 

See #100 for details

Note: this PR requires https://github.com/deepset-ai/haystack/pull/6714 that's not present in beta4, need to wait for beta5 to let the tests pass


vblagoje:
@masci Do we have to do this for all integrations, and this is a prototype of how to do it for other integrations?

masci:
@vblagoje we have to do similar changes to all the integrations, see #100 it contains the list

PR #214, opened 2024-01-16
Move package under haystack_integrations

silvanocerza:
Closes #204

PR #215, opened 2024-01-16
Add `collection_name` parameter and creation

silvanocerza:
Depends on #214.

Now `WeaviateDocumentStore` can be initialized specifying a collection name, if that collection doesn't exist it will be created.


anakin87 (approved):
looks good!

Issue #210, opened 2024-01-15
llama_cpp: mount integration under haystack_integrations.*


masci:
## Checklist

- [x] Update the structure, the code, and examples on haystack-core-integrations
- [x] Release a new version of the package
- [x] Update the integration tile on https://github.com/deepset-ai/haystack-integrations (@bilgeyucel is taking care in https://github.com/deepset-ai/haystack-integrations/pull/118)
- [X] Update the docs at https://docs.haystack.deepset.ai/
- [x] Update the notebooks on https://github.com/deepset-ai/haystack-cookbook (if needed)

PR #217, opened 2024-01-16
Mount llama_cpp in haystack_integrations

anakin87:
Part of #210


shadeMe (approved):
LGTM, but I'm not super familiar with all the details. Having another set of eyes could be useful here; tagging @masci.

masci:
Looks good! 👍

Issue #142, opened 2023-12-22
Add PGVector DocumentStore


mathislucka:
## Summary and motivation
PGVector is a popular request from our community. We should have it in 2.0.

## Detailed design

TBD

## Checklist

If the request is accepted, ensure the following checklist is complete before closing this issue.

- [x] The code was merged in the `main` branch
- [x] Docs are published at https://docs.haystack.deepset.ai/
- [x] There is a Github workflow running the tests for the integration nightly and at every PR
- [x] A label named like `integration:<your integration name>` has been added to this repo
- [x] The [labeler.yml](https://github.com/deepset-ai/haystack-core-integrations/blob/main/.github/labeler.yml) file has been updated
- [x] The package has been released on PyPI
- [x] An integration tile has been added to https://github.com/deepset-ai/haystack-integrations (https://github.com/deepset-ai/haystack-integrations/pull/162)
- [x] The integration has been listed in the [Inventory section](https://github.com/deepset-ai/haystack-core-integrations#inventory) of this repo README
- [x] There is an example available to demonstrate the feature
- [x] The feature was announced through social media

sahusiddharth:
@mathislucka I have started working on this, can you clarify something for me?

When it comes to metadata filtering there is way suggested in the [template](https://github.com/deepset-ai/document-store), but the one followed in [chroma docstore](https://github.com/deepset-ai/haystack-core-integrations/blob/main/integrations/chroma/src/chroma_haystack/document_store.py#L55) is quite different, how should I proceed from here?

anakin87:
Hey @sahusiddharth, sorry for the late reply.

I recently developed the [Pinecone Document Store](https://github.com/deepset-ai/haystack-core-integrations/tree/main/integrations/pinecone), so I try to share some suggestions.

- the official metadata filtering is that of the template. You can find more information [in the docs](https://docs.haystack.deepset.ai/v2.0/docs/metadata-filtering).
I think filters in Chroma Document Store are the legacy filters.

- as explained in [Creating Custom Document Stores](https://docs.haystack.deepset.ai/v2.0/docs/creating-custom-document-stores), each Document Store must respect the Document Store protocol, by implementing the 4 standard methods (respecting their signature). For functions that are specific to your Document Store, you are free to implement any other methods you need.

[...]

PR #239, opened 2024-01-18
Pgvector Document Store - minimal implementation

anakin87:
Part of #142

Minimal implementation of the Document Store logic using psycopg3.

Missing parts (to be added in future PRs):
- filters
- embedding retrieval


integrations/pgvector/src/haystack_integrations/document_stores/pgvector/document_store.py
```diff
@@ -0,0 +1,377 @@
...
+
+
+class PgvectorDocumentStore:
+    def __init__(
+        self,
+        *,
+        connection_string: str,
+        table_name: str = "haystack_documents",
+        embedding_dimension: int = 768,
+        embedding_similarity_function: Literal[
```
masci, line 52: I'm not sure about this... We're introducing new names like `cosine_distance` that don't exist in pgvector, so they need to be documented, won't be part of any docs outside Haystack and won't be familiar to pgvector users coming to Haystack. I think changing this parameter to `postgresql_ops` or `pgvector_ops`  and having the literal set to `"vector_cosine_ops", "vector_ip_ops", "vector_l2_ops",`  would be more explicit.

integrations/pgvector/src/haystack_integrations/document_stores/pgvector/document_store.py
```diff
@@ -0,0 +1,377 @@
...
+            self._connection.rollback()
+            raise DocumentStoreError(error_msg) from e
+        return result
+
+    def _create_table_if_not_exists(self):
+        """
+        Creates the table to store Haystack documents if it doesn't exist yet.
+        """
+
+        table_structure_str = ", ".join(f"{name} {dtype} {constraint}" for name, dtype, constraint in TABLE_DEFINITION)
```
masci, line 157: Why not storing the SQL statement as a string? Something like
```py
CREATE_TABLE_STMT = """
CREATE TABLE IF NOT EXISTS %s (
'id' VARCHAR(128) PRIMARY KEY,
'embedding' VECTOR(%s),
...
);
"""

self._execute_sql(CREATE_TABLE_STMT, params={...}, error_msg="Could not create table in PgvectorDocumentStore")
```

To me this would have several advantages:
- Code is easier to read than this for-loop
- I can go and copy the query to paste it in my pg console for testing
- If you know SQL you know what's going to happen at glance

integrations/pgvector/src/haystack_integrations/document_stores/pgvector/document_store.py
```diff
@@ -0,0 +1,377 @@
...
+        self._execute_sql(create_sql, error_msg="Could not create table in PgvectorDocumentStore")
+
+    def delete_table(self):
+        """
+        Deletes the table used to store Haystack documents.
+        """
+
+        delete_sql = SQL("DROP TABLE IF EXISTS {table_name}").format(table_name=Identifier(self.table_name))
+
+        self._execute_sql(delete_sql, error_msg="Could not delete table in PgvectorDocumentStore")
```
masci, line 172: Let's put the name of the table in the error message

integrations/pgvector/src/haystack_integrations/document_stores/pgvector/document_store.py
```diff
@@ -0,0 +1,377 @@
...
+                "SELECT 1 FROM pg_indexes WHERE tablename = %s AND indexname = %s",
+                (self.table_name, HNSW_INDEX_NAME),
+                "Could not check if HNSW index exists",
+            ).fetchone()
+        )
+
+        if index_esists and not self.hnsw_recreate_index_if_exists:
+            logger.warning(
+                "HNSW index already exists and won't be recreated. "
+                "If you want to recreate it, set hnsw_recreate_index=True"
```
masci, line 197: "If you want to recreate it, pass 'hnsw_recreate_index_if_exists=True' to the document store constructor" 

or something like that...

integrations/pgvector/src/haystack_integrations/document_stores/pgvector/document_store.py
```diff
@@ -0,0 +1,377 @@
...
+
+        self._create_hnsw_index()
+
+    def _create_hnsw_index(self):
+        """
+        Internal method to create the HNSW index.
+        """
+
+        pg_ops = SIMILARITY_FUNCTION_TO_POSTGRESQL_OPS[self.embedding_similarity_function]
+        effective_hnsw_index_creation_kwargs = {
```
masci, line 212: I think you mean `actual_hnsw_index_creation_kwargs`?

masci (changes requested):
took a first pass

integrations/pgvector/src/haystack_integrations/document_stores/pgvector/document_store.py
anakin87: - these names should be familiar for pgvector users: see https://github.com/pgvector/pgvector?tab=readme-ov-file#vector-functions

- regarding `embedding_similarity_function`, it is more of a Haystack concept... Can we call it `vector_distance`?

integrations/pgvector/src/haystack_integrations/document_stores/pgvector/document_store.py
anakin87: will do!

integrations/pgvector/src/haystack_integrations/document_stores/pgvector/document_store.py
masci: What about `vector_function` with reference to https://github.com/pgvector/pgvector?tab=readme-ov-file#vector-functions ?

integrations/pgvector/src/haystack_integrations/document_stores/pgvector/document_store.py
```diff
@@ -0,0 +1,380 @@
...
+            policy = DuplicatePolicy.FAIL
+
+        db_documents = self._from_haystack_to_pg_documents(documents)
+
+        columns_str = "(" + ", ".join(col for col in COLUMNS) + ")"
+        values_placeholder_str = "VALUES (" + ", ".join(f"%({col})s" for col in COLUMNS) + ")"
+
+        insert_statement = SQL("INSERT INTO {table_name} " + columns_str + " " + values_placeholder_str).format(
+            table_name=Identifier(self.table_name)
+        )
```
masci, line 284: This is very hard to read, I would explicit the query in a string variable `INSERT_STATEMENT` if possible

integrations/pgvector/src/haystack_integrations/document_stores/pgvector/document_store.py
```diff
@@ -0,0 +1,380 @@
...
+
+        if policy == DuplicatePolicy.OVERWRITE:
+            update_statement = SQL(
+                "ON CONFLICT (id) DO UPDATE SET "
+                + ", ".join(f"{col} = EXCLUDED.{col}" for col in COLUMNS if col != "id")
```
masci, line 289: Same as above, this could be a `UPDATE_STATEMENT` variable where I can easily read what will happen

integrations/pgvector/src/haystack_integrations/document_stores/pgvector/document_store.py
```diff
@@ -0,0 +1,380 @@
...
+        """
+
+        haystack_documents = []
+        for document in documents:
+            haystack_dict = dict(document)
+            blob_data = haystack_dict.pop("blob_data")
+            blob_meta = haystack_dict.pop("blob_meta")
+            blob_mime_type = haystack_dict.pop("blob_mime_type")
+
+            if not haystack_dict["meta"]:
```
masci, line 351: `db_document["meta"]` contains `Json(...)`, what does `Document.meta` contain after `Document.from_dict`?

integrations/pgvector/src/haystack_integrations/document_stores/pgvector/document_store.py
```diff
@@ -0,0 +1,380 @@
...
+        written_docs = 0
+        while True:
+            if self._cursor.fetchone():
+                written_docs += 1
+            if not self._cursor.nextset():
+                break
+
+        return written_docs
+
+    def _from_haystack_to_pg_documents(self, documents: List[Document]) -> List[Dict[str, Any]]:
```
masci, line 317: can we add a unit test for this function?

integrations/pgvector/src/haystack_integrations/document_stores/pgvector/document_store.py
```diff
@@ -0,0 +1,380 @@
...
+            db_document["blob_mime_type"] = blob.mime_type if blob and blob.mime_type else None
+
+            db_document["dataframe"] = Json(db_document["dataframe"]) if db_document["dataframe"] else None
+            db_document["meta"] = Json(db_document["meta"])
+
+            db_documents.append(db_document)
+
+        return db_documents
+
+    def _from_pg_to_haystack_documents(self, documents: List[Dict[str, Any]]) -> List[Document]:
```
masci, line 339: can we add a unit test for this function?

integrations/pgvector/src/haystack_integrations/document_stores/pgvector/document_store.py
```diff
@@ -0,0 +1,380 @@
...
+        if actual_hnsw_index_creation_kwargs:
+            actual_hnsw_index_creation_kwargs_str = ", ".join(
+                f"{key} = {value}" for key, value in actual_hnsw_index_creation_kwargs.items()
+            )
+            sql_add_creation_kwargs = SQL("WITH ({creation_kwargs_str})").format(
+                creation_kwargs_str=SQL(actual_hnsw_index_creation_kwargs_str)
+            )
+            sql_create_index += sql_add_creation_kwargs
+
+        self._execute_sql(sql_create_index, error_msg="Could not create HNSW index")
```
masci, line 231: Do we know why `Could not create HNSW index`? Any way we can add the actual error to these messages?

integrations/pgvector/src/haystack_integrations/document_stores/pgvector/document_store.py
anakin87: OK. Let's do that

integrations/pgvector/src/haystack_integrations/document_stores/pgvector/document_store.py
anakin87: sorry this condition is useless.
`haystack_dict['meta']` is always a valid dict and not None.
I'll remove this condition.

integrations/pgvector/src/haystack_integrations/document_stores/pgvector/document_store.py
anakin87: In this case, we don't know...

this expression in `_execute_sql` should also show the actual error:
```python
try:
    result = cursor.execute(sql_query, params)
except Error as e:
    self._connection.rollback()
    raise DocumentStoreError(error_msg) from e
return result
```

Or did you mean something else?

masci: Ah no, shame on me, `from e` does what I want

masci (approved):
Noice! :shipit:

PR #240, opened 2024-01-18
Remove ChromaSingleQueryRetriever

masci:
The retriever was originally introduced when Haystack components expected to receive `List[List[Document]]`, but we lately standardized the format to pass single queries and receive one single list of results. For this reason `ChromaSingleQueryRetriever` is being removed and `ChromaQueryRetriever` changed so it looks like the other retrievers and can be used as a drop-in replacement in existing pipelines.

### DIFF
--- a/integrations/chroma/src/haystack_integrations/components/retrievers/chroma/retriever.py
+++ b/integrations/chroma/src/haystack_integrations/components/retrievers/chroma/retriever.py
@@ -40,8 +40,8 @@ class ChromaQueryRetriever:
 
         :raises ValueError: If the specified document store is not found or is not a MemoryDocumentStore instance.
         """
-        if not top_k:
-            top_k = 3
+        top_k = top_k or self.top_k
+
         return {"documents": self.document_store.search([query], top_k)[0]}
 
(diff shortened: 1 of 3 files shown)


integrations/chroma/src/haystack_integrations/components/retrievers/chroma/retriever.py
```diff
@@ -25,24 +25,24 @@ def __init__(self, document_store: ChromaDocumentStore, filters: Optional[Dict[s
...
         Run the retriever on the given input data.
 
-        :param queries: The input data for the retriever. In this case, a list of queries.
+        :param query: The input data for the retriever. In this case, a plain-text query.
         :return: The retrieved documents.
 
         :raises ValueError: If the specified document store is not found or is not a MemoryDocumentStore instance.
         """
         if not top_k:
             top_k = 3
```
anakin87, line 44: unrelated to this PR.

Why not `top_k = top_k or self.top_k`?

anakin87 (approved):
Looks good.

I left an unrelated comment...

Issue #241, opened 2024-01-18
UnstructuredFileConverter: support custom metadata dict


lambda-science:
**Is your feature request related to a problem? Please describe.**
Converter such as `PyPDFToDocument` take as input `file_paths` and an optional `meta` list of dict such as `indexing_pipeline.run({"PDFToTextConverter": {"sources": file_paths, "meta": file_metas}})`  
Currently `UnstructuredFileConverter` only support `paths` parameter such as: `indexing_pipeline.run({"UnstructuredFileConverter": {"paths": file_paths}})`, adding `meta` list of dict results in a crash. Also it automatically extract it's own `meta` (which is good, but I wish to extend them)
![image](https://github.com/deepset-ai/haystack-core-integrations/assets/20109584/4afd30c7-9962-4f3b-947d-95ead832cd3b)

**Describe the solution you'd like**
Modify the `UnstructuredFileConverter` to support an optional `meta` list of dict to be added to the Document.  

**Describe alternatives you've considered**
Maybe a custom component that is like `MetadataWritter` that could add `meta` dict after the `converter` and before the `writer`. But it could be maybe messy (because `UnstructuredFileConverter` chunk documents, so the `meta` list of dict that has the length of your `file_path`, will not be the same length as output documents from `UnstructuredFileConverter` and then you have to merge two dict because the meta dict already exist in the doc

lambda-science:
> This sounds like a reasonable feature request!
> 
> Unfortunately, the implementation might be non-trivial because the `UnstructuredFileConverter` also accepts directories as `paths` and automatically searches for files.
> 
> However, if you want to open a PR that covers the simplest use cases, feel free to do so. Otherwise, we'll take a look in the future...

Working on it right now, I just have to fix a test, but it looks very doable :)

lambda-science:
Proposed MR: https://github.com/deepset-ai/haystack-core-integrations/pull/242

PR #242, opened 2024-01-18
Feat: UnstructuredFileConverter meta field

lambda-science:
Fix: https://github.com/deepset-ai/haystack-core-integrations/issues/241

This pull request add a new optional `meta` field to add custom metadata to Unstructured documents. 
From:
```python
class UnstructuredFileConverter:

    @component.output_types(documents=List[Document])
    def run(self, paths: Union[List[str], List[os.PathLike]]):
```
To:
```python
class UnstructuredFileConverter:

    @component.output_types(documents=List[Document])
    def run(
        self,
        paths: Union[List[str], List[os.PathLike]],
        meta: Optional[Union[Dict[str, Any], List[Dict[str, Any]]]] = None,
    ):
```

The implementation is inspired from the handling of metadata in the `PyPDFToDocument` component from Haystack main repo.

Ping @anakin87


lambda-science:
And maybe I should add more test case because all test case only have one file with one metadata dict. Maybe a test case with two files and a `meta` list of dict

anakin87:
@lambda-science tests are failing because of mypy. Something in the code should be fixed.

Thanks for your PR. I will take a look later or tomorrow...

lambda-science:
> @lambda-science tests are failing because of mypy. Something in the code should be fixed.
> 
> Thanks for your PR. I will take a look later or tomorrow...

No worries, I realized I could use my own fix before it is merged and released with a simple requirement.txt such as:
```
unstructured-fileconverter-haystack @ git+https://github.com/deepset-ai/haystack-core-integrations@3bc51a788356545dc692183591cf91bdf5391713#subdirectory=integrations/unstructured
```
That's very cool ! So it's not urgent at all 🌞

lambda-science:
I have the feeling that something is off with metadata, because the auto-generated page_number is always the last page with `one doc per elem`, investigating...

integrations/unstructured/src/haystack_integrations/components/converters/unstructured/converter.py
```diff
@@ -107,19 +117,20 @@ def run(self, paths: Union[List[str], List[os.PathLike]]):
 
         # currently, the files are converted sequentially to gently handle API failures
         documents = []
+        meta_list = normalize_metadata(meta, sources_count=len(all_filepaths))
```
anakin87, line 120: If `len(all_filepaths) != len meta`, `normalize_metadata` method fails with `ValueError`.
This is expected and right for common converters, but this component also loops all files in the provided directories, so this case would be quite common and I would not want the converter to fail.

I would propose a slightly different approach. Something similar to this (untested):

```python
try:
    meta_list = normalize_metadata(meta, sources_count=len(all_filepaths))
except ValueError as e:
    logger.warning("The length of the metadata list does not match the length of all filepaths (also files inside directories are included in all filepaths). The metadata list will be ignored. ")
    meta_list = [{}] * all_filepaths
```

anakin87 (commented):
Hey @lambda-science, thanks for the good work!

- I left a comment about metadata handling
- I would add a test where the path is a directory

integrations/unstructured/src/haystack_integrations/components/converters/unstructured/converter.py
lambda-science: > If `len(all_filepaths) != len meta`, `normalize_metadata` method fails with `ValueError`. This is expected and right for common converters, but this component also loops all files in the provided directories, so this case would be quite common and I would not want the converter to fail.
> 
> I would propose a slightly different approach. Something similar to this (untested):
> 
> ```python
> meta_list = [{}] * all_filepaths
> try:
>     meta_list = normalize_metadata(meta, sources_count=len(all_filepaths))
> except ValueError as e:
>     logger.warning("The length of the metadata list does not match the length of all filepaths (also files inside directories are included in all filepaths). The metadata list will be ignored. ")
> ```

I've added a test where the path is a directory.
Related to metadata handling hm, I disagree a little bit.
How it works currently (with my merge request): if there are no meta provided, it works without error and doesn't add custom metadata to documents. If metadata are provided by the user, they should match the number of files given or the number of files in the directory path, otherwiser it raises an error.

As a user of Haystack, I would much rather have a component crash and tell me "you're making a mistake" than silently working, but not the way I expected. Realising after indexing files that now my document store is full of documents without metadata because the component decided to handle missing metadata by ignoring them sounds painful to me. Because then I would have to clean my docstore and re-index my files: but cleaning part of the docstore when documents doesn't have metadata is... hard). 

Also I feel like if it only throw a Warning and doesn't raise error, I would only realise the issue with my indexation way later, through a query pipeline throwing weird results or metadata filtering not working as expected.

anakin87: Thinking about it more, I agree with you...
Let's just improve the docstring to make the behavior more explicit.

integrations/unstructured/src/haystack_integrations/components/converters/unstructured/converter.py
```diff
@@ -136,13 +148,16 @@ def _create_documents(
...
+            metadata = meta.copy()
+            metadata["name"] = str(filepath)
+            docs = [Document(content=text, meta=metadata)]
 
         elif document_creation_mode == "one-doc-per-page":
             texts_per_page: defaultdict[int, str] = defaultdict(str)
             meta_per_page: defaultdict[int, dict] = defaultdict(dict)
             for el in elements:
-                metadata = {"name": str(filepath)}
+                metadata = meta.copy()
```
anakin87, line 159:
```suggestion
                metadata = copy.deepcopy(meta)
```
[To avoid problems with nested dictionaries](https://stackoverflow.com/a/2465951/10883094), let's make a deep copy.

integrations/unstructured/src/haystack_integrations/components/converters/unstructured/converter.py
```diff
@@ -89,12 +90,21 @@ def to_dict(self) -> Dict[str, Any]:
...
+    ):
         """
         Convert files to Haystack Documents using the Unstructured API (hosted or running locally).
 
         :param paths: List of paths to convert. Paths can be files or directories.
             If a path is a directory, all files in the directory are converted. Subdirectories are ignored.
+        :param meta: Optional metadata to attach to the Documents.
+          This value can be either a list of dictionaries or a single dictionary.
+          If it's a single dictionary, its content is added to the metadata of all produced Documents.
+          If it's a list, the length of the list must match the number of sources, because the two lists will be zipped.
```
anakin87, line 106:
```suggestion
          If it's a list, the length of the list must match the number of paths, because the two lists will be zipped.
          Please note that if the paths contain directories, the length of the meta list must match the actual number of files contained.
```

this should clarify things...

anakin87 (commented):
Hey... Sorry for the long wait.

I found some small opportunities for improvement, then we can merge this good PR!

integrations/unstructured/src/haystack_integrations/components/converters/unstructured/converter.py
```diff
@@ -136,13 +148,16 @@ def _create_documents(
 
         if document_creation_mode == "one-doc-per-file":
             text = separator.join([str(el) for el in elements])
-            docs = [Document(content=text, meta={"name": str(filepath)})]
+            metadata = meta.copy()
+            metadata["name"] = str(filepath)
```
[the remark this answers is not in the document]
lambda-science: I'm very fine with this, I also found it weird. I kept it as it was to not make "breaking" changes. Like if someone already use the component and expect to be able to do `["name"]` for the `file_path`, but we can change it for sure !

integrations/unstructured/src/haystack_integrations/components/converters/unstructured/converter.py
lambda-science: Did not know about that, good catch !

anakin87 (approved):
Looks good...

Thanks @lambda-science!

I'm going to merge this PR and release a new version...

anakin87:
New version: https://pypi.org/project/unstructured-fileconverter-haystack/0.3.0/

PR #245, opened 2024-01-18
mount import path under haystack_integrations

masci:
part of https://github.com/deepset-ai/haystack-core-integrations/issues/201


anakin87 (approved):
looks good!

PR #252, opened 2024-01-22
Ollama Text Embedder with new format

AlistairLR112:
Integrating Embeddings generated from Ollama into Haystack


AlistairLR112:
@anakin87

integrations/ollama/src/haystack_integrations/components/embedders/ollama/text_embedder.py
```diff
@@ -0,0 +1,58 @@
...
+        url: str = "http://localhost:11434/api/embeddings",
+        generation_kwargs: Optional[Dict[str, Any]] = None,
+        timeout: int = 120,
+    ):
+        """
+        :param model: The name of the model to use. The model should be available in the running Ollama instance.
+            Default is "orca-mini".
+        :param url: The URL of the chat endpoint of a running Ollama instance.
+            Default is "http://localhost:11434/api/embeddings".
+        :param generation_kwargs: Optional arguments to pass to the Ollama generation endpoint, such as temperature,
```
julian-risch, line 21: These `generation_kwargs` look exactly like the ones supported by the generator. Can they really be used for the embedder too?

integrations/ollama/src/haystack_integrations/components/embedders/ollama/text_embedder.py
```diff
@@ -0,0 +1,58 @@
...
+        :return: A dictionary with the key "embedding" and a list of floats as the value
+        """
+
+        payload = self._create_json_payload(text, generation_kwargs)
+
+        response = requests.post(url=self.url, json=payload, timeout=self.timeout)
+
+        response.raise_for_status()
+
+        return response.json()
```
julian-risch, line 58: In addition to the embedding, the response of some other embedders contains the model name and other meta data. We could make the OllamaTextEmbedder more consistent with the other embedders by adding  `"meta": {"model": self.model, ...}` here. However, not all embedders currently do that so it doesn't block the PR in my opinion.

julian-risch (changes requested):
@AlistairLR112 Thank you for opening this pull request. It looks very good to me already! I just have a question about the `generation_kwargs`. What options are supported here? From the user perspective it looks the same options as for the generators but I doubt that. Can you explain the options?

integrations/ollama/src/haystack_integrations/components/embedders/ollama/text_embedder.py
AlistairLR112: From what I have seen from the ollama docs, yes -> unfortunately they haven't said if they can't

integrations/ollama/src/haystack_integrations/components/embedders/ollama/text_embedder.py
AlistairLR112: this can be added!

integrations/ollama/src/haystack_integrations/components/embedders/ollama/text_embedder.py
anakin87: I was actually dubious about this, but as @AlistairLR112 says, there are no obvious and documented differences in Ollama between the text generation endpoints and the embedding creation endpoint.

For now, let's keep the parameter as is.

integrations/ollama/src/haystack_integrations/components/embedders/ollama/text_embedder.py
AlistairLR112: @julian-risch - do you have an example? I was looking at a few and could only find "embedding" being returned

anakin87: This may be a good example: https://github.com/deepset-ai/haystack/blob/461556cca2d0a5990be05c362e8f1079422a5614/haystack/components/embedders/openai_text_embedder.py#L90

AlistairLR112: added

anakin87 (approved):
:rocket:

PR #257, opened 2024-01-23
Pgvector - filters

anakin87:


integrations/pgvector/src/haystack_integrations/document_stores/pgvector/document_store.py
```diff
@@ -162,7 +165,10 @@ def _execute_sql(
             result = cursor.execute(sql_query, params)
         except Error as e:
             self._connection.rollback()
-            raise DocumentStoreError(error_msg) from e
+            sql_query_str = sql_query.as_string(cursor) if not isinstance(sql_query, str) else sql_query
+            detailed_error_msg = f"{error_msg}.\nSQL query: {sql_query_str} \nParameters: {params}"
+            raise DocumentStoreError(detailed_error_msg) from e
```
anakin87, line 170: unrelated to this PR, but should show useful info when the query fails

integrations/pgvector/tests/test_filters.py
```diff
@@ -0,0 +1,63 @@
...
+            [
+                d
+                for d in filterable_docs
+                if (d.meta.get("number") == 100 and d.meta.get("chapter") == "intro")
+                or (d.meta.get("page") == "90" and d.meta.get("chapter") == "conclusion")
+            ],
+        )
+
+    @pytest.mark.skip(reason="NOT operator is not supported in PgVectorDocumentStore")
+    def test_not_operator(self, document_store, filterable_docs): ...
```
anakin87, line 63: This bothers me a bit.

Since we magically handle NULL values, as required by most of our base tests, it is difficult to properly support the NOT operator.
(For example, `a != 3` becomes `a IS DISTINCT FROM 3` to also include NULL values in the result.)

I also tried to support the NOT operator, but this involved inverting all nested conditions (which can also be logical) and started to get messy.

So the simplest solutions are:
- do not support the NOT operator
- drop magic handling of NULL values, consistent with Postgres behavior (but requires overriding many of our base tests).

Happy to discuss this...

integrations/pgvector/tests/test_document_store.py
```diff
@@ -14,27 +14,6 @@
 
 
 class TestDocumentStore(CountDocumentsTest, WriteDocumentsTest, DeleteDocumentsTest):
-    @pytest.fixture
```
masci, line 17: nit: to ease the review process, this could've gone later in another PR

integrations/pgvector/src/haystack_integrations/document_stores/pgvector/document_store.py
masci: Beware printing raw SQL queries, they can leak secrets and sensitive data.

integrations/pgvector/src/haystack_integrations/document_stores/pgvector/document_store.py
```diff
@@ -307,7 +331,12 @@ def write_documents(self, documents: List[Document], policy: DuplicatePolicy = D
             raise DuplicateDocumentError from ie
         except Error as e:
             self._connection.rollback()
-            raise DocumentStoreError from e
+            sql_query_str = sql_insert.as_string(self._cursor)
+            error_msg = (
+                f"Could not write documents to PgvectorDocumentStore. \n"
+                f"SQL query: {sql_query_str} \nParameters: {db_documents}"
+            )
+            raise DocumentStoreError(error_msg) from e
```
masci, line 339: Same security considerations

integrations/pgvector/src/haystack_integrations/document_stores/pgvector/filters.py
```diff
@@ -0,0 +1,228 @@
...
+}
+
+
+def _convert_filters_to_where_clause_and_params(filters: Dict[str, Any]) -> tuple[SQL, tuple]:
+    """
+    Convert Haystack filters to a WHERE clause and a tuple of params to query PostgreSQL.
+    """
+    if not isinstance(filters, dict):
+        msg = "Filters must be a dictionary"
+        raise FilterError(msg)
```
masci, line 29: Not sure you need this, you already assumed it was a dict in the caller

integrations/pgvector/src/haystack_integrations/document_stores/pgvector/filters.py
```diff
@@ -0,0 +1,228 @@
...
+    query_parts, values = [], []
+    for c in conditions:
+        query_parts.append(c[0])
+        values.append(c[1])
+    if isinstance(values[0], list):
+        values = list(chain.from_iterable(values))
+
+    operator = condition["operator"]
+    if operator == "AND":
+        sql_query = "(" + " AND ".join(query_parts) + ")"
```
masci, line 60: why not an f-string?

integrations/pgvector/src/haystack_integrations/document_stores/pgvector/filters.py
```diff
@@ -0,0 +1,228 @@
...
+        raise FilterError(msg)
+
+    return sql_query, values
+
+
+def _parse_comparison_condition(condition: Dict[str, Any]) -> tuple[str, List[Any]]:
+    if "field" not in condition:
+        # 'field' key is only found in comparison dictionaries.
+        # We assume this is a logic dictionary since it's not present.
+        return _parse_logical_condition(condition)
```
masci, line 74: This is confusing to me: something called `_parse_comparison_condition` might return the parsing of a logical condition... Shouldn't be the caller making this check?

integrations/pgvector/src/haystack_integrations/document_stores/pgvector/filters.py
```diff
@@ -0,0 +1,228 @@
...
+        field = _treat_meta_field(field, value)
+
+    field, value = COMPARISON_OPERATORS[operator](field, value)
+    return field, [value]
+
+
+def _treat_meta_field(field: str, value: Any) -> str:
+    """
+    Internal method that modifies the field str
+    to make the meta JSONB field queryable.
```
masci, line 102: Can you add an example of an input and resulting output for this function?

integrations/pgvector/src/haystack_integrations/document_stores/pgvector/filters.py
```diff
@@ -0,0 +1,228 @@
...
+
+    if type_value:
+        field = f"({field})::{type_value}"
+
+    return field
+
+
+def _equal(field: str, value: Any) -> tuple[str, Any]:
+    if value is None:
+        # no_value is a placeholder that will be removed in _convert_filters_to_where_clause_and_params
```
masci, line 123: If that's the case, let's declare it as a constant `NO_VALUE="no_value"` at the module level so it's easier to change if needed

masci (changes requested):
Took a first pass and left some minor comments. Can we add unit tests for the private methods in `filters.py`? Base tests mostly rely on happy paths, and I suspect this module might benefit from direct coverage, specially for the faulty corner cases.

integrations/pgvector/src/haystack_integrations/document_stores/pgvector/document_store.py
anakin87: I've been thinking about it a bit. What do you suggest to show useful information without risk of leaks?

integrations/pgvector/src/haystack_integrations/document_stores/pgvector/filters.py
anakin87: you are right...
(copy-pasted from another Document Store, but the caller already checks this)

integrations/pgvector/src/haystack_integrations/document_stores/pgvector/document_store.py
anakin87: I've been thinking about it a bit. What do you suggest to show useful information without risk of leaks?

integrations/pgvector/src/haystack_integrations/document_stores/pgvector/filters.py
anakin87: agree!

anakin87:
I have tried to address most of the comments and added some unit tests (some of these overlap with existing ones, but run very fast).

integrations/pgvector/src/haystack_integrations/document_stores/pgvector/document_store.py
masci: We could print the SQL query in the debug logs and do something like:
```
detailed_error_msg = f"{error_msg}.\nYou can find the SQL query and the parameters in the debug logs"
```
It's not perfect but it should be ok for most of the cases

integrations/pgvector/tests/test_filters.py
masci: Let's start by not supporting NOT, if there's demand we can proceed with option number 2, knowing that the investment in patching the base tests is worth it.

masci (changes requested):
Let's remove the SQL query from the error and we're good to go

masci (approved):
Nice!

PR #260, opened 2024-01-24
Support more collection settings when creating a new `WeaviateDocumentStore`

silvanocerza:
This PR adds support for more collection settings when creating a new `WeaviateDocumentStore`.

It also adds a `docker-compose.yml` to start a Weaviate instance for testing purposes.
The `document_store` fixture also is added but not used as of now.


integrations/weaviate/src/haystack_integrations/document_stores/weaviate/document_store.py
```diff
@@ -25,6 +25,20 @@
...
+# It's a list of properties that will be created on the collection.
+# These are extremely similar to the Document dataclass, but with a few differences:
+# - `id` is renamed to `_original_id` as the `id` field is reserved by Weaviate.
+# - `blob` is split into `blob_data` and `blob_mime_type` as it's more efficient to store them separately.
+DOCUMENT_COLLECTION_PROPERTIES = [
+    {"name": "_original_id", "dataType": ["text"]},
+    {"name": "content", "dataType": ["text"]},
+    {"name": "dataframe", "dataType": ["text"]},
+    {"name": "blob_data", "dataType": ["blob"]},
+    {"name": "blob_mime_type", "dataType": ["text"]},
```
anakin87, line 38: what if the blob has a [meta field](https://github.com/deepset-ai/haystack/blob/c47b82c54f313cf269ab0c6d4989e6857fdb082e/haystack/dataclasses/byte_stream.py#L13)?
do we need another field to map it?

integrations/weaviate/src/haystack_integrations/document_stores/weaviate/document_store.py
```diff
@@ -49,6 +63,16 @@ def __init__(
...
         :param url: The URL to the weaviate instance, defaults to None.
+        :param collection_settings: The collection settings to use, defaults to None.
+            If None it will use a collection named `default` with the following properties:
+            - _original_id: text
+            - content: text
+            - dataframe: text
+            - blob_data: blob
+            - blob_mime_type: text
+            - score: number
+            See the official `Weaviate documentation<https://weaviate.io/developers/weaviate/manage-data/collections>`_
```
anakin87, line 74: I'm just curious about this format for links...
Is it Sphinx? Is it supported in ReadMe?
Should it be `Link text <link URL>`?

integrations/weaviate/src/haystack_integrations/document_stores/weaviate/document_store.py
```diff
@@ -25,6 +25,20 @@
...
+# - `blob` is split into `blob_data` and `blob_mime_type` as it's more efficient to store them separately.
+DOCUMENT_COLLECTION_PROPERTIES = [
+    {"name": "_original_id", "dataType": ["text"]},
+    {"name": "content", "dataType": ["text"]},
+    {"name": "dataframe", "dataType": ["text"]},
+    {"name": "blob_data", "dataType": ["blob"]},
+    {"name": "blob_mime_type", "dataType": ["text"]},
+    {"name": "score", "dataType": ["number"]},
+]
+
```
anakin87, line 41: Is `meta` field missing? Is it intentional?

anakin87 (commented):
That seems like a reasonable change.

I left some comments.

integrations/weaviate/src/haystack_integrations/document_stores/weaviate/document_store.py
silvanocerza: Yep, it's intentional. Ideally I would specify it like this:
```
{"name": "meta", "dataType": ["object"]}
```
Problem is that that makes it mandatory for `meta` to contain some field. Trying to save a `Document` with `meta == {}` would fail.

If I omit it from the schema and let Weaviate automatically pick the type, that makes it work. I need to verify what happens with `auto-schema` off though.

integrations/weaviate/src/haystack_integrations/document_stores/weaviate/document_store.py
silvanocerza: I omitted it for the same reason I omitted `meta`, though I forgot to properly name the key when converting the Document. I'll handle that.

integrations/weaviate/src/haystack_integrations/document_stores/weaviate/document_store.py
silvanocerza: This should be handle by `pydoc` when converting the docstrings to Markdown. 

In any case we need to standardise the docstrings sooner than later.

integrations/weaviate/src/haystack_integrations/document_stores/weaviate/document_store.py
silvanocerza: Ah! Nothing to handle in this PR actually, I still have the code locally. 👀

anakin87 (approved):

PR #266, opened 2024-01-25
CI: sync API docs with readme.com

masci:
When changes to `integrations/*` are merged on `main`:
- build the API docs for the affected integrations
- publish the docs to https://docs.haystack.deepset.ai/v2.0/reference/integrations-chroma

Note for the reviewer:
- I don't particularly like https://github.com/deepset-ai/haystack-core-integrations/pull/266/files#diff-2c60cbcedf68c89cdc4eac5c906c44e16dd4e7766b616a18efe556267c8e0f77R49-R56 any suggestion about how to improve much appreciated
- If we merge this workflow, it would fail for integrations that don't support `hatch run docs` yet. Not sure if we should wait, let it fail, or add safeguards to skip the job when `hatch run docs` is not available

### DIFF
--- a/.github/workflows/CI_readme_sync.yml
+++ b/.github/workflows/CI_readme_sync.yml
@@ -47,10 +47,8 @@ jobs:
           README_API_KEY: ${{ secrets.README_API_KEY }}
           ALL_CHANGED_DIRS: ${{ steps.changed-files.outputs.all_changed_files }}
         run: |
-          for d in "$ALL_CHANGED_DIRS"; do
-            cd $d
-            hatch run docs
-            cd -
+          for d in $ALL_CHANGED_DIRS; do
+            (cd $d; hatch run docs)
           done
           mkdir tmp
           find . -name "_readme_*.md" -exec cp "{}" tmp \;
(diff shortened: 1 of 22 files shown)


.github/workflows/CI_readme_sync.yml
```diff
@@ -0,0 +1,64 @@
...
+          # from Readme.io as we need them to associate the slug
+          # in config files with their id.
+          README_API_KEY: ${{ secrets.README_API_KEY }}
+          ALL_CHANGED_DIRS: ${{ steps.changed-files.outputs.all_changed_files }}
+        run: |
+          for d in "$ALL_CHANGED_DIRS"; do
+            cd $d
+            hatch run docs
+            cd -
+          done
```
shadeMe, line 53:
```suggestion
          for d in "$ALL_CHANGED_DIRS"; do
          	(cd $d; hatch run docs)
          done
```
Not much of an improvement, but this might work too?

shadeMe (approved):
LGTM! The suggestion's optional. Feel free to merge (or not).

> If we merge this workflow, it would fail for integration that don't support hatch run docs yet. Not sure if we should wait, let it fail, or add safeguards to skip the job when hatch run docs is not available

Just a thought: A probable outcome of skipping the job will be our procrastinating adding support for hatch in those integrations 😄 If it's not a lot of work, we might want to prioritize it.

masci:
Do not merge until https://github.com/deepset-ai/haystack-core-integrations/issues/274 is complete

PR #267, opened 2024-01-25
Implement `count_document` for WeaviateDocumentStore

silvanocerza:
A test is skipped since we have a working `write_documents()` as of now.


silvanocerza:
I updated CI in this same PR as the change was minimal and this is the first PR that actually needs a running Weaviate instance.

I also had to ditch Windows and Mac tests in CI as Weaviate provides no Docker image for them. 🤷

anakin87 (approved):
:+1:

PR #272, opened 2024-01-25
feat: Implement `UpTrainEvaluator`

shadeMe:
Related to #248.

We introduce `UpTrainEvaluator`, a component that uses the [UpTrain LLM evaluation framework](https://docs.uptrain.ai/getting-started/introduction) to calculate evaluation metrics for RAG pipelines (among others). Refer https://github.com/deepset-ai/haystack/pull/6784 for an overview of the API design.

This PR introduces the following user-facing classes:
- `UpTrainMetric` - A enumeration that lists the supported UpTrain metrics. 
- `UpTrainEvaluator` - The pipeline component interfaces with the evaluation framework. It accepts a single metric and its optional parameters. It also provides extra optional parameters to configure the API client. The inputs to the pipeline are dynamically configured depending on the metric. This is done with help of a metric descriptor table that contains metadata concerning input/output conversion formats, expected inputs/outputs, etc.
- `UpTrainEvaluatorOutput` - Encapsulates one or more metric results for a single datapoint in the input. The evaluator outputs a list of these, one for each element in the input(s).


integrations/uptrain/src/uptrain_haystack/evaluator.py
```diff
@@ -0,0 +1,194 @@
...
+        expected_inputs = self.descriptor.input_parameters
+        component.set_input_types(self, **expected_inputs)
+
+    @component.output_types(output=List["UpTrainEvaluatorOutput"])
+    def run(self, **inputs) -> Dict[str, Any]:
+        """
+        Run the UpTrain evaluator.
+
+        :param inputs:
+            The inputs to evaluate. Match the input parameters of the metric.
```
julian-risch, line 74: Giving one example here would help users to get started.

integrations/uptrain/tests/test_evaluator.py
```diff
@@ -0,0 +1,372 @@
...
+    for r, o in zip(results, outputs):
+        assert len(r.result) == len(o)
+
+        expected = {(name if name is not None else str(metric), score, exp) for name, score, exp in o}
+        got = {(x.name, x.score, x.explanation) for x in r.result}
+        assert got == expected
+
+
+@pytest.mark.skipif("OPENAI_API_KEY" not in os.environ, reason="OPENAI_API_KEY not set")
+@pytest.mark.parametrize(
```
julian-risch, line 318: I see you are heavily using parameterization in the tests, which has some advantages and disadvantages. It helps to  avoid code duplication but makes it a bit harder to read and understand what's being tested here. Let's add a comment here.

integrations/uptrain/src/uptrain_haystack/evaluator.py
```diff
@@ -0,0 +1,194 @@
...
+
+    :param input:
+        The input that was evaluated.
+    :param result:
+        The result of the evaluation. Can contain
+        multiple results depending on the metric.
+    """
+
+    input: Dict[str, Any]
+    result: List[UpTrainMetricResult]
```
julian-risch, line 194: The PR description doesn't list UpTrainMetricResult as user facing but it is as part of the UpTrainEvaluatorOutput

integrations/uptrain/src/uptrain_haystack/evaluator.py
```diff
@@ -0,0 +1,194 @@
...
+            backend_client = EvalLLM(openai_api_key=api_key)
+        elif self.api == "uptrain":
+            backend_client = APIClient(uptrain_api_key=api_key)
+
+        self._backend_metric = backend_metric
+        self._backend_client = backend_client
+
+
+@dataclass(frozen=True)
+class UpTrainEvaluatorOutput:
```
julian-risch, line 182: This type of output is very specific and makes it hard to connect it to other components in a pipeline. I suggest we use four separate output edges instead: `input`, `name`, `score`, `explanation`. That way we get rid of UpTrainEvaluatorOutput and UpTrainMetricResult. We don't really need `input` in that case.

integrations/uptrain/src/uptrain_haystack/evaluator.py
```diff
@@ -0,0 +1,194 @@
...
+
+    @component.output_types(output=List["UpTrainEvaluatorOutput"])
+    def run(self, **inputs) -> Dict[str, Any]:
+        """
+        Run the UpTrain evaluator.
+
+        :param inputs:
+            The inputs to evaluate. Match the input parameters of the metric.
+        :returns:
+            A list of :class:`UpTrainEvaluatorOutput` objects, each containing
```
julian-risch, line 76: As discussed, the run method should better run on a single input sample and return a single output sample rather than lists of input and output samples. 
The list processing I would expect in a what we called `run_batch` method in the past. It's best to sync here with team members working on the pipeline improvements. @silvanocerza 
My last status was that each component should have a run method and optionally they can define a "run_batch" method that takes a list of each of the inputs of the run method.

julian-risch (changes requested):
This PR is in really great shape! I have commented what we also briefly talked about. 
Could you please also add an example of how the new component can be used in a pipeline? Here is an example of another integration's example: https://github.com/deepset-ai/haystack-core-integrations/blob/main/integrations/chroma/example/example.py

julian-risch (commented):
You can add an entry about UpTrain also to the inventory in the readme as part of this PR. https://github.com/deepset-ai/haystack-core-integrations/tree/main?tab=readme-ov-file#inventory

integrations/uptrain/src/uptrain_haystack/evaluator.py
shadeMe: Added. Also updated the docstring to point to the metric enum.

integrations/uptrain/tests/test_evaluator.py
shadeMe: Added.

integrations/uptrain/src/uptrain_haystack/evaluator.py
shadeMe: As discussed elsewhere, the output has now been modified to return a nested list of `Dict` objects, each of which has the above attributes. The `input` field has been removed.

integrations/uptrain/src/uptrain_haystack/evaluator.py
shadeMe: Since this is an ongoing discussion, we'll continue with the current implementation for the time being.

README.md
```diff
@@ -60,21 +60,22 @@ deepset-haystack
...
+| [unstructured-fileconverter-haystack](integrations/unstructured/)   | File converter      | [![PyPI - Version](https://img.shields.io/pypi/v/unstructured-fileconverter-haystack.svg)](https://pypi.org/project/unstructured-fileconverter-haystack) | [![Test / unstructured](https://github.com/deepset-ai/haystack-core-integrations/actions/workflows/unstructured.yml/badge.svg)](https://github.com/deepset-ai/haystack-core-integrations/actions/workflows/unstructured.yml)                         |
+| [uptrainr-haystack](integrations/uptrain/)                          | Evaluator           | [![PyPI - Version](https://img.shields.io/pypi/v/uptrain-haystack.svg)](https://pypi.org/project/uptrain-haystack)                                       | [![Test / uptrain](https://github.com/deepset-ai/haystack-core-integrations/actions/workflows/uptrain.yml/badge.svg)](https://github.com/deepset-ai/haystack-core-integrations/actions/workflows/uptrain.yml)                                        |
```
julian-risch, line 81: There is a typo in uptrainr-haystack

shadeMe: Fixed.

julian-risch (approved):
LGTM! 👍 The example is really helpful. The topic of list processing we can postpone and for the output format, let's see whether there are any use cases that would benefit from having separate edges instead of one dict. We talked about the pipeline visualization that unfortunately hides the contents of the dict from the user in its current implementation. Let's get this merged fast and collect feedback from users! Thanks for the fruitful discussions and great job! 🙂

julian-risch:
And let's see whether somebody from UpTrain can help with the integration test of the response matching metric that fails with a 500 Internal Server Error

PR #278, opened 2024-01-26
Implement filtering for `WeaviateDocumentStore`

silvanocerza:
This PR add support for `filters` in `WeaviateDocumentStore.filter_documents()` method.

All tests are passing except for `test_comparison_not_equal_with_dataframe`. 
For some reason that test is not returning what we expect from Weaviate. 
I decided to skip it for the time being as I wanted to get this out of the door.

### DIFF
--- a/integrations/weaviate/pyproject.toml
+++ b/integrations/weaviate/pyproject.toml
@@ -10,9 +10,7 @@ readme = "README.md"
 requires-python = ">=3.8"
 license = "Apache-2.0"
 keywords = []
-authors = [
-  { name = "deepset GmbH", email = "info@deepset.ai" },
-]
+authors = [{ name = "deepset GmbH", email = "info@deepset.ai" }]
 classifiers = [
   "Development Status :: 4 - Beta",
   "Programming Language :: Python",
@@ -28,6 +26,7 @@ dependencies = [
   "haystack-ai",
   "weaviate-client==3.*",
   "haystack-pydoc-tools",
+  "python-dateutil",
 ]
 
 [project.urls]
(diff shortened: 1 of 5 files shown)


integrations/weaviate/pyproject.toml
```diff
@@ -28,6 +26,7 @@ dependencies = [
   "haystack-ai",
   "weaviate-client==3.*",
   "haystack-pydoc-tools",
+  "python-dateutil",
```
silvanocerza, line 29: I had to add this to have proper parsing of ISO 8601 dates as [`datetime.fromisoformat()`](https://docs.python.org/3/library/datetime.html#datetime.date.fromisoformat) supports it properly only from `3.11`.

integrations/weaviate/src/haystack_integrations/document_stores/weaviate/_filters.py
```diff
@@ -0,0 +1,290 @@
...
+        # if it has a field set and not set at the same time.
+        # This will cause the filter to match no Document.
+        # This way we keep the behavior consistent with other Document Stores.
+        return {
+            "operator": "And",
+            "operands": [
+                {"path": field, "operator": "IsNull", "valueBoolean": False},
+                {"path": field, "operator": "IsNull", "valueBoolean": True},
+            ],
+        }
```
anakin87, line 137: If I understood correctly, this is a trick to create a filter that returns no Documents.
It is done to keep the global filtering logic unaffected. Right?

If so, can we centralize that logic in a utility function?

integrations/weaviate/src/haystack_integrations/document_stores/weaviate/_filters.py
```diff
@@ -0,0 +1,290 @@
...
+    "in": _in,
+    "not in": _not_in,
+}
+
+
+def _parse_comparison_condition(condition: Dict[str, Any]) -> Dict[str, Any]:
+    if "field" not in condition:
+        # 'field' key is only found in comparison dictionaries.
+        # We assume this is a logic dictionary since it's not present.
+        return _parse_logical_condition(condition)
```
anakin87, line 268: Some time ago, in another PR, Massi made me notice that this might be better handled in the caller. :slightly_smiling_face:

anakin87 (commented):
A great but challenging job!

I left two minor comments.

anakin87 (approved):
:+1:

Issue #283, opened 2024-01-29
Cohere Embedders are not functioning correctly


bilgeyucel:
**Describe the bug**
Cohere embedders don't work as expected. Somehow the retrieved documents are unexpected compared to the OpenAI embeddings. You can find the detailed information in this [discord thread](https://discord.com/channels/993534733298450452/1200452127072456836)

**To Reproduce**
See the code snippet and the example docs in this [discord thread](https://discord.com/channels/993534733298450452/1200452127072456836)
I could reproduce both with `InMemoryDocumentStore` and `ChromaDocumentStore`
**Updated version for cohere-haystack 0.3.0** 👇 
```python
from haystack.document_stores.types import DuplicatePolicy
from haystack.components.embedders.openai_document_embedder import OpenAIDocumentEmbedder
from haystack.components.embedders.openai_text_embedder import OpenAITextEmbedder
from haystack.document_stores.in_memory import InMemoryDocumentStore
from haystack.components.writers import DocumentWriter
from haystack.components.retrievers.in_memory import InMemoryEmbeddingRetriever
from haystack import Pipeline
from haystack_integrations.components.embedders.cohere import CohereDocumentEmbedder, CohereTextEmbedder

import pickle

with open('docs.pkl', 'rb') as f:
    docs = pickle.load(f)
print(len(docs))

embedding = "Cohere"

if embedding == "OpenAI":
    doc_embedder = OpenAIDocumentEmbedder(model="text-embedding-3-small")
    query_embedder = OpenAITextEmbedder(model="text-embedding-3-small")
elif embedding == "Cohere":
    doc_embedder = CohereDocumentEmbedder(model="embed-multilingual-v3.0", input_type="search_document")
    query_embedder = CohereTextEmbedder(model="embed-multilingual-v3.0", input_type="search_query")

# Ingestion
doc_store = InMemoryDocumentStore(embedding_similarity_function="cosine")  # tried with the default `dot_product` and the scores were the same 
doc_writer = DocumentWriter(document_store = doc_store, policy = DuplicatePolicy.SKIP)

ingestion_pipe = Pipeline()
ingestion_pipe.add_component("doc_embedder", doc_embedder)
ingestion_pipe.add_component("doc_writer", doc_writer)

ingestion_pipe.connect("doc_embedder.documents", "doc_writer.documents")
ingestion_pipe.run({"doc_embedder": {"documents": docs}})

# Retrieval
[...]

PR #284, opened 2024-01-29
fix: Cohere inconsistent embeddings and documents lengths

sjrl:
I am looking into this issue https://github.com/deepset-ai/haystack-core-integrations/issues/283 and found a bug that I think is unrelated. 

**Update:**
Running the code in issue https://github.com/deepset-ai/haystack-core-integrations/issues/283 shows that the changes in this branch do seem to help. See comment [here](https://github.com/deepset-ai/haystack-core-integrations/issues/283#issuecomment-1914775276).

Basically we were copying the embeddings so we'd end up with twice as many embeddings as documents. I don't think this will fix the error brought up in issue https://github.com/deepset-ai/haystack-core-integrations/issues/283, but we should avoid doing this. 

Additional:
* also fixed some failing integration tests


sjrl:
Hey @ZanSara I'm noticing that the integration tests that require a Cohere API key are not running in CI. Is this on purpose or should we be able to run them?

ZanSara:
Hey @sjrl, thank you for this PR! Let's split the two things though:
1. the bugfix in one PR, so I can approve it right away, and
2. issue + PR for the Cohere tests, we need to address that separately and other tests might need fixing
Thank you for your help!

ZanSara (approved):
Testing out the example code the improvement is evident! Thank you for fixing this one! :hugs:

illorca-verbi:
Thanks for this fix. When will the release be cut for https://pypi.org/project/cohere-haystack/?

anakin87:
released: https://pypi.org/project/cohere-haystack/0.3.1/

PR #298, opened 2024-01-30
Pgvector - embedding retrieval

anakin87:
Implementation of the embedding retrieval for Pgvector Document Store.

The Embedding Retriever will be implemented in a future PR.

### DIFF
--- a/integrations/pgvector/src/haystack_integrations/document_stores/pgvector/document_store.py
+++ b/integrations/pgvector/src/haystack_integrations/document_stores/pgvector/document_store.py
@@ -52,8 +52,10 @@ blob_mime_type = EXCLUDED.blob_mime_type,
 meta = EXCLUDED.meta
 """
 
+VALID_VECTOR_FUNCTIONS = ["cosine_similarity", "inner_product", "l2_distance"]
+
 VECTOR_FUNCTION_TO_POSTGRESQL_OPS = {
-    "cosine_distance": "vector_cosine_ops",
+    "cosine_similarity": "vector_cosine_ops",
     "inner_product": "vector_ip_ops",
     "l2_distance": "vector_l2_ops",
 }
@@ -70,7 +72,7 @@ class PgvectorDocumentStore:
         connection_string: str,
         table_name: str = "haystack_documents",
         embedding_dimension: int = 768,
-        vector_function: Literal["cosine_distance", "inner_product", "l2_distance"] = "cosine_distance",
+        vector_function: Literal["cosine_similarity", "inner_product", "l2_distance"] = "cosine_similarity",
         recreate_table: bool = False,
         search_strategy: Literal["exact_nearest_neighbor", "hnsw"] = "exact_nearest_neighbor",
         hnsw_recreate_index_if_exists: bool = False,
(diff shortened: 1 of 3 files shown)


integrations/pgvector/src/haystack_integrations/document_stores/pgvector/document_store.py
```diff
@@ -87,12 +89,16 @@ def __init__(
         :param table_name: The name of the table to use to store Haystack documents. Defaults to "haystack_documents".
         :param embedding_dimension: The dimension of the embedding. Defaults to 768.
         :param vector_function: The similarity function to use when searching for similar embeddings.
-            Defaults to "cosine_distance". Set it to one of the following values:
-        :type vector_function: Literal["cosine_distance", "inner_product", "l2_distance"]
+            Defaults to "cosine_similarity". "cosine_similarity" and "inner_product" are similarity functions,
+            so the most similar documents are the ones with the lowest score.
+            "l2_distance" is a distance function, so the most similar documents are the ones with the smallest score.
```
masci, line 94:
```suggestion
            Defaults to "cosine_similarity". "cosine_similarity" and "inner_product" are similarity functions and
            higher scores indicate greater similarity between the documents.
            "l2_distance" returns the straight-line distance between vectors, and the most similar documents are the ones with the smallest score.
```

integrations/pgvector/src/haystack_integrations/document_stores/pgvector/document_store.py
```diff
@@ -87,12 +89,16 @@ def __init__(
...
+            When using the "hnsw" search strategy, the vector_function value is used to build an appropriate index.
```
masci, line 95:
```suggestion
            
            Important: when using the "hnsw" search strategy, an index will be created that depends on the `vector_function` passed here. Make sure subsequent queries will keep using the same vector similarity function in order to take advantage of the index.
```

integrations/pgvector/src/haystack_integrations/document_stores/pgvector/document_store.py
```diff
@@ -87,12 +89,16 @@ def __init__(
...
+        :type vector_function: Literal["cosine_similarity", "inner_product", "l2_distance"]
         :param recreate_table: Whether to recreate the table if it already exists. Defaults to False.
         :param search_strategy: The search strategy to use when searching for similar embeddings.
             Defaults to "exact_nearest_neighbor". "hnsw" is an approximate nearest neighbor search strategy,
             which trades off some accuracy for speed; it is recommended for large numbers of documents.
+            When using the "hnsw" search strategy, the vector_function value is used to build an appropriate index.
```
masci, line 101:
```suggestion
            Important: when using the "hnsw" search strategy, an index will be created that depends on the `vector_function` passed here. Make sure subsequent queries will keep using the same vector similarity function in             order to take advantage of the index.
```

integrations/pgvector/src/haystack_integrations/document_stores/pgvector/document_store.py
```diff
@@ -423,3 +432,98 @@ def delete_documents(self, document_ids: List[str]) -> None:
...
+        :param top_k: Maximum number of Documents to return, defaults to 10
+        :param vector_function: The similarity function to use when searching for similar embeddings.
+            Defaults to the PgvectorDocumentStore's vector_function.
+            Since vector_function is used to build the HNSW index (when using the "hnsw" search strategy),
+            if a vector_function other than the one used to build the index is chosen,
+            the index will not be used and the search will be slower.
+            "cosine_similarity" and "inner_product" are similarity functions,
+            so the most similar documents are the ones with the lowest score.
+            "l2_distance" is a distance function, so the most similar documents are the ones with the smallest score.
+        :type vector_function: Literal["cosine_similarity", "inner_product", "l2_distance"]
```
masci, line 466: Since this is a private method, let's not re-document the parameters, this way it'll be more maintainable
```suggestion

        This method is not meant to be part of the public interface of
        `PgvectorDocumentStore` and it should not be called directly.
        `PgvectorEmbeddingRetriever` uses this method directly and is the public interface for it.
```

masci (changes requested):

PR #301, opened 2024-01-31
Pinecone - change dummy vector

### LINKED ISSUE #300: Pinecone: Dense vectors must contain at least one non-zero value.
**Describe the bug**
This limit was introduced very recently: see [Pinecone limits](https://docs.pinecone.io/docs/limits)

This affects our tests: https://github.com/deepset-ai/haystack-core-integrations/actions/runs/7719142884/job/21052234350
but also **affects the core functioning of the Document Store** (we are relying on a dummy vector, which contains only zeros)

anakin87:
Fixes #300.

I only changed the dummy vector from `[0.0] * self.dimension` to `[-10.0] * self.dimension`.

It is important to choose a highly improbable value for the user-supplied vectors, because this value is removed during retrieval.


vblagoje (approved):
LGTM

PR #323, opened 2024-02-01
Pinecone - decrease concurrency in tests

anakin87:
[Pinecone tests keep failing](https://github.com/deepset-ai/haystack-core-integrations/actions/runs/7734112772/job/21087510975), despite my recent attempt to fix them in #288.

The error is "Too Many Requests".
So I am doing two things:
- decrease the concurrency in tests (lowering the number of workers from 3 to 2)
- decrease the sleep time a bit, since it does not seem to be responsible for this error


ZanSara (approved):
As discussed, let's hope this is enough to at least reduce the flakyness of the test.

Issue #302, opened 2024-01-31
amazon_bedrock: generate api docs


dfokina:
Follow the pattern of PR deepset-ai/haystack#262
The `order` for this pydoc will be `10`

PR #326, opened 2024-02-01
amazon bedrock: generate api docs

dfokina:
Fixes https://github.com/deepset-ai/haystack-core-integrations/issues/302

### DIFF
--- a/integrations/amazon_bedrock/pydoc/config.yml
+++ b/integrations/amazon_bedrock/pydoc/config.yml
@@ -3,6 +3,9 @@ loaders:
     search_path: [../src]
     modules: [
       "haystack_integrations.components.generators.amazon_bedrock.generator",
+      "haystack_integrations.components.generators.amazon_bedrock.adapters",
+      "haystack_integrations.components.generators.amazon_bedrock.errors",
+      "haystack_integrations.components.generators.amazon_bedrock.handlers",
     ]
     ignore_when_discovered: ["__init__"]
 processors:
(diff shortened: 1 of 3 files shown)


integrations/amazon_bedrock/pydoc/config.yml
```diff
@@ -0,0 +1,28 @@
+loaders:
+  - type: haystack_pydoc_tools.loaders.CustomPythonLoader
+    search_path: [../src]
+    modules: [
+      "haystack_integrations.components.generators.amazon_bedrock.generator",
```
ZanSara, line 5: I think we should document all files in the repo:

```suggestion
      "haystack_integrations.components.generators.amazon_bedrock.generator",
      "haystack_integrations.components.generators.amazon_bedrock.adapter",
      "haystack_integrations.components.generators.amazon_bedrock.errors",
      "haystack_integrations.components.generators.amazon_bedrock.handler",
```

ZanSara (changes requested):
I think some modules are missing from the config

integrations/amazon_bedrock/pydoc/config.yml
dfokina: @ZanSara added! (I added the import paths with the file names, e.g. adapter**s** and handler**s** (multiple), is that right 🤔 ?)

ZanSara: Oh that's right!

ZanSara (approved):
LGTM

PR #221, opened 2024-01-16
Mount unstructured in haystack_integrations and refactor the structure of the project

anakin87:
Part of #205

I also took the opportunity to improve the structure of the project,
which had been created when the core-integrations were less standardized.

In particular, when trying to fix #218, I couldn't properly release a new version.
This problem should be fixed by this PR.

PR #329, opened 2024-02-02
Labeler: add new entry for pgvector

anakin87:
- This PR updates labeler.yml to also take pgvector into account.
- I also changed the paths for the unstructured integration (refactored in #221).


masci (approved):
Thanks!

PR #336, opened 2024-02-04
unstructured: fix metadata order mixed up

### LINKED ISSUE #331: unstructured: metadata get mixed up
**Describe the bug**
When indexing files with Unstructured component. Metadata get mixed (?)
For example with these data:
```python
    files = ["test/samples/sample1.pdf", "test/samples/sample2.pdf", "test/samples/sample3.pdf" ]
    meta = [
        {"meta1": "value1", "source_test": "pytest_api"},
        {"meta2": "value2", "source_test": "pytest_api"},
        {"meta3": "value3", "source_test": "pytest_api"},
    ]
```
Then these data are converted to json to make a call to my API
I get these results 
:
```json
    {
        "id": "b89ce3cc7ed9839459d1606018cf6beb720df0424515cd4cc9442b51970f72b8",
        "content": "blablablalba",
        "meta": {
            "meta2": "value2",
            "source_test": "pytest_api",
            "filename": "5f27_sample1.pdf",
            "s3_key": "bc6c_sample2.pdf",
            "file_path": "C:\\Users\\cmeyer\\code-project\\llm-ale-chatbot\\haystack_api\\rest_api\\file-upload\\5f27_sample1.pdf",
            "languages": [
                "fra"
            ],
            "page_number": 1,
            "filetype": "application/pdf",
            "category": "UncategorizedText"
        },
        "score": 0.0
    },
```  
Here the content correspond to the right file_path & filename (generated by unstructured) BUT my CUSTOM metadata that are not generated by Unstructured processing are mixed up (meta2, and s3_key are wrong) !  
Sorry for the not very reproducible example. I'm just writing to know if someone already had similar issue. I make a detailed report this week-end. This doesn't happen with PyPDF, so it's weird.

 - OS: Windows running a LINUX Docker
 - Haystack version: 2.0.0 beta5
 - Integration version: 0.3.0

lambda-science:
Fix #331 

Because unstructured integration code use python `sets`, file orders are not preserved leading to metadata attribution to wrong filepaths. Using lists fix this as list are ordered.

Also C/C from the issue, I added:

> And actually I'm not sure why we need a set logic here to make filepath unique. I feel like it's up to the user to provide unique paths ? For example what happens if a user provide 10 path and 10 metadata but then some filepath are duplicated so then we have 8 filepath and 10 metadata ? It will raise error from normalize_metadata I guess
> 
> What I think is that:
> 
> We can support directories as path BUT then metadata should max be of a length of 1 (same metadata for all files in directory). Because I'm not sure it's clear how path.glob() orders files (leading to metadata attribution confusion)
> If direct paths to files are provided: don't make them unique with sets.


lambda-science:
I'm sorry for the long commit list it's weird. It took all my previous history also from https://github.com/deepset-ai/haystack-core-integrations/pull/242
Maybe I should have created a new branch :/

integrations/unstructured/src/haystack_integrations/components/converters/unstructured/converter.py
```diff
@@ -109,15 +109,18 @@ def run(
...
-
+        paths_obj = [Path(path) for path in paths]
+        filepaths = [path for path in paths_obj if path.is_file()]
+        filepaths_in_directories = [
+            filepath for path in paths_obj if path.is_dir() for filepath in path.glob("*.*") if filepath.is_file()
+        ]
+        if filepaths_in_directories != [] and isinstance(meta, list):
+            error = """For directories as path metadata can only be a unique dictionary, not a list.
+                To provide different metadata (list) for each files,
+                please provide an explicit list of direct paths instead."""
```
anakin87, line 120:
```suggestion
        if filepaths_in_directories and isinstance(meta, list):
            error = """"If providing directories in the `paths` parameter,
             `meta` can only be a dictionary (metadata applied to every file),
             and not a list. To specify different metadata for each file,
             provide an explicit list of direct paths instead."""
```

anakin87 (commented):
Hey... Thanks for the fix!

I generally agree with your approach.

Let's also update the docstrings about `meta` to reflect the changes...

lambda-science:
Sounds good to me ! :)

anakin87 (approved):
:rocket:

PR #343, opened 2024-02-05
refactor!: `UptrainEvaluator` contexts are now passed as nested lists.

shadeMe:
The evaluator will automatically flatten them. This change allows other evaluators that take nested lists for contexts (like DeepEval) share the same, basic interface.

### DIFF
--- a/integrations/uptrain/example/example.py
+++ b/integrations/uptrain/example/example.py
@@ -8,8 +8,12 @@ QUESTIONS = [
     "Who created the Python language?",
 ]
 CONTEXTS = [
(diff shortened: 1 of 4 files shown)


julian-risch (changes requested):
We need to update also the example code to use nested lists. Similar to that update @TuanaCelik and I can then update the tutorial and the integration tile https://github.com/deepset-ai/haystack-integrations/pull/149 Otherwise looks good to me. 👍

julian-risch (approved):
LGTM! 👍

Issue #250, opened 2024-01-19
Add DeepEval evaluation framework integration


julian-risch:
## Summary and motivation

Integrate the DeepEval evaluation framework by implementing a new DeepEvalEvaluator component

## Detailed design

Based on the proposal https://github.com/deepset-ai/haystack/pull/6784 we need to implement an DeepEvalEvaluator component that is initialized with one metric and an OpenAI API key. Depending on the particular metric, the component expects for example question, answer and context as input and outputs a dictionary with the calculated score.

## Checklist

If the request is accepted, ensure the following checklist is complete before closing this issue.

- [x] The code is documented with docstrings and was merged in the `main` branch
- [x] Docs are published at https://docs.haystack.deepset.ai/
- [x] There is a Github workflow running the tests for the integration nightly and at every PR
- [x] A label named like `integration:<your integration name>` has been added to this repo
- [x] The [labeler.yml](https://github.com/deepset-ai/haystack-core-integrations/blob/main/.github/labeler.yml) file has been updated
- [x] The package has been released on PyPI
- [x] An integration tile has been added to https://github.com/deepset-ai/haystack-integrations
- [x] The integration has been listed in the [Inventory section](https://github.com/deepset-ai/haystack-core-integrations#inventory) of this repo README
- [x] There is an example available to demonstrate the feature
- [x] The feature was announced through social media

sahusiddharth:
@julian-risch can I work on this?

julian-risch:
With the social media announcement, this issue is closed now: https://twitter.com/Haystack_AI/status/1760726816526925839

PR #346, opened 2024-02-06
feat: Implement `DeepEvalEvaluator`

shadeMe:
Related to #250.

We introduce `DeepEvalEvaluator`, a component that uses the [DeepEval LLM evaluation framework](https://docs.confident-ai.com/docs/getting-started) to calculate evaluation metrics for RAG pipelines (among others). Refer https://github.com/deepset-ai/haystack/pull/6784 for an overview of the API design.

This PR introduces the following user-facing classes:
- `DeepEvalMetric` - A enumeration that lists the supported DeepEval metrics. Currently, only those metrics that are related to RAG pipelines are supported.
- `DeepEvalEvaluator` - Th pipeline component interfaces with the evaluation framework. It accepts a single metric and its optional parameters. The inputs to the pipeline are dynamically configured depending on the metric. This is done with help of a metric descriptor table that contains metadata concerning input/output conversion formats, expected inputs/outputs, etc.

The output of the component is a nested list of metric results. Each input can have one or more results, depending on the metric. Each result is a dictionary containing the following keys and values:

- `name` - The name of the metric.
- `score` - The score of the metric.
- `explanation` - An optional explanation of the score.


integrations/deepeval/pyproject.toml
```diff
@@ -0,0 +1,156 @@
...
+git_describe_command = 'git describe --tags --match="integrations/deepeval-v[0-9]*"'
+
+[tool.hatch.envs.default]
+dependencies = ["coverage[toml]>=6.5", "pytest"]
+[tool.hatch.envs.default.scripts]
+test = "pytest {args:tests}"
+test-cov = "coverage run -m pytest {args:tests}"
+cov-report = ["- coverage combine", "coverage report"]
+cov = ["test-cov", "cov-report"]
+
```
masci, line 49: The `readme_sync` CI job will fail when we merge this on `main`. Should we add at least a `hatch run docs` no-op command in this PR?

masci (commented):
I'm not familiar with deepeval but code looks good to me. I left a question about api docs now that we merged the CI job that will attempt to run `hatch run docs` on integrations' changes merged on `main`. Another option might be making that CI job more resilient.

integrations/deepeval/pyproject.toml
shadeMe: Good point, I'll add that.

masci (approved):
Thanks for adding the docs, LGTM

PR #347, opened 2024-02-06
refactor!: Use `Secret` for API key in `UpTrainEvaluator`

shadeMe:
And some other misc build fixes.


integrations/uptrain/src/haystack_integrations/components/evaluators/uptrain_evaluator.py
```diff
@@ -185,15 +186,10 @@ def _init_backend(self):
...
             raise ValueError(msg)
 
-        api_key = os.environ.get(self.api_key_env_var)
-        if api_key is None:
-            msg = f"Missing API key environment variable '{self.api_key_env_var}' for UpTrain evaluator"
-            raise ValueError(msg)
-
         if self.api == "openai":
-            backend_client = EvalLLM(openai_api_key=api_key)
+            backend_client = EvalLLM(openai_api_key=self.api_key.resolve_value())  # type: ignore
```
anakin87, line 190: Why do we need `# type: ignore`?
If it is unavoidable, can we add a comment explaining the reason?

anakin87 (commented):
Generally, it looks good.
I only left a comment.

integrations/uptrain/src/haystack_integrations/components/evaluators/uptrain_evaluator.py
shadeMe: Actually, that should be avoidable. Fixed.

anakin87 (approved):
:+1:

PR #348, opened 2024-02-07
fix: update to latest haystack-ai version

masci:
Fixing failing tests after the release of Haystack 2-beta6


anakin87 (approved):
Looks good!

PR #379, opened 2024-02-08
feat: Add Optimum Embedders

### LINKED ISSUE #137: Add OptimumEmbedder
Huggin Face's Optimum library provides faster inference through ONNX and TensorRT. This can be used to create blazing fast embedding components. The concepts used in Optimum also play well with some of the concepts that we have in Haystack. For example:

Loading non-ONNX checkpoints requires a conversion step, this takes some time. We can do that step in our warmup function (https://huggingface.co/docs/optimum/onnxruntime/usage_guides/models#loading-a-vanilla-transformers-model).

- fast embedders based on Optimum
- apply other tricks such as sorting by sequence length and dynamic padding to bring down inference time (see https://github.com/UKPLab/sentence-transformers/blob/c006921e9e9977bc107b05676266b581091688a2/sentence_transformers/SentenceTransformer.py#L179)

- Not having a fast embedder component
- https://github.com/qdrant/fastembed > supported models are limited

**Additional context**
https://colab.research.google.com/drive/10UAtpz26Gv2LtamT8j33LmI5UFQFwF4T?usp=sharing
https://github.com/huggingface/optimum-benchmark/tree/main/examples/fast-mteb

```[tasklist]
### Tasks
- [x] The code is documented with docstrings and was merged in the `main` branch
- [ ] https://github.com/deepset-ai/haystack-core-integrations/issues/742
- [x] There is a Github workflow running the tests for the integration nightly and at every PR
- [x] A label named like `integration:<your integration name>` has been added to this repo
- [x] The [labeler.yml](https://github.com/deepset-ai/haystack-core-integrations/blob/main/.github/labeler.yml) file has been updated
- [x] The package has been released on PyPI
- [x] An integration tile has been added to https://github.com/deepset-ai/haystack-integrations
- [x] The integration has been listed in the [Inventory section](https://github.com/deepset-ai/haystack-core-integrations#inventory) of this repo README
- [x] There is an example available to demonstrate the feature
- [x] The feature was announced through social media
```

awinml:
### Related Issues

- fixes #137 

### Proposed Changes:

Add support for inferencing embedding models using the [Hugging Face Optimum Library](https://huggingface.co/docs/optimum/index). These components are designed to seamlessly inference models using the high speed ONNX runtime.

Introduces two components:
- `OptimumTextEmbedder`, a component for embedding strings.
- `OptimumDocumentEmbedder`, a component for computing Document embeddings.

Additional optimizations implemented to bring down inference time:
- Sorting by sequence length: The text sequences are sorted in descending order based on their length, before creating embeddings. Please see the [Sentence Transformers Implementation](https://github.com/UKPLab/sentence-transformers/blob/454cb376cb54b7e17538dc85983e4c1e36fa4a99/sentence_transformers/SentenceTransformer.py#L260C23-L260C23) for reference.
- Dynamic Padding: The text sequences are padded to the longest sequence in the batch. This is achieved by setting `padding=True` for the  `AutoTokenizer`. Please see the [Transformers Documentation](https://huggingface.co/docs/transformers/en/pad_truncation) for reference.

For the [TensorRT ONNX runtime](https://huggingface.co/docs/optimum/onnxruntime/usage_guides/gpu#tensorrt-engine-build-and-warmup), we cache the TensorRT engine since it takes time to build. For the other runtimes, the conversion to ONNX is fairly quick so we don't cache the engine by default. If the user wishes to cache the other runtimes, the necessary parameters can be passed using the `model_kwargs` parameter.

The implementation for Mean Pooling is based on the implementation provided in the [Sentence Transformers documentation](https://www.sbert.net/examples/applications/computing-embeddings/README.html#sentence-embeddings-with-transformers).

### How did you test it?

Tests were added in `optimum_document_embedder.py` and `optimum_text_embedder.py`.

### Notes for the reviewer

- The versions of `optimum` and `transformers` has been pinned to `optimum==1.15.0` and `transformers==4.36.2`, due to bugs in the latest optimum release. Please refer to the following issues for more information: https://github.com/huggingface/optimum/issues/1673 and https://github.com/huggingface/optimum/issues/1675.

 - For using `optimum` with GPU based runtimes, the `optimum[onnxruntime-gpu]` package is required. To install this package `optimum` has be fully uninstalled first. To overcome this limitation, two optional dependencies have been set: 
      For the CPU version: `pip install optimum-haystack[cpu]`, which installs `optimum[onnxruntime]`.
      For the GPU version: `pip install optimum-haystack[gpu]`, which installs `optimum[onnxruntime-gpu]`.
      
      This approach is not very user friendly, since the component is not usable when using just `pip install optimum-haystack`. 
      I am not very happy with this approach, please suggest a better approach.

### Usage Examples:

These examples demonstrate how the embedders can be used to inference the `BAAI/bge-small-en-v1.5` embedding model using different ONNX runtimes.

On CPU:

```python
from haystack_integrations.components.embedders import OptimumTextEmbedder

text_to_embed = "I love pizza!"

text_embedder = OptimumTextEmbedder(model="BAAI/bge-small-en-v1.5")
text_embedder.warm_up()

print(text_embedder.run(text_to_embed))

# {'embedding': [0.017020374536514282, -0.023255806416273117, ...],
```

On GPU using the `CUDAExecutionProvider`:

```python
from haystack_integrations.components.embedders import OptimumTextEmbedder

text_to_embed = "I love pizza!"

text_embedder = OptimumTextEmbedder(
    model="BAAI/bge-small-en-v1.5", 
    onnx_execution_provider="CUDAExecutionProvider"
)
text_embedder.warm_up()

print(text_embedder.run(text_to_embed))

# [0.017020374536514282, -0.023255806416273117, ...]
```

### DIFF
--- a/integrations/optimum/pyproject.toml
+++ b/integrations/optimum/pyproject.toml
@@ -1,0 +1,201 @@
+[build-system]
+requires = ["hatchling", "hatch-vcs"]
+build-backend = "hatchling.build"
+
+[project]
+name = "optimum-haystack"
+dynamic = ["version"]
+description = "Component to embed strings and Documents using models loaded with the HuggingFace Optimum library. This component is designed to seamlessly inference models using the high speed ONNX runtime."
+readme = "README.md"
+requires-python = ">=3.8"
+license = "Apache-2.0"
+keywords = []
+authors = [
+  { name = "deepset GmbH", email = "info@deepset.ai" },
+  { name = "Ashwin Mathur", email = "" },
+]
+classifiers = [
+  "Development Status :: 4 - Beta",
+  "Programming Language :: Python",
+  "Programming Language :: Python :: 3.8",
+  "Programming Language :: Python :: 3.9",
+  "Programming Language :: Python :: 3.10",
+  "Programming Language :: Python :: 3.11",
+  "Programming Language :: Python :: Implementation :: CPython",
+  "Programming Language :: Python :: Implementation :: PyPy",
+]
+dependencies = [
+  "haystack-ai",
+  "transformers[sentencepiece]==4.36.2",
+  "sentence-transformers>=2.2.0",
+]
+
+[project.optional-dependencies]
+cpu = [
+  "optimum[onnxruntime]==1.15.0",
+]
+gpu = [
+  "optimum[onnxruntime-gpu]==1.15.0",
+]
+
+[project.urls]
+Documentation = "https://github.com/deepset-ai/haystack-core-integrations/tree/main/integrations/optimum#readme"
+Issues = "https://github.com/deepset-ai/haystack-core-integrations/issues"
+Source = "https://github.com/deepset-ai/haystack-core-integrations/tree/main/integrations/optimum"
+
+[tool.hatch.build.targets.wheel]
+packages = ["src/haystack_integrations"]
+
+[tool.hatch.version]
+source = "vcs"
+tag-pattern = 'integrations\/optimum-v(?P<version>.*)'
+
+[tool.hatch.version.raw-options]
+root = "../.."
+git_describe_command = 'git describe --tags --match="integrations/optimum-v[0-9]*"'
+
+[tool.hatch.envs.default]
+dependencies = [
+  "coverage[toml]>=6.5",
+  "pytest",
+]
+[tool.hatch.envs.default.scripts]
+test = "pytest {args:tests}"
+test-cov = "coverage run -m pytest {args:tests}"
+cov-report = [
+    "- coverage combine",
+    "coverage report",
+]
+cov = [
+    "test-cov",
+    "cov-report",
+]
+
+[[tool.hatch.envs.all.matrix]]
+python = ["3.8", "3.9", "3.10", "3.11", "3.12"]
+
+
+[tool.hatch.envs.lint]
+detached = true
+dependencies = [
+    "black>=23.1.0",
+    "mypy>=1.0.0",
+    "ruff>=0.0.243",
+]
+
+[tool.hatch.envs.lint.scripts]
+typing = "mypy --install-types --non-interactive --explicit-package-bases {args:src/ tests}"
+style = [
+    "ruff {args:.}",
+    "black --check --diff {args:.}",
+]
+fmt = [
+    "black {args:.}",
+    "ruff --fix {args:.}",
+    "style",
+]
+all = [
+    "style",
+    "typing",
+]
+
+[tool.hatch.metadata]
+allow-direct-references = true
+
+[tool.ruff.lint.isort]
+known-first-party = ["src"]
+
+[tool.black]
+target-version = ["py37"]
+line-length = 120
+skip-string-normalization = true
+
+[tool.ruff]
+target-version = "py37"
+line-length = 120
+select = [
+    "A",
+    "ARG",
+    "B",
+    "C",
+    "DTZ",
+    "E",
+    "EM",
+    "F",
+    "I",
+    "ICN",
+    "ISC",
+    "N",
+    "PLC",
+    "PLE",
+    "PLR",
+    "PLW",
+    "Q",
+    "RUF",
+    "S",
+    "T",
+    "TID",
+    "UP",
+    "W",
+    "YTT",
+]
+ignore = [
+    # Allow non-abstract empty methods in abstract base classes
+    "B027",
+    # Ignore checks for possible passwords
+    "S105", "S106", "S107",
+    # Ignore complexity
+    "C901", "PLR0911", "PLR0912", "PLR0913", "PLR0915",
+]
+unfixable = [
+    # Don't touch unused imports
+    "F401",
+]
+
+[tool.ruff.lint.flake8-tidy-imports]
+ban-relative-imports = "parents"
+
+[tool.ruff.lint.per-file-ignores]
+# Tests can use magic values, assertions, and relative imports
+"tests/**/*" = ["PLR2004", "S101", "TID252"]
+# Examples can print their output
+"examples/**" = ["T201"]
+"tests/**" = ["T201"]
+
+[tool.coverage.run]
+source_pkgs = ["optimum", "tests"]
+branch = true
+parallel = true
+
+[tool.coverage.paths]
+optimum = ["src/haystack_integrations", "*/optimum/src/haystack_integrations"]
+tests = ["tests", "*/optimum/tests"]
+
+[tool.coverage.report]
+exclude_lines = [
+    "no cov",
+    "if __name__ == .__main__.:",
+    "if TYPE_CHECKING:",
+]
+
+[[tool.mypy.overrides]]
+module = [
+  "haystack.*",
+  "haystack_integrations.*",
+  "pytest.*",
+  "numpy.*",
+  "optimum.*",
+  "torch.*",
+  "transformers.*",
+  "huggingface_hub.*"
+]
+ignore_missing_imports = true
+
+[tool.pytest.ini_options]
+addopts = ["--strict-markers", "-vv"]
+markers = [
+  "integration: integration tests",
+  "unit: unit tests",
+  "embedders: embedders tests",
+]
+log_cli = true
--- a/integrations/optimum/src/haystack_integrations/components/embedders/optimum_document_embedder.py
+++ b/integrations/optimum/src/haystack_integrations/components/embedders/optimum_document_embedder.py
@@ -1,0 +1,269 @@
+from typing import Any, Dict, List, Optional
+
+import numpy as np
+import torch
+from haystack import Document, component, default_from_dict, default_to_dict
+from haystack.utils import Secret, deserialize_secrets_inplace
+from haystack.utils.hf import HFModelType, check_valid_model, deserialize_hf_model_kwargs, serialize_hf_model_kwargs
+from optimum.onnxruntime import ORTModelForFeatureExtraction
+from tqdm import tqdm
+from transformers import AutoTokenizer
+
+
+class OptimumDocumentEmbedder:
+    """
+    A component for computing Document embeddings using models loaded with the HuggingFace Optimum library.
+    This component is designed to seamlessly inference models using the high speed ONNX runtime.
+
+    The embedding of each Document is stored in the `embedding` field of the Document.
+
+    Usage example:
+    ```python
+    from haystack.dataclasses import Document
+    from haystack_integrations.components.embedders import OptimumDocumentEmbedder
+
+    doc = Document(content="I love pizza!")
+
+    document_embedder = OptimumDocumentEmbedder(model="BAAI/bge-small-en-v1.5")
+    document_embedder.warm_up()
+
+    result = document_embedder.run([doc])
+    print(result["documents"][0].embedding)
+
+    # [0.017020374536514282, -0.023255806416273117, ...]
+    ```
+
+    Key Features and Compatibility:
+        - **Primary Compatibility**: Designed to work seamlessly with any embedding model present on the Hugging Face
+        Hub.
+        - **Conversion to ONNX**: The models are converted to ONNX using the HuggingFace Optimum library. This is
+        performed in real-time, during the warm-up step.
+        - **Accelerated Inference on GPU**: Supports using different execution providers such as CUDA and TensorRT, to
+        accelerate ONNX Runtime inference on GPUs.
+        Simply pass the execution provider as the onnx_execution_provider parameter. Additonal parameters can be passed
+        to the model using the model_kwargs parameter.
+        For more details refer to the HuggingFace documentation:
+        https://huggingface.co/docs/optimum/onnxruntime/usage_guides/gpu.
+    """
+
+    def __init__(
+        self,
+        model: str = "BAAI/bge-small-en-v1.5",
+        token: Optional[Secret] = Secret.from_env_var("HF_API_TOKEN", strict=False),  # noqa: B008
+        prefix: str = "",
+        suffix: str = "",
+        normalize_embeddings: bool = True,
+        onnx_execution_provider: str = "CPUExecutionProvider",
+        model_kwargs: Optional[Dict[str, Any]] = None,
+        batch_size: int = 32,
+        progress_bar: bool = True,
+        meta_fields_to_embed: Optional[List[str]] = None,
+        embedding_separator: str = "\n",
+    ):
+        """
+        Create a OptimumDocumentEmbedder component.
+
+        :param model: A string representing the model id on HF Hub. Default is "BAAI/bge-small-en-v1.5".
+        :param token: The HuggingFace token to use as HTTP bearer authorization.
+        :param prefix: A string to add to the beginning of each text.
+        :param suffix: A string to add to the end of each text.
+        :param normalize_embeddings: Whether to normalize the embeddings to unit length.
+        :param onnx_execution_provider: The execution provider to use for ONNX models. Defaults to
+        "CPUExecutionProvider".
+        :param model_kwargs: Dictionary containing additional keyword arguments to pass to the model.
+            In case of duplication, these kwargs override `model`, `onnx_execution_provider`, and `token` initialization
+            parameters.
+        :param batch_size: Number of Documents to encode at once.
+        :param progress_bar: Whether to show a progress bar or not. Can be helpful to disable in production deployments
+            to keep the logs clean.
+        :param meta_fields_to_embed: List of meta fields that should be embedded along with the Document text.
+        :param embedding_separator: Separator used to concatenate the meta fields to the Document text.
+        """
+        check_valid_model(model, HFModelType.EMBEDDING, token)
+        self.model = model
+
+        self.token = token
+        token = token.resolve_value() if token else None
+
+        self.prefix = prefix
+        self.suffix = suffix
+        self.normalize_embeddings = normalize_embeddings
+        self.onnx_execution_provider = onnx_execution_provider
+        self.batch_size = batch_size
+        self.progress_bar = progress_bar
+        self.meta_fields_to_embed = meta_fields_to_embed or []
+        self.embedding_separator = embedding_separator
+
+        model_kwargs = model_kwargs or {}
+
+        # Check if the model_kwargs contain the parameters, otherwise, populate them with values from init parameters
+        model_kwargs.setdefault("model_id", model)
+        model_kwargs.setdefault("provider", onnx_execution_provider)
+        model_kwargs.setdefault("use_auth_token", token)
+
+        self.model_kwargs = model_kwargs
+        self.embedding_model = None
+        self.tokenizer = None
+
+    def warm_up(self):
+        """
+        Convert the model to ONNX.
+
+        The model is cached if the "TensorrtExecutionProvider" is used, since it takes a while to to build the TensorRT
+        engine.
+        """
+        if self.embedding_model is None:
+            self.tokenizer = AutoTokenizer.from_pretrained(self.model)
+
+            if self.onnx_execution_provider == "TensorrtExecutionProvider":
+                # Cache engine for TensorRT
+                provider_options = {
+                    "trt_engine_cache_enable": True,
+                    "trt_engine_cache_path": f"tmp/trt_cache_{self.model}",
+                }
+                self.embedding_model = ORTModelForFeatureExtraction.from_pretrained(
+                    **self.model_kwargs, use_cache=False, provider_options=provider_options
+                )
+            else:
+                # export=True converts the model to ONNX on the fly
+                self.embedding_model = ORTModelForFeatureExtraction.from_pretrained(**self.model_kwargs, export=True)
+
+    def to_dict(self) -> Dict[str, Any]:
+        """
+        Serialize this component to a dictionary.
+        """
+        serialization_dict = default_to_dict(
+            self,
+            model=self.model,
+            prefix=self.prefix,
+            suffix=self.suffix,
+            normalize_embeddings=self.normalize_embeddings,
+            onnx_execution_provider=self.onnx_execution_provider,
+            batch_size=self.batch_size,
+            progress_bar=self.progress_bar,
+            meta_fields_to_embed=self.meta_fields_to_embed,
+            embedding_separator=self.embedding_separator,
+            model_kwargs=self.model_kwargs,
+            token=self.token.to_dict() if self.token else None,
+        )
+
+        model_kwargs = serialization_dict["init_parameters"]["model_kwargs"]
+        model_kwargs.pop("token", None)
+
+        serialize_hf_model_kwargs(model_kwargs)
+        return serialization_dict
+
+    @classmethod
+    def from_dict(cls, data: Dict[str, Any]) -> "OptimumDocumentEmbedder":
+        """
+        Deserialize this component from a dictionary.
+        """
+        deserialize_secrets_inplace(data["init_parameters"], keys=["token"])
+        deserialize_hf_model_kwargs(data["init_parameters"]["model_kwargs"])
+        return default_from_dict(cls, data)
+
+    def mean_pooling(self, model_output: torch.tensor, attention_mask: torch.tensor) -> torch.tensor:
+        """
+        Perform Mean Pooling on the output of the Embedding model.
+
+        :param model_output: The output of the embedding model.
+        :param attention_mask: The attention mask of the tokenized text.
+        :return: The embeddings of the text after mean pooling.
+        """
+        # First element of model_output contains all token embeddings
+        token_embeddings = model_output[0]
+        input_mask_expanded = attention_mask.unsqueeze(-1).expand(token_embeddings.size()).float()
+        sum_embeddings = torch.sum(token_embeddings * input_mask_expanded, 1)
+        sum_mask = torch.clamp(input_mask_expanded.sum(1), min=1e-9)
+        return sum_embeddings / sum_mask
+
+    def _prepare_texts_to_embed(self, documents: List[Document]) -> List[str]:
+        """
+        Prepare the texts to embed by concatenating the Document text with the metadata fields to embed.
+        """
+        texts_to_embed = []
+        for doc in documents:
+            meta_values_to_embed = [
+                str(doc.meta[key]) for key in self.meta_fields_to_embed if key in doc.meta and doc.meta[key] is not None
+            ]
+
+            text_to_embed = (
+                self.prefix + self.embedding_separator.join([*meta_values_to_embed, doc.content or ""]) + self.suffix
+            )
+
+            texts_to_embed.append(text_to_embed)
+        return texts_to_embed
+
+    def _embed_batch(self, texts_to_embed: List[str], batch_size: int) -> List[List[float]]:
+        """
+        Embed a list of texts in batches.
+        """
+        # Determine device for tokenizer output
+        device = (
+            "cuda"
+            if self.onnx_execution_provider
+            in ["CUDAExecutionProvider", "ROCMExecutionProvider", "TensorrtExecutionProvider"]
+            else "cpu"
+        )
+
+        # Sorting by length
+        length_sorted_idx = np.argsort([-len(sen) for sen in texts_to_embed])
+        sentences_sorted = [texts_to_embed[idx] for idx in length_sorted_idx]
+
+        all_embeddings = []
+        for i in tqdm(
+            range(0, len(sentences_sorted), batch_size), disable=not self.progress_bar, desc="Calculating embeddings"
+        ):
+            batch = sentences_sorted[i : i + batch_size]
+            encoded_input = self.tokenizer(batch, padding=True, truncation=True, return_tensors="pt").to(device)  # type: ignore
+
+            # Compute token embeddings
+            with torch.no_grad():
+                model_output = self.embedding_model(**encoded_input)  # type: ignore
+
+            # Perform mean pooling
+            sentence_embeddings = self.mean_pooling(model_output, encoded_input["attention_mask"].cpu())
+
+            all_embeddings.extend(sentence_embeddings.tolist())
+
+        # Reorder embeddings according to original order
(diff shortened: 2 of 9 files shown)


integrations/optimum/src/haystack_integrations/components/embedders/optimum_document_embedder.py
```diff
@@ -0,0 +1,269 @@
...
+        ):
+            batch = sentences_sorted[i : i + batch_size]
+            encoded_input = self.tokenizer(batch, padding=True, truncation=True, return_tensors="pt").to(device)  # type: ignore
+
+            # Compute token embeddings
+            with torch.no_grad():
+                model_output = self.embedding_model(**encoded_input)  # type: ignore
+
+            # Perform mean pooling
+            sentence_embeddings = self.mean_pooling(model_output, encoded_input["attention_mask"].cpu())
```
shadeMe, line 225: Why is the attention mask being moved to the host here? Shouldn't it be moved to `device` instead?

integrations/optimum/src/haystack_integrations/components/embedders/optimum_text_embedder.py
```diff
@@ -0,0 +1,200 @@
...
+
+import torch
+from haystack import component, default_from_dict, default_to_dict
+from haystack.utils import Secret, deserialize_secrets_inplace
+from haystack.utils.hf import HFModelType, check_valid_model, deserialize_hf_model_kwargs, serialize_hf_model_kwargs
+from optimum.onnxruntime import ORTModelForFeatureExtraction
+from transformers import AutoTokenizer
+
+
+class OptimumTextEmbedder:
```
shadeMe, line 11: It looks like a lot of the code below is essentially the same as the one in the document embedder. Let's move them to a common sub-module and call them instead.

shadeMe (changes requested):
Thanks for the comprehensive PR! I've added a few comments.

integrations/optimum/src/haystack_integrations/components/embedders/optimum_text_embedder.py
```diff
@@ -0,0 +1,200 @@
...
+        For more details refer to the HuggingFace documentation:
+        https://huggingface.co/docs/optimum/onnxruntime/usage_guides/gpu.
+    """
+
+    def __init__(
+        self,
+        model: str = "BAAI/bge-small-en-v1.5",
+        token: Optional[Secret] = Secret.from_env_var("HF_API_TOKEN", strict=False),  # noqa: B008
+        prefix: str = "",
+        suffix: str = "",
```
sjrl, line 48: Not sure what standard practice should be here, but the model that is specified by default recommends that we prepend an instruction to the query (resource [here](https://huggingface.co/BAAI/bge-small-en-v1.5#frequently-asked-questions))

> For the bge-*-v1.5, we improve its retrieval ability when not using instruction. No instruction only has a slight degradation in retrieval performance compared with using instruction. So you can generate embedding without instruction in all cases for convenience.
>
> For a retrieval task that uses short queries to find long related documents, it is recommended to add instructions for these short queries. The best method to decide whether to add instructions for queries is choosing the setting that achieves better performance on your task. In all cases, the documents/passages do not need to add the instruction.

so I wonder if we choose this model as default then we should also by default populate the `prefix` parameter?

integrations/optimum/src/haystack_integrations/components/embedders/optimum_text_embedder.py
```diff
@@ -0,0 +1,200 @@
...
+    @classmethod
+    def from_dict(cls, data: Dict[str, Any]) -> "OptimumTextEmbedder":
+        """
+        Deserialize this component from a dictionary.
+        """
+        deserialize_secrets_inplace(data["init_parameters"], keys=["token"])
+        deserialize_hf_model_kwargs(data["init_parameters"]["model_kwargs"])
+        return default_from_dict(cls, data)
+
+    def mean_pooling(self, model_output: torch.tensor, attention_mask: torch.tensor) -> torch.tensor:
```
sjrl, line 142: Another potential issue I see. The default model BAAI/bge-small-en-v1.5 shows in their code sample on HF that we should be using cls pooling and not mean pooling. I think we need to allow the user to specify which pooling method they would like to use since it can often change with the picked model. 

See here: https://huggingface.co/BAAI/bge-small-en-v1.5#using-huggingface-transformers

Relevant code snippet:
```python
from transformers import AutoTokenizer, AutoModel
import torch
# Sentences we want sentence embeddings for
sentences = ["样例数据-1", "样例数据-2"]

# Load model from HuggingFace Hub
tokenizer = AutoTokenizer.from_pretrained('BAAI/bge-large-zh-v1.5')
model = AutoModel.from_pretrained('BAAI/bge-large-zh-v1.5')
model.eval()

# Tokenize sentences
encoded_input = tokenizer(sentences, padding=True, truncation=True, return_tensors='pt')
# for s2p(short query to long passage) retrieval task, add an instruction to query (not add instruction for passages)
# encoded_input = tokenizer([instruction + q for q in queries], padding=True, truncation=True, return_tensors='pt')

# Compute token embeddings
with torch.no_grad():
    model_output = model(**encoded_input)
    # Perform pooling. In this case, cls pooling.
    sentence_embeddings = model_output[0][:, 0]
# normalize embeddings
sentence_embeddings = torch.nn.functional.normalize(sentence_embeddings, p=2, dim=1)
print("Sentence embeddings:", sentence_embeddings)
```

sjrl: If the model is on the HF Hub has the necessary Sentence Transformer files (e.g. https://huggingface.co/BAAI/bge-small-en-v1.5/blob/main/1_Pooling/config.json) then it would be possible to auto detect with pooling method to use.

integrations/optimum/src/haystack_integrations/components/embedders/optimum_text_embedder.py
```diff
@@ -0,0 +1,200 @@
...
+            "cuda"
+            if self.onnx_execution_provider
+            in ["CUDAExecutionProvider", "ROCMExecutionProvider", "TensorrtExecutionProvider"]
+            else "cpu"
+        )
+
+        encoded_input = self.tokenizer([text_to_embed], padding=True, truncation=True, return_tensors="pt").to(device)
+
+        # Compute token embeddings
+        with torch.no_grad():
```
sjrl, line 188: I believe it is generally recommend to use `torch.inference_mode()` instead of `torch.no_grad()` since I think it contains additional speed ups. 

However, I could be wrong here and that this does not mesh well with ONNX runtime. Do you know if we should switch to `torch.inference_mode()`?

integrations/optimum/pyproject.toml
```diff
@@ -0,0 +1,194 @@
...
+  "Programming Language :: Python :: 3.9",
+  "Programming Language :: Python :: 3.10",
+  "Programming Language :: Python :: 3.11",
+  "Programming Language :: Python :: Implementation :: CPython",
+  "Programming Language :: Python :: Implementation :: PyPy",
+]
+dependencies = [
+  "haystack-ai",
+  "transformers[sentencepiece]==4.36.2",
+  "sentence-transformers>=2.2.0",
```
sjrl, line 30: I think we can remove the `sentence-transformers` dependency. I ran the tests locally with this removed and everything still passed.

sjrl: Or is there a scenario in which this dependency is needed?

integrations/optimum/src/haystack_integrations/components/embedders/optimum_text_embedder.py
shadeMe: Now that you mention it, I think it's a no-op in either case as PyTorch won't be used to execute the model. The tokenizer's encode operation could still make use of it, but I'd wager that the cost/memory savings are negligible at best. 

We can just remove it.

sjrl:
I took a look into this 

> The versions of optimum and transformers has been pinned to optimum==1.15.0 and transformers==4.36.2, due to bugs in the latest optimum release. Please refer to the following issues for more information: https://github.com/huggingface/optimum/issues/1673 and https://github.com/huggingface/optimum/issues/1675.

to see if there was a way we could overcome pinning the dependencies. 

Thankfully, it looks like https://github.com/huggingface/optimum/issues/1675 has been resolved with version 1.16.2, but the other issue still remains. 

Another error that I found while running the tests with versions `optimum>=1.16.2` and `transformers>=4.37.2` was 
```bash
onnxruntime.capi.onnxruntime_pybind11_state.InvalidArgument: [ONNXRuntimeError] : 2 : INVALID_ARGUMENT : Invalid input name: token_type_ids
```
which can be solved with the following code block added right before calling the model 
```python
# Only pass required inputs otherwise onnxruntime can raise an error
inputs_to_remove = set(encoded_input.keys()).difference(self.embedding_model.inputs_names)
for key in inputs_to_remove:
    encoded_input.pop(key)
```
Even though this isn't needed with the current pinned dependencies (at least for the tested models) I think this would be good to add to avoid needing to solve this error in the future.

integrations/optimum/pyproject.toml
sjrl: I took a deeper dive into this, since I found I was getting errors like this 
```python
../../../../../../Library/Application Support/hatch/env/virtual/optimum-haystack/oY5Xs40Q/optimum-haystack/lib/python3.10/site-packages/optimum/exporters/tasks.py:1240: in get_model_class_for_task
    loaded_library = importlib.import_module(library)
/opt/homebrew/Caskroom/miniforge/base/envs/haystack-core-integrations/lib/python3.10/importlib/__init__.py:126: in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
<frozen importlib._bootstrap>:1050: in _gcd_import
    ???
<frozen importlib._bootstrap>:1027: in _find_and_load
    ???
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

name = 'sentence_transformers', import_ = <function _gcd_import at 0x10115f490>

>   ???
E   ModuleNotFoundError: No module named 'sentence_transformers'
```
when running the tests with versions `transformers[sentencepiece]>=4.37.2` and `optimum[onnxruntime]>=1.16.2`. 

The culprit it appears is that the [main export function](https://github.com/huggingface/optimum/blob/8651c0ca1cccf095458bc80329dec9df4601edb4/optimum/exporters/onnx/__main__.py#L164) of Optimum into ONNX has a option called `library_name`
```python
        library_name (`Optional[str]`, defaults to `None`):
            The library of the model (`"transformers"` or `"diffusers"` or `"timm"` or `"sentence_transformers"`). If not provided, will attempt to automatically detect the library name for the checkpoint.
``` 
which auto detects that some of the models in the test should be loaded with the `sentence_transformers` library. I'm not entirely sure why this error doesn't occur with the current pinned dependencies (I was able to get tests to pass without sentence transformers installed), but it looks like Optimum basically has some hidden dependencies (namely these three additional libraries `diffusers`, `timm` and `sentence_transformers`) depending on which model you load from the HF Hub. 

Not entirely sure how to proceed here, but it seems like keeping `sentence_transformers` as a dependency is probably a good idea due to the popularity of the format.

integrations/optimum/src/haystack_integrations/components/embedders/optimum_text_embedder.py
```diff
@@ -0,0 +1,200 @@
...
+        # Determine device for tokenizer output
+        device = (
+            "cuda"
+            if self.onnx_execution_provider
+            in ["CUDAExecutionProvider", "ROCMExecutionProvider", "TensorrtExecutionProvider"]
+            else "cpu"
+        )
+
+        encoded_input = self.tokenizer([text_to_embed], padding=True, truncation=True, return_tensors="pt").to(device)
+
```
sjrl, line 186: Based on the details of this [comment](https://github.com/deepset-ai/haystack-core-integrations/pull/379#issuecomment-1945480803), I'd recommend adding this code block here
```python
# Only pass required inputs otherwise onnxruntime can raise an error
inputs_to_remove = set(encoded_input.keys()).difference(self.embedding_model.inputs_names)
for key in inputs_to_remove:
    encoded_input.pop(key)
```

shadeMe:
@awinml Just wanted to give you an update that we'll move this PR on to our plate to expedite its merging.

integrations/optimum/src/haystack_integrations/components/embedders/optimum_document_embedder.py
```diff
@@ -0,0 +1,269 @@
...
+
+    def warm_up(self):
+        """
+        Convert the model to ONNX.
+
+        The model is cached if the "TensorrtExecutionProvider" is used, since it takes a while to to build the TensorRT
+        engine.
+        """
+        if self.embedding_model is None:
+            self.tokenizer = AutoTokenizer.from_pretrained(self.model)
```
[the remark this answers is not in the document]
awinml: Yes, I will add the `token` parameter to it.

integrations/optimum/src/haystack_integrations/components/embedders/optimum_document_embedder.py
```diff
@@ -0,0 +1,269 @@
...
+
+            if self.onnx_execution_provider == "TensorrtExecutionProvider":
+                # Cache engine for TensorRT
+                provider_options = {
+                    "trt_engine_cache_enable": True,
+                    "trt_engine_cache_path": f"tmp/trt_cache_{self.model}",
```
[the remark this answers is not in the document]
awinml: By default, the ONNX model is cached similar to how other transformers model are cached. I don't think introducing an additional caching parameter is necessary here.

We could highlight that it takes a while to build the TensorRT engine in the docs and suggest the user to pass the parameters for caching it, in the following manner:
```python
embedder = OptimumTextEmbedder(
    model_kwargs={
        "provider_options": {
            "trt_engine_cache_enable": True,
            "trt_engine_cache_path": f"tmp/trt_cache_model",
        }
    }
)
```

integrations/optimum/src/haystack_integrations/components/embedders/optimum_document_embedder.py
awinml: You're right, we can put it directly on the device here. I will make that change.

integrations/optimum/src/haystack_integrations/components/embedders/optimum_text_embedder.py
awinml: I have moved the model loading code into a separate backend sub-module (similar to how we have done with the Sentence Transformers Embedders). It has an `embed()` method that can be used for both the embedders.

integrations/optimum/src/haystack_integrations/components/embedders/optimum_text_embedder.py
awinml: I think setting a default prefix is a bad idea. I have changed the default model to `sentence-transformers/all-mpnet-base-v2`, which does not require a prefix or suffix by default. This is also the default for the Sentence Transformers Embedders, so it maintains consistency.

integrations/optimum/src/haystack_integrations/components/embedders/optimum_text_embedder.py
awinml: You're right, I will remove it, its not required here.

integrations/optimum/pyproject.toml
awinml: You're right, this is the reason I added `sentence_transformers` as a dependency.

awinml:
Thanks! @sjrl 

I added this code block and unpinned the `optimum` and `transformers` versions.
```python 
# Only pass required inputs otherwise onnxruntime can raise an error
inputs_to_remove = set(encoded_input.keys()).difference(self.embedding_model.inputs_names)
for key in inputs_to_remove:
    encoded_input.pop(key)
```

All the tests pass successfully!

integrations/optimum/src/haystack_integrations/components/embedders/optimum_document_embedder.py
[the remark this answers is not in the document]
shadeMe: Adding it as a suggestion sounds better!

integrations/optimum/pyproject.toml
sjrl: @shadeMe should we add this as a developer comment somewhere? I'm thinking that future users or devs might run into this issue for `timm` or `diffusers` (or any other libraries they might add) and be confused.

integrations/optimum/src/haystack_integrations/components/embedders/optimum_document_embedder.py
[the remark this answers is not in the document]
awinml: That's good, I will add it to the docstrings.

integrations/optimum/pyproject.toml
shadeMe: Definitely! We can directly add it to `pyproject.toml`, I think?

awinml: Sounds good! I will add it to these comments to the `pyproject.toml`.

integrations/optimum/src/haystack_integrations/components/embedders/optimum_text_embedder.py
awinml: Yes, we can do that. I have added a `pooling` submodule with classes to fetch the sentence-transformers config from HF Hub. The other pooling methods (CLS Pooling, Max Pooling, Mean Square Root - Length Pooling, Weighted Mean Pooling and Last Token Pooling) are also now supported. 

The implementation for the pooling methods has been adapted from the [sentence-transformers pooling implementation](https://github.com/UKPLab/sentence-transformers/blob/master/sentence_transformers/models/Pooling.py).

integrations/optimum/src/haystack_integrations/components/embedders/optimum_text_embedder.py
```diff
@@ -0,0 +1,201 @@
...
+        self.token = token
+        token = token.resolve_value() if token else None
+
+        if isinstance(pooling_mode, str):
+            self.pooling_mode = PoolingMode.from_str(pooling_mode)
+        # Infer pooling mode from model config if not provided,
+        if pooling_mode is None:
+            self.pooling_mode = HFPoolingMode.get_pooling_mode(model, token)
+        # Set default to "mean" if not found in model config and not specified by user
```
sjrl, line 115: I actually think it would be better to raise an error here if no pooling mode was provided or detected. This will still fail fast since it is in the init method and by providing a default we allow the user to make a silent error (e.g. the config download from HF could have failed due to a server issue or loss of wifi) which then would lead to sub-par results when using the embeddings.

awinml: Makes sense! I will update it so that it raises an error when no pooling mode is provided or detected.

integrations/optimum/src/haystack_integrations/components/embedders/pooling.py
```diff
@@ -0,0 +1,231 @@
...
+                pooling_mode = pooling_modes_map.get(pooling_mode_from_config)
+            # If no True pooling modes or more than one True pooling mode is found, return None
+            else:
+                pooling_mode = None
+            return pooling_mode
+        except Exception:
+            return None
+
+
+class Pooling:
```
shadeMe, line 82: The code in this class is essentially a duplication of the [`Pooling` class in `sentence-transformers`](https://github.com/UKPLab/sentence-transformers/blob/ecdda29859d6a651326addd11bd237870b39866a/sentence_transformers/models/Pooling.py#L100). As such, I see no need for it, especially since the latter is a dependency for this integration.

Let's just reuse their code instead. We can still have the enum and a wrapper class to convert the inputs/outputs to the right formats.

shadeMe (changes requested):

integrations/optimum/src/haystack_integrations/components/embedders/backends/optimum_backend.py
```diff
@@ -0,0 +1,109 @@
...
+import numpy as np
+import torch
+from haystack.utils.auth import Secret
+from haystack_integrations.components.embedders.pooling import Pooling, PoolingMode
+from optimum.onnxruntime import ORTModelForFeatureExtraction
+from tqdm import tqdm
+from transformers import AutoTokenizer
+
+
+class _OptimumEmbeddingBackendFactory:
```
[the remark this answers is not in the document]
awinml: Currently, this class is just being used to reuse instances of the model when its loaded. Since the ONNX runtime backend management is being handled by `ORTModelForFeatureExtraction`, I guess we don't need this abstraction.

I will simplify the implementation so that we directly use `OptimumEmbeddingBackend` for both the embedders.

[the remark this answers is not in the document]
shadeMe: Yeah, let's directly use the backend class for now. Also, the cache key was under-specified, leading to subtle conflicts when the same model was used with different model args.

integrations/optimum/src/haystack_integrations/components/embedders/pooling.py
```diff
@@ -0,0 +1,231 @@
...
+            if len(true_pooling_modes) == 1:
+                pooling_mode_from_config = true_pooling_modes[0]
+                pooling_mode = pooling_modes_map.get(pooling_mode_from_config)
+            # If no True pooling modes or more than one True pooling mode is found, return None
+            else:
+                pooling_mode = None
+            return pooling_mode
+        except Exception:
+            return None
+
```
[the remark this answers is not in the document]
awinml: Yes, we can scope the exception here. Updated it to raise a `ValueError` with the exception.

integrations/optimum/src/haystack_integrations/components/embedders/backends/optimum_backend.py
```diff
@@ -0,0 +1,109 @@
...
+        texts_to_embed: Union[str, List[str]],
+        normalize_embeddings: bool,
+        pooling_mode: PoolingMode = PoolingMode.MEAN,
+        progress_bar: bool = False,
+        batch_size: int = 1,
+    ) -> Union[List[List[float]], List[float]]:
+        """
+        Embed text or list of texts using the Optimum model.
+
+        :param texts_to_embed: T
```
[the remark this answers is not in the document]
awinml: The docstring has been updated.

integrations/optimum/src/haystack_integrations/components/embedders/backends/optimum_backend.py
```diff
@@ -0,0 +1,109 @@
...
+                pooling_mode=pooling_mode,
+                attention_mask=encoded_input["attention_mask"].to(device),
+                model_output=model_output,
+            )
+            sentence_embeddings = pooling.pool_embeddings()
+
+            all_embeddings.extend(sentence_embeddings.tolist())
+
+        # Reorder embeddings according to original order
+        all_embeddings = [all_embeddings[idx] for idx in np.argsort(length_sorted_idx)]
```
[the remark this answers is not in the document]
awinml: Thanks! I have updated the code and also added a unit test for the same.

integrations/optimum/src/haystack_integrations/components/embedders/optimum_document_embedder.py
```diff
@@ -63,13 +64,47 @@ def __init__(
         """
         Create a OptimumDocumentEmbedder component.
 
-        :param model: A string representing the model id on HF Hub. Default is "BAAI/bge-small-en-v1.5".
+        :param model: A string representing the model id on HF Hub. Defaults to
+            "sentence-transformers/all-mpnet-base-v2".
```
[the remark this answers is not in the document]
awinml: The docstring has been updated.

integrations/optimum/src/haystack_integrations/components/embedders/optimum_text_embedder.py
```diff
@@ -42,24 +44,59 @@ class OptimumTextEmbedder:
...
         onnx_execution_provider: str = "CPUExecutionProvider",
+        pooling_mode: Optional[Union[str, PoolingMode]] = None,
         model_kwargs: Optional[Dict[str, Any]] = None,
     ):
         """
         Create a OptimumTextEmbedder component.
 
-        :param model: A string representing the model id on HF Hub. Default is "BAAI/bge-small-en-v1.5".
+        :param model: A string representing the model id on HF Hub. Defaults to
+            "sentence-transformers/all-mpnet-base-v2".
```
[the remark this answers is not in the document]
awinml: The docstring has been updated.

integrations/optimum/src/haystack_integrations/components/embedders/backends/optimum_backend.py
```diff
@@ -0,0 +1,109 @@
...
+
+            # Pool Embeddings
+            pooling = Pooling(
+                pooling_mode=pooling_mode,
+                attention_mask=encoded_input["attention_mask"].to(device),
+                model_output=model_output,
+            )
+            sentence_embeddings = pooling.pool_embeddings()
+
+            all_embeddings.extend(sentence_embeddings.tolist())
```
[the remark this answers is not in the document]
awinml: The code has been updated to convert to a list after the normalization has been done.

integrations/optimum/src/haystack_integrations/components/embedders/optimum_document_embedder.py
```diff
@@ -107,26 +151,12 @@ def __init__(
...
-                    "trt_engine_cache_enable": True,
-                    "trt_engine_cache_path": f"tmp/trt_cache_{self.model}",
-                }
-                self.embedding_model = ORTModelForFeatureExtraction.from_pretrained(
-                    **self.model_kwargs, use_cache=False, provider_options=provider_options
-                )
-            else:
-                # export=True converts the model to ONNX on the fly
-                self.embedding_model = ORTModelForFeatureExtraction.from_pretrained(**self.model_kwargs, export=True)
+        if not hasattr(self, "embedding_backend"):
```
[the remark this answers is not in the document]
awinml: Okay, I will do that.

integrations/optimum/src/haystack_integrations/components/embedders/optimum_document_embedder.py
```diff
@@ -251,7 +227,7 @@ def run(self, documents: List[Document]):
             )
             raise TypeError(msg)
 
-        if not (self.embedding_model and self.tokenizer):
+        if not hasattr(self, "embedding_backend"):
```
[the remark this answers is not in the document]
awinml: Updated. It is now set to `None` in init and initialized in `warm_up`.

integrations/optimum/src/haystack_integrations/components/embedders/optimum_text_embedder.py
```diff
@@ -168,33 +188,14 @@ def run(self, text: str):
             )
             raise TypeError(msg)
 
-        if not (self.embedding_model and self.tokenizer):
+        if not hasattr(self, "embedding_backend"):
```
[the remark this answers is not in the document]
awinml: Updated.

integrations/optimum/src/haystack_integrations/components/embedders/pooling.py
awinml: Makes sense, now that we are keeping `sentence-transformers` as a dependency.

The individual pooling methods have been removed. The `pool_embeddings` method now converts the inputs and outputs to the right format.

integrations/optimum/src/haystack_integrations/components/embedders/optimum_document_embedder.py
```diff
@@ -85,6 +120,15 @@ def __init__(
...
         token = token.resolve_value() if token else None
 
+        if isinstance(pooling_mode, str):
+            self.pooling_mode = PoolingMode.from_str(pooling_mode)
+        # Infer pooling mode from model config if not provided,
+        if pooling_mode is None:
+            self.pooling_mode = HFPoolingMode.get_pooling_mode(model, token)
+        # Set default to "mean" if not found in model config and not specified by user
+        if self.pooling_mode is None:
+            self.pooling_mode = PoolingMode.MEAN
```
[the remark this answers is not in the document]
awinml: Updated. An error is raised if the Pooling mode not found in model config and not specified by user.

.github/workflows/optimum.yml
```diff
@@ -0,0 +1,56 @@
...
+      - uses: actions/checkout@v4
+
+      - name: Set up Python ${{ matrix.python-version }}
+        uses: actions/setup-python@v5
+        with:
+          python-version: ${{ matrix.python-version }}
+
+      - name: Install Hatch
+        run: pip install --upgrade hatch
+
```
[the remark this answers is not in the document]
awinml: Yes, I will add it. What should I set the `order` value?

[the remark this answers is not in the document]
shadeMe: @dfokina What should the `order` be?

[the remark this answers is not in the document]
awinml: Okay, I will do that!

integrations/optimum/pydoc/config.yml
```diff
@@ -0,0 +1,32 @@
...
+    expression:
+    documented_only: true
+    do_not_filter_modules: false
+    skip_empty_modules: true
+  - type: smart
+  - type: crossref
+renderer:
+  type: haystack_pydoc_tools.renderers.ReadmePreviewRenderer
+  excerpt: Optimum integration for Haystack
+  category_slug: haystack-integrations
```
dfokina, line 23: Apologies, we just recently changed the required category slug.

```suggestion
  category_slug: integrations-api
```

shadeMe (approved):
Thanks again for the contribution!

Issue #378, opened 2024-02-08
feat(unstructured): adding element index as metadata when partitioning by elem


lambda-science:
**Is your feature request related to a problem? Please describe.**
With unstructured we can do partition: one doc per file, one doc per page or one doc per elem.
When doing one doc per elem there is no way to track orders of element coming from a same doucment.
This informations can be usefull for example for a `ContextExpander` component.
Let's say you retrieve an element, you can retrieve also the previous and next element to expand the current context .

**Describe the solution you'd like**
Just automatically add index metadata.
I will do a PR.

anakin87:
done in #382 
new release: https://pypi.org/project/unstructured-fileconverter-haystack/0.4.0/

PR #382, opened 2024-02-08
feat(unstructured): add element index as metadata

lambda-science:
Adding element index in the metadata when partition mode is by element
Close feature request: https://github.com/deepset-ai/haystack-core-integrations/issues/378

> Is your feature request related to a problem? Please describe.
> With unstructured we can do partition: one doc per file, one doc per page or one doc per elem.
> When doing one doc per elem there is no way to track orders of element coming from a same doucment.
> This informations can be usefull for example for a ContextExpander component.
> Let's say you retrieve an element, you can retrieve also the previous and next element to expand the current context .
> 
> Describe the solution you'd like
> Just automatically add index metadata.
> I will do a PR.


silvanocerza (approved):
Cool, thanks! 🚀

lambda-science:
> Cool, thanks! 🚀

Happy about the merge ! :)
Is it possible to do a small bump of the version to release the feature ? :)

Issue #370, opened 2024-02-08
Update core integration secret managment - Cohere


shadeMe:
- [x] Code (c.f https://github.com/deepset-ai/haystack-core-integrations/pull/347)
- [x] Documentation
- [x] Tutorials
- [x] Cookbooks
- [x] Articles
- [x] Integration Tiles

bilgeyucel:
We don't have any articles around Cohere, this issue is resolved

PR #386, opened 2024-02-09
refactor!: Use `Secret` for API keys in Cohere components

shadeMe:
Related to #370.

### DIFF
--- a/integrations/cohere/pyproject.toml
+++ b/integrations/cohere/pyproject.toml
@@ -10,9 +10,7 @@ readme = "README.md"
 requires-python = ">=3.7"
 license = "Apache-2.0"
 keywords = []
-authors = [
-  { name = "deepset GmbH", email = "info@deepset.ai" },
-]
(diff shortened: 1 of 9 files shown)


julian-risch (approved):
LGTM! 👍 I thought about whether we should add a link to a documentation page about secret management in Haystack in the docstrings where we currently just say: `:param api_key: The API key for the Cohere API.`. However, it's such a core concept that we can  mention it in getting started pages, etc. rather than in every docstring of components with api keys.

shadeMe:
> LGTM! 👍 I thought about whether we should add a link to a documentation page about secret management in Haystack in the docstrings where we currently just say: `:param api_key: The API key for the Cohere API.`. However, it's such a core concept that we can mention it in getting started pages, etc. rather than in every docstring of components with api keys.

Yeah, I've already drafted a explainer doc for it. It'll appear on the sidebar under the "Haystack Concepts" section once it's merged.

Issue #358, opened 2024-02-08
Fastembed Integration


nickprock:
## Summary and motivation

Add the Fastembed embedding functionality to Haystack

## Detailed design

FastEmbed is a lightweight, fast, Python library built for embedding generation. It doesn't need a lot of dependencies, embedders are quantized and use ONNX runtime.

## Checklist

If the request is accepted, ensure the following checklist is complete before closing this issue.
```[tasklist]
### Tasks
- [x] The code is documented with docstrings and was merged in the `main` branch
- [x] Docs are published at https://docs.haystack.deepset.ai/ (docs in review)
- [x] There is a Github workflow running the tests for the integration nightly and at every PR
- [x] A label named like `integration:<your integration name>` has been added to this repo
- [x] The [labeler.yml](https://github.com/deepset-ai/haystack-core-integrations/blob/main/.github/labeler.yml) file has been updated
- [x] The package has been released on PyPI
- [x] An integration tile has been added to https://github.com/deepset-ai/haystack-integrations
- [x] The integration has been listed in the [Inventory section](https://github.com/deepset-ai/haystack-core-integrations#inventory) of this repo README
- [x] There is an example available to demonstrate the feature
- [x] The feature was announced through social media
```

anakin87:
Hey, @nickprock...

 We are happy to have this integration! :heart_decoration: 
 
To understand how to start, you can look at similar integrations (instructor-embedders).
If you struggle with the machinery, ping me and I will provide some help!
(Please also create small PRs...)

anakin87:
First implementation merged.
Missing:
- support for prefix/suffix ([see for example OpenAI Embedders](https://github.com/deepset-ai/haystack/blob/549021d2fcf26feac772731ae1e5607365be242b/haystack/components/embedders/openai_text_embedder.py#L37))
- support for FastEmbed parallel

nickprock:
Fix to use it in pipeline:
* convert value in embeddings from numpy float to float

dfokina:
Hi, the documentation is published:
https://docs.haystack.deepset.ai/v2.0/docs/fastembedtextembedder
https://docs.haystack.deepset.ai/v2.0/docs/fastembeddocumentembedder

PR #388, opened 2024-02-10
Fastembed: fix readme and examples

anakin87:
Related to #358 
README and some usage examples have wrong import paths.


anakin87:
done in #389

PR #390, opened 2024-02-11
fastembed fix: added prefix and suffix

nickprock:
As requested [here](https://github.com/deepset-ai/haystack-core-integrations/issues/358#issuecomment-1936965853) I have added suffix and prefix.

refer to issue #358


anakin87 (approved):
Nice! :star2:

PR #391, opened 2024-02-12
fix: from numpy float to float

nickprock:
To use the library into pipeline I must change the type of elements in mebeddings from numpy.float to float.

related to issue #358


nickprock:
at the moment I can't fix it, I thought it was more immediate, I will work on it as soon as I can

nickprock:
*"I put on a colored patch"* (we say in Italy)

anakin87:
Changing the backend should do the trick :smiley:

anakin87 (approved):

nickprock:
> Changing the backend should do the trick :smiley: 

I didn't have the courage to touch it from the GitHub online editor. 😅

PR #397, opened 2024-02-12
docs: update paths and titles

dfokina:
Updated:

- Correct order for the pydocs (to follow alphabetical order)
- Paths to docstrings
- Distinct names for Embedders

### DIFF
--- a/integrations/instructor_embedders/pydoc/config.yml
+++ b/integrations/instructor_embedders/pydoc/config.yml
@@ -3,8 +3,9 @@ loaders:
     search_path: [../src]
     modules:
       [
-        "haystack_integrations.components.embedders.instructor_embedders",
-        "haystack_integrations.components.embedders.instructor_embedders.embedding_backend",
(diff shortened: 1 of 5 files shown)


integrations/instructor_embedders/pydoc/config.yml
```diff
@@ -17,9 +18,9 @@ processors:
   - type: crossref
 renderer:
   type: haystack_pydoc_tools.renderers.ReadmePreviewRenderer
-  excerpt: Embedders integration for Haystack
+  excerpt: Instructor integration for Haystack
```
anakin87, line 21: Probably, it's better to specify Instructor embedders.
There is a different project called just "Instructor."

integrations/instructor_embedders/pydoc/config.yml
```diff
@@ -17,9 +18,9 @@ processors:
...
   category_slug: haystack-integrations
-  title: Embedders
+  title: Instructor
```
anakin87, line 23: same as above

PR #402, opened 2024-02-13
Adopt `Secret` to pgvector

davidsbatista:
PostgreSQL connection string, read from an environment variable with the `Secret` component:
- adapted the tests to use the secret component
- added a custom `from_dict` method to the doc store class
- set some methods as `staticmethod`, since the self was not being used
- change the `coverage` config so that it actually shows the coverage over code and not test
- improved the documentation
- fixed some typos


anakin87:
Hey! I'm not still an expert on the new [Secret management](https://www.notion.so/deepsetai/Haystack-Concepts-Secret-Management-256078841b14498987ec7d3cd2fdb933).

I think your implementation should follow in some way that of [`UptrainEvaluator`](https://github.com/deepset-ai/haystack-core-integrations/blob/main/integrations/uptrain/src/haystack_integrations/components/evaluators/uptrain/evaluator.py).

In the current PR, I see that the `connection_string` is resolved at initialization,
then the `to_dict` method includes it in serialization, leaking it.

I don't know what the best way to approach this is, and I would leave it to you to experiment/decide.
One possibility is to have a public `connection_string` that is serialized as Secret, without leaking it and another internal one,  `_connection_string`, that is not serialized/is excluded by the `to_dict` method.

Let me know what you think and if you need help.

davidsbatista:
thanks for your comments @anakin87 - I did some changes

integrations/pgvector/src/haystack_integrations/document_stores/pgvector/document_store.py
```diff
@@ -162,6 +163,13 @@ def to_dict(self) -> Dict[str, Any]:
             hnsw_ef_search=self.hnsw_ef_search,
         )
 
+    @classmethod
+    def from_dict(cls, init_parameters: Dict[str, Any]) -> "PgvectorDocumentStore":
+        conn_str_data = init_parameters["init_parameters"]["connection_string"]
+        conn_str = Secret.from_dict(conn_str_data) if conn_str_data is not None else None
+        init_parameters["init_parameters"]["connection_string"] = conn_str
+        return default_from_dict(cls, init_parameters)
```
anakin87, line 171: Not sure if this part can be simplified
(something like [this](https://github.com/deepset-ai/haystack-core-integrations/blob/9c069c948983a4b725a056c46cd347ab52349656/integrations/uptrain/src/haystack_integrations/components/evaluators/uptrain/evaluator.py#L157-158))

anakin87 (approved):
Great job!

I left a comment with a doubt. if it makes sense, please consider it.

integrations/pgvector/src/haystack_integrations/document_stores/pgvector/document_store.py
davidsbatista: nice! thanks, well spotted :) 👍🏽

PR #409, opened 2024-02-13
Adding mistral generator and embedders

TuanaCelik:
This PR adds the `MistralChatGenerator`, the `MistralDocumentEmbedder` and the `MistralTextEmbedder`.

There are still 2 thing I'd like to finish:

- [x] Add tests
- [x] Add examples


TuanaCelik:
Notes: 
- I created the chat generator tests from the openai chat generator tests
- let me know if this was not best practice as I re-made the mode chat chunks etc inside the test files here.

TuanaCelik:
Additional note: 
- I noticed that I cannot test for `api_bae_url` in `MistralTextEmbedder` because of the way `OpenAITextEmbedder` is set up.
- I created this PR to resolve that, which has been merged now: https://github.com/deepset-ai/haystack/pull/7004
- Until it's released, the tests here won't pass with `api_base_url` being looked at in the tests for the text generator.

shadeMe (changes requested):
Just a couple of small changes. Once those are added and the tests pass, this will look good to merge!

integrations/mistral/tests/test_mistral_chat_generator.py
```diff
@@ -0,0 +1,314 @@
...
+from typing import Iterator
+from unittest.mock import patch
+
+import pytest
+import pytz
+from haystack.components.generators.utils import print_streaming_chunk
+from haystack.dataclasses import ChatMessage, StreamingChunk
+from haystack.utils.auth import Secret
+from haystack_integrations.components.generators.mistral.chat.chat_generator import MistralChatGenerator
+from openai import OpenAIError, Stream
```
davidsbatista, line 12: `openai` is not a dependency on `pyproject.toml`

davidsbatista: wait, I'm wrong, it's another issue - I think it should be added here

```
[[tool.mypy.overrides]]
module = [
  "mistral.*",
  "haystack.*",
  "haystack_integrations.*",
  "pytest.*",
  "numpy.*",
]
```

I get this error when running the tests locally:

```
tests/test_mistral_chat_generator.py:12: error: Cannot find implementation or library stub for module named "openai"  [import-not-found]
tests/test_mistral_chat_generator.py:13: error: Cannot find implementation or library stub for module named "openai.types.chat"  [import-not-found]
tests/test_mistral_chat_generator.py:14: error: Cannot find implementation or library stub for module named "openai.types.chat.chat_completion_chunk"  [import-not-found]

```

TuanaCelik: Hey @davidsbatista - Just so that I'm understsanding correctly, should I do the following? This one confused me because isn't openai a dependency of haystack-ai anyway? Why would we need to install it separately. Educate me pls 🙏 
```
[[tool.mypy.overrides]]
module = [
  "mistral.*",
  "haystack.*",
  "haystack_integrations.*",
  "openai.*",
  "pytest.*",
  "numpy.*",
]
```

davidsbatista: It's not related to the dependencies; my bad, I thought it was.

This is `mypy`, saying that it could not find type annotations for that lib; it often happens with 3rd party libs, so we are telling `mypy` to ignore it.

.github/workflows/mistral.yml
```diff
@@ -51,7 +51,11 @@ jobs:
 
       - name: Lint
         if: matrix.python-version == '3.9' && runner.os == 'Linux'
-        run: hatch run lint:all
+        run: hatch run lint:all\
```
shadeMe, line 54:
```suggestion
        run: hatch run lint:all
```

shadeMe (changes requested):

shadeMe (approved):

PR #411, opened 2024-02-14
Jina: update secrets management

anakin87:
part of #366

### DIFF
--- a/integrations/jina/src/haystack_integrations/components/embedders/jina/document_embedder.py
+++ b/integrations/jina/src/haystack_integrations/components/embedders/jina/document_embedder.py
@@ -1,11 +1,11 @@
 # SPDX-FileCopyrightText: 2023-present deepset GmbH <info@deepset.ai>
 #
 # SPDX-License-Identifier: Apache-2.0
-import os
 from typing import Any, Dict, List, Optional, Tuple
 
 import requests
-from haystack import Document, component, default_to_dict
+from haystack import Document, component, default_from_dict, default_to_dict
+from haystack.utils import Secret, deserialize_secrets_inplace
 from tqdm import tqdm
 
 JINA_API_URL: str = "https://api.jina.ai/v1/embeddings"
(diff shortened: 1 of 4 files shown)


integrations/jina/src/haystack_integrations/components/embedders/jina/text_embedder.py
```diff
@@ -48,22 +48,22 @@ def __init__(
...
         """
 
-        api_key = api_key or os.environ.get("JINA_API_KEY")
-        # we check whether api_key is None or an empty string
-        if not api_key:
+        resolved_api_key = api_key.resolve_value()
+        if resolved_api_key is None:
             msg = (
-                "JinaTextEmbedder expects an API key. "
+                "JinaDocumentEmbedder expects an API key. "
```
julian-risch, line 54: Needs to be changed back to `JinaTextEmbedder`

julian-risch (changes requested):
One copy-paste issue that needs to be fixed. Otherwise looks good to me. 👍

integrations/jina/src/haystack_integrations/components/embedders/jina/text_embedder.py
anakin87: fixed. Thanks!

julian-risch (approved):
LGTM! 👍

integrations/jina/src/haystack_integrations/components/embedders/jina/document_embedder.py
```diff
@@ -57,16 +56,15 @@ def __init__(
...
-        api_key = api_key or os.environ.get("JINA_API_KEY")
-        # we check whether api_key is None or an empty string
-        if not api_key:
+        resolved_api_key = api_key.resolve_value()
+        if resolved_api_key is None:
             msg = (
                 "JinaDocumentEmbedder expects an API key. "
                 "Set the JINA_API_KEY environment variable (recommended) or pass it explicitly."
             )
             raise ValueError(msg)
```
shadeMe, line 65: `resolve_value()` will raise an error if `strict=True`, which it is by default, and the env var is not set, so this is dead code. Let's remove it and directly call the method when updating the headers. Same for the other component too.

PR #412, opened 2024-02-14
Add `WeaviateEmbeddingRetriever`

silvanocerza:
Part of #131. Depends on #410.


anakin87 (approved):
Looks good!

_Unrelated to this PR:_
as a user, I would like to understand how I can set a **different [distance metric](https://weaviate.io/developers/weaviate/config-refs/distances#available-distance-metrics)** when I create my Document Store/Index. 
Can we expose this configuration and make it visible?

PR #416, opened 2024-02-14
Adopt Secret to Amazon Bedrock

davidsbatista:
- added `Secret` to `__init__` AWS params
- updated the tests
- moved all the fixtures to a `conftest.py`
- added an extra fixture to define env vars for testing
- changed the `pyproject.toml`
  - to reflect test coverage over the code
  - added an ignore to allow for calling functions as params


davidsbatista:
nit: @shadeMe do you know about a more concise/clean way to define the AWS params on the `__init__()` with `Secret`?

integrations/amazon_bedrock/src/haystack_integrations/components/generators/amazon_bedrock/generator.py
```diff
@@ -72,11 +73,21 @@ class AmazonBedrockGenerator:
...
+        ),
+        aws_secret_access_key: Optional[Secret] = EnvVarSecret(
+            env_vars=["AWS_SECRET_ACCESS_KEY"], strict=False
+        ).from_env_var("AWS_SECRET_ACCESS_KEY"),
+        aws_session_token: Optional[Secret] = EnvVarSecret(env_vars=["AWS_SESSION_TOKEN"], strict=False).from_env_var(
+            "AWS_SESSION_TOKEN"
+        ),
+        aws_region_name: Optional[Secret] = EnvVarSecret(env_vars=["AWS_DEFAULT_REGION"], strict=False).from_env_var(
+            "AWS_DEFAULT_REGION"
+        ),
```
shadeMe, line 87: Neither `EnvVarSecret` nor `TokenSecret` are meant to be used directly. Let's go with this:
```python
Secret.from_evn_var(["env1", "env2"], strict=False)
```

integrations/amazon_bedrock/src/haystack_integrations/components/generators/amazon_bedrock/generator.py
```diff
@@ -72,11 +73,21 @@ class AmazonBedrockGenerator:
...
+        aws_profile_name: Optional[str] = EnvVarSecret(env_vars=["AWS_PROFILE"], strict=False).from_env_var(
```
shadeMe, line 88: Type annotation needs to be updated.

shadeMe (changes requested):

integrations/amazon_bedrock/src/haystack_integrations/components/generators/amazon_bedrock/generator.py
davidsbatista: thanks 👍🏽

integrations/amazon_bedrock/src/haystack_integrations/components/generators/amazon_bedrock/chat/chat_generator.py
```diff
@@ -61,11 +62,13 @@ class AmazonBedrockChatGenerator:
...
-        aws_session_token: Optional[str] = None,
-        aws_region_name: Optional[str] = None,
-        aws_profile_name: Optional[str] = None,
+        aws_access_key_id: Optional[Secret] = Secret.from_env_var(["AWS_ACCESS_KEY_ID"], strict=False),  # noqa: B008
+        aws_secret_access_key: Optional[Secret] = Secret.from_env_var(  # noqa: B008
+            ["AWS_SECRET_ACCESS_KEY"], strict=False
+        ),
+        aws_session_token: Optional[Secret] = Secret.from_env_var(["AWS_SESSION_TOKEN"], strict=False),  # noqa: B008
+        aws_region_name: Optional[Secret] = Secret.from_env_var(["AWS_DEFAULT_REGION"], strict=False),  # noqa: B008
+        aws_profile_name: Optional[Secret] = Secret.from_env_var(["AWS_PROFILE"], strict=False),  # noqa: B008
```
shadeMe, line 71: Single env vars can be passed as strings:
```python
Secret.from_env_var("AWS_ACCESS_KEY")
```

integrations/amazon_bedrock/tests/test_amazon_chat_bedrock.py
```diff
@@ -38,17 +38,12 @@ def mock_prompt_handler():
...
     """
     Test that the to_dict method returns the correct dictionary without aws credentials
     """
     generator = AmazonBedrockChatGenerator(
         model="anthropic.claude-v2",
-        aws_access_key_id="some_fake_id",
-        aws_secret_access_key="some_fake_key",
-        aws_session_token="some_fake_token",
-        aws_profile_name="some_fake_profile",
-        aws_region_name="fake_region",
```
shadeMe, line 51: Let's actually check for these in the expected dict instead of setting them to `None`. Same with the `from_dict` test.

shadeMe (changes requested):
Looks good! Just a couple of nitpicks.

integrations/amazon_bedrock/src/haystack_integrations/components/generators/amazon_bedrock/chat/chat_generator.py
davidsbatista: 👍🏽

integrations/amazon_bedrock/tests/test_amazon_chat_bedrock.py
davidsbatista: the `to_dict()` method doesn't save any of the AWS env variables/Secret - should it?

shadeMe: Ah, I didn't notice that! Yeah, it definitely should - consider a situation where the user passes their own secret (with a different env var). If we don't save it (and load it again), the reconstructed component will try to read from the default env var, which might or might not be populated.

C.f https://github.com/deepset-ai/haystack-core-integrations/pull/386/files

davidsbatista: OK, I still need to make one more change then: store the variables names in the class instance as parameters. 

Sorry for this confusion. I just assumed that for core-integration components that rely on AWS, nothing is ever stored since everything is standardized (i.e.: en variables names, config files). I would just use the `Secret` component to read the environmental variables and set up the connection, being that the only scope of the variables.

But I will make it look like all the other components.

shadeMe (approved):

Issue #365, opened 2024-02-08
Update core integration secret managment - Instructor Embedders


shadeMe:
- [x] Code (c.f https://github.com/deepset-ai/haystack-core-integrations/pull/347)
- [x] Documentation
- [x] Tutorials
- [x] Cookbooks
- [x] Articles
- [x] Integration Tiles

PR #417, opened 2024-02-14
instructor - new secret management

anakin87:
part of #365

I followed the same approach used for Sentence Transformers embedders in the Haystack repo


davidsbatista (approved):

integrations/instructor_embedders/src/haystack_integrations/components/embedders/instructor_embedders/instructor_document_embedder.py
```diff
@@ -62,7 +63,7 @@ def __init__(
         self,
         model: str = "hkunlp/instructor-base",
         device: Optional[str] = None,
-        use_auth_token: Union[bool, str, None] = None,
+        token: Optional[Secret] = Secret.from_env_var("HF_API_TOKEN", strict=False),  # noqa: B008
```
davidsbatista, line 66: that's actually a better idea, to have those `ignore` statements inline, instead of globally defined 👍🏽

davidsbatista (approved):
LGTM

PR #419, opened 2024-02-15
Choere - remove matching error message from tests

anakin87:
- generally, we don't want to match the error message
- it causes errors like [this](https://github.com/deepset-ai/haystack-core-integrations/actions/runs/7909096256/job/21589539454): the error message of the Cohere API can easily change over time

(I also removed the match clause from other tests)


masci (approved):
🔥

silvanocerza (approved):

Issue #373, opened 2024-02-08
Update core integration secret managment - Astra


shadeMe:
- [ ] Code (c.f https://github.com/deepset-ai/haystack-core-integrations/pull/347)
- [ ] Documentation
- [ ] Tutorials
- [ ] Cookbooks
- [ ] Articles
- [ ] Integration Tiles

PR #423, opened 2024-02-15
[Astra] Change authentication parameters

masci:
Part of #373 this PR also makes the document store constructor more practical for datastax users


julian-risch (approved):
LGTM! 👍 We will need to update the documentation page https://haystack.deepset.ai/integrations/astradb and use the updated example code there too.

alexleventer:
Thanks! 👍

PR #425, opened 2024-02-15
Update WeaviateDocumentStore authentication to use new Secret class

### LINKED ISSUE #369: Update core integration secret managment - Weaviate
- [ ] Code (c.f https://github.com/deepset-ai/haystack-core-integrations/pull/347)
- [x] Documentation
- [x] Tutorials
- [x] Cookbooks
- [x] Articles
- [x] Integration Tiles

silvanocerza:
Fixes #369.

Weaviate supports different manner of authentication via some utility dataclasses. 
Cause of this it wasn't feasible to use an init parameterer for each possible secret, so we mirrored a bit their utility classes structures to use the `Secret` class.

You'll notice that I used a `default_factory` in all dataclasses fields cause we can't use `default` as `Secret` is not a frozen dataclass in the latest beta. When that changes we should be able to use `default`.

### DIFF
--- a/integrations/weaviate/src/haystack_integrations/document_stores/weaviate/auth.py
+++ b/integrations/weaviate/src/haystack_integrations/document_stores/weaviate/auth.py
@@ -1,5 +1,6 @@
 from abc import ABC, abstractmethod
 from dataclasses import dataclass, field, fields
+from enum import Enum
 from typing import Any, Dict, Type
 
 from haystack.core.errors import DeserializationError
@@ -10,6 +11,30 @@ from weaviate.auth import AuthApiKey as WeaviateAuthApiKey
 from weaviate.auth import AuthBearerToken as WeaviateAuthBearerToken
 from weaviate.auth import AuthClientCredentials as WeaviateAuthClientCredentials
 from weaviate.auth import AuthClientPassword as WeaviateAuthClientPassword
+
+
+class SupportedAuthTypes(Enum):
+    """
+    Supported auth credentials for WeaviateDocumentStore.
+    """
+
+    API_KEY = "api_key"
+    BEARER = "bearer"
+    CLIENT_CREDENTIALS = "client_credentials"
+    CLIENT_PASSWORD = "client_password"
+
+    def __str__(self):
+        return self.value
+
+    @staticmethod
+    def from_class(auth_class) -> "SupportedAuthTypes":
+        auth_types = {
+            AuthApiKey: SupportedAuthTypes.API_KEY,
+            AuthBearerToken: SupportedAuthTypes.BEARER,
+            AuthClientCredentials: SupportedAuthTypes.CLIENT_CREDENTIALS,
+            AuthClientPassword: SupportedAuthTypes.CLIENT_PASSWORD,
+        }
+        return auth_types[auth_class]
 
 
 @dataclass
@@ -30,10 +55,7 @@ class AuthCredentials(ABC):
             else:
                 _fields[field.name] = getattr(self, field.name)
 
-        return default_to_dict(
-            self,
-            **_fields,
-        )
+        return {"type": str(SupportedAuthTypes.from_class(self.__class__)), "init_parameters": _fields}
 
     @staticmethod
     def from_dict(data: Dict[str, Any]) -> "AuthCredentials":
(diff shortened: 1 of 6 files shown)


integrations/weaviate/src/haystack_integrations/document_stores/weaviate/auth.py
```diff
@@ -0,0 +1,170 @@
...
+    AuthCredentials for Bearer token authentication.
+    By default it will load `access_token` from the environment variable `WEAVIATE_ACCESS_TOKEN`,
+    `expires_in` from the environment variable `WEAVIATE_EXPIRES_IN`, and `refresh_token` from the environment variable
+    `WEAVIATE_REFRESH_TOKEN`.
+    `WEAVIATE_EXPIRES_IN` environment variable is optional, if set must be an integer.
+    `WEAVIATE_REFRESH_TOKEN` environment variable is optional.
+    """
+
+    access_token: Secret = field(default_factory=lambda: Secret.from_env_var(["WEAVIATE_ACCESS_TOKEN"]))
+    expires_in: Secret = field(default_factory=lambda: Secret.from_env_var(["WEAVIATE_EXPIRES_IN"], strict=False))
```
shadeMe, line 90: This doesn't look like a secret. Let's just make it a regular int that defaults to 60.

integrations/weaviate/src/haystack_integrations/document_stores/weaviate/auth.py
```diff
@@ -0,0 +1,170 @@
...
+
+    @abstractmethod
+    def resolve_value(self):
+        """
+        Resolves all the secrets in the auth credentials object and returns the corresponding Weaviate object.
+        All subclasses must implement this method.
+        """
+
+
+@dataclass
```
shadeMe, line 60: Let's freeze all dataclasses.

integrations/weaviate/src/haystack_integrations/document_stores/weaviate/auth.py
```diff
@@ -0,0 +1,170 @@
...
+
+    @staticmethod
+    def from_dict(data: Dict[str, Any]) -> "AuthCredentials":
+        """
+        Converts a dictionary representation to an auth credentials object.
+        """
+        if "type" not in data:
+            msg = "Missing 'type' in serialization data"
+            raise DeserializationError(msg)
+        return _AUTH_CLASSES[data["type"]]._from_dict(data)
```
shadeMe, line 42: Let's just move the map into the function.

integrations/weaviate/src/haystack_integrations/document_stores/weaviate/document_store.py
```diff
@@ -124,7 +116,7 @@ def __init__(
         """
         self._client = weaviate.Client(
             url=url,
-            auth_client_secret=auth_client_secret,
+            auth_client_secret=auth_client_secret.resolve_value() if auth_client_secret else None,
```
shadeMe, line 119: Does it actually make sense to make the credentials optional? Does Weaviate offer any kind of unauthenticated access?

integrations/weaviate/src/haystack_integrations/document_stores/weaviate/auth.py
```diff
@@ -0,0 +1,170 @@
...
+    """
+    Base class for all auth credentials supported by WeaviateDocumentStore.
+    Can be used to deserialize from dict any of the supported auth credentials.
+    """
+
+    def to_dict(self) -> Dict[str, Any]:
+        """
+        Converts the object to a dictionary representation for serialization.
+        """
+        # We assume all fields are Secret instances
```
shadeMe, line 26: We should check instead of assuming, especially if the `expires_in` field is going to be an `int` (see below.)

integrations/weaviate/src/haystack_integrations/document_stores/weaviate/auth.py
```diff
@@ -0,0 +1,170 @@
...
+        _fields = {f.name: getattr(self, f.name).to_dict() for f in fields(self)}
+
+        return default_to_dict(
+            self,
+            **_fields,
+        )
```
shadeMe, line 32: I'd be more comfortable if we used an explicit enum instead of the full type name. That way, it won't break if we move/rename the underlying classes.

shadeMe (changes requested):

integrations/weaviate/src/haystack_integrations/document_stores/weaviate/document_store.py
silvanocerza: Yep, if you check out the tests we do exactly that. Obviously not recommended for production use but it's possible to skip authentication.

shadeMe (approved):

PR #428, opened 2024-02-16
Update the Astra DB Integration to fit latest conventions

erichare:
This PR updates the Astra DB integration by depending on `astrapy`, a client library that abstracts away components of the Data API. It also cleans up the naming conventions for environment variables to match the standard Astra convention.


.github/workflows/astra.yml
```diff
@@ -59,6 +59,6 @@ jobs:
 
     - name: Run tests
       env:
-        ASTRA_API_ENDPOINT: ${{ secrets.ASTRA_API_ENDPOINT }}
-        ASTRA_TOKEN: ${{ secrets.ASTRA_TOKEN }}
+        ASTRA_DB_API_ENDPOINT: ${{ secrets.ASTRA_DB_API_ENDPOINT }}
+        ASTRA_DB_APPLICATION_TOKEN: ${{ secrets.ASTRA_DB_APPLICATION_TOKEN }}
```
masci, line 63: Let's keep the old secrets name for now, it's not user-facing and it'll spare setting new secrets for now

```suggestion
        ASTRA_DB_API_ENDPOINT: ${{ secrets.ASTRA_API_ENDPOINT }}
        ASTRA_DB_APPLICATION_TOKEN: ${{ secrets.ASTRA_TOKEN }}
```

integrations/astra/pyproject.toml
```diff
@@ -51,6 +52,7 @@ dependencies = [
   "coverage[toml]>=6.5",
   "pytest",
   "haystack-pydoc-tools",
+  "astrapy>=0.7.0",
```
masci, line 55: The jobs depends on the integration already, that depends on `astra`, no need for the extra dep here
```suggestion
```

integrations/astra/pyproject.toml
```diff
@@ -75,6 +77,7 @@ dependencies = [
   "black>=23.1.0",
   "mypy>=1.0.0",
   "ruff>=0.0.243",
+  "astrapy>=0.7.0",
```
masci, line 80: The jobs depends on the integration already, that depends on `astra`, no need for the extra dep here
```suggestion
```

masci (changes requested):
Thanks for this PR! I left a couple of suggestions related to package definition and we can merge

.github/workflows/astra.yml
erichare: Great point, that was a silly oversight on my part! Yeah, since these are just repo secrets absolutely no reason they need to match. updating.

erichare:
Thanks so much @masci . I think I addressed your comments. Let me know if there's anything else you see!

masci (approved):

PR #413, opened 2024-02-14
feat: MongoDBAtlas Document Store

ZanSara:
Related to #143 

Add support for MongoDB Atlas in the form of `MongoDBAtlasDocumentStore`.

Note that:
- Filters are not implemented yet
- No retrievers are implemented yet

The missing features will be added in later PRs.

PR #430, opened 2024-02-16
fix: remove filters from `MongoDBAtlasDocumentStore.count()` method

ZanSara:
Fixes https://github.com/deepset-ai/haystack-core-integrations/pull/413#discussion_r1491426418

Removed the `filters` parameter from the `MongoDBAtlasDocumentStore.count()` method.


davidsbatista (approved):
LGTM

PR #432, opened 2024-02-16
Adopt Secret to AmazonSagemaker

davidsbatista:
The SagemakerGenerator now uses `Secret` to manage credentials

`pyproject.toml`:
- fixed test coverage
- fixed mypy issues with missing overrides
- added missing pytest markers

`tests`:
- updated and refactored the tests
- tried to follow a similar approach as in [test_amazon_bedrock.py](https://github.com/deepset-ai/haystack-core-integrations/blob/main/integrations/amazon_bedrock/tests/test_amazon_bedrock.py):


integrations/amazon_sagemaker/src/haystack_integrations/components/generators/amazon_sagemaker/sagemaker.py
```diff
@@ -138,27 +162,47 @@ def from_dict(cls, data) -> "SagemakerGenerator":
...
         """
+        deserialize_secrets_inplace(
+            data["init_parameters"],
+            ["aws_access_key_id", "aws_secret_access_key", "aws_session_token", "aws_region_name", "aws_profile_name"],
+        )
         return default_from_dict(cls, data)
 
-    def warm_up(self):
+    @classmethod
+    def get_aws_session(
```
silvanocerza, line 172: I would change this to `staticmethod` since we don't really need the class in this case.
Also let's rename it to `_get_aws_session`, this should be part of the private API.

integrations/amazon_sagemaker/src/haystack_integrations/components/generators/amazon_sagemaker/sagemaker.py
```diff
@@ -138,27 +162,47 @@ def from_dict(cls, data) -> "SagemakerGenerator":
...
+        cls,
+        aws_access_key_id: Optional[str] = None,
+        aws_secret_access_key: Optional[str] = None,
+        aws_session_token: Optional[str] = None,
+        aws_region_name: Optional[str] = None,
+        aws_profile_name: Optional[str] = None,
+        **kwargs,
```
silvanocerza, line 179: Why do we need `kwargs` here? 🤔

integrations/amazon_sagemaker/src/haystack_integrations/components/generators/amazon_sagemaker/sagemaker.py
```diff
@@ -138,27 +162,47 @@ def from_dict(cls, data) -> "SagemakerGenerator":
...
-                f"Make sure the Endpoint exists and AWS environment is configured."
+            return boto3.Session(
+                aws_access_key_id=aws_access_key_id,
+                aws_secret_access_key=aws_secret_access_key,
+                aws_session_token=aws_session_token,
+                region_name=aws_region_name,
+                profile_name=aws_profile_name,
             )
+        except BotoCoreError as e:
+            provided_aws_config = {k: v for k, v in kwargs.items() if k in AWS_CONFIGURATION_KEYS}
```
silvanocerza, line 204: `kwargs` will never contain any of the fields in `AWS_CONFIGURATION_KEYS` as they're all explicit in `get_aws_session`.

This should be changed.

integrations/amazon_sagemaker/src/haystack_integrations/components/generators/amazon_sagemaker/sagemaker.py
```diff
@@ -59,25 +67,30 @@ class SagemakerGenerator:
...
-        :param aws_secret_access_key_var: The name of the env var where the AWS secret access key is stored.
-        :param aws_session_token_var: The name of the env var where the AWS session token is stored.
-        :param aws_region_name_var: The name of the env var where the AWS region name is stored.
-        :param aws_profile_name_var: The name of the env var where the AWS profile name is stored.
+
+        :param aws_access_key_id: The name of the env var where the AWS access key ID is stored.
+        :param aws_secret_access_key: The name of the env var where the AWS secret access key is stored.
+        :param aws_session_token: The name of the env var where the AWS session token is stored.
+        :param aws_region_name: The name of the env var where the AWS region name is stored.
+        :param aws_profile_name: The name of the env var where the AWS profile name is stored.
```
silvanocerza, line 89: We should update these, they're not the name of env vars anymore.

integrations/amazon_sagemaker/src/haystack_integrations/components/generators/amazon_sagemaker/sagemaker.py
```diff
@@ -36,14 +42,16 @@ class SagemakerGenerator:
     ```bash
     export AWS_ACCESS_KEY_ID=<your_access_key_id>
     export AWS_SECRET_ACCESS_KEY=<your_secret_access_key>
+
+    export AWS_SECRET_ACCESS_KEY=<your_secret_access_key>
+
```
silvanocerza, line 47: A bit nitpicky but I don't the need for those lines really. 😅 

```suggestion
    export AWS_SECRET_ACCESS_KEY=<your_secret_access_key>
```

integrations/amazon_sagemaker/src/haystack_integrations/components/generators/amazon_sagemaker/sagemaker.py
```diff
@@ -1,24 +1,30 @@
...
 
 logger = logging.getLogger(__name__)
 
+AWS_CONFIGURATION_KEYS = [
+    "aws_access_key_id",
+    "aws_secret_access_key",
+    "aws_session_token",
+    "aws_region_name",
+    "aws_profile_name",
+]
```
silvanocerza, line 26: Would be better if we could do without this constant, if we want to print them in the error message in `get_aws_session` we should reference the arguments explicitly.

integrations/amazon_sagemaker/tests/test_sagemaker.py
```diff
@@ -1,243 +1,175 @@
...
+        "aws_session_token": {"type": "env_var", "env_vars": ["AWS_SESSION_TOKEN"], "strict": False},
+        "aws_region_name": {"type": "env_var", "env_vars": ["AWS_DEFAULT_REGION"], "strict": False},
+        "aws_profile_name": {"type": "env_var", "env_vars": ["AWS_PROFILE"], "strict": False},
+        "aws_custom_attributes": {"accept_eula": True},
+        "generation_kwargs": {"max_new_tokens": 10},
+    },
+}
+
+
+@pytest.mark.unit
```
silvanocerza, line 25: No need to mark tests as `unit`, we only marm `integration` ones.

integrations/amazon_sagemaker/tests/test_sagemaker.py
```diff
@@ -1,243 +1,175 @@
...
+        "model": "model",
+        "aws_access_key_id": {"type": "env_var", "env_vars": ["AWS_ACCESS_KEY_ID"], "strict": False},
+        "aws_secret_access_key": {"type": "env_var", "env_vars": ["AWS_SECRET_ACCESS_KEY"], "strict": False},
+        "aws_session_token": {"type": "env_var", "env_vars": ["AWS_SESSION_TOKEN"], "strict": False},
+        "aws_region_name": {"type": "env_var", "env_vars": ["AWS_DEFAULT_REGION"], "strict": False},
+        "aws_profile_name": {"type": "env_var", "env_vars": ["AWS_PROFILE"], "strict": False},
+        "aws_custom_attributes": {"accept_eula": True},
+        "generation_kwargs": {"max_new_tokens": 10},
+    },
+}
```
silvanocerza, line 22: I would put this directly in the tests that use it, even if means duplicating it. 
Makes it easier to maintain and reason about.

As an example see this 👉  https://github.com/deepset-ai/haystack-core-integrations/blob/d28ffa1b7ab4a6e8363da2644996b70150145faa/integrations/google_ai/tests/generators/test_gemini.py#L97-L117

integrations/amazon_sagemaker/tests/test_sagemaker.py
```diff
@@ -1,243 +1,175 @@
...
+
+
+@pytest.mark.unit
+@pytest.mark.usefixtures("set_env_variables", "mock_boto3_session")
```
silvanocerza, line 26: I see no benefits in using `usefixtures` in this case. If we wrap all the tests that need them in a class and decorate the class I'd be ok with that.
Otherwise just them as parameters.

silvanocerza (changes requested):
Left some comments.

integrations/amazon_sagemaker/tests/test_sagemaker.py
davidsbatista: for some weird reason, having them as parameters fails - some config issue, I suspect, I will look into it

integrations/amazon_sagemaker/src/haystack_integrations/components/generators/amazon_sagemaker/sagemaker.py
davidsbatista: I will just remove it since it is omitted, or the user must pass a `Secret` - this is confusing.

I already stated in the example that the AWS credentials should be set up as environment variables or defined in a credentials file (which is then read by boto3)- should be clear enough, no?

silvanocerza: Maybe instead of removing it we could go with something like ```The `Secret` for AWS access key ID```?

davidsbatista: 👍🏽

silvanocerza (approved):
Nice 👍

PR #436, opened 2024-02-16
Llama.cpp: change wrong links and imports

anakin87:
- README inventory had wrong links and was not alphabetically sorted
- llama.cpp README had wrong imports


julian-risch (approved):
LGTM! 👍 🙌

PR #438, opened 2024-02-18
Fix Readme sync

masci:
- The env variable seemed to mess up with the API key contents, now passing the secret directly to `--key`, as per Readme's offical docs
- Also pin the action less aggressively, so we can pull updates more frequently (dependabot was late and made us make a big version jump)

Test workflow passed with this new configuration: https://github.com/deepset-ai/haystack-core-integrations/actions/runs/7948402876/job/21698363580


anakin87 (approved):
:raised_hands:

PR #439, opened 2024-02-19
Amazon Sagemaker - create API docs

anakin87:
From a [recent failed workflow](https://github.com/deepset-ai/haystack-core-integrations/actions/runs/7957150243/job/21719290510), I noticed that the API docs machinery was missing for Amazon Sagemaker integration.

I'm adding it.


davidsbatista (approved):
LGTM - thanks for fixing it @anakin87 👍🏽

PR #440, opened 2024-02-19
Feat: Add filters to run function in retrievers of elasticsearch

sebastian-weisshaar:
The retrievers of elasticsearch only accept filters on initialization. They should also accept filters as argument to the `run` function.


anakin87 (commented):
Hey! Thanks for the contribution.

The PR looks good.
Please also add `filters` to the docstring. [You can take Pgvector as an example](https://github.com/deepset-ai/haystack-core-integrations/blob/df0090f501897ddc99b3295866795c87600750db/integrations/pgvector/src/haystack_integrations/components/retrievers/pgvector/embedding_retriever.py#L32).

anakin87 (approved):
:raised_hands:

PR #445, opened 2024-02-20
Allow passing in required project_name

lbux:
UpTrain requires a project_name when using its API whereas openai does not. The integration will crash if no project_name is included.

This will allow for an eval declaration like so:

evaluator = UpTrainEvaluator(
    metric=UpTrainMetric.METRIC,
    api="uptrain",
    project_name="uptrain-project",
)

This was the fix I found that resolved my issue. It might not be a full fix as I have not read too deep into the source code.

### DIFF
--- a/integrations/uptrain/src/haystack_integrations/components/evaluators/uptrain/evaluator.py
+++ b/integrations/uptrain/src/haystack_integrations/components/evaluators/uptrain/evaluator.py
@@ -36,6 +36,7 @@ class UpTrainEvaluator:
         api: str = "openai",
         api_key: Secret = Secret.from_env_var("OPENAI_API_KEY"),
         api_params: Optional[Dict[str, Any]] = None,
+        project_name: Optional[str] = None,
     ):
         """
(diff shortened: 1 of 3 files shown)


masci (approved):
Thanks for the PR! Patch looks like a solid fix to me.

integrations/uptrain/src/haystack_integrations/components/evaluators/uptrain/evaluator.py
```diff
@@ -193,6 +197,9 @@ def _init_backend(self):
         if self.api == "openai":
             backend_client = EvalLLM(openai_api_key=api_key)
         elif self.api == "uptrain":
+            if not self.project_name:
+                msg = f"project_name not provided. UpTrain API requires a project name."
```
masci, line 201:
```suggestion
                msg = "project_name not provided. UpTrain API requires a project name."
```

masci:
FTR https://docs.uptrain.ai/tutorials/uptrain-api-client#step-2-run-the-evaluation

PR #447, opened 2024-02-20
fix order of API docs

dfokina:
This PR will also push the docs to Readme


silvanocerza:
All docs jobs are green. 👍 

Mongo tests are failing as expected. Astra now is failing too for some reason. 😢

dfokina:
@silvanocerza shall we wait for the astra fix, or can approve & merge this?

silvanocerza:
Should be fine merging without waiting for the fine. I'll handle that.

PR #449, opened 2024-02-20
replace token api str with Secret

davidsbatista:
- fixed coverage configuration in `pyproject.toml`
- `api_key` is now a `Optional[Secret]`
- updated tests
- fixed typos + convert method to static

### DIFF
--- a/integrations/unstructured/tests/test_converter.py
+++ b/integrations/unstructured/tests/test_converter.py
@@ -2,7 +2,6 @@
 #
 # SPDX-License-Identifier: Apache-2.0
 import pytest
-from haystack.utils import Secret
 from haystack_integrations.components.converters.unstructured import UnstructuredFileConverter
 
 


integrations/unstructured/src/haystack_integrations/components/converters/unstructured/converter.py
```diff
@@ -29,7 +30,7 @@ class UnstructuredFileConverter:
     def __init__(
         self,
         api_url: str = UNSTRUCTURED_HOSTED_API_URL,
-        api_key: Optional[str] = None,
+        api_key: Optional[Secret] = Secret.from_env_var("UNSTRUCTURED_API_KEY"),  # noqa: B008
```
anakin87, line 33:
```suggestion
        api_key: Optional[Secret] = Secret.from_env_var("UNSTRUCTURED_API_KEY", strict=False),  # noqa: B008
```

To make the Secret optional, you can adopt this approach, as done for example in the [`TransformersSimilarityRanker`](https://github.com/deepset-ai/haystack/blob/327c2d260d19ced93fd50e78b63b1465d089c597/haystack/components/rankers/transformers_similarity.py#L45). When you resolve an env var which is not set, you get None. With `strict=True` (default), you get an error.

You can also follow a different approach, as shown [in the docs](https://docs.haystack.deepset.ai/v2.0/docs/secret-management#example).

integrations/unstructured/src/haystack_integrations/components/converters/unstructured/converter.py
```diff
@@ -64,12 +65,11 @@ def __init__(
 
         is_hosted_api = api_url == UNSTRUCTURED_HOSTED_API_URL
 
-        api_key = api_key or os.environ.get("UNSTRUCTURED_API_KEY")
         # we check whether api_key is None or an empty string
         if is_hosted_api and not api_key:
```
anakin87, line 69: This line should be changed to actually test if the `api_key` is `None`
If you follow the previous suggestion, you can use `resolve_value`.

anakin87 (changes requested):
I requested some changes.

Please also use the resolved value of the `api_key` in the ` _partition_file_into_elements` method.

integrations/unstructured/src/haystack_integrations/components/converters/unstructured/converter.py
davidsbatista: upps, this API KEY is not mandatory, I missed that - thanks 👍🏽

integrations/unstructured/src/haystack_integrations/components/converters/unstructured/converter.py
davidsbatista: yes, you are right, thanks 👍🏽

anakin87 (approved):
LGTM!

PR #455, opened 2024-02-21
Bedrock refactoring

anakin87:
Move errors and some utility functions to a new package: `haystack_integrations.common.amazon_bedrock`.

These classes and functions are already used by both Generators and will also be used in the Embedders.

I also cleaned up more unused stuff...


masci (approved):
Thanks!

PR #456, opened 2024-02-21
Bedrock - remove `supports` method

anakin87:
This method and the related tests were ported from Haystack 1.x.
The method seems to be unused.


masci (approved):
🔥

PR #457, opened 2024-02-21
Elasticsearch: add user-agent header

maxjakob:
[Elastic](https://www.elastic.co/) wants to provide a great developer experience with 3rd party integrations. In order to be able to prioritize maintenance and feature development efforts, we keep track of which integrations are popular by setting the `user-agent` header. We can then see usage statistics in the [Elastic Cloud](https://cloud.elastic.co/) metrics.

We want to gauge the usage of the Haystack integration and hence add a `user-agent` header here.

### DIFF
--- a/integrations/elasticsearch/src/haystack_integrations/document_stores/elasticsearch/document_store.py
+++ b/integrations/elasticsearch/src/haystack_integrations/document_stores/elasticsearch/document_store.py
@@ -13,6 +13,7 @@ from haystack.dataclasses import Document
 from haystack.document_stores.errors import DocumentStoreError, DuplicateDocumentError
 from haystack.document_stores.types import DuplicatePolicy
 from haystack.utils.filters import convert
+from haystack.version import __version__ as haystack_version
 
 from elasticsearch import Elasticsearch, helpers  # type: ignore[import-not-found]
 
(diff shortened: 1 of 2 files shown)


silvanocerza (approved):
Nice, thanks Max. Would be cool to have some insights in the future. :)

serenachou:
@maxjakob what kind of usage are we seeing here ?

maxjakob:
@serenachou Some usage, not too much. LlamaIndex usage peaked at 7 clusters in January, now it's about twice as popular as Haystack:
image
https://stack-telemetry.elastic.dev/app/dashboards#/view/9c6b0fe0-6dbe-11ee-9785-813562888be7

maxjakob:
@serenachou We see some usage in our Elastic Cloud metrics, around handful in the two weeks. This might still go up if people haven't updated to the latest version yet.

serenachou:
@maxjakob  what would you say is a meaningful threshold of usage where we would want to take action? Can we instrument an alert to remind us to return to Haystack on that in the telemetry cluster?

maxjakob:
@serenachou Great idea. I have set up an email alert for when Cloud usage goes beyond 5 clusters so we can have a deeper look at the usage patterns.

Issue #415, opened 2024-02-14
chore: Update all generator integrations to use new generic callable (de)serializers for their callback handlers


vblagoje:
We have migrated all the (de)serialize callback handlers in Haystack 2.x to use the new generic utils methods serialize_callable and deserialize_callable. Now, we need to do that for all generators in [haystack-core-integrations](https://github.com/deepset-ai/haystack-core-integrations). The best time to migrate these will likely be after the next Haystack beta is released.

masci:
@ZanSara I think the two linked PRs do something different - this issue was open to keep track of the changes needed after replacing `haystack.components.generators.utils.serialize_callback_handler` in `main` with `haystack.utils.callable_serialization.serialize_callable`. The course of action I imagine would involve a new beta release to ship `serialize_callback_handler` and change the integrations (or merging your current PRs but change the integration again later after we release the change). In that sense, you should add `ollama` to the list of integrations that we need to change.

PR #460, opened 2024-02-21
chore: Use `serialize_callable` instead of `serialize_callback_handler` in Cohere 

ZanSara:
Related to https://github.com/deepset-ai/haystack-core-integrations/issues/415

NOTE: requires a new beta release for the tests to pass. Don't merge!


davidsbatista:
not sure if I'm seeing the whole thing here, but, instead of a new beta release, wouldn't the tests pass with this [issue fixed](https://github.com/deepset-ai/haystack-core-integrations/issues/435), i.e: test `core-integrations` against the `main` branch of haystack core?

ZanSara:
@davidsbatista that would be a nightly workflow that prevents the opposite scenario: a Haystack release that breaks existing integrations. 

I agree that the whole workflow needs work, it would make sense that the main of integrations work on the main of Haystack, but for now it's safer to test everything primarily against the released Haystack.

davidsbatista (approved):

PR #463, opened 2024-02-22
weaviate: migrate from weaviate python client v3  to v4

### LINKED ISSUE #458: Migrate Weaviate python client from v3 to v4
The v4 python client uses GRPC so it is faster when ingesting and querying data. See [this](https://weaviate.io/blog/py-client-v4-release#speed-improvements) blog post.

Upgrade the weaviate-client-python dependency to > 4.0 and rewrite the client calls to use the new API.  

Not migrating

**Additional context**
* Weaviate v3 to v4 [migration guide](https://weaviate.io/developers/weaviate/client-libraries/python/v3_v4_migration)

hsm207:
Fixes #458


integrations/weaviate/src/haystack_integrations/document_stores/weaviate/document_store.py
```diff
@@ -238,72 +233,42 @@ def _to_document(self, data: Dict[str, Any]) -> Document:
...
+        return list(result)
+
+    def _query_with_filters(self, filters: weaviate.collections.classes.filters.Filter) -> List[Dict[str, Any]]:
+
+        try:
+            # this is the default value for max number of objects to retrieve in Weaviate
+            # see QUERY_MAXIMUM_RESULTS
+            # see https://weaviate.io/developers/weaviate/config-refs/env-vars#overview
+            # and https://weaviate.io/developers/weaviate/api/graphql/additional-operators#pagination-with-offset
+            limit = 10_000
```
silvanocerza, line 252: Is this actually the max possibile value in Weaviate or just the default? 
Cause if it's the default it would be better to not hardcode this and use the actual value.

integrations/weaviate/src/haystack_integrations/document_stores/weaviate/document_store.py
```diff
@@ -238,72 +233,42 @@ def _to_document(self, data: Dict[str, Any]) -> Document:
...
+            result = self._collection.query.fetch_objects(
+                filters=convert_filters(filters), include_vector=True, limit=limit
             )
-            .with_additional(["id vector"])
-            .with_limit(100)
-        )
+        except weaviate.exceptions.WeaviateQueryError as e:
+            msg = f"Failed to query documents in Weaviate. Error: {e.message}"
+            raise DocumentStoreError(msg) from None
```
silvanocerza, line 258: Shouldn't we reraise?
```suggestion
            raise DocumentStoreError(msg) from e
```

integrations/weaviate/src/haystack_integrations/document_stores/weaviate/document_store.py
hsm207: It is is the default value in weaviate.

To change it the user would need to specify it when setting up their weaviate instance and the client can't query this value from the server.

integrations/weaviate/src/haystack_integrations/document_stores/weaviate/document_store.py
```diff
@@ -216,6 +193,24 @@ def _to_data_object(self, document: Document) -> Dict[str, Any]:
 
         return data
 
+    def _convert_weaviate_v4_object_to_v3_object(self, data: Object) -> Dict[str, Any]:
```
silvanocerza, line 196: What's the need of this function? Can't we do it this in `_to_document()`?

integrations/weaviate/src/haystack_integrations/document_stores/weaviate/document_store.py
hsm207: I felt the traceback from WeaviateQueryError is an implementation detail that the user does not need to know, since the problem would already be clear from `msg`

silvanocerza: I would leave it, I think it's useful in any case.

integrations/weaviate/tests/test_document_store.py
```diff
@@ -145,67 +146,80 @@ def assert_documents_are_equal(self, received: List[Document], expected: List[Do
...
-        mock_client.schema.create_class.assert_called_once_with(
+        # mock_client.schema.get.assert_called_once()
+        # mock_client.schema.exists.assert_called_once_with("My_collection")
+        mock_client.collections.exists.assert_called_once_with("My_collection")
+        mock_client.collections.create_from_dict.assert_called_once_with(
             {"class": "My_collection", "properties": DOCUMENT_COLLECTION_PROPERTIES}
         )
+        # mock_client.schema.create_class.assert_called_once_with(
+        #     {"class": "My_collection", "properties": DOCUMENT_COLLECTION_PROPERTIES}
+        # )
```
silvanocerza, line 202: Should the commented lines be removed?

integrations/weaviate/tests/test_document_store.py
```diff
@@ -307,21 +318,21 @@ def test_from_dict(self, _mock_weaviate, monkeypatch):
...
         assert document_store._embedded_options.binary_path == DEFAULT_BINARY_PATH
         assert document_store._embedded_options.version == "1.23.0"
         assert document_store._embedded_options.port == DEFAULT_PORT
         assert document_store._embedded_options.hostname == "127.0.0.1"
         assert document_store._embedded_options.additional_env_vars is None
         assert document_store._embedded_options.grpc_port == DEFAULT_GRPC_PORT
-        assert document_store._additional_config.grpc_port_experimental == 12345
-        assert document_store._additional_config.connection_config.session_pool_connections == 20
-        assert document_store._additional_config.connection_config.session_pool_maxsize == 20
+        # assert document_store._additional_config.grpc_port_experimental == 12345
```
silvanocerza, line 333: Other commented lines to remove.

integrations/weaviate/src/haystack_integrations/document_stores/weaviate/document_store.py
hsm207: `_to_document()` assumes the result from a weaviate query is a dict. However, in v4, it is now an `Object`, so I wrote this function to do the conversion instead of modifying  `_to_document()`. But we could certainly do it in `_to_document()`.

Do you prefer the conversion to be done in `_to_document()` or update the sig so that `_to_document()` accepts `Object`? I'm not sure what impact the latter would have on the rest of the integration.

silvanocerza: Feel free to change it's signature if necessary, it's only meant to be used in the internals.

Ideally `_to_document()` should take whatever is returned by the Weaviate client and return a `Document`. If in v4 it's an `Object` that should be what it expects.

integrations/weaviate/src/haystack_integrations/document_stores/weaviate/document_store.py
hsm207: I agree.

Then should we introduce another param to the init function or an env var to control this value?

silvanocerza: Can't we use `limit` and `offset` similar to the previous implementation? 
We can decide on a sane default for `limit`.

integrations/weaviate/src/haystack_integrations/document_stores/weaviate/document_store.py
```diff
@@ -238,72 +233,42 @@ def _to_document(self, data: Dict[str, Any]) -> Document:
...
-                collection_name,
-                properties,
+    def _query_paginated(self, properties: List[str]):
+
+        try:
+            result = self._collection.iterator(include_vector=True, return_properties=properties)
+        except weaviate.exceptions.WeaviateQueryError as e:
+            msg = f"Failed to query documents in Weaviate. Error: {e.message}"
+            raise DocumentStoreError(msg) from None
+        return list(result)
```
silvanocerza, line 243: I think we can avoid making it a list here? Should a bit more efficient.

integrations/weaviate/src/haystack_integrations/document_stores/weaviate/document_store.py
hsm207: the previous `_query_with_filters` implementation did not have `limit` or `offset`

are you referring to a different function?

integrations/weaviate/src/haystack_integrations/document_stores/weaviate/document_store.py
hsm207: I've evaluated the impact of changing `data` into `Object` and see that the change will be more complicated than it needs to be.

The reason is that this integration flattens documents before storing them in weaviate.

Since this is just an internal method, why don't we keep it like this until weaviate has full support for nested objects? We can include this change as part of a PR to update this part of the code:

https://github.com/deepset-ai/haystack-core-integrations/blob/a430cd630d15b9589c554d0e16d1cb7fac027a91/integrations/weaviate/src/haystack_integrations/document_stores/weaviate/_filters.py#L247-L251

silvanocerza: I don't get it, why is it more complex?

Right now you're calling with `_to_document()` passing the result of `_convert_weaviate_v4_object_to_v3_object()` to it. Where's the problem in making `_to_document()` take an `Object` and do the whole work of converting it to `Document`?

integrations/weaviate/src/haystack_integrations/document_stores/weaviate/document_store.py
silvanocerza: Yeah, I got confused with `_query_paginated()`. My comment still stand though, can't we use pagination?

integrations/weaviate/src/haystack_integrations/document_stores/weaviate/document_store.py
hsm207: you're right, it's actually easy. I thought I had to unflatten the metadata field too but turns out there's already another function to do that

integrations/weaviate/src/haystack_integrations/document_stores/weaviate/document_store.py
hsm207: Sure. What default values should we set for `limit` and `offset` and do you want to make it user configurable?

silvanocerza: I would go with a starting `limit` of `1000`, `offset` obviously at `0` and increment it by `limit` each time.

silvanocerza: And let's not make it configurable as of now, if people in the future ask for that to be customizable we can always change it then.

silvanocerza (approved):

PR #467, opened 2024-02-22
fix: Revert "Allow passing in required project_name (#445)"

shadeMe:
This reverts commit ffe86a9 as it adds a superfluous parameter - API-specific params like `project_name` can be already be passed with the `api_params` init parameter.

This PR additionally clarifies this in the component's docstring and adds a sanity check to make sure the required API params are correctly passed.


masci (approved):
Nice, thanks for the extra docs, I didn't notice you could already pass the param

lbux:
Thanks for the fix 🙏

PR #477, opened 2024-02-23
refactor!: Simplify Optimum backend impl

shadeMe:
In preparation for the optimizer/quantizer integration.


shadeMe:
> I went through the diff and it mostly looks like rearranging code, so LGTM. Let me know if that's a wrong assumption and I should pay attention to any particular area instead.

That's correct - It's just regular refactoring.

PR #481, opened 2024-02-23
Fix example code for Astra DB pipeline

erichare:
This cleans up the examples for the Astra DB integration to ensure that the variables passed match


silvanocerza:
Thanks! 🙏

PR #487, opened 2024-02-26
Mistral: refactor tests

anakin87:


integrations/mistral/tests/test_mistral_chat_generator.py
```diff
@@ -174,27 +176,6 @@ def test_run_with_params(self, chat_messages, mock_chat_completion):
         assert len(response["replies"]) == 1
         assert [isinstance(reply, ChatMessage) for reply in response["replies"]]
 
-    def test_run_with_params_streaming(self, chat_messages):
```
julian-risch, line 177: Could you please explain why we remove this test?

anakin87: yes.

It seems to me a duplicate of https://github.com/deepset-ai/haystack-core-integrations/blob/7a8a29248da75d69fd66e7962165158e291333e6/integrations/mistral/tests/test_mistral_chat_generator.py#L254 where there is a callback involved.

The removed test is not a unit test and requires an API key to run.
So it seems to have a very similar purpose to `test_live_run_streaming`.

julian-risch: Okay, agreed. We could have made the `test_run_with_params_streaming` a unit test similar to `test_run_with_params` as an alternative.

PR #490, opened 2024-02-27
Changed Default Ollama Embedding models to supported model: nomic-embed-text

jmdevita:
Ollama has officially released embedding models with their 0.26 release. Before, we used a lightweight llm model orca-mini to embed, which is much slower. The Nomic-Embed-Text model is much more efficient and should be the standard for an embedded model. This requires the nomic-embed-text model to be pulled from the repo via "ollama pull nomic-embed-text".

https://ollama.com/library/nomic-embed-text


anakin87 (approved):
Thanks for the contribution!

Looks good...

PR #501, opened 2024-02-29
docs: review chroma integration

wochinge:
**Proposed Changes**
- fix https://github.com/deepset-ai/haystack/issues/7130
- fix https://github.com/deepset-ai/haystack/issues/7183


integrations/chroma/src/haystack_integrations/document_stores/chroma/document_store.py
```diff
@@ -177,15 +186,19 @@ def write_documents(self, documents: List[Document], policy: DuplicatePolicy = D
     def delete_documents(self, document_ids: List[str]) -> None:
         """
         Deletes all documents with a matching document_ids from the document store.
+
         Fails with `MissingDocumentError` if no document with this id is present in the store.
```
davidsbatista, line 190: is this an exception being thrown? if so I would say added to the docstring with the `:raises:` keyword

wochinge: great catch - let me have a look

wochinge: dropped it

PR #502, opened 2024-02-29
pgvector - review docstrings and API reference

anakin87:
- fixes https://github.com/deepset-ai/haystack/issues/7146
- fixes https://github.com/deepset-ai/haystack/issues/7198 (small change)


integrations/pgvector/pydoc/config.yml
```diff
@@ -4,7 +4,6 @@ loaders:
     modules: [
       "haystack_integrations.components.retrievers.pgvector.embedding_retriever",
       "haystack_integrations.document_stores.pgvector.document_store",
-      "haystack_integrations.document_stores.pgvector.filters",
```
anakin87, line 7: this module contains the filtering logic for Pgvector.
Should not be user-facing.

PR #505, opened 2024-02-29
fix: make write_documents compatible with the DocumentStore protocol

masci:
Document store API was outdated


julian-risch (approved):
LGTM! 👍 Double checked the DuplicatePolicy and it looks all good. Only DuplicatePolicy.FAIL is supported so the number of written documents is really always equal to the number of documents in the input.

PR #529, opened 2024-03-04
docs: Update `ragas-haystack` docstrings

shadeMe:
- fixes https://github.com/deepset-ai/haystack/issues/7149
- fixes https://github.com/deepset-ai/haystack/issues/7201


julian-risch (approved):
Looks good to me. 👍 The fact that an OPENAI_API_KEY env variable needs to be set is not mentioned in the docstrings anymore but we have it in the documentation page.

PR #537, opened 2024-03-05
feat: Add `NvidiaTextEmbedder`, `NvidiaDocumentEmbedder` and co.

shadeMe:
- fixes https://github.com/deepset-ai/haystack-private/issues/70

### DIFF
--- a/integrations/nvidia/pydoc/config.yml
+++ b/integrations/nvidia/pydoc/config.yml
@@ -1,7 +1,12 @@
 loaders:
   - type: haystack_pydoc_tools.loaders.CustomPythonLoader
     search_path: [../src]
-    modules: []
+    modules:
+      [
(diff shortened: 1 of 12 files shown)


silvanocerza (approved):

davidsbatista:
I was doing a review, do you still want my comments?

davidsbatista:
Overall seems to me a very good job, and kudos for already implementing the batch processing!

My only comment is regarding the `assert` in the code. I think assertions are better suited for tests or debugging proposes.
- If they are needed due to the type checker limitations, I would suggest a `# type: ignore`
- Otherwise, why not test the condition and raise an Exception?
- Also, asserts can be ignored by passing the `-O` flag to the interpreter

NOTE: this is not a blocker for me, I am just curious why you use them.

PR #551, opened 2024-03-06
Ollama: change testing workflow

anakin87:
- testing Ollama using Docker started failing: https://github.com/deepset-ai/haystack-core-integrations/actions/runs/8164727809
- the integrations works correctly (locally tested)
- Github services are sometimes unreliable and Ollama can be installed without Docker

So I am dropping the installation via Docker and I'm installing it following [the official instructions](https://github.com/ollama/ollama?tab=readme-ov-file#linux).

### DIFF
--- a/.github/workflows/ollama.yml
+++ b/.github/workflows/ollama.yml
@@ -41,7 +41,20 @@ jobs:
         run: |
           curl -fsSL https://ollama.com/install.sh | sh
           ollama serve &
-          sleep 10
+          
+          # Check if the service is up and running with a timeout of 60 seconds
+          timeout=60
+          while [ $timeout -gt 0 ] && ! curl -sSf http://localhost:11434/ > /dev/null; do
+            echo "Waiting for Ollama service to start..."
+            sleep 5
+            ((timeout-=5))
+          done
+
+          if [ $timeout -eq 0 ]; then
+            echo "Timed out waiting for Ollama service to start."
+            exit 1
+          fi
+          
           ollama pull ${{ env.LLM_FOR_TESTS }}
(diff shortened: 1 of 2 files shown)


integrations/ollama/pyproject.toml
```diff
@@ -157,20 +157,17 @@ ban-relative-imports = "parents"
...
 [tool.coverage.report]
+omit = ["*/tests/*", "*/__init__.py"]
+show_missing=true
 exclude_lines = [
-    "no cov",
-    "if __name__ == .__main__.:",
-    "if TYPE_CHECKING:",
+  "no cov",
+  "if __name__ == .__main__.:",
+  "if TYPE_CHECKING:",
```
anakin87, line 170: unrelated to this PR.
I just want the right coverage to be shown.

.github/workflows/ollama.yml
```diff
@@ -33,16 +33,18 @@ jobs:
...
-          - 11434:11434
         
     steps:   
       - uses: actions/checkout@v4
 
+      - name: Install Ollama and pull the required models
+        run: |
+          curl -fsSL https://ollama.com/install.sh | sh
+          ollama serve &
+          sleep 10
```
wochinge, line 44: is there a health check for serve which we can use to see if the server is up? That might be a bit more robust than the hard coded sleep. Up to you though

wochinge (approved):
One tiny question. Otherwise looking good.

.github/workflows/ollama.yml
anakin87: good idea. Done!

PR #555, opened 2024-03-07
[RAGAS] fix: Metric parameter validation and metric descriptors

### LINKED ISSUE #528: [RAGAS] Update metric parameter validation and metric descriptors
- The validate_metric_parameters function doesn't actually check if an empty dict is passed to it because of the set subset function (which always returns true for empty sets). We should make this function similar to [DeepEval's](https://github.com/deepset-ai/haystack-core-integrations/blob/4df46344df2d1ef4f19187d69d31edf0cef05c28/integrations/deepeval/src/haystack_integrations/components/evaluators/deepeval/metrics.py#L174).
  - Also we should not convert Nones passed as the metric_param to an empty dict - The none should handled explicitly.
  - Our unit tests don't catch this atm - needs to be fixed.
  - Furthermore, i see that the unit test relies on the Ragas backend raising an error (for aspect critique, for example) - We shouldn't do this, especially when upstream breaks so often.
    - Let's modify the impl so that our scaffolding checks the init parameters and our unit tests reliably catch those.
    - We'll also need to test the other metrics that require init params.
      - Since those are currently missing, we missed the above issue with the validation logic.
- The init parameters in the descriptors seem to be incorrect. AFAICT, these are the only metrics that need extra init params:
  - Answer correctness - weights
  - Answer similarity - threshold
  - Aspect critique - name, definition, strictness
  - Answer relevancy - strictness
- The rest need to be removed. Keep in mind that the docstrings in Ragas seem to be out of sync.

julian-risch:
fixes #528 

- Extends metric param validation, so that it is done by our integration framework and not Ragas

### DIFF
--- a/integrations/ragas/src/haystack_integrations/components/evaluators/ragas/evaluator.py
+++ b/integrations/ragas/src/haystack_integrations/components/evaluators/ragas/evaluator.py
@@ -80,7 +80,7 @@ class RagasEvaluator:
 
     def _init_metric(self):
         MetricParamsValidator.validate_metric_parameters(
-            self.metric, self.descriptor.init_parameters, self.metric_params
+            self.metric, self.descriptor.required_init_parameters, self.descriptor.optional_init_parameters, self.metric_params
         )
         self._backend_metric = self.descriptor.backend(**self.metric_params)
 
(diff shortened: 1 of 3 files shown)


julian-risch:
@shadeMe now we don't distinguish required and optional parameters. `name` is used only for `ASPECT_CRITIQUE`, where it is needed.

Note that the implementation of  `def _init_metric(self):` in deepeval and now in ragas checks that all metric init parameters provided by the user are in the list of expected metric init parameters. In addition, it checks that if the metric expects any init parameters then the user must provide at least one of those. It doesn’t check that all of them are provided by the user. So it’s not fully checking required parameters but also not checking only allowed parameters.

Here is the part of code that I am referring to: https://github.com/deepset-ai/haystack-core-integrations/pull/555/files#diff-c0c023551bfa22d8d23e1e3f6cefcae07b3819de9b9811e313126f6bae1c5a53R81

PR #559, opened 2024-03-07
Add missing import in NvidiaGenerator docstring

silvanocerza:


integrations/nvidia/src/haystack_integrations/components/generators/nvidia/generator.py
```diff
@@ -20,6 +20,7 @@ class NvidiaGenerator:
     Usage example:
     ```python
     from haystack_integrations.components.generators.nvidia import NvidiaGenerator
+    from haystack_integrations.components.generators.nvidia.models import NvidiaGeneratorModel
```
shadeMe, line 23: Let's reexport this symbol in `nvidia` and use that instead.

shadeMe (changes requested):

shadeMe (approved):

PR #566, opened 2024-03-11
Make tests show coverage

anakin87:
The tests for some integrations showed the coverage of test files instead of the coverage of other files in the integration.

I tried to fix this in a standardized way, by changing the respective pyproject files.


davidsbatista (approved):
LGTM, thanks for this @anakin87 👍🏽 

maybe at some point in the future we can agree at a minimum test coverage, or at least which methods must be always tested

PR #568, opened 2024-03-11
Astra DB: Add integration usage tracking

erichare:
This pull request passes in the haystack name and version to Astra's constructor, allowing Astra to track how much usage is occurring of this particular integration. Please let me know if there's any other / more idiomatic way to do so! Thank you!


integrations/astra/src/haystack_integrations/document_stores/astra/astra_client.py
```diff
@@ -5,6 +5,10 @@
 
 from astrapy.api import APIRequestError
 from astrapy.db import AstraDB
+
+# For version tracking
+from haystack import __name__ as integration_name
```
masci, line 10: I would just hardcode the string `haystack` at the module level and save an import call, something like
```py
CALLER_NAME = 'haystack'

masci (changes requested):
One small nit otherwise looks good to me 👍

masci (approved):
LGTM!

PR #570, opened 2024-03-12
feat: Qdrant datetime filtering support

Anush008:
Qdrant v1.8 introduced support for [datetime filtering](https://qdrant.tech/documentation/concepts/filtering/#datetime-range).

This PR updates the Qdrant integration to support the same in Haystack.


Anush008:
Blocked by https://github.com/qdrant/qdrant-client/issues/532.

Anush008:
Hi @anakin87. Could you please rerun the CI?

`qdrant-client==1.8.1` should now fix this.

integrations/qdrant/src/haystack_integrations/document_stores/qdrant/filters.py
```diff
@@ -141,34 +133,47 @@ def _build_nin_condition(self, key: str, value: List[models.ValueVariants]) -> m
             must_not=[
                 (
                     models.FieldCondition(key=key, match=models.MatchText(text=item))
-                    if isinstance(item, str) and " " not in item
+                    if isinstance(item, str) and " " in item
```
anakin87, line 136: can you explain this change?

integrations/qdrant/src/haystack_integrations/document_stores/qdrant/filters.py
```diff
@@ -141,34 +133,47 @@ def _build_nin_condition(self, key: str, value: List[models.ValueVariants]) -> m
...
 
-    def _build_lt_condition(self, key: str, value: float) -> models.Condition:
+    def _build_lt_condition(self, key: str, value: Union[str, float, int]) -> models.Condition:
+        if isinstance(value, str) and is_datetime_string(value):
+            return models.FieldCondition(key=key, range=models.DatetimeRange(lt=value))
+
         if not isinstance(value, (int, float)):
-            msg = f"Value {value} is not an int or float"
+            msg = f"Value {value} is not an int or float or datetime string"
             raise FilterError(msg)
```
anakin87, line 149: I would do something like this:
```python
    msg = f"Value {value} is not an int or float or datetime string"
    if isinstance(value, str):
        if is_datetime_string(value):
            return models.FieldCondition(key=key, range=models.DatetimeRange(lt=value))
       raise FilterError(msg)

        if not isinstance(value, (int, float)):
          raise FilterError(msg)
```

It is important that if the string is not recognized as a datetime object, an error is raised.

anakin87 (commented):
I'm happy that the tests pass.
Thanks for this contribution!

I left two comments.

integrations/qdrant/src/haystack_integrations/document_stores/qdrant/filters.py
Anush008: This fixes the previous faulty implementation.
If the string has a whitespace we do a [full-text match](https://qdrant.tech/documentation/concepts/filtering/#full-text-match) else a [keyword/number match](https://qdrant.tech/documentation/concepts/filtering/#match).

integrations/qdrant/src/haystack_integrations/document_stores/qdrant/filters.py
Anush008: Would https://github.com/deepset-ai/haystack-core-integrations/pull/570/commits/dd821d05269183a899b3b97cd88828790f9931e5 do?

anakin87 (approved):
LGTM!

anakin87:
Released: https://pypi.org/project/qdrant-haystack/3.2.0/

Issue #486, opened 2024-02-26
Notify the team if nightly tests fail


silvanocerza:
As of now nightly test never notify the Core team if there's any failure.

We should add a job to let the team know that there are failures so we can fix it promptly.

masci:
Let's do this after https://github.com/deepset-ai/haystack-core-integrations/issues/435

PR #577, opened 2024-03-13
Notify nightly failures

anakin87:
Part of #486
Based on #574. **Do not merge before the base has been merged.** 

As discussed, I added a notification step to each integration workflow:
if the test is a nightly run (== is scheduled) and fails, a corresponding event is sent to Datadog, which then takes care of notifications using a specifically configured Monitor.

To reduce duplication, I created a composite action to send events to Datadog.
As you may have noticed, I ran several manual tests.


masci (approved):
Nice! 🚢

Issue #549, opened 2024-03-06
Qdrant: Support Sparse Vectors


lambda-science:
**Is your feature request related to a problem? Please describe.**
Qdrant v1.7.0 introduced sparse verctors (with SPLADE) and hybrid retrieval.
Could be cool to implement. https://qdrant.tech/articles/sparse-vectors/

**Describe the solution you'd like**
Allow to create collection with optional sparse vector and add a retrieve for hybrid search (and a SPLADE only ?)
Current:
```python
    def _recreate_collection(self, collection_name: str, distance, embedding_dim: int):
        self.client.recreate_collection(
            collection_name=collection_name,
            vectors_config=rest.VectorParams(
                size=embedding_dim,
                distance=distance,
            ),
            shard_number=self.shard_number,
            replication_factor=self.replication_factor,
            write_consistency_factor=self.write_consistency_factor,
            on_disk_payload=self.on_disk_payload,
            hnsw_config=self.hnsw_config,
            optimizers_config=self.optimizers_config,
            wal_config=self.wal_config,
            quantization_config=self.quantization_config,
            init_from=self.init_from,
        )
```
could become as in the example article above:
```python
    def _recreate_collection(self, collection_name: str, distance, embedding_dim: int):
        self.client.recreate_collection(
            collection_name=collection_name,
            vectors_config={
        "text-dense": rest.VectorParams(
            size=embedding_dim,
            distance=distance,
        )
    },
    sparse_vectors_config={
        "text-sparse": rest.SparseVectorParams(
            index=models.SparseIndexParams(
                on_disk=False,
            )
        )
    },
            shard_number=self.shard_number,
            replication_factor=self.replication_factor,
            write_consistency_factor=self.write_consistency_factor,
            on_disk_payload=self.on_disk_payload,
            hnsw_config=self.hnsw_config,
            optimizers_config=self.optimizers_config,
            wal_config=self.wal_config,
            quantization_config=self.quantization_config,
            init_from=self.init_from,
        )
```

However this requiere a number of component:
* a SPLADE query embedder to embed the user question at query time => ISSUES if integrated inside this integration package at it probably requiere some big machine-learning libs ?
* a SPLADE encoder to write document sparse vectors during indexation => ISSUES if integrated inside this integration package at it probably requiere some big machine-learning libs ?
* a new hybrid retriever that can do a hybrid search such as:
```python
client.search_batch(
    collection_name=collection_name,
    requests=[
[...]

PR #578, opened 2024-03-13
feat(Qdrant): start to work on sparse vector integration

lambda-science:
Related to issue #549 
CC @Anush008 if you want to start having an eye on this and feedback / guidelines 👀

This is a first pull request towards integration of Sparse vector capabilities from Qdrant v1.7/v1.8  (https://qdrant.tech/articles/qdrant-1.7.x/).
I'm working on this following this article: https://qdrant.tech/articles/sparse-vectors/

Main changes / work (70%) has to be done in `document_store.py`, 25% in the Haystack-Qdrant converter and 5% just to add the Retriever that simply do a `document_store.py` call. Also fixing test will take time.  

CHANGES:
* NEW `query_by_sparse()` in `document_store.py` to be able to query with a sparse embedding input.
* MODIFIED `_recreate_collection()` in `document_store.py` to add Sparse Vector in `vectors_config`
* NEW `QdrantSparseRetriever` in `retriever.py` that calls the `self._document_store.query_by_sparse()`
* 
TODO:
* How we handle previous collections without this feature ?

Main Concerns / Questions:
* Afraid that this create some breaking changes with old collections as our vectors are now a dict with both unnamed dense vector and named sparse vectors. 
* What will happens for already existing collection withtout sparse ? Recreated/Errors ? Should we maybe make it optional ? Modify `_set_up_collection` logic for warnings or safeguards ?


Anush008:
I want to ask @anakin87, if we'll have Haystack integrations to generate sparse vectors? Since @lambda-science's work relies on this. A clarification would be helpful.

integrations/qdrant/src/haystack_integrations/document_stores/qdrant/document_store.py
```diff
@@ -298,6 +298,50 @@ def get_documents_by_id(
...
+        results = [self.qdrant_to_haystack.point_to_document(point) for point in points]
+        if scale_score:
+            for document in results:
+                score = document.score
+                if self.similarity == "cosine":
+                    score = (score + 1) / 2
+                else:
+                    score = float(1 / (1 + np.exp(-score / 100)))
+                document.score = score
+        return results
```
lambda-science, line 343: Maybe we can't scale score like for Dense Vectors when using Sparse vectors, should be changed !

lambda-science:
> relies

I was planning to create custom component or public Haystack component contribution to create Sparse embedding after doing this integration in Qdrant, specifically since FastEmbed will release their embedder soon ! (I know it's a Qdrant maintained lib :p ) https://github.com/qdrant/fastembed/pull/144

Qdrant integration could support Sparse embedding before we can generate them in Haystack, it's not an issue. We would just not be able to use the retriever for now :)

But maybe this integration could requiere to modify the Document Dataclass to not only have the `embedding` field for dense vectors, but also have an optional `sparse_embedding` for sparse :) Here, just a supplementary field: https://github.com/deepset-ai/haystack/blob/main/haystack/dataclasses/document.py#L69

Anush008:
Wouldn't it be better to have a sparse embedding component first? Will help in validating the new Qdrant sparse retriever.

Also, Just in.
FastEmbed will now have a sparse embedding model.

https://github.com/qdrant/fastembed/pull/146

lambda-science:
> Wouldn't it be better to have a sparse embedding component first? Will help in validating the new Qdrant sparse retriever.
> 
> Also, Just in. FastEmbed will now have a sparse embedding model.
> 
> [qdrant/fastembed#146](https://github.com/qdrant/fastembed/pull/146)

I can look into modifying the FastEmbed integration in Haystack to create a SparseEmbedding component then with this freshly relased version of FastEmbed 😁

Anush008:
Great. Please let me know if you need a hand with the code.

lambda-science:
I think I finished my PR for a Working Sparse Retriever in Qdrant based on my previous PR with Sparse Embedder from FastEmbed @Anush008 and @anakin87 I guess both PR are compatible to try to work with and allow do a complete round trip of testing (without Haytack document class modification).

integrations/qdrant/src/haystack_integrations/document_stores/qdrant/converters.py
```diff
@@ -19,7 +19,15 @@ def documents_to_batch(
...
-            vector = payload.pop(embedding_field) or {}
+            vector = {}
+            if embedding_field in payload and payload[embedding_field] is not None:
+                dense_vector = payload.pop(embedding_field) or []
+                vector["text-dense"] = dense_vector
+            # TODO: Adapt to Haystack Modification of the Document Dataclass
+            if "_sparse_vector" in payload["meta"]:
+                sparse_vector = payload["meta"].pop("_sparse_vector", {"indices": [], "values": []})
+                sparse_vector_instance = rest.SparseVector(**sparse_vector)
+                vector["text-sparse"] = sparse_vector_instance
```
lambda-science, line 30: before haystack document class modification, we have to store it in the meta field

integrations/qdrant/src/haystack_integrations/components/retrievers/qdrant/retriever.py
```diff
@@ -120,3 +120,248 @@ def run(
...
+            top_k=top_k or self._top_k,
+            scale_score=scale_score or self._scale_score,
+            return_embedding=return_embedding or self._return_embedding,
+        )
+
+        return {"documents": docs}
+
+
+# @component
+# class QdrantHybridRetriever:
```
lambda-science, line 244: Commented because not really working. Search_Batch return ScoredPoint and not Point, so they are not convertible to Haystack documents (no payload)

integrations/qdrant/src/haystack_integrations/document_stores/qdrant/document_store.py
```diff
@@ -298,6 +298,117 @@ def get_documents_by_id(
             documents.append(self.qdrant_to_haystack.point_to_document(record))
         return documents
 
+    # def query_hybrid(
```
lambda-science, line 301: Commented because not really working. Search_Batch return ScoredPoint and not Point, so they are not convertible to Haystack documents (no payload)

integrations/qdrant/tests/test_retriever.py
```diff
@@ -112,3 +113,253 @@ def test_run(self, filterable_docs: List[Document]):
...
+            query_sparse_embedding=self._generate_mocked_sparse_embedding(1)[0], top_k=5, return_embedding=False
+        )
+
+        assert len(results["documents"]) == 5  # type: ignore
+
+        for document in results["documents"]:  # type: ignore
+            assert document.embedding is None
+
+
+# class TestQdrantHybridRetriever(FilterableDocsFixtureMixin):
```
lambda-science, line 240: Commented because not really working. Search_Batch return ScoredPoint and not Point, so they are not convertible to Haystack documents (no payload)

lambda-science (commented):
Might be a breaking change PR because I switched from unnamed dense vector to named dense vector "dense-text"

lambda-science:
Updated the PR to support the futur version of Haystack 2.1.0 that integrates sparse document in the document dataclass :) https://github.com/deepset-ai/haystack/pull/7382
Test passing in local as well as linting.

Also fixed hybrid retrieval :)

lambda-science:
Latest question I have:

* Afraid that this create some breaking changes with old collections as our vectors are now a dict with both unnamed dense vector and named sparse vectors.
* What will happens for already existing collection withtout sparse ? Recreated/Errors ? Should we maybe make it optional ? Modify _set_up_collection logic for warnings or safeguards ?

integrations/qdrant/src/haystack_integrations/document_stores/qdrant/converters.py
```diff
@@ -44,13 +54,26 @@ def convert_id(self, _id: str) -> str:
...
+            payload["embedding"] = None
         payload["score"] = point.score if hasattr(point, "score") else None
+        if hasattr(point, "vector") and point.vector is not None and "text-sparse" in point.vector:
+            parse_vector_dict = {
+                "indices": point.vector["text-sparse"].indices,
+                "values": point.vector["text-sparse"].values,
+            }
+            payload["sparse_embedding"] = parse_vector_dict
+        else:
+            # TODO: does this work with this https://github.com/deepset-ai/haystack/pull/7382#discussion_r1530515595 ?
```
lambda-science, line 77: Check this

integrations/qdrant/src/haystack_integrations/document_stores/qdrant/converters.py
```diff
@@ -15,11 +15,21 @@ def documents_to_batch(
...
         points = []
         for document in documents:
             payload = document.to_dict(flatten=False)
-            vector = payload.pop(embedding_field) or {}
+            vector = {}
+            if embedding_field in payload and payload[embedding_field] is not None:
+                dense_vector = payload.pop(embedding_field) or []
+                vector["text-dense"] = dense_vector
+            # TODO: does this work with this https://github.com/deepset-ai/haystack/pull/7382#discussion_r1530515595 ?
+            # TODO: maybe check if not empty string also on top of None ?
```
lambda-science, line 28: Check this

lambda-science (commented):
2 Todo left on how sparse are handled by default by the dataclass (empty string and not none ?)

integrations/qdrant/tests/test_retriever.py
lambda-science: fixed

integrations/qdrant/src/haystack_integrations/document_stores/qdrant/document_store.py
lambda-science: it's dot product so I keep only the else !

integrations/qdrant/src/haystack_integrations/document_stores/qdrant/converters.py
```diff
@@ -18,11 +18,21 @@ def documents_to_batch(
...
+            if sparse_embedding_field in payload and payload[sparse_embedding_field] is not None:
+                sparse_vector = payload.pop(sparse_embedding_field, {"indices": [], "values": []})
+                sparse_vector_instance = rest.SparseVector(**sparse_vector)
+                vector["text-sparse"] = sparse_vector_instance
```
Anush008, line 35: Lines 38 to 48 to be removed?

integrations/qdrant/src/haystack_integrations/document_stores/qdrant/document_store.py
```diff
@@ -385,9 +492,8 @@ def _set_up_collection(
             # Create Payload index if payload_fields_to_index is provided
             self._create_payload_index(collection_name, payload_fields_to_index)
             return
-
-        current_distance = collection_info.config.params.vectors.distance
-        current_vector_size = collection_info.config.params.vectors.size
+        current_distance = collection_info.config.params.vectors["text-dense"].distance
```
Anush008, line 495: A constant for the vector names would be better since the literals have several occurrences.

Anush008:
Left some file comments.
This would be a breaking release due to the named vectors.

anakin87:
@Anush008 I can't see your comments...

integrations/qdrant/src/haystack_integrations/components/retrievers/qdrant/retriever.py
```diff
@@ -120,3 +120,248 @@ def run(
...
+
+    document_store = QdrantDocumentStore(
+        ":memory:",
+        recreate_index=True,
+        return_sparse_embedding=True,
+        wait_result_from_api=True,
+    )
+    retriever = QdrantSparseRetriever(document_store=document_store)
+
+    # using a fake sparse vector to keep the example simple
```
Anush008, line 143:
```suggestion
```

anakin87:
nowhere :thinking:

Anush008:
Not sure if we'd want to append dense and sparse results as a hybrid search result.

Since hybrid search usually employs some sort of fusion of the 2.

integrations/qdrant/src/haystack_integrations/document_stores/qdrant/document_store.py
```diff
@@ -298,6 +297,111 @@ def get_documents_by_id(
             documents.append(self.qdrant_to_haystack.point_to_document(record))
         return documents
 
+    def query_hybrid(
```
Anush008, line 300: Adding to https://github.com/deepset-ai/haystack-core-integrations/pull/578#issuecomment-2012977944, we can accept a fusion function for dense and sparse scores.

An [example implementation](https://github.com/run-llama/llama_index/blob/4394c7f11e907c4a7c9926ae98eb53e6d60a1619/llama-index-integrations/vector_stores/llama-index-vector-stores-qdrant/llama_index/vector_stores/qdrant/base.py#L471-L510).

integrations/qdrant/src/haystack_integrations/document_stores/qdrant/converters.py
lambda-science: the 
```python
            # TODO: remove as soon as we introduce the support for sparse embeddings in Qdrant
            if "sparse_embedding" in payload:
                sparse_embedding = payload.pop("sparse_embedding", None)
                if sparse_embedding:
                    logger.warning(
                        "Document %s has the `sparse_embedding` field set,"
                        "but storing sparse embeddings in Qdrant is not currently supported."
                        "The `sparse_embedding` field will be ignored.",
                        payload["id"],
                    )
```
I think yes but it was after my commits so I guess it's not in my PR @anakin87

integrations/qdrant/src/haystack_integrations/document_stores/qdrant/document_store.py
lambda-science: agree, constant for "dense embedding" vector name and for "sparse embedding" vector name. Not entierly sure where/how to put this in the class

integrations/qdrant/src/haystack_integrations/document_stores/qdrant/document_store.py
lambda-science: Yeah I'm a little bit un-educated on how to do proper hybrid search I was thinking just appending both topK, never thought about proper function to do this.

lambda-science: Could be good to provide a default fusion function like appending or something like Relative Score Fusion https://docs.llamaindex.ai/en/stable/examples/vector_stores/qdrant_hybrid.html#customizing-hybrid-fusion-fn

integrations/qdrant/src/haystack_integrations/document_stores/qdrant/document_store.py
Anush008: Replacing all occurrences of `"text-dense"` and `"text-dense"` literals in this file with a constant.

integrations/qdrant/src/haystack_integrations/document_stores/qdrant/converters.py
anakin87: yes. I added these lines to not break Document Stores when we release the change to Document, but should be removed in this PR.

lambda-science:
> Not sure if we'd want to append dense and sparse results as a hybrid search result.
> 
> Since hybrid search usually employs some sort of fusion of the 2.

I was thinking about hybrid search as looking both in sparse and dense in one request. It makes sens to me now that hybrid means actually a mix of both scores (through a fusion function) ! Could be cool to have both maybe

anakin87:
TBH, the scope of this PR is getting big and if you agree (@lambda-science) I would prefer to introduce Hybrid retrieval in a future PR, after a bit of study. Obviously, we should make sure the Document Store is compatible.

Anush008:
We can remove `hybrid_search()` then. Would only be useful with a fusion function.

Anush008:
Separate PRs would work better though.

lambda-science:
Latest commits
* remove Hybrid stuff
* add constant for `text-dense` `text-sparse` in converters and documents store
* remove old code that should not be put in PR (compatibility before 2.1.0)

Last concerns/TODO before ready for merge:  
1. Take care to say that it's breaking change and needs new collection (or migration script) because change of vectors naming
2. Last check:
```python
def documents_to_batch [....]
# TODO: does this work with this https://github.com/deepset-ai/haystack/pull/7382#discussion_r1530515595 ?
# TODO: maybe check if not empty string also on top of None ?
  if sparse_embedding_field in payload and payload[sparse_embedding_field] is not None:
      sparse_vector = payload.pop(sparse_embedding_field, {"indices": [], "values": []})
      sparse_vector_instance = rest.SparseVector(**sparse_vector)
      vector[SPARSE_VECTORS_NAME] = sparse_vector_instance
      
[...]

def point_to_document(self, point: QdrantPoint) -> Document:
    payload = {**point.payload}
    if hasattr(point, "vector") and point.vector is not None and SPARSE_VECTORS_NAME in point.vector:
        parse_vector_dict = {
            "indices": point.vector[SPARSE_VECTORS_NAME].indices,
            "values": point.vector[SPARSE_VECTORS_NAME].values,
        }
        payload["sparse_embedding"] = parse_vector_dict
    else:
        # TODO: does this work with this https://github.com/deepset-ai/haystack/pull/7382#discussion_r1530515595 ?
        payload["sparse_embedding"] = None
```

When doing `document_to_batch` shoulf we check if sparse vector is `None` or `""` ?
When doing point to document should we put sparse vector to `None` or `""` ? @anakin87 
(concerning the small catch you did about ids that would maybe change)

lambda-science:
EDIT:
By trying with latest commit of Qdrant Converter and FastEmbed, although test are passing on both im getting this:
```python
 File "/code/haystack_api/controller/files.py", line 254, in index_file_task
result = indexing_pipeline.run(
^^^^^^^^^^^^^^^^^^^^^^
File "/usr/local/lib/python3.11/site-packages/haystack/core/pipeline/pipeline.py", line 817, in run
res = comp.run(**last_inputs[name])
^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
File "/usr/local/lib/python3.11/site-packages/haystack/components/writers/document_writer.py", line 111, in run
documents_written = self.document_store.write_documents(documents=documents, policy=policy)
^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
File "/usr/local/lib/python3.11/site-packages/haystack_integrations/document_stores/qdrant/document_store.py", line 213, in write_documents
batch = self.haystack_to_qdrant_converter.documents_to_batch(
^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
File "/usr/local/lib/python3.11/site-packages/haystack_integrations/document_stores/qdrant/converters.py", line 28, in documents_to_batch
payload = document.to_dict(flatten=False)
^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
File "/usr/local/lib/python3.11/site-packages/haystack/dataclasses/document.py", line 142, in to_dict
data["sparse_embedding"] = sparse_embedding.to_dict()
^^^^^^^^^^^^^^^^^^^^^^^^
AttributeError: 'dict' object has no attribute 'to_dict'
```

integrations/qdrant/src/haystack_integrations/components/retrievers/qdrant/retriever.py
```diff
@@ -120,3 +120,120 @@ def run(
...
+            Deserialized component.
+        """
+        document_store = QdrantDocumentStore.from_dict(data["init_parameters"]["document_store"])
+        data["init_parameters"]["document_store"] = document_store
+        return default_from_dict(cls, data)
+
+    @component.output_types(documents=List[Document])
+    def run(
+        self,
+        query_sparse_embedding: Dict[str, Union[List[int], List[float]]],
```
anakin87, line 213: Let's use the new `SparseEmbedding` object

integrations/qdrant/src/haystack_integrations/document_stores/qdrant/converters.py
```diff
@@ -18,24 +21,23 @@ def documents_to_batch(
...
+        sparse_embedding_field: str,
     ) -> List[rest.PointStruct]:
         points = []
         for document in documents:
             payload = document.to_dict(flatten=False)
-            vector = payload.pop(embedding_field) or {}
+            vector = {}
+            if embedding_field in payload and payload[embedding_field] is not None:
+                dense_vector = payload.pop(embedding_field) or []
+                vector[DENSE_VECTORS_NAME] = dense_vector
```
anakin87, line 32:
```suggestion
            vector[DENSE_VECTORS_NAME] = payload.pop(embedding_field, None) or []
```

integrations/qdrant/src/haystack_integrations/document_stores/qdrant/converters.py
```diff
@@ -18,24 +21,23 @@ def documents_to_batch(
...
+            # TODO: does this work with this https://github.com/deepset-ai/haystack/pull/7382#discussion_r1530515595 ?
+            # TODO: maybe check if not empty string also on top of None ?
+            if sparse_embedding_field in payload and payload[sparse_embedding_field] is not None:
+                sparse_vector = payload.pop(sparse_embedding_field, {"indices": [], "values": []})
+                sparse_vector_instance = rest.SparseVector(**sparse_vector)
+                vector[SPARSE_VECTORS_NAME] = sparse_vector_instance
```
anakin87, line 38:
```suggestion
            sparse_vector = payload.pop(sparse_embedding_field, {"indices": [], "values": []})
            sparse_vector_instance = rest.SparseVector(**sparse_vector)
            vector[SPARSE_VECTORS_NAME] = sparse_vector_instance
```

Since you are providing a default for pop, this should work.

integrations/qdrant/src/haystack_integrations/document_stores/qdrant/document_store.py
```diff
@@ -298,6 +300,44 @@ def get_documents_by_id(
             documents.append(self.qdrant_to_haystack.point_to_document(record))
         return documents
 
+    def query_by_sparse(
+        self,
+        query_sparse_embedding: Dict[str, Union[List[int], List[float]]],
```
anakin87, line 305: Let's use the new SparseEmbedding object

integrations/qdrant/src/haystack_integrations/document_stores/qdrant/document_store.py
```diff
@@ -298,6 +300,44 @@ def get_documents_by_id(
...
+        scale_score: bool = True,  # noqa: FBT001, FBT002
+        return_embedding: bool = False,  # noqa: FBT001, FBT002
+    ) -> List[Document]:
+        qdrant_filters = self.qdrant_filter_converter.convert(filters)
+
+        query_indices = query_sparse_embedding["indices"]
+        query_values = query_sparse_embedding["values"]
+        if len(query_indices) != len(query_values):
+            error_message = "The indices and values of the sparse embedding query must have the same length."
+            raise ValueError(error_message)
```
anakin87, line 317: using the SparseEmbedding object, we don't need this check

integrations/qdrant/tests/test_retriever.py
```diff
@@ -112,3 +117,126 @@ def test_run(self, filterable_docs: List[Document]):
...
+        for doc in filterable_docs:
+            doc.sparse_embedding = SparseEmbedding.from_dict(self._generate_mocked_sparse_embedding(1)[0])
+
+        document_store.write_documents(filterable_docs)
+
+        retriever = QdrantSparseRetriever(document_store=document_store)
+
+        results: List[Document] = retriever.run(query_sparse_embedding=self._generate_mocked_sparse_embedding(1)[0])
+
+        assert len(results["documents"]) == 10  # type: ignore
```
anakin87, line 233: why type:ignore?

anakin87 (commented):
I left some comments

integrations/qdrant/src/haystack_integrations/document_stores/qdrant/converters.py
lambda-science:
```python
            # TODO: does this work with this https://github.com/deepset-ai/haystack/pull/7382#discussion_r1530515595 ?
            # TODO: maybe check if not empty string also on top of None ?
            sparse_vector = payload.pop(sparse_embedding_field, {"indices": [], "values": []})
>           sparse_vector_instance = rest.SparseVector(**sparse_vector)
E           TypeError: qdrant_client.http.models.models.SparseVector() argument after ** must be a mapping, not NoneType

src/haystack_integrations/document_stores/qdrant/converters.py:34: TypeError
```

anakin87: ok you are right.
let's add a check somewhere that raises an error if `sparse_embedding_field` has 0 length.
The same for `embedding_field`

integrations/qdrant/tests/test_retriever.py
lambda-science: I used the same template as previous retriever (dense)
Without:
```python
cmd [3] | mypy --install-types --non-interactive --explicit-package-bases src/ tests
tests/test_retriever.py:233: error: No overload variant of "__getitem__" of "list" matches argument type "str"  [call-overload]
tests/test_retriever.py:233: note: Possible overload variants:
tests/test_retriever.py:233: note:     def __getitem__(self, SupportsIndex, /) -> Any
tests/test_retriever.py:233: note:     def __getitem__(self, slice, /) -> list[Any]
Found 1 error in 1 file (checked 13 source files)

```

integrations/qdrant/src/haystack_integrations/document_stores/qdrant/converters.py
lambda-science: > ok you are right. let's add a check somewhere that raises an error if `sparse_embedding_field` has 0 length. The same for `embedding_field`

Sooooooo I reverted it to this and test are good:

```python
            if embedding_field in payload and payload[embedding_field] is not None:
                dense_vector = payload.pop(embedding_field) or []
                vector[DENSE_VECTORS_NAME] = dense_vector
            if (
                sparse_embedding_field in payload
                and payload[sparse_embedding_field] is not None
                and payload[sparse_embedding_field] != ""
            ):
                sparse_vector = payload.pop(sparse_embedding_field, {"indices": [], "values": []})
                sparse_vector_instance = rest.SparseVector(**sparse_vector)
                vector[SPARSE_VECTORS_NAME] = sparse_vector_instance
```
But I guess maybe you don't like the `and payload[sparse_embedding_field] != ""` check ?

Also I'm still wondering what to do with this: (put to none or empty string ?)
```
class QdrantToHaystack:
    def __init__(self, content_field: str, name_field: str, embedding_field: str, sparse_embedding_field: str):
        self.content_field = content_field
        self.name_field = name_field
        self.embedding_field = embedding_field
        self.sparse_embedding_field = sparse_embedding_field

    def point_to_document(self, point: QdrantPoint) -> Document:
        payload = {**point.payload}
        if hasattr(point, "vector") and point.vector is not None and DENSE_VECTORS_NAME in point.vector:
            payload["embedding"] = point.vector[DENSE_VECTORS_NAME]
        else:
            payload["embedding"] = None
        payload["score"] = point.score if hasattr(point, "score") else None
        if hasattr(point, "vector") and point.vector is not None and SPARSE_VECTORS_NAME in point.vector:
            parse_vector_dict = {
                "indices": point.vector[SPARSE_VECTORS_NAME].indices,
                "values": point.vector[SPARSE_VECTORS_NAME].values,
            }
            payload["sparse_embedding"] = parse_vector_dict
        else:
            # TODO: does this work with this https://github.com/deepset-ai/haystack/pull/7382#discussion_r1530515595 ?
            payload["sparse_embedding"] = None
        return Document.from_dict(payload)
```

lambda-science:
Did an integration test on my haystack app 
with FastEmbed https://github.com/deepset-ai/haystack-core-integrations/pull/579/commits/bc3f555a4b169cd582d879f5c38c6baa331710d7
and Qdrant https://github.com/deepset-ai/haystack-core-integrations/pull/578/commits/7cf1882e7f02f23ffec1e233de47622f0b85b087
Looks like everything is working

integrations/qdrant/src/haystack_integrations/document_stores/qdrant/converters.py
lambda-science: > In Document Store **init**, I would raise an error if embedding_field or sparse_embedding_field are ""

I just did an end-to-end test on my haystack app and everything is working fine now.
If  you want to add a raised error can you do it ? I think we are not talking about the same thing and I'm not sure I understand what you want to check :)
Your concerns are about the name of fields being empty
My concerns were about the content of document's sparsembedding being an empty string (and not none) as defined in the dataclass (trick to not change the existing IDs https://github.com/deepset-ai/haystack/pull/7382#discussion_r1530515595)

anakin87: I simplified a bit this function.

A bit unrelated: I see several things that can be simplified in this Document Store (not related to the Sparse feature)

anakin87:
@lambda-science We (mostly you) made good progress.
Feel free to start thinking about a possible migration script/utility.

(I will dedicate myself to something else this afternoon).

anakin87:
I was thinking that probably this can become a non breaking change if we introduce a use_sparse_embeddings parameter at init time (default False) and if False we fall back to the previous mechanism.
(Also, I am not sure that dense vectors need to be named vectors.)

@lambda-science can you try to experiment in this direction?

lambda-science:
> I was thinking that probably this can become a non breaking change if we introduce a use_sparse_embeddings parameter at init time (default False) and if False we fall back to the previous mechanism. (Also, I am not sure that dense vectors need to be named vectors.)
> 
> @lambda-science can you try to experiment in this direction?

So I integrated your previous comments.
Related to named/unamed vector, I think it's just the clean-code way to put explicit name to the vectors instead of using empty string, but yeah it's breaking.
Related to `non breaking change if we introduce a use_sparse_embeddings parameter at init time (default False) and if False we fall back to the previous mechanism.`. It could be possible but I'm some kind of lazy about it because it would requiere a lot of more code and testing for both cases.
I can totally work on migration script though, I already have one prepare (more or less)

anakin87:
I understand your concerns. I would prefer not to introduce breaking changes, unless it's absolutely necessary. 
The use of non-named vectors for dense embeddings should make the work simpler, especially after examining the implementations in other frameworks.
If you are unable or unwilling to explore this direction, please let me know, and I will attempt it in the coming days. :heart_decoration:

lambda-science:
> use_sparse_embeddings

I will try to have a look this afternoon to support both mode :)

integrations/qdrant/src/haystack_integrations/document_stores/qdrant/converters.py
```diff
@@ -18,23 +21,27 @@ def documents_to_batch(
...
-                        payload["id"],
-                    )
+                sparse_vector = payload.pop(sparse_embedding_field, None)
+                if sparse_vector is not None:
+                    sparse_vector_instance = rest.SparseVector(**sparse_vector)
+                    vector[SPARSE_VECTORS_NAME] = sparse_vector_instance
+
+            if not use_sparse_embeddings:
+                vector = payload.pop(embedding_field) or {}
+            _id = self.convert_id(payload.get("id"))
```
lambda-science, line 44: Handle previous collection without named dense vector (and without sparse) so that there is not breaking change.

integrations/qdrant/src/haystack_integrations/document_stores/qdrant/converters.py
```diff
@@ -58,13 +65,38 @@ def convert_id(self, _id: str) -> str:
...
+                payload["embedding"] = None
+
+            if hasattr(point, "vector") and point.vector is not None and SPARSE_VECTORS_NAME in point.vector:
+                parse_vector_dict = {
+                    "indices": point.vector[SPARSE_VECTORS_NAME].indices,
+                    "values": point.vector[SPARSE_VECTORS_NAME].values,
+                }
+                payload["sparse_embedding"] = parse_vector_dict
+            else:
+                payload["sparse_embedding"] = None
```
lambda-science, line 101: Handle previous collection without named dense vector (and without sparse) so that there is not breaking change.

integrations/qdrant/src/haystack_integrations/document_stores/qdrant/document_store.py
```diff
@@ -298,24 +308,75 @@ def get_documents_by_id(
...
+                with_vectors=return_embedding,
+            )
+        if not self.use_sparse_embeddings:
+            points = self.client.search(
+                collection_name=self.index,
+                query_vector=query_embedding,
+                query_filter=qdrant_filters,
+                limit=top_k,
+                with_vectors=return_embedding,
+            )
```
lambda-science, line 379: Handle previous collection without named dense vector (and without sparse) so that there is not breaking change.

integrations/qdrant/src/haystack_integrations/document_stores/qdrant/document_store.py
```diff
@@ -381,13 +443,33 @@ def _set_up_collection(
...
+                f"to use Named Dense Vectors (`text-sparse`) and Named Sparse Vectors (`text-dense`)."
+            )
+            raise ValueError(msg)
+        if not self.use_sparse_embeddings and isinstance(collection_info.config.params.vectors, dict):
+            msg = (
+                f"Collection '{collection_name}' already exists in Qdrant, "
+                f"but it has been originaly created with sparse embedding vectors."
+                f"If you want to use that collection, please set `use_sparse_embeddings=True`"
+            )
+            raise ValueError(msg)
```
lambda-science, line 472: Explicitly throw an error if there is a possible imcompatibility between legacy vector type and the new vector type (dict) that support sparse embedding

integrations/qdrant/tests/test_converters.py
```diff
@@ -48,5 +81,6 @@ def test_point_to_document_reverts_proper_structure_from_record(
     assert "my-id" == document.id
     assert "Lorem ipsum" == document.content
     assert "text" == document.content_type
+    assert document.sparse_embedding is None
     assert {"test_field": 1} == document.meta
     assert 0.0 == np.sum(np.array([1.0, 0.0, 0.0, 0.0]) - document.embedding)
```
lambda-science, line 86: Test with sparse deactivated (legacy)

integrations/qdrant/tests/test_converters.py
```diff
@@ -28,7 +31,37 @@ def test_convert_id_is_deterministic(haystack_to_qdrant: HaystackToQdrant):
...
+            "text-sparse": {"indices": [7, 1024, 367], "values": [0.1, 0.98, 0.33]},
+        },
+    )
+    document = qdrant_to_haystack.point_to_document(point)
+    assert "my-id" == document.id
+    assert "Lorem ipsum" == document.content
+    assert "text" == document.content_type
+    assert {"indices": [7, 1024, 367], "values": [0.1, 0.98, 0.33]} == document.sparse_embedding.to_dict()
+    assert {"test_field": 1} == document.meta
+    assert 0.0 == np.sum(np.array([1.0, 0.0, 0.0, 0.0]) - document.embedding)
```
lambda-science, line 60: Test with sparse activated (new)

lambda-science (commented):
Hello @anakin87 @Anush008  sorry for this commits and comments and notification again.
As suggested, this commit implement the `use_sparse_embeddings` parameter for the document store so that changes are not breaking. It's set to true automatically, but throw an error if previous collection is not compatible. (True by default because it's better for the futur, if someone create a collection without looking the params. Doesn't want to run migration after !)
If false: the behaviour is the same as before (Unnamed Dense Vector only)
if true: the document store now used Named Dense Vector and Named Sparse Vectors

If a mismatch between existing collection and this parameters: an error is throw with either (i) suggesting to do migration (legacy to new) (ii) suggestion to set use_sparse_embeddings to True (new collection set that is loaded with use_sparse_embeddings=False)

This possibly needs more testing. I'm not entirely sure how to make sure there is no breaking change :)
EDIT: I will do complete integration test with a previous and new collection to see if it breaks my app. Maybe today or tomorrow

integrations/qdrant/src/haystack_integrations/document_stores/qdrant/document_store.py
```diff
@@ -381,13 +443,34 @@ def _set_up_collection(
...
+
+        if self.use_sparse_embeddings:
+            current_distance = collection_info.config.params.vectors[DENSE_VECTORS_NAME].distance
+            current_vector_size = collection_info.config.params.vectors[DENSE_VECTORS_NAME].size
+        if not self.use_sparse_embeddings:
+            current_distance = collection_info.config.params.vectors.distance
+            current_vector_size = collection_info.config.params.vectors.size
```
lambda-science, line 473: Explicitly throw an error if there is a possible imcompatibility between legacy vector type and the new vector type (dict) that support sparse embedding

integrations/qdrant/src/haystack_integrations/document_stores/qdrant/converters.py
```diff
@@ -18,23 +21,27 @@ def documents_to_batch(
...
-                        "but storing sparse embeddings in Qdrant is not currently supported."
-                        "The `sparse_embedding` field will be ignored.",
-                        payload["id"],
-                    )
+                sparse_vector = payload.pop(sparse_embedding_field, None)
+                if sparse_vector is not None:
+                    sparse_vector_instance = rest.SparseVector(**sparse_vector)
+                    vector[SPARSE_VECTORS_NAME] = sparse_vector_instance
+
+            if not use_sparse_embeddings:
```
Anush008, line 42:
```suggestion
            else:
```

integrations/qdrant/src/haystack_integrations/document_stores/qdrant/converters.py
```diff
@@ -58,13 +65,38 @@ def convert_id(self, _id: str) -> str:
...
+        self.sparse_embedding_field = sparse_embedding_field
 
     def point_to_document(self, point: QdrantPoint) -> Document:
         payload = {**point.payload}
-        payload["embedding"] = point.vector if hasattr(point, "vector") else None
         payload["score"] = point.score if hasattr(point, "score") else None
+        if not self.use_sparse_embeddings:
+            payload["embedding"] = point.vector if hasattr(point, "vector") else None
+
+        if self.use_sparse_embeddings:
```
Anush008, line 88:
```suggestion
        else:
```

integrations/qdrant/src/haystack_integrations/document_stores/qdrant/converters.py
```diff
@@ -58,13 +65,38 @@ def convert_id(self, _id: str) -> str:
...
+            if hasattr(point, "vector") and point.vector is not None and DENSE_VECTORS_NAME in point.vector:
+                payload["embedding"] = point.vector[DENSE_VECTORS_NAME]
+            else:
+                payload["embedding"] = None
+
```
Anush008, line 93:
```suggestion
            if hasattr(point, "vector") and point.vector is not None:
                payload["embedding"] = point.vector.get(DENSE_VECTORS_NAME)

```

integrations/qdrant/src/haystack_integrations/document_stores/qdrant/document_store.py
```diff
@@ -298,24 +308,75 @@ def get_documents_by_id(
...
+                collection_name=self.index,
+                query_vector=rest.NamedVector(
+                    name=DENSE_VECTORS_NAME,
+                    vector=query_embedding,
+                ),
+                query_filter=qdrant_filters,
+                limit=top_k,
+                with_vectors=return_embedding,
+            )
+        if not self.use_sparse_embeddings:
```
Anush008, line 372:
```suggestion
        else:
```

integrations/qdrant/src/haystack_integrations/document_stores/qdrant/document_store.py
```diff
@@ -407,24 +490,59 @@ def _set_up_collection(
...
+                replication_factor=self.replication_factor,
+                write_consistency_factor=self.write_consistency_factor,
+                on_disk_payload=self.on_disk_payload,
+                hnsw_config=self.hnsw_config,
+                optimizers_config=self.optimizers_config,
+                wal_config=self.wal_config,
+                quantization_config=self.quantization_config,
+                init_from=self.init_from,
+            )
+        if not use_sparse_embeddings:
```
Anush008, line 528:
```suggestion
        else:
```

integrations/qdrant/src/haystack_integrations/document_stores/qdrant/document_store.py
```diff
@@ -381,13 +443,34 @@ def _set_up_collection(
...
+                f"Collection '{collection_name}' already exists in Qdrant, "
+                f"but it has been originaly created with sparse embedding vectors."
+                f"If you want to use that collection, please set `use_sparse_embeddings=True`"
+            )
+            raise ValueError(msg)
+
+        if self.use_sparse_embeddings:
+            current_distance = collection_info.config.params.vectors[DENSE_VECTORS_NAME].distance
+            current_vector_size = collection_info.config.params.vectors[DENSE_VECTORS_NAME].size
+        if not self.use_sparse_embeddings:
```
Anush008, line 471:
```suggestion
        else:
```

integrations/qdrant/src/haystack_integrations/document_stores/qdrant/document_store.py
Anush008, line 379:
```suggestion
        points = self.client.search(
            collection_name=self.index,
            query_vector=rest.NamedVector(
                name=DENSE_VECTORS_NAME if self.use_sparse_embeddings else "",
                vector=query_embedding,
            ),
            query_filter=qdrant_filters,
            limit=top_k,
            with_vectors=return_embedding,
        )
```

integrations/qdrant/src/haystack_integrations/document_stores/qdrant/document_store.py
```diff
@@ -407,24 +490,59 @@ def _set_up_collection(
...
+                shard_number=self.shard_number,
+                replication_factor=self.replication_factor,
+                write_consistency_factor=self.write_consistency_factor,
+                on_disk_payload=self.on_disk_payload,
+                hnsw_config=self.hnsw_config,
+                optimizers_config=self.optimizers_config,
+                wal_config=self.wal_config,
+                quantization_config=self.quantization_config,
+                init_from=self.init_from,
+            )
```
Anush008, line 545:
```suggestion
        dense_vector_name = DENSE_VECTORS_NAME if use_sparse_embeddings else ""

        vectors_config = {
            dense_vector_name: rest.VectorParams(
                size=embedding_dim,
                on_disk=on_disk,
                distance=distance,
            ),
        }

        if use_sparse_embeddings:
            vectors_config[SPARSE_VECTORS_NAME] = rest.SparseVectorParams(
                index=rest.SparseIndexParams(
                    on_disk=on_disk,
                )
            )

        self.client.recreate_collection(
            collection_name=collection_name,
            vectors_config=vectors_config,
            shard_number=self.shard_number,
            replication_factor=self.replication_factor,
            write_consistency_factor=self.write_consistency_factor,
            on_disk_payload=self.on_disk_payload,
            hnsw_config=self.hnsw_config,
            optimizers_config=self.optimizers_config,
            wal_config=self.wal_config,
            quantization_config=self.quantization_config,
            init_from=self.init_from,
        )
```

lambda-science: I think this broke the test.

lambda-science: @Anush008  Fixed with late commit, hope this is still not breaking change.

lambda-science:
Yet another bug
```python
  File "/code/haystack_api/pipeline/custom_pipelines.py", line 67, in <module>
2024-03-25T09:49:35.806089309Z     document_store = QdrantDocumentStore(
2024-03-25T09:49:35.806091313Z                      ^^^^^^^^^^^^^^^^^^^^
2024-03-25T09:49:35.806093317Z   File "/usr/local/lib/python3.11/site-packages/haystack_integrations/document_stores/qdrant/document_store.py", line 145, in __init__
2024-03-25T09:49:35.806095561Z     self._set_up_collection(
2024-03-25T09:49:35.806097535Z   File "/usr/local/lib/python3.11/site-packages/haystack_integrations/document_stores/qdrant/document_store.py", line 451, in _set_up_collection
2024-03-25T09:49:35.806099799Z     if not self.use_sparse_embeddings and DENSE_VECTORS_NAME in collection_info.config.params.vectors.keys():
2024-03-25T09:49:35.806103326Z                                                                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
2024-03-25T09:49:35.806105971Z   File "/usr/local/lib/python3.11/site-packages/pydantic/main.py", line 767, in __getattr__
2024-03-25T09:49:35.806108365Z     raise AttributeError(f'{type(self).__name__!r} object has no attribute {item!r}')
2024-03-25T09:49:35.806110479Z AttributeError: 'VectorParams' object has no attribute 'keys'
```

This is getting really annoying to test as there is no test to load a previously created collection. I have to clone in my haystack app and rebuild my image => 10 min of work just to test a modification end-to-end. 
My previous commits were working and now we're back to more compatiblity work 🫠

integrations/qdrant/src/haystack_integrations/document_stores/qdrant/document_store.py
lambda-science: This modification of code now force use to handle 3 cases:

1. Before this PR vectors and not dict just a VectorParams
2. After this PR + use sparse = True: vectors are dict with named dense "text-dense"
3. After this PR + use sprase = False: vectors are dict with named dense ""

Too many case to cover, my head will go boom.

anakin87: Feel free to pause your work. I'll look into it later

lambda-science:
Thanks @Anush008 @anakin87 I think I've done enough I will stop my part of contribution here because it's has been kind of exhausting to fix the test and test everything, I spent way too much hours fixing stuff. Specifically because end-to-end testing is messy (importing an old collection to see if nothing is breaking).
I reverted part of your suggestion Anush because it would be still breaking change / add a supplementary case to support.
Now the PR is compatible with two things:

* Old collection before this PR that used dense unnamed vectors.
* New collection after this PR that use both dense and sparse named vectors by default

The only thing to do is that maybe if it's an old collection, it should set use_sparse to false automatically instead of throwing the error asking the user to do it. 

But please, please, please, test deeply your suggestion before doing so (with old collections for example), because then I spend a lot of time fixing stuff

integrations/qdrant/src/haystack_integrations/document_stores/qdrant/document_store.py
Anush008: Yikes. Sorry. `The sparse_vector_config` is a different property.

It's looking good now.

.github/workflows/qdrant.yml
```diff
@@ -58,7 +58,9 @@ jobs:
 
       - name: Run tests
         id: tests
-        run: hatch run cov
+        run: |
+          hatch run pip install git+https://github.com/deepset-ai/haystack.git #TODO: rm before merging
```
Anush008, line 62: A reminder.

Anush008 (approved):
Thank you for the contribution @lambda-science.

integrations/qdrant/src/haystack_integrations/document_stores/qdrant/document_store.py
lambda-science: > Yikes. Sorry. `The sparse_vector_config` is a different property.
> 
> It's looking good now.

No worries, sometimes I get frustrated while fighting with code ahah

anakin87:
@lambda-science thanks for all your work!

I've pushed some little simplifications, but looks good.

The only thing that still leaves me doubtful is having `use_sparse_embeddings=True` by default, but you provided good reasons. I will think about it...

At this point let's wait for 2.0.1!

lambda-science:
> @lambda-science thanks for all your work!
> 
> I've pushed some little simplifications, but looks good.
> 
> The only thing that still leaves me doubtful is having `use_sparse_embeddings=True` by default, but you provided good reasons. I will think about it...
> 
> At this point let's wait for 2.0.1!

One solution would be to have `use_sparse_embeddings=True` by default, and if we detect an old collection (vector are not dict with sparse), then we can set `self.use_sparse_embeddings=False` and log a warning. So that no breaking/blocking/modification needed for old collection with dense only vector. And all new collections use dense+sparse named vectors by default.
I tried to do it but it was mess, but I guess it can be easy to implement, maybe you could do that @anakin87 ? (After 2.0.1 maybe ahah)

anakin87:
Next week we will release 2.0.1!

We discussed with @masci about this new feature.

We would like to introduce it gradually:
- set `use_sparse_embeddings=False` by default in qdrant-haystack 3.3.0
- (see how much the new feature is used/what the reception is)
- later (in qdrant-haystack 4.0.0?), switch to `use_sparse_embeddings=True`

@lambda-science do you want to make this change or would you prefer me to make it?

lambda-science:
> Next week we will release 2.0.1!
> 
> We discussed with @masci about this new feature.
> 
> We would like to introduce it gradually:
> 
> * set `use_sparse_embeddings=False` by default in qdrant-haystack 3.3.0
> * (see how much the new feature is used/what the reception is)
> * later (in qdrant-haystack 4.0.0?), switch to `use_sparse_embeddings=True`
> 
> @lambda-science do you want to make this change or would you prefer me to make it?

Hey

I let you commit the modification to put  `use_sparse_embeddings=False` by default.

Happy to hear haystack 2.0.1 is coming soon.

However is still think people will be quite unhappy when they will discover that the default behavior of the integration is not compatible with all features (from haystack or qdrant) and requires manual scripting for migration.

anakin87:
I am working on these changes.

@lambda-science if you have it at hand, can you share the migration script you used?

lambda-science:
> I am working on these changes.
> 
> @lambda-science if you have it at hand, can you share the migration script you used?

In my case I used something like this, with `pinecone-text` lib for a BM25 encoder (instead of FastEmbed because ressources consomptions). Also note that `old_collection` refers to collection created before this PR and `new_collection` refers to a collection created after this PR with `use_sparse_embeddings=True`

Something like this (a bit modified compared to the real one I use for simplicity):
```python
"""
Script to help upgrade a Qdrant collection that was created <1.7.0
(before sparse vector support), to add sparse vectors in a new collection.

The script simply scrolls through all points in the old collection, creates
a sparse vector for those points, and upserts them into a new collection.
"""

from qdrant_client import QdrantClient
from qdrant_client.http import models
from pinecone_text.sparse import BM25Encoder

import time

bm25 = BM25Encoder.default()
client = QdrantClient(url="http://localhost:6333")

# In my case I create the new_collection with proper configuration (sparse) by booting my Haystack instance
# before running this script. 
old_collection = "pythia"
new_collection = "pythia-bm25"

total_points = client.count(
    collection_name=old_collection,
    exact=True,
).count

next_page_offset = "first"
points_transmitted = 0
start = time.time()

# Disable indexing while adding points so it's faster
client.update_collection(
    collection_name=new_collection,
    optimizer_config=models.OptimizersConfigDiff(indexing_threshold=0),
)

while next_page_offset:
    if next_page_offset == "first":
        offset = None
    else:
        offset = next_page_offset

    # get the records
    records = client.scroll(
        collection_name=old_collection,
        limit=100,
        with_payload=True,
        with_vectors=True,
        offset=offset,
    )

    next_page_offset = records[1]
    current_records = records[0]

    points = []

    # Calculate Sparse Embeddings for each document and set the vector dict
    for record in current_records:
        vector = {}

        content_without_newlines = record.payload["content"].replace("\n", " ").replace(
            "\r", " "
        )
        bm25_encoded = bm25.encode_documents(content_without_newlines)

        # Set Vectors
        vector["text-sparse"] = models.SparseVector(**bm25_encoded)
        vector["text-dense"] = record.vector

        point = {"id": record.id, "payload": record.payload, "vector": vector}

        embedding_point = models.PointStruct(**point)
        points.append(embedding_point)

    client.upsert(collection_name=new_collection, points=points)

    points_transmitted += len(points)
    points_remaining = total_points - points_transmitted

    # Print progress
    message = f"Points transmitted: {points_transmitted}/{total_points}\nPercent done {points_transmitted/total_points*100:.2f}%\nTime elapsed: {time.time() - start:.2f} seconds\nTime remaining: {(((time.time() - start) / points_transmitted) * points_remaining) / 60:.2f} minutes\nCurrent offset: {next_page_offset}\n"
    print(message)

# Re-enable indexing once all points are added with my own indexing threshold
client.update_collection(
    collection_name=new_collection,
    optimizer_config=models.OptimizersConfigDiff(indexing_threshold=20000),
)

```

anakin87 (approved):
Thanks also for your script! In a future PR, I will add a similar one to this project.

PR #594, opened 2024-03-18
Fix opensearch errors bulk write

### LINKED ISSUE #547: OpenSearch Bulk error doesn't contain "create" key
**Describe the bug**
When calling `write_documents` from OpenSearch vector store, if something goes wrong the error is treated with this line:
```python
        if errors:
            duplicate_errors_ids = []
            other_errors = []
            for e in errors:
                error_type = e["create"]["error"]["type"]
```
https://github.com/deepset-ai/haystack-core-integrations/blob/4998a7a6d840070d87f838c73cdf9010aba16865/integrations/opensearch/src/haystack_integrations/document_stores/opensearch/document_store.py#L184

However today in production I come across a problem with `write_documents` and the `bulk` funciton where the error code doesn't actually contains the `"create"` key.

CloudWatch (AWS) logs:
```
2024-03-06T09:10:38.444+01:00 Creating embeddings: 100%\|██████████\| 1/1 [00:00<00:00,  1.02it/s]

2024-03-06T09:10:38.475+01:00 03/06/2024 08:10:38 AM 'create'
```
**To Reproduce**
No idea how to reproduce as I don't have the real error due to this logic.

 - OS: Ubuntu Docker Container
 - Haystack version: 2.0 beta 8
 - Integration version: OpenSearch v0.2.0

silvanocerza:
Fix #547

### DIFF
--- a/integrations/opensearch/src/haystack_integrations/document_stores/opensearch/document_store.py
+++ b/integrations/opensearch/src/haystack_integrations/document_stores/opensearch/document_store.py
@@ -181,6 +181,9 @@ class OpenSearchDocumentStore:
             duplicate_errors_ids = []
             other_errors = []
             for e in errors:
+                if "create" not in e:
+                    other_errors.append(e)
+                    continue
                 error_type = e["create"]["error"]["type"]
                 if policy == DuplicatePolicy.FAIL and error_type == "version_conflict_engine_exception":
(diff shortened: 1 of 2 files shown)


silvanocerza:
Merging as I already reviewed the change and added a test to verify it's working properly.

Issue #595, opened 2024-03-18
Integration tests are skipped


anakin87:
Currently, some of the tests for these integrations are skipped because the expected env var is not set:
- ~ragas~ 
- ~uptrain~
- ~deepeval~ 
- [x] nvidia
- [x] gradient https://github.com/deepset-ai/haystack-core-integrations/pull/625
- [x] astra https://github.com/deepset-ai/haystack-core-integrations/pull/623

It happens for different reasons. We should understand why and fix this issue.

silvanocerza:
API key for Nvidia integration added, integration tests should run now.

anakin87:
@silvanocerza yes!
Those with nvcf backend run fine: https://github.com/deepset-ai/haystack-core-integrations/actions/runs/8444638551/job/23146314844

PR #596, opened 2024-03-18
evaluation components - use API keys in tests

anakin87:
part of #595


masci (approved):
Some tests expected to fail for unrelated reasons

PR #597, opened 2024-03-18
Add NIM backend support

silvanocerza:


integrations/nvidia/src/haystack_integrations/components/embedders/nvidia/_nvcf_backend.py
```diff
@@ -12,11 +50,7 @@ class EmbeddingsRequest:
     encoding_format: Literal["float", "base64"] = "float"
 
     def __post_init__(self):
-        if isinstance(self.input, list):
-            if len(self.input) > MAX_INPUTS:
-                msg = f"The number of inputs should not exceed {MAX_INPUTS}"
-                raise ValueError(msg)
-        else:
+        if not isinstance(self.input, list):
```
shadeMe, line 53: Why did we remove the max inputs check?

integrations/nvidia/src/haystack_integrations/components/generators/nvidia/_nvcf_backend.py
```diff
@@ -0,0 +1,120 @@
+from dataclasses import asdict, dataclass
+from typing import Any, Dict, List, Optional, Tuple
+
+from haystack.utils.auth import Secret
+from haystack_integrations.utils.nvidia import NvidiaCloudFunctionsClient
+
+from .backend import GeneratorBackend
+
+MAX_INPUT_STRING_LENGTH = 2048
```
shadeMe, line 9: We don't use these to perform any checks. We should do that.

shadeMe (changes requested):

integrations/nvidia/src/haystack_integrations/components/generators/nvidia/_nvcf_backend.py
shadeMe: Not required for generators, apparently.

shadeMe (approved):
Ship it far!

PR #605, opened 2024-03-20
CI: push docs when releasing a new version

### LINKED ISSUE #479: Push integration docs only when releasing a new version
As of now the docs for any integration are updated whenever a change is merged in `main`. 

This is wrong as we might publish documentation for features or changes that have yet to be released.
We should change the workflow to push documentation only when there is a new release for that integration.

anakin87:
fixes #479

I modified the existing "Core / Sync docs with Readme" workflow
- does not run on main
- runs when pushing release tags or is manually activated
- only generates docs for the specified integration

I tested the workflow in this PR and seems to work properly: https://github.com/deepset-ai/haystack-core-integrations/actions/runs/8356934800/job/22875054669

### DIFF
--- a/.github/workflows/CI_readme_sync.yml
+++ b/.github/workflows/CI_readme_sync.yml
@@ -2,8 +2,18 @@ name: Core / Sync docs with Readme
 
 on:
   push:
+    tags:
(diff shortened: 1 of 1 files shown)


masci (approved):
LGTM!

PR #606, opened 2024-03-20
Make Document Stores initially skip `SparseEmbedding`

### LINKED ISSUE #604: Make Document Stores initially skip `SparseEmbedding`
This field will come with the next release of Haystack.

In the meantime, we should make sure that Document Stores initially skip this field
and do not break when calling `write_documents`.

Later, the support for Sparse Embedding Retrieval will be added for some compatible Document Stores, starting from Qdrant.

anakin87:
Fixes #604: see that for details


davidsbatista (approved):
LGTM

Issue #608, opened 2024-03-21
Update the Cohere client to version 5


masci:
The Cohere integration pins the version of the client to 4.x, because version 5.0.0 contained several breaking changes.
They provide a migration guide https://github.com/cohere-ai/cohere-python/blob/main/4.0.0-5.0.0-migration-guide.md to help with the transition.

anakin87:
done in #702 and #711

PR #609, opened 2024-03-21
fix: Pin cohere version

masci:
Cohere just released a major release of their client, and there are several breaking changes. Pinning the version to <5 for now.

Tracking issue: https://github.com/deepset-ai/haystack-core-integrations/issues/608


anakin87 (approved):
👍

PR #612, opened 2024-03-22
fix readme_sync workflow

anakin87:
The workflow is failing when manually triggered: https://github.com/deepset-ai/haystack-core-integrations/actions/runs/8381537136/job/22967317316

This happens because the manually entered tag is ignored.
I am trying to fix this.


masci (approved):
👍

PR #618, opened 2024-03-23
Bug fix for document_store.py

crtr0:
Fixed a bug where we ensure that a doc isn't already in our new_documents array before we consider adding it.


masci (changes requested):
Thanks for your contribution! Any chance you can add a unit test to cover this piece of logic?

masci (approved):

PR #630, opened 2024-03-28
Disable nightly tests for uptrain

masci:
Tests are currently expected to fail, disabling nightlies


vblagoje (approved):
Seems fine @masci, failures are expected, right?

PR #638, opened 2024-04-02
LlamaCPP: fix failing tests

anakin87:
Try to fix errors like this: https://github.com/deepset-ai/haystack-core-integrations/actions/runs/8515449565/job/23345387417
Segmentation fault (core dumped)

- Using a smaller model or changing the batch size do not help
- Pinning `llama-cpp-python<0.2.58` fixes the failing tests

I also opened an issue on llama-cpp-python: https://github.com/abetlen/llama-cpp-python/issues/1319

@awinml do you have any ideas?

### DIFF
--- a/integrations/llama_cpp/pyproject.toml
+++ b/integrations/llama_cpp/pyproject.toml
@@ -27,7 +27,7 @@ classifiers = [
 ]
 dependencies = [
   "haystack-ai",
-  "llama-cpp-python<0.2.56"
+  "llama-cpp-python<0.2.58"
 ]
 
 [project.urls]
--- a/integrations/llama_cpp/pyproject.toml
+++ b/integrations/llama_cpp/pyproject.toml
@@ -27,7 +27,7 @@ classifiers = [
 ]
 dependencies = [
   "haystack-ai",
-  "llama-cpp-python"
+  "llama-cpp-python<0.2.56"
 ]
 
 [project.urls]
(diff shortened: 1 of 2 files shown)


awinml:
Hi @anakin87,

Thanks for reaching out. I couldn't figure out a solution, but I conducted some tests to isolate the issue.

- I executed both the examples and the code in the `test_run` and `test_run_rag_pipeline` functions. The component runs smoothly when called from a Python file (both standalone and in a Haystack pipeline).

- I tried running it in a shell spawned through Hatch, and the issue doesn't seem to be related to Hatch. The error only occurs when running with Pytest, and I was able to replicate the same error locally.

By the way, I noticed that the `rag_pipeline_example.py` file in the examples directory still has the old imports for `InMemoryDocumentStore` and `InMemoryEmbeddingRetriever` (unrelated to this issue, though).

Please let me know if you need any additional information or if you'd like me to try something else. I'm happy to assist further.

Here's my system information for reference:

- OS: Linux - Ubuntu 20.04
- Python version: 3.8.10
- `llama-cpp-python` version: 0.2.58

anakin87:
Hey...
Thanks for having a look!

My bug report on llama-cpp-python was labeled as high priority, so I would wait for them to fix it.

anakin87:
The bug was fixed in https://github.com/abetlen/llama-cpp-python/commit/8649d7671bd1a7c0d9cc6a5ad91c6ca286512ab3

`llama-cpp-python==0.2.59` is ok

PR #640, opened 2024-04-03
fix: handle opensearch write_documents error handling

paualarco:
Fix error handling in opensearch `write_documents`

It seems only to be failing on haystack 2.0 🤔 
Error:

```
packages/haystack_integrations/document_stores/opensearch/document_store.py\", line 171, in write_documents\n    error_type = e[\"create\"][\"error\"][\"type\"]\nKeyError: 'create'",
```

### DIFF
--- a/integrations/opensearch/src/haystack_integrations/document_stores/opensearch/document_store.py
+++ b/integrations/opensearch/src/haystack_integrations/document_stores/opensearch/document_store.py
@@ -193,8 +193,6 @@ class OpenSearchDocumentStore:
                     error_type = None 
                 if policy == DuplicatePolicy.FAIL and error_type == "version_conflict_engine_exception":
                     duplicate_errors_ids.append(e["create"]["_id"])
-                if policy == DuplicatePolicy.FAIL and error_type == "version_conflict_engine_exception":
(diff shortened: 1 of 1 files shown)


davidsbatista:
Hi @paualarco thanks for the contribution.

I'm looking into this right now,  a suggestion: always try to add a reproducible example. It helps the reviewers get into context quickly.

davidsbatista:
@paualarco I suspect this was already been fixed by this https://github.com/deepset-ai/haystack-core-integrations/issues/547 - we just need to make a new release

but still, can you provide me with a snippet of code that replicates the error you are having?

davidsbatista:
new release: https://pypi.org/project/opensearch-haystack/0.3.0/

paualarco:
@davidsbatista yes, apologies for not providing a snippet. This happened when trying to insert a string field in opensearch that was already indexed as an integer. 
We reverted the haystack version to see the actual error stacktrace and found out. 
I guess then we could close this issue?

davidsbatista:
thanks for the feedback @paualarco just confirm me that's everything is solved from your side with the new release and I will close it

PR #641, opened 2024-04-04
fix: disabling SSL for local OpenSearch docker

davidsbatista:
- disabling the SSL for the local Docker Image since I was getting this error:

```
ConnectionError([SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed: unable to get local issuer certificate (_ssl.c:1129)) caused by: SSLError([SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed: unable to get local issuer certificate (_ssl.c:1129))
```

### DIFF
--- a/integrations/opensearch/docker-compose.yml
+++ b/integrations/opensearch/docker-compose.yml
@@ -7,6 +7,7 @@ services:
     restart: on-failure
     environment:
       - discovery.type=single-node
+      - plugins.security.disabled=true
       - "ES_JAVA_OPTS=-Xms1024m -Xmx1024m"
     healthcheck:
         test: curl --fail https://localhost:9200/_cat/health -ku admin:admin || exit 1


julian-risch (commented):
You can also turn off certificate verification when you initialize the document store. I wouldn't want to change the docker compose file. It's also the one that is used in the tests so we know it works that way.
```python
document_store = OpenSearchDocumentStore(
    hosts=["https://localhost:9200"],
    index="test_index",
    http_auth=("admin", "admin"),
    verify_certs=False,
    embedding_dim=4,
    method={"space_type": "cosinesimil", "engine": "nmslib", "name": "hnsw"},
)
```
What about extending the example on the integration page instead?

davidsbatista:
ok - makes sense, did not knew it was being used in tests - thanks 👍🏽

PR #644, opened 2024-04-05
feat: Implement keyword retrieval for pgvector integration

jlonge4:


jlonge4:
@anakin87 Sorry about the mess the other PR turned into lol. Hoping to add a keyword retriever test and call this one done soon 👍

integrations/pgvector/src/haystack_integrations/document_stores/pgvector/document_store.py
```diff
@@ -80,9 +80,11 @@ def __init__(
         vector_function: Literal["cosine_similarity", "inner_product", "l2_distance"] = "cosine_similarity",
         recreate_table: bool = False,
         search_strategy: Literal["exact_nearest_neighbor", "hnsw"] = "exact_nearest_neighbor",
+        hybrid_search: bool = False,
```
anakin87, line 83:
```suggestion
```

this is not used currently, so let's remove this.

integrations/pgvector/src/haystack_integrations/document_stores/pgvector/document_store.py
```diff
@@ -80,9 +80,11 @@ def __init__(
...
         hnsw_recreate_index_if_exists: bool = False,
         hnsw_index_creation_kwargs: Optional[Dict[str, int]] = None,
         hnsw_ef_search: Optional[int] = None,
+        language: Optional[str] = "english",
```
anakin87, line 87:
```suggestion
        language: str = "english",
```

integrations/pgvector/src/haystack_integrations/document_stores/pgvector/document_store.py
```diff
@@ -117,6 +119,7 @@ def __init__(
         :param hnsw_ef_search: The `ef_search` parameter to use at query time. Only used if search_strategy is set to
             `"hnsw"`. You can find more information about this parameter in the
             [pgvector documentation](https://github.com/pgvector/pgvector?tab=readme-ov-file#hnsw)
+        :param language: The language to use for the full-text/hybrid search.
```
anakin87, line 122: let's add some information on where to find the supported languages.
I haven't found much
https://stackoverflow.com/questions/39751892/get-full-list-of-full-text-search-configuration-languages

integrations/pgvector/src/haystack_integrations/document_stores/pgvector/document_store.py
```diff
@@ -415,16 +433,6 @@ def _from_haystack_to_pg_documents(documents: List[Document]) -> List[Dict[str,
...
-            if "sparse_embedding" in db_document:
-                sparse_embedding = db_document.pop("sparse_embedding", None)
-                if sparse_embedding:
-                    logger.warning(
-                        "Document %s has the `sparse_embedding` field set,"
-                        "but storing sparse embeddings in Pgvector is not currently supported."
-                        "The `sparse_embedding` field will be ignored.",
-                        db_document["id"],
-                    )
-
```
anakin87, line 427: this should not be removed.
It is important for our upcoming SparseEmbedding support.
Since PgVector does not support this feature, this code block prevents failures.

integrations/pgvector/src/haystack_integrations/document_stores/pgvector/document_store.py
```diff
@@ -475,6 +483,56 @@ def delete_documents(self, document_ids: List[str]) -> None:
 
         self._execute_sql(delete_sql, error_msg="Could not delete documents from PgvectorDocumentStore")
 
+    def _keyword_retrieval(
+        self,
+        user_query: str,
+        top_k: int = 10,
+        filters: Optional[Dict[str, Any]] = None,
+        language: Optional[str] = "english",
```
anakin87, line 491:
```suggestion
```

I would remove this parameter.
Once the language is set, you cannot search using other languages.
Am I wrong?

anakin87 (commented):
I left some comments to improve the implementation.

Thank you for your work!

Ping me when some tests are in place.

jlonge4:
@anakin87 alright so I made the suggested edits from your comments, and I moved the create_keyword_index method into the keyword search function so that it wouldn't be applied unless needed. I feel like this is really close but something is up with my test. May need a bail out lol. Lastly, not quite sure how you would like to handle filters in this case since the keyword query is very different.

integrations/pgvector/src/haystack_integrations/document_stores/pgvector/document_store.py
jlonge4: In regard to this, I can't really find a web/documentation reference either. However when executing the query "SELECT cfgname FROM pg_ts_config;" I got the list - > 
cfgname
--
simple
danish
dutch
english
finnish
french
german
hungarian
italian
norwegian
portuguese
romanian
russian
spanish
swedish
turkish

jlonge4:
@anakin87 Scratch that I figured out what I had done wrong! Now all that I would need you to check out is how you prefer to handle filters.

jlonge4:
@anakin87 was also curious if it would be much easier to just perform hybrid retrieval by using each of the PGVector Retrievers once this is merged and combining results with the current Joiner to merge docs. Same result, much less code lol.

anakin87:
@jlonge4 sorry for the long wait
- I would create the keyword index when creating the document store (if the index does not exist) and not at retrieval time
- filters are converted into a WHERE clause, so in keyword retrieval, I would do something similar to embedding retrieval https://github.com/deepset-ai/haystack-core-integrations/blob/6a8297501d42c0a2b64dfc395da455530bdde013/integrations/pgvector/src/haystack_integrations/document_stores/pgvector/document_store.py#L542
- about hybrid retriever: not urgent; in the future, in another PR, we can create a `PgvectorHybridRetriever`, which would probably be more efficient, but I agree that a similar result can be achieved in Haystack with the Joiner

jlonge4:
@anakin87 take a look but I *think* we are good to move forward. Lmk what you are thinking!

jlonge4:
@anakin87 quite stumped on this one... even though the meta field is added, it appears there is no meta key at all in the result ```cls = <class 'haystack.dataclasses.document.Document'>
data = {'content': 'The quick brown fox chased the dog', 'dataframe': None, 'embedding': [0.1, 0.1, 0.1, 0.1, 0.1, 0.1, ...], 'id': '1f526f4f11b5ba7dd76b20ee314eb8fd688402fed738ed7fbeec33f433ec8811'}```

anakin87:
I will have a look... Thanks!

anakin87 (approved):
Everything looks good.
I'm going to merge this PR.

jlonge4:
@anakin87 thank you sir! And thanks for that last push!

PR #649, opened 2024-04-09
fix `haystack-ai` pin

anakin87:


masci (approved):
LGTM when tests pass (uptrain excluded, _ça va sans dire_)

Issue #642, opened 2024-04-04
QdrantDocumentStore issue


marygriffus:
**Describe the bug**
When creating a QdrantDocumentStore, if there is already a preexisting qdrant instance with data in it, and the params do not match precisely, `recreate_index` will destroy the old index and create a new one, blowing away the old data and making it so that incoming data does not match. Even after turning this off and updating settings to match our Qdrant params, I ran into an issue that seemed to be a mismatch between the QdrantDocumentStore and the QdrantClient.

**To Reproduce**
Bring up a qdrant instance and start a collection with the config:
```
{
  "params": {
    "vectors": {
      "fast-all-minilm-l6-v2": {
        "size": 384,
        "distance": "Cosine"
      }
    },
    "shard_number": 1,
    "replication_factor": 1,
    "write_consistency_factor": 1,
    "on_disk_payload": true
  },
  "hnsw_config": {
    "m": 16,
    "ef_construct": 100,
    "full_scan_threshold": 10000,
    "max_indexing_threads": 0,
    "on_disk": false
  },
  "optimizer_config": {
    ...
  },
  "wal_config": {
    ...
  },
  "quantization_config": null
}
```

Then instantiate a QdrantDocumentStore like below and use it in a pipeline.

```
        document_store = QdrantDocumentStore(
            url=qdrant_host,
            port=qdrant_port,
            index=fusion_payload.datastore,
            embedding_dim=384,
            similarity="cosine",
            recreate_index=False,
            hnsw_config={"m": 16, "ef_construct": 100}
        )
```

At first, I had `recreate_index=True` and left embedding_dim and hnsw_config blank, which blew my collection away and any new data failed to be added; ideally I think the document store would default to the settings discovered through the qdrant client. However, when I switched to `recreate_index=False`, I continued to have issues.

The first error I ran into with this setup was this:
```
  File "/src/app/.venv/lib/python3.11/site-packages/haystack_integrations/document_stores/qdrant/document_store.py", line 138, in __init__
    self._set_up_collection(index, embedding_dim, recreate_index, similarity, on_disk, payload_fields_to_index)
  File "/src/app/.venv/lib/python3.11/site-packages/haystack_integrations/document_stores/qdrant/document_store.py", line 389, in _set_up_collection
    current_distance = collection_info.config.params.vectors.distance
                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #666, opened 2024-04-16
Qdrant: add embedding retrieval example

anakin87:
Examples are missing for this integration.

I'm adding this simple example, also to help users figure out how to use Qdrant Document Store (see #642).


vblagoje (approved):
🚢

PR #672, opened 2024-04-22
add HF_TOKEN var to access gated models

masci:


anakin87 (approved):
https://github.com/deepset-ai/haystack-core-integrations/actions/runs/8776553405/job/24080220144

The tests began to fail because Mistral models became gated.

PR #675, opened 2024-04-22
Qdrant - add hybrid retriever

### LINKED ISSUE #664: Qdrant: hybrid retrieval
Follows #549. (@lambda-science) 
The following are just thoughts, nothing is set in stone.

I imagine having a `QdrantHybridRetriever` that takes `query_embedding` and `sparse_query_embedding` as input; returns Documents.

This can serve as an inspiration for implementation (note the Reciprocal Rank Fusion)
https://github.com/qdrant/qdrant-client/blob/8e3ea58f781e4110d11c0a6985b5e6bb66b85d33/qdrant_client/qdrant_fastembed.py#L519

anakin87:
fixes #664

Introduce a simple `QdrantHybridRetriever` that takes `query_embedding` and `query_sparse_embedding` as input; returns Documents.
Applies Reciprocal Rank Fusion, as defined in the Qdrant Python Client.

### DIFF
--- a/integrations/qdrant/src/haystack_integrations/document_stores/qdrant/document_store.py
+++ b/integrations/qdrant/src/haystack_integrations/document_stores/qdrant/document_store.py
@@ -16,6 +16,7 @@ from haystack.utils.filters import convert as convert_legacy_filters
 from qdrant_client import grpc
 from qdrant_client.http import models as rest
 from qdrant_client.http.exceptions import UnexpectedResponse
+from qdrant_client.hybrid.fusion import reciprocal_rank_fusion
 from tqdm import tqdm
 
 from .converters import (
@@ -307,7 +308,7 @@ class QdrantDocumentStore:
             )
         return documents
 
-    def query_by_sparse(
+    def _query_by_sparse(
         self,
         query_sparse_embedding: SparseEmbedding,
         filters: Optional[Dict[str, Any]] = None,
@@ -349,7 +350,7 @@ class QdrantDocumentStore:
                 document.score = score
         return results
 
-    def query_by_embedding(
+    def _query_by_embedding(
         self,
         query_embedding: List[float],
         filters: Optional[Dict[str, Any]] = None,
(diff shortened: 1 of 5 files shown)


integrations/qdrant/src/haystack_integrations/document_stores/qdrant/document_store.py
```diff
@@ -383,6 +384,77 @@ def query_by_embedding(
...
+        top_k: int = 10,
+        return_embedding: bool = False,
+    ) -> List[Document]:
+        """
+        Retrieves documents based on both dense and sparse embeddings
+         and fuses the results using Reciprocal Rank Fusion.
+
+        This method is not mean to be part of the public interface of
+        `QdrantDocumentStore` nor called directly.
+        `QdrantHybridRetriever` uses this method directly and is the public interface for it.
```
masci, line 401:
```suggestion
        This method is not part of the public interface of `QdrantDocumentStore` and shouldn't be used directly.
        Use the `QdrantHybridRetriever` instead.
```

integrations/qdrant/src/haystack_integrations/document_stores/qdrant/document_store.py
```diff
@@ -383,6 +384,77 @@ def query_by_embedding(
...
+        :param query_embedding: Dense embedding of the query.
+        :param query_sparse_embedding: Sparse embedding of the query.
+        :param filters: Filters applied to the retrieved Documents.
+        :param top_k: Maximum number of Documents to return.
+        :param return_embedding: Whether to return the embeddings of the retrieved documents.
+
+        :returns: List of Document that are most similar to `query_embedding` and `query_sparse_embedding`.
+        """
+
+        # this implementation is based on that of Python Qdrant client
```
masci, line 412:
```suggestion
        # This implementation is based on the code from the Python Qdrant client:
```

integrations/qdrant/src/haystack_integrations/document_stores/qdrant/document_store.py
```diff
@@ -383,6 +384,77 @@ def query_by_embedding(
...
+        if not self.use_sparse_embeddings:
+            message = (
+                "You are trying to query using sparse embeddings, but the Document Store "
+                "was initialized with `use_sparse_embeddings=False`. "
+            )
+            raise QdrantStoreError(message)
+
+        qdrant_filters = convert_filters_to_qdrant(filters)
+
+        sparse_request = rest.SearchRequest(
```
masci, line 423: Should we catch exceptions here, and `raise from`? Not sure if a finer grain is helpful so feel free to ignore

integrations/qdrant/tests/test_document_store.py
```diff
@@ -1,25 +1,41 @@
...
     DeleteDocumentsTest,
     WriteDocumentsTest,
+    _random_embeddings,
 )
-from haystack_integrations.document_stores.qdrant import QdrantDocumentStore
+from haystack_integrations.document_stores.qdrant.document_store import QdrantDocumentStore, QdrantStoreError
 
 
-class TestQdrantStoreBaseTests(CountDocumentsTest, WriteDocumentsTest, DeleteDocumentsTest):
+def _generate_mocked_sparse_embedding(n):
```
masci, line 18: let's move this to `conftest.py` and make it a fixture, we shouldn't import from test modules

masci (changes requested):
Left a few comments, mostly docs edits

integrations/qdrant/tests/conftest.py
```diff
@@ -0,0 +1,18 @@
...
+    This fixture generates a random SparseEmbedding object each time it is used.
+    """
+
+    def _generate_random_sparse_embedding():
+        random_indice_length = np.random.randint(3, 15)
+        indices = list(range(random_indice_length))
+        values = [np.random.random_sample() for _ in range(random_indice_length)]
+        return SparseEmbedding(indices=indices, values=values)
+
+    return _generate_random_sparse_embedding
```
masci, line 18: Why returning a callable instead of the `SparseEmbedding` directly?

integrations/qdrant/tests/conftest.py
```diff
@@ -0,0 +1,18 @@
+import numpy as np
+import pytest
+from haystack.dataclasses import SparseEmbedding
+
+
+@pytest.fixture
```
masci, line 6: I think we can scope it by session? Any reason why you would want different vectors for each different test function?

integrations/qdrant/src/haystack_integrations/document_stores/qdrant/document_store.py
anakin87: Now I tried doing this for the cases that make the most sense to me

integrations/qdrant/tests/conftest.py
masci: Resolved offline

PR #679, opened 2024-04-23
chore: move uptrain integration to staging

masci:
The `uptrain` integration will be adopted by Uptrain. During the transition the PyPI package will be available via `pip install uptrain-haystack` as usual.


julian-risch (approved):
LGTM! 👍

PR #680, opened 2024-04-23
chore: add license classifiers

### LINKED ISSUE #650: Add license for core integrations on their respective PYPI package website
Add license in sidebar on https://pypi.org/project/llama-cpp-haystack/
instead of adding it in description of it.

It will be helpful in automation process to find the license from there instead of having it in description.
Pls pls

masci:
Fixes #650 

- Once this PR gets merged, the change will be visible after the first release of each package.
- I propose we don't release the packages only for this reason

### DIFF
--- a/integrations/amazon_bedrock/pyproject.toml
+++ b/integrations/amazon_bedrock/pyproject.toml
@@ -14,6 +14,7 @@ authors = [
   { name = "deepset GmbH", email = "info@deepset.ai" },
 ]
 classifiers = [
+  "License :: OSI Approved :: Apache Software License",
   "Development Status :: 4 - Beta",
   "Programming Language :: Python",
   "Programming Language :: Python :: 3.8",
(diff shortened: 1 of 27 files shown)


integrations/amazon_bedrock/pyproject.toml
```diff
@@ -14,6 +14,7 @@ authors = [
   { name = "deepset GmbH", email = "info@deepset.ai" },
 ]
 classifiers = [
+  "License :: OSI Approved :: Apache Software License",
```
shadeMe, line 17: Does it make sense to mention the version (v2.0) here as well?

shadeMe (approved):
LGTM! One minor comment.

integrations/amazon_bedrock/pyproject.toml
masci: That wouldn't be standard, see https://pypi.org/classifiers/

PR #683, opened 2024-04-24
fix: pass namespace in the docstore init

### LINKED ISSUE #682: Add namespace parameter to AstraDB integration
When you use a database for different clients, the correct approach is to create a different namespace for each client. This is extremely useful for scaling and keeping the clients isolated.

**Additional context**
I have noticed in the code that you have an AstraClient class where the namespace already exists. I think it would be relatively quick to include namespace parameters in the AstraDocumentStore class.

masci:
Fixes #682 

It's now possible to pass the namespace to the document store constructor.

### DIFF
--- a/integrations/astra/src/haystack_integrations/document_stores/astra/document_store.py
+++ b/integrations/astra/src/haystack_integrations/document_stores/astra/document_store.py
@@ -55,6 +55,7 @@ class AstraDocumentStore:
         embedding_dimension: int = 768,
         duplicates_policy: DuplicatePolicy = DuplicatePolicy.NONE,
         similarity: str = "cosine",
+        namespace: Optional[str] = None,
     ):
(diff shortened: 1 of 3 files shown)


integrations/astra/src/haystack_integrations/document_stores/astra/document_store.py
```diff
@@ -99,13 +100,16 @@ def __init__(
         self.embedding_dimension = embedding_dimension
         self.duplicates_policy = duplicates_policy
         self.similarity = similarity
+        if namespace:
+            self.namespace = namespace
```
anakin87, line 104: should we serialize this param in `to_dict`?

anakin87 (approved):
Looks good.

Feel free to merge if the tests pass

PR #688, opened 2024-04-24
`FastembedSparseTextEmbedder` - remove `batch_size`

anakin87:
`FastembedTextEmbedder` can only accept and process a string at a time.
`batch_size` parameter is practically ignored and misleading -> I'm removing it.

(Other minor docstrings improvements)


shadeMe:
@masci Do we have a deprecation policy for core integrations? This is a breaking change.

anakin87:
I agree that formally this is a breaking change.

However, this component was released 2 weeks ago and it has not yet been documented or promoted.
So it should not impact many users.

In any case, let's wait for Massi's opinion.

anakin87:
I talked with Massi: breaking changes are allowed in major releases.

Once this PR is merged, I will create a major release of `fastembed-haystack` (1.0.0).

When we have release notes for core integrations, everything will be clearer.

shadeMe (approved):

PR #692, opened 2024-04-25
Allow vanilla qdrant filters

jdb78:
Allow passing a qdrant filter object instead of a haystack filter object - this is useful when the haystack filter object does not support the use case, e.g. when to use full text search and when to match exactly


silvanocerza:
I'm a bit unsure about this change to be fair. 

I understand the need obviously, though we'd need to change also the `filter_documents()` signature to reflect this. The retrievers too I guess. It's not enough to change the conversion function.

I'm not sure whether linting would be happy though, changing the signature like this might mean breaking the [Document Store Protocol](https://docs.haystack.deepset.ai/docs/creating-custom-document-stores#documentstore-protocol) for some linters probably. 🤔 

I would probably do this only for the retrievers as most of the times they don't use `filter_documents()` directly to retriever documents.

jdb78:
updated the signatures

silvanocerza (approved):

PR #700, opened 2024-04-30
Feature/bump chromadb dep to 0.5.0

jongirard:
## Description

Bump version of chromadb to latest (0.5.0) which contains several improvements, for example, the addition of new [embedding functions](https://github.com/chroma-core/chroma/blob/0.5.0/chromadb/utils/embedding_functions.py#L966) such as `OllamaEmbeddingFunction`.

Previously version was pinned to `<0.4.20` due to broken filtering tests. This appears to have been caused due to a variance in `assert_documents_are_equal` expected/received data (unsorted list) which has now been resolved in this PR.


integrations/chroma/pyproject.toml
```diff
@@ -26,7 +26,7 @@ classifiers = [
 ]
 dependencies = [
   "haystack-ai",
-  "chromadb<0.4.20",  #  FIXME: investigate why filtering tests broke on 0.4.20
+  "chromadb>=0.5.0",
```
anakin87, line 29:
```suggestion
  "chromadb",
```

I would remove the pin to a specific version.

masci (approved):
Thanks!

PR #701, opened 2024-04-30
Update Nvidia integration to support new endpoints

### LINKED ISSUE #696: Add support for new NVidia APIs
NVidia integration should be able to be used with their new OpenAI compatible API on https://build.nvidia.com/explore/discover

silvanocerza:
Fixes #696.

This PR updates both Nvidia generator and embedders to support the new [API catalog endpoints](https://build.nvidia.com).

These new endpoints require a different API key from the previous ones so I opted to use an `NVIDIA_CATALOG_API_KEY` env var in tests to differentiate it from the previous one.

### DIFF
--- a/integrations/nvidia/pyproject.toml
+++ b/integrations/nvidia/pyproject.toml
@@ -117,7 +117,6 @@ unfixable = [
   # Don't touch unused imports
   "F401",
 ]
-extend-exclude = ["tests", "example"]
 
 [tool.ruff.isort]
 known-first-party = ["src"]
--- a/integrations/nvidia/src/haystack_integrations/components/embedders/nvidia/_nim_backend.py
+++ b/integrations/nvidia/src/haystack_integrations/components/embedders/nvidia/_nim_backend.py
@@ -1,6 +1,7 @@
 from typing import Any, Dict, List, Optional, Tuple
 
 import requests
+from haystack.utils.auth import Secret
 
 from .backend import EmbedderBackend
 
@@ -12,12 +13,17 @@ class NimBackend(EmbedderBackend):
         self,
         model: str,
         api_url: str,
+        api_key: Optional[Secret] = None,
         model_kwargs: Optional[Dict[str, Any]] = None,
     ):
         headers = {
             "Content-Type": "application/json",
             "accept": "application/json",
         }
+
+        if api_key:
+            headers["authorization"] = f"Bearer {api_key.resolve_value()}"
+
         self.session = requests.Session()
         self.session.headers.update(headers)
 
--- a/integrations/nvidia/src/haystack_integrations/components/embedders/nvidia/_nim_backend.py
+++ b/integrations/nvidia/src/haystack_integrations/components/embedders/nvidia/_nim_backend.py
@@ -1,7 +1,7 @@
 from typing import Any, Dict, List, Optional, Tuple
 
 import requests
-from haystack.utils.auth import Secret
+from haystack.utils import Secret
 
 from .backend import EmbedderBackend
 
--- a/integrations/nvidia/src/haystack_integrations/components/embedders/nvidia/_nvcf_backend.py
+++ b/integrations/nvidia/src/haystack_integrations/components/embedders/nvidia/_nvcf_backend.py
@@ -1,3 +1,4 @@
+import warnings
 from dataclasses import asdict, dataclass
 from typing import Any, Dict, List, Literal, Optional, Tuple, Union
 
@@ -17,6 +18,7 @@ class NvcfBackend(EmbedderBackend):
         api_key: Secret,
         model_kwargs: Optional[Dict[str, Any]] = None,
     ):
+        warnings.warn("Nvidia NGC is deprecated, use Nvidia NIM instead.", DeprecationWarning, stacklevel=2)
         if not model.startswith("playground_"):
             model = f"playground_{model}"
 
--- a/integrations/nvidia/src/haystack_integrations/components/embedders/nvidia/text_embedder.py
+++ b/integrations/nvidia/src/haystack_integrations/components/embedders/nvidia/text_embedder.py
@@ -77,7 +77,9 @@ class NvidiaTextEmbedder:
 
             self.backend = NvcfBackend(self.model, api_key=self.api_key, model_kwargs={"model": "query"})
         else:
-            self.backend = NimBackend(self.model, api_url=self.api_url, model_kwargs={"input_type": "query"})
+            self.backend = NimBackend(
+                self.model, api_url=self.api_url, api_key=self.api_key, model_kwargs={"input_type": "query"}
+            )
 
         self._initialized = True
 
--- a/integrations/nvidia/src/haystack_integrations/components/embedders/nvidia/text_embedder.py
+++ b/integrations/nvidia/src/haystack_integrations/components/embedders/nvidia/text_embedder.py
@@ -1,4 +1,4 @@
-from typing import Any, Dict, List, Literal, Optional
+from typing import Any, Dict, List, Optional
 
 from haystack import component, default_from_dict, default_to_dict
 from haystack.utils import Secret, deserialize_secrets_inplace
@@ -6,6 +6,7 @@ from haystack.utils import Secret, deserialize_secrets_inplace
 from ._nim_backend import NimBackend
 from ._nvcf_backend import NvcfBackend
 from .backend import EmbedderBackend
+from .truncate import TruncateMode
 
 
 @component
@@ -38,7 +39,7 @@ class NvidiaTextEmbedder:
         api_url: Optional[str] = None,
         prefix: str = "",
         suffix: str = "",
-        truncate: Literal["NONE", "START", "END"] = "NONE",
+        truncate: Optional[TruncateMode] = None,
     ):
         """
         Create a NvidiaTextEmbedder component.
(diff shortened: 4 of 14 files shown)


mattf (commented):
the concept of NIM is a service + model that can be deployed anywhere. we host some on https://build.nvidia.com and users can export them to run on their own infrastructure. we happen to use NVIDIA Cloud Functions (nvcf) to serve the hosted NIMs.

the generator nims follow a `/chat/completion` api (e.g. https://docs.api.nvidia.com/nim/reference/meta-llama3-8b-infer), the embedding nims follow https://docs.api.nvidia.com/nim/reference/nvidia-embedding-2b-infer

1. api keys
   - we aim to let users provide param `nvidia_api_key` -> param `api_key` -> env `NVIDIA_API_KEY`, where `nvidia_api_key` takes precedence over `api_key` which takes precedence over `NVIDIA_API_KEY`
   - local nim does not require an api key
   - it is recommended to use environment for keys
2. mode switching for hosted and local nim w/ `mode()`
   - parameters are a mode name, `literal["nvidia", "nim"]`, an api key for nvidia mode and `base_url` for nim mode
   - nvidia is the default mode
   - e.g. NVIDIAGenerator().mode("nim", base_url="http://.../v1") - we're including `/v1`, but arguably shouldn't. we're not including `/chat/completion` or `/embedding` because the service should provide `/models` at the same base url
3. https://catalog.ngc.nvidia.com/ai-foundation-models are deprecated, e.g. `nvolveqa_40k`, users should use models from https://build.nvidia.com, e.g. https://docs.api.nvidia.com/nim/reference/nvidia-embed-qa-4
4. Embedder.prefix/suffix, these aren't common for NIM embedding models, are they idiomatic for haystack?
5. embedding models support a `truncate` parameter for service-side truncation, see https://docs.api.nvidia.com/nim/reference/nvidia-embedding-2b-infer

integrations/nvidia/pyproject.toml
```diff
@@ -117,7 +117,6 @@ unfixable = [
   # Don't touch unused imports
   "F401",
 ]
-extend-exclude = ["tests", "example"]
```
shadeMe, line 120: Why are we including tests in the linting? It's more of a hassle than anything else.

integrations/nvidia/src/haystack_integrations/components/embedders/nvidia/_nim_backend.py
```diff
@@ -1,6 +1,7 @@
 from typing import Any, Dict, List, Optional, Tuple
 
 import requests
+from haystack.utils.auth import Secret
```
shadeMe, line 4:
```suggestion
from haystack.utils import Secret
```

integrations/nvidia/src/haystack_integrations/components/generators/nvidia/_nim_backend.py
```diff
@@ -1,6 +1,7 @@
 from typing import Any, Dict, List, Optional, Tuple
 
 import requests
+from haystack.utils.auth import Secret
```
shadeMe, line 4:
```suggestion
from haystack.utils import Secret
```

integrations/nvidia/src/haystack_integrations/components/generators/nvidia/_nim_backend.py
```diff
@@ -26,8 +32,9 @@ def __init__(
         self.model_kwargs = model_kwargs or {}
 
     def generate(self, prompt: str) -> Tuple[List[str], List[Dict[str, Any]]]:
-        # We're using the chat completion endpoint as the local containers don't support
+        # We're using the chat completion endpoint as the NVidia API doesn't support
```
shadeMe, line 35:
```suggestion
        # We're using the chat completion endpoint as the NIM API doesn't support
```

shadeMe (commented):
Just a couple of minor changes.

integrations/nvidia/pyproject.toml
silvanocerza: I think it makes sense to have some linting in tests, at least to keep formatting consistent. If I'm not mistaken all other integration have it enabled for tests too, same for core.

Also I enabled it cause [tests were failing in CI](https://github.com/deepset-ai/haystack-core-integrations/actions/runs/8892387873/job/24416337070#step:6:38) and I didn't catch that with auto formatting.

We could disable checks for tests but I would very much prefer to keep tests formatting tidy at least. I do agree that type checking can go out of the window for tests. This just enables `ruff` though.

shadeMe:
@mattf Thanks very much for your comments. A few questions/comments from my side:

>     1. api keys
>        
>        * we aim to let users provide param `nvidia_api_key` -> param `api_key` -> env `NVIDIA_API_KEY`, where `nvidia_api_key` takes precedence over `api_key` which takes precedence over `NVIDIA_API_KEY`
>        * local nim does not require an api key
>        * it is recommended to use environment for keys

Could you expand on that a bit more? Why is there a need for two parameters that accept an authentication secret?


>     2. mode switching for hosted and local nim w/ `mode()`
>        
>        * parameters are a mode name, `literal["nvidia", "nim"]`, an api key for nvidia mode and `base_url` for nim mode
>        * nvidia is the default mode
>        * e.g. NVIDIAGenerator().mode("nim", base_url="http://.../v1") - we're including `/v1`, but arguably shouldn't. we're not including `/chat/completion` or `/embedding` because the service should provide `/models` at the same base url
> 

Correct me if I'm wrong, but it sounds like you're suggesting the above to make the `api_key` and `base_url` parameters mutually exclusive? Haystack components don't modify their (user-facing) state after initialization, so a `mode` method would not be suitable here. 

>     3. [catalog.ngc.nvidia.com/ai-foundation-models](https://catalog.ngc.nvidia.com/ai-foundation-models) are deprecated, e.g. `nvolveqa_40k`, users should use models from [build.nvidia.com](https://build.nvidia.com), e.g. [docs.api.nvidia.com/nim/reference/nvidia-embed-qa-4](https://docs.api.nvidia.com/nim/reference/nvidia-embed-qa-4)
> 

Okay, so we can safely remove support for that "backend" then? We call it the `NvCfBackend` in our code.

>     4. Embedder.prefix/suffix, these aren't common for NIM embedding models, are they idiomatic for haystack?
> 

This depends on how the embedding model is trained. Some wrap their inputs in special markers/meta tokens to indicate their semantics, i.e., is the text part of a query or a passage. As such, those parameters are standard in Haystack embedding components.

>     5. embedding models support a `truncate` parameter for service-side truncation, see [docs.api.nvidia.com/nim/reference/nvidia-embedding-2b-infer](https://docs.api.nvidia.com/nim/reference/nvidia-embedding-2b-infer)

Thanks for the tip; we can add support for that.

integrations/nvidia/pyproject.toml
shadeMe: Okay if we just stick to `black`/formatting and keep `mypy`/`pylint` disabled.

integrations/nvidia/src/haystack_integrations/components/embedders/nvidia/_nvcf_backend.py
```diff
@@ -17,6 +18,7 @@ def __init__(
         api_key: Secret,
         model_kwargs: Optional[Dict[str, Any]] = None,
     ):
+        warnings.warn("Nvidia NGC is deprecated, use Nvidia NIM instead.", DeprecationWarning, stacklevel=2)
```
silvanocerza, line 21: I'll add a link to the documentation after it's published.

integrations/nvidia/src/haystack_integrations/components/embedders/nvidia/text_embedder.py
```diff
@@ -52,13 +54,20 @@ def __init__(
             A string to add to the beginning of each text.
         :param suffix:
             A string to add to the end of each text.
+        :param truncate:
+            Specifies how inputs longer that the maximum token length should be truncated.
+            If START, the input will be truncated from the start.
+            If END, the input will be truncated from the end.
```
shadeMe, line 60: Let's move this to the enum's docstring(s).

integrations/nvidia/src/haystack_integrations/components/embedders/nvidia/text_embedder.py
```diff
@@ -52,13 +54,20 @@ def __init__(
...
+            If None an error will be raised if the input is too long.
+            Defaults to None.
```
shadeMe, line 62:
```suggestion
```

integrations/nvidia/src/haystack_integrations/components/embedders/nvidia/truncate.py
```diff
@@ -0,0 +1,30 @@
+from enum import Enum
+
+
+class TruncateMode(Enum):
```
shadeMe, line 4:
```suggestion
class EmbeddingTruncateMode(Enum):
```
Since we don't use it for the generator.

integrations/nvidia/src/haystack_integrations/components/embedders/nvidia/truncate.py
```diff
@@ -0,0 +1,30 @@
...
+    """
+    Enum used to specify truncation mode for embeddings.
```
shadeMe, line 6:
```suggestion
    Specifies how inputs to the NVIDIA embedding components are truncated.
```

integrations/nvidia/src/haystack_integrations/components/embedders/nvidia/text_embedder.py
```diff
@@ -38,6 +39,7 @@ def __init__(
         api_url: Optional[str] = None,
         prefix: str = "",
         suffix: str = "",
+        truncate: Optional[TruncateMode] = None,
```
shadeMe, line 42: Let's make this a `str|TruncateMode` union and convert the former to the latter.

shadeMe (changes requested):

shadeMe (approved):

PR #702, opened 2024-04-30
Update Cohere integration to use Cohere SDK v5

malexw:
Hello! 👋 

This PR updates the Cohere integration to use v5 of the Cohere SDK. No behaviour should be changed, though I'm not sure of the best way to perform integration tests to verify that. I've added a few comments to the PR to explain the reasons for some of the changes.


integrations/cohere/src/haystack_integrations/components/embedders/cohere/document_embedder.py
```diff
@@ -8,7 +8,7 @@
 from haystack.utils import Secret, deserialize_secrets_inplace
 from haystack_integrations.components.embedders.cohere.utils import get_async_response, get_response
 
-from cohere import COHERE_API_URL, AsyncClient, Client
+from cohere import AsyncClient, Client
```
malexw, line 11: `COHERE_API_URL` is no longer exported from the SDK. The Cohere team recommends using a string literal in the code to replace it.

integrations/cohere/src/haystack_integrations/components/embedders/cohere/utils.py
```diff
@@ -24,20 +24,14 @@ async def get_async_response(cohere_async_client: AsyncClient, texts: List[str],
...
+    response = await cohere_async_client.embed(texts=texts, model=model_name, input_type=input_type, truncate=truncate)
+    if response.meta is not None:
+        metadata = response.meta
+    for emb in response.embeddings:
+        all_embeddings.append(emb)
 
-    except CohereError as error_response:
-        msg = error_response.message
-        raise ValueError(msg) from error_response
+    return all_embeddings, metadata
```
malexw, line 34: I've removed the `try / except` block around this code entirely since the SDK no longer exports `CohereError`. Is there a reason for trapping exceptions from the SDK and converting them to `ValueError`s, or is it OK to let those exceptions bubble up?

julian-risch: The .embed code can [raise a couple of different exceptions:](https://github.com/cohere-ai/cohere-python/blob/main/src/cohere/base_client.py#L2649) BadRequestError, TooManyRequestsError, InternalServerError, JSONDecodeError, ApiError. They are all not ValueErrors but inherit from ApiError and Exception: https://github.com/cohere-ai/cohere-python/blob/437fdc65e969bf6e69c2639ae906671912436c5f/src/cohere/core/api_error.py#L6
An application built with this Haystack integration now needs to look out for other types of exceptions but this breaking change is the only disadvantage I see. Your change makes the raised exceptions more informative, which is helpful. It wasn't an actual ValueError before.

integrations/cohere/src/haystack_integrations/components/embedders/cohere/document_embedder.py
julian-risch: I checked the examples we have in our documentation pages and none of them sets `COHERE_API_URL` so there are no changes required in the documentation pages.

julian-risch (approved):
Looks very good to me! 👍 @malexw congrats on your first contribution to this repo! Regarding integration tests: We do have them so I am confident that the upgrade does not change the behavior. For example, here is one of the integration tests: https://github.com/deepset-ai/haystack-core-integrations/blob/50634337a86290f7eb935daffc245d45d1508b66/integrations/cohere/tests/test_cohere_ranker.py#L327 They run if you have COHERE_API_KEY or CO_API_KEY environment variables set.

PR #704, opened 2024-04-30
Type hints in pgvector document store updated for 3.8 compability

Ezhvsalate:
Without this fix code can not be run on Py3.8 although it's in list of supported versions


anakin87 (approved):
Thanks for the contribution!

PR #705, opened 2024-05-01
Implement filters for chromaQueryTextRetriever via existing haystack filters logic

jongirard:
## Description

Implement filters for chromaQueryTextRetriever via existing haystack filters logic. Utilizes existing strategy/methods to implement haystack filters via https://docs.trychroma.com/reference/Collection#query as the API is the same (both `search` and `search_embeddings` use collection.query directly with `query_texts`/`query_embeddings` argument set respectively).


shadeMe (dismissed):
Thanks for the PR! Could you please add a few tests for the new functionality?

masci (approved):

PR #707, opened 2024-05-01
fix: Weaviate schema class name conversion which preserves PascalCase

### LINKED ISSUE #706: Weaviate document store schema mangled by enforced capitalization
**Describe the bug**
Weaviate document store schema mangled by enforced capitalization of schema class name. The class name specified in the Weaviate schema is [capitalized in the WeaviateDocumentStore constructor](https://github.com/deepset-ai/haystack-core-integrations/blob/main/integrations/weaviate/src/haystack_integrations/document_stores/weaviate/document_store.py#L169) which changes the casing of the class name. The resulting Weaviate schema is then inconsistent with the expected result of the class name exactly matching the input.

This is an overreach of responsibilities by the `WeaviateDocumentStore` constructor. It should not be mutating the user input. If the input is invalid, i.e., the class name does not start with a capital letter, then a `ValueError` should be raised instead.

**To Reproduce**
Pass a schema in the Weaviate collection settings with a class name in PascalCase such as `CaseDocument`. This results in a class being added to the Weaviate schema with the name `Casedocument` instead of `CaseDocument` as expected.

 - OS: Ubuntu-22.04 in WSL2
 - Haystack version: 2.0.1
 - Integration version: weaviate-haystack 2.0.0

chrisk314:
Fixes #706 by implementing a class name conversion which capitalizes the first letter but otherwise preserves casing of the remainder of the class name string.

It should be noted that raising a `ValueError` for class names which do not start with a capital letter would be preferable; the `WeaviateDocumentStore` constructor should not really be mutating the user input. However, this would technically be a breaking change at this point so is probably less appealing.

### DIFF
--- a/integrations/weaviate/src/haystack_integrations/document_stores/weaviate/document_store.py
+++ b/integrations/weaviate/src/haystack_integrations/document_stores/weaviate/document_store.py
@@ -166,7 +166,9 @@ class WeaviateDocumentStore:
             }
         else:
             # Set the class if not set
-            collection_settings["class"] = collection_settings.get("class", "default").capitalize()
+            _class_name = collection_settings.get("class", "Default")
+            _class_name = _class_name[0].upper() + _class_name[1:]
+            collection_settings["class"] = _class_name
             # Set the properties if they're not set
             collection_settings["properties"] = collection_settings.get("properties", DOCUMENT_COLLECTION_PROPERTIES)
 


davidsbatista:
Hi @chrisk314,

Thanks for your contribution, looks good; I'm just going to add a test.

davidsbatista (approved):
LGTM

PR #711, opened 2024-05-03
Follow up: update Cohere integration to use Cohere SDK v5

masci:
https://github.com/deepset-ai/haystack-core-integrations/pull/702 was not complete and integration tests were failing.

I removed the mocked unit tests, they are not helpful and in this case hid the underlying problem.


julian-risch (approved):
Looks good to me. 👍 Thanks again! I ran the integration tests locally now too.

PR #713, opened 2024-05-03
chore: sync integrations docs with all the available Haystack versions

masci:
Sync docs with all the Haystack versions available on Readme.

### DIFF
--- a/.github/workflows/CI_readme_sync.yml
+++ b/.github/workflows/CI_readme_sync.yml
@@ -18,6 +18,11 @@ env:
 jobs:
   sync:
     runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      max-parallel: 1
+      matrix:
+        hs-docs-version: ['2.0', '2.1', '2.2-unstable']
     steps:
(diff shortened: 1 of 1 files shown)


.github/workflows/CI_readme_sync.yml
```diff
@@ -16,8 +16,26 @@ env:
...
+          echo "versions=$VERSIONS" >> "$GITHUB_OUTPUT"
+
   sync:
     runs-on: ubuntu-latest
+    needs: get-versions
+    strategy:
+      fail-fast: false
+      max-parallel: 1
+      matrix:
+        hs-docs-version: ${{ needs.get-versions.outputs.versions }}
```
silvanocerza, line 38: I think this is necessary. 🤔 

```suggestion
        hs-docs-version: ${{ fromJSON(needs.get-versions.outputs.versions) }}
```

masci: Ah yeah, that's JSON

silvanocerza (approved):

PR #716, opened 2024-05-06
fix: Fix langfuse nightly tests

vblagoje:
* Fixes langfuse nightly tests

The only test, so far, we have is integration test so let's run all unit/integration tests

Resolves https://github.com/deepset-ai/haystack-core-integrations/actions/runs/8962156252


masci:
Why did we merge an integration with no tests in the first place?

shadeMe (approved):

PR #719, opened 2024-05-06
chore: pass the haystack docs version when generating docs

masci:
Pass the Haystack docs version on readme when generating docs.


anakin87 (approved):

julian-risch:
I guess my review is not needed here anymore

PR #731, opened 2024-05-13
fix: align tests and docs on NVIDIA_API_KEY (instead of NVIDIA_CATALOG_API_KEY)

mattf:
README.md says `Note: integration tests will be skipped unless the env var NVIDIA_API_KEY is set.`

with this MR, the statement is now true for all cases.

### DIFF
--- a/integrations/nvidia/tests/test_document_embedder.py
+++ b/integrations/nvidia/tests/test_document_embedder.py
@@ -388,15 +388,15 @@ class TestNvidiaDocumentEmbedder:
             assert isinstance(doc.embedding[0], float)
 
     @pytest.mark.skipif(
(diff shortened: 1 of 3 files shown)


silvanocerza:
I think this is going to break integration tests if merged. They're not going to run in this PR since it comes from your fork and env vars won't be set.

The different env var is necessary since we're testing different services that need different API keys. This is the only way to differentiate between them as of now.

Best thing would be to update the `README.md` to reflect the requirement.

silvanocerza:
Oh cool, nice to see this change. I remember trying it in the past and it wasn't working.

I'll approve and merge this as soon as you sign the CLA. 👍

dglogo:
lgtm! Thanks @mattf

silvanocerza (approved):

PR #734, opened 2024-05-14
make hosted nim default

mattf:
all inference calls should be directed to integrate.api.nvidia.com. the nvcf inference is deprecated and will be removed.

this allows for users to get on integrate.api.nvidia.com now and avoid any disruptions.


dglogo:
lgtm @mattf! Thanks!

PR #739, opened 2024-05-18
feat: Improve `OpenSearchDocumentStore.__init__` arguments

### LINKED ISSUE #671: Improve init arguments of `OpenSearchDocumentStore`

I think that the current implementation of `OpenSearchDocumentStore` should be more specific about which arguments it accepts in the `__init__` method. 

- Some specific keys extracted from the `kwargs`(e.g. `"embedding_dim"`, `"method"`, `"settings"`) are not mentioned in the docstring. The docstring says that the `kwargs` are the ones that the `OpenSearch` client takes. AFAIK (please correct me if I'm wrong) the ones extracted are not common `kwargs` required by the client. For example, they are used when creating an index (`client.indices.create()`).
- It'd be nice to have also  a [`max_chunk_bytes` argument](https://github.com/opensearch-project/opensearch-py/blob/main/opensearchpy/helpers/actions.py#L308) to pass to the internal `bulk` function. The default 100MB may be too big for certain Openserach instances. In these cases, the current implementation raises an error.

I suggest adding these arguments explicitly to the `__init__` and to improve its docstring.

e.g.
```python
def __init__(
        self,
        *,
        index: str = "default",
        embedding_dim: int = ...,
        method: ...,
        max_chunk_bytes: ...,
        **kwargs,
    ):
```

**Additional context**
Happy to make a PR if you agree on this proposal :)

EdAbati:
Closes #671 

As explained in the issue, I am proposing to change a bit the arguments that `OpenSearchDocumentStore` accepts.

The goal is:
1. First, to allow the user to set the most appropriate `max_chunk_bytes`. (The current default makes this component fail for certain deployments)
2. To be more transparent on which arguments are required and which `**kwargs` are passed to the `OpenSearch` client. Currently the docstring does not specify which `**kwargs` are passed to the client, and which are passed to `self._client.indices.create`.

Happy to hear your thoughts on how I can improve this PR :)


vblagoje:
Thanks for this contribution @EdAbati we'll be getting back to you shortly. @tstadel how does this proposal sound to you?

integrations/opensearch/src/haystack_integrations/document_stores/opensearch/document_store.py
```diff
@@ -101,6 +127,7 @@ def to_dict(self) -> Dict[str, Any]:
             self,
             hosts=self._hosts,
             index=self._index,
+            max_chunk_bytes=self._max_chunk_bytes,
             **self._kwargs,
         )
```
tstadel, line 132: As the now explicitly added kwargs `embedding_dim`, `method`, `mappings` and `settings` are not part of `self._kwargs` anymore, we have to treat them separately here.
Can we please add them here too and add a proper test that ensures any init args are serialized correctly?

integrations/opensearch/src/haystack_integrations/document_stores/opensearch/document_store.py
```diff
@@ -174,6 +201,7 @@ def write_documents(self, documents: List[Document], policy: DuplicatePolicy = D
             refresh="wait_for",
             index=self._index,
             raise_on_error=False,
+            max_chunk_bytes=self._max_chunk_bytes,
```
tstadel, line 204: `bulk` is also used in `delete_documents`. Can you please add `max_chunk_bytes` there for consistency too?

tstadel:
@vblagoje contribution looks like a nice improvement. @EdAbati thanks for raising this. I left two comments that should be tackled before merging.

vblagoje:
@EdAbati have you had a chance to look at the potential issues pointed by Thomas?

EdAbati:
Hi both thank you for the review. :) I am a bit busy this week, I'll go back to this PR next week

integrations/opensearch/src/haystack_integrations/document_stores/opensearch/document_store.py
EdAbati: good point! I added them now

EdAbati:
Hi, I made some changes to fix what was suggested :)
I can see a mypy error but only in the python 3.9 CI. do you know what is causing it and/or how I could solve it?

integrations/opensearch/tests/test_document_store.py
```diff
@@ -161,7 +199,7 @@ def test_bm25_retrieval_pagination(self, document_store: OpenSearchDocumentStore
                 Document(content="Exilir is a functional programming language"),
                 Document(content="F# is a functional programming language"),
                 Document(content="C# is a functional programming language"),
-                Document(content="C++ is an object oriented programming language"),
```
tstadel, line 164: Was this changed by accident?

masci:
Tests are green 👍

integrations/opensearch/tests/test_document_store.py
EdAbati: Oops, definetly 😅 apologies

vblagoje:
🙏 @masci and thanks a lot for this contribution @EdAbati , looks ok to you now @tstadel ?

tstadel (approved):
@EdAbati thanks for contributing 
@vblagoje Yes, ready to merge 🎉

PR #766, opened 2024-05-28
feat: defer the database connection to when it's needed

### LINKED ISSUE #765: `ElasticsearchDocumentStore`: avoid connecting to the server in the class constructor

masci:
### Related Issues

- fixes #765 

### Proposed Changes:

Make the database connection lazy, deferred to first usage.

### Checklist

### DIFF
--- a/integrations/elasticsearch/src/haystack_integrations/document_stores/elasticsearch/document_store.py
+++ b/integrations/elasticsearch/src/haystack_integrations/document_stores/elasticsearch/document_store.py
@@ -93,48 +93,58 @@ class ElasticsearchDocumentStore:
         :param **kwargs: Optional arguments that `Elasticsearch` takes.
         """
         self._hosts = hosts
-        self._client = Elasticsearch(
-            hosts,
-            headers={"user-agent": f"haystack-py-ds/{haystack_version}"},
-            **kwargs,
-        )
+        self._client = None
         self._index = index
         self._embedding_similarity_function = embedding_similarity_function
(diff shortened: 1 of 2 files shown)


integrations/elasticsearch/src/haystack_integrations/document_stores/elasticsearch/document_store.py
```diff
@@ -93,48 +93,60 @@ def __init__(
...
+                self._hosts,
+                headers={"user-agent": f"haystack-py-ds/{haystack_version}"},
+                **self._kwargs,
+            )
+            # Check client connection, this will raise if not connected
+            client.info()
+
+            if self._custom_mapping:
+                mappings = self._custom_mapping
+            else:
```
vblagoje, line 119: The only things that pops out is perhaps having `default_mappings` property and then simplifying indices creation with:

```
mappings = self._custom_mapping if self._custom_mapping else self.default_mappings
```
Makes it more obvious what's going on and easier to understand. I don't know however if this is pythonic or not. Just 2 cents.

vblagoje (approved):
Looks good 👍

integrations/elasticsearch/src/haystack_integrations/document_stores/elasticsearch/document_store.py
masci: It does make sense indeed, just in these PRs I prefer to keep the logic changes to the bare minimum to achieve the original goal.

PR #775, opened 2024-05-30
fix: pass empty dict to filter instead of None

masci:
### Related Issues

- fixes failing CI

### Proposed Changes:

Not sure where this comes from, but in absence of filters we used to pass `None` in the query pipeline and now it raises an error. Changed to passing an empty dict instead.

### Checklist


julian-risch (approved):
LGTM. 
Can't have ever worked with None before. We didn't notice because when we call _embedding_retrieval in the retriever we use the `filters` setting from the init, which is `{}`: https://github.com/deepset-ai/haystack-core-integrations/blob/f68b0dc48d282f7abdf05e83fb3251cb09b1a26f/integrations/mongodb_atlas/src/haystack_integrations/components/retrievers/mongodb_atlas/embedding_retriever.py#L63

Issue #657, opened 2024-04-12
add return_embedding parameter to OpenSearchEmbeddingRetriever


nickprock:
**Is your feature request related to a problem? Please describe.**
In `InMemoryEmbeddingRetriever` there is the parameter `return_embedding` that returns also embedding in `Document`, with OpenSearchEmbeddingRetriever I must recalculates embeddings after the retrieve step.

nickprock:
Thanks @ArzelaAscoIi 🙏

PR #784, opened 2024-06-04
feat: return_embeddings flag for opensearch

ArzelaAscoIi:
### Related Issues

- fixes https://github.com/deepset-ai/haystack-core-integrations/issues/657

### Proposed Changes:
- add flag to Opensearch init to return or skip embeddings

### How did you test it?
- integration test

### Checklist

### DIFF
--- a/integrations/opensearch/src/haystack_integrations/document_stores/opensearch/document_store.py
+++ b/integrations/opensearch/src/haystack_integrations/document_stores/opensearch/document_store.py
@@ -39,6 +39,7 @@ class OpenSearchDocumentStore:
         index: str = "default",
         max_chunk_bytes: int = DEFAULT_MAX_CHUNK_BYTES,
         embedding_dim: int = 768,
+        return_embedding: bool = False,
         method: Optional[Dict[str, Any]] = None,
         mappings: Optional[Dict[str, Any]] = None,
         settings: Optional[Dict[str, Any]] = DEFAULT_SETTINGS,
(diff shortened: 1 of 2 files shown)


integrations/opensearch/src/haystack_integrations/document_stores/opensearch/document_store.py
```diff
@@ -326,6 +330,11 @@ def _bm25_retrieval(
         if filters:
             body["query"]["bool"]["filter"] = normalize_filters(filters)
 
+        # For some applications not returning the embedding can save a lot of bandwidth
+        # if you don't need this data not retrieving it can be a good idea
+        if not self._return_embedding:
+            body["_source"] = {"excludes": ["embedding"]}
```
shadeMe, line 336: Curious: Does it make sense to return the embedding at all for keyword retrieval?

shadeMe (approved):

integrations/opensearch/src/haystack_integrations/document_stores/opensearch/document_store.py
ArzelaAscoIi: I know a pipeline in production that does bm25 retrieval (I think) and then uses the returned document vectors to issue an other search via embedding retrieval. With this, the application does not require to compute embeddings and therefore no GPU since the embedding already exist for the first retrieved doc. 

I think we can keep this. Not sure how valid this is.

wochinge:
Nice work, @ArzelaAscoIi !

PR #787, opened 2024-06-05
chore: Drop support for Python 3.8 for Unstructured integration

silvanocerza:
Unstructured library doesn't support Python 3.8 as can be seen in [latest nightly test failure](https://github.com/deepset-ai/haystack-core-integrations/actions/runs/9376015101/job/25815211259#step:9:58).

This is also made clear by [their `setup.py`](https://github.com/Unstructured-IO/unstructured/blob/fdb27378cb7fc25bcae66bf52bcfc16e051a42a8/setup.py#L96).

### DIFF
--- a/.github/workflows/unstructured.yml
+++ b/.github/workflows/unstructured.yml
@@ -30,7 +30,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-latest]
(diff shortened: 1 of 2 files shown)


anakin87 (approved):
:raised_hands:

PR #806, opened 2024-06-12
pinecone tests: wait for index creation

anakin87:
### Related Issues

- fixes #issue-number

### Checklist

### DIFF
--- a/integrations/pinecone/tests/test_document_store.py
+++ b/integrations/pinecone/tests/test_document_store.py
@@ -166,6 +166,8 @@ def test_serverless_index_creation_from_scratch(sleep_time):
     # Trigger the connection
     _ = ds.index
 
+    time.sleep(sleep_time)
+
     index_description = client.describe_index(name=index_name)
     assert index_description["name"] == index_name
     assert index_description["dimension"] == 30


integrations/pinecone/pyproject.toml
```diff
@@ -145,7 +145,8 @@ ban-relative-imports = "parents"
 [tool.coverage.run]
 source = ["haystack_integrations"]
 branch = true
-parallel = false
+concurrency = ["multiprocessing"]
+parallel = true
```
anakin87, line 149: unrelated. I am trying to collect the real coverage.

integrations/pinecone/tests/conftest.py
```diff
@@ -7,7 +7,7 @@
 from haystack_integrations.document_stores.pinecone import PineconeDocumentStore
 
 # This is the approximate time it takes for the documents to be available
-SLEEP_TIME = 10
+SLEEP_TIME = 15
```
anakin87, line 10: unrelated. Tests were not stable, so Increased this value.

shadeMe: Is that in seconds? Let's add the unit to the comment (or, even better, the identifier).

shadeMe (changes requested):

integrations/pinecone/tests/conftest.py
anakin87: done.

anakin87:
Massi made me notice that I was introducing a tight coupling between CI and code and suggested a better way to fix the problem.

Now I am defining the index name to use in the workflow matrix.

masci (approved):
LGTM!

PR #809, opened 2024-06-14
feat: support Claude v3, Llama3 and Command R models on Amazon Bedrock

tstadel:
### Related Issues

- model support update for Amazon Bedrock

### Proposed Changes:
- added support for
  - Claude V3
  - Llama3
  - Command R

### How did you test it?
- added tests

### Notes for the reviewer
- using messages API by default for Claude models is fine as all models on Bedrock support it (we did the same in Haystack v1 already)

### Checklist

### DIFF
--- a/integrations/amazon_bedrock/src/haystack_integrations/components/generators/amazon_bedrock/adapters.py
+++ b/integrations/amazon_bedrock/src/haystack_integrations/components/generators/amazon_bedrock/adapters.py
@@ -408,7 +408,7 @@ class AmazonTitanAdapter(BedrockModelAdapter):
         return chunk.get("outputText", "")
 
 
-class MetaLlama2ChatAdapter(BedrockModelAdapter):
+class MetaLlamaAdapter(BedrockModelAdapter):
     """
     Adapter for Meta's Llama2 models.
     """
--- a/integrations/amazon_bedrock/src/haystack_integrations/components/generators/amazon_bedrock/adapters.py
+++ b/integrations/amazon_bedrock/src/haystack_integrations/components/generators/amazon_bedrock/adapters.py
@@ -98,6 +98,10 @@ class AnthropicClaudeAdapter(BedrockModelAdapter):
     Adapter for the Anthropic Claude models.
     """
 
+    def __init__(self, model_kwargs: Dict[str, Any], max_length: Optional[int]) -> None:
+        self.use_messages_api = model_kwargs.get("use_messages_api", True)
+        super().__init__(model_kwargs, max_length)
+
     def prepare_body(self, prompt: str, **inference_kwargs) -> Dict[str, Any]:
         """
         Prepares the body for the Claude model
@@ -108,16 +112,30 @@ class AnthropicClaudeAdapter(BedrockModelAdapter):
             - `prompt`: The prompt to be sent to the model.
             - specified inference parameters.
         """
-        default_params = {
-            "max_tokens_to_sample": self.max_length,
-            "stop_sequences": ["\n\nHuman:"],
-            "temperature": None,
-            "top_p": None,
-            "top_k": None,
-        }
-        params = self._get_params(inference_kwargs, default_params)
+        if self.use_messages_api:
+            default_params: Dict[str, Any] = {
+                "anthropic_version": "bedrock-2023-05-31",
+                "max_tokens": self.max_length,
+                "system": None,
+                "stop_sequences": None,
+                "temperature": None,
+                "top_p": None,
+                "top_k": None,
+            }
+            params = self._get_params(inference_kwargs, default_params)
 
-        body = {"prompt": f"\n\nHuman: {prompt}\n\nAssistant:", **params}
+            body = {"messages": [{"role": "user", "content": prompt}], **params}
+        else:
+            default_params = {
+                "max_tokens_to_sample": self.max_length,
+                "stop_sequences": ["\n\nHuman:"],
+                "temperature": None,
+                "top_p": None,
+                "top_k": None,
+            }
+            params = self._get_params(inference_kwargs, default_params)
+
+            body = {"prompt": f"\n\nHuman: {prompt}\n\nAssistant:", **params}
         return body
 
     def _extract_completions_from_response(self, response_body: Dict[str, Any]) -> List[str]:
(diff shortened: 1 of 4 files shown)


integrations/amazon_bedrock/src/haystack_integrations/components/generators/amazon_bedrock/adapters.py
```diff
@@ -98,6 +98,10 @@ class AnthropicClaudeAdapter(BedrockModelAdapter):
     Adapter for the Anthropic Claude models.
     """
 
+    def __init__(self, model_kwargs: Dict[str, Any], max_length: Optional[int]) -> None:
+        self.use_messages_api = model_kwargs.get("use_messages_api", True)
```
sjrl, line 102: I know this is how we handled it in v1, but I wonder if exposing `use_messages_api` as an actual param instead of inside `model_kwargs` would be a good idea so we could add a docstring for it.

integrations/amazon_bedrock/src/haystack_integrations/components/generators/amazon_bedrock/chat/chat_generator.py
```diff
@@ -49,7 +49,7 @@ class AmazonBedrockChatGenerator:
 
     SUPPORTED_MODEL_PATTERNS: ClassVar[Dict[str, Type[BedrockModelChatAdapter]]] = {
         r"anthropic.claude.*": AnthropicClaudeChatAdapter,
-        r"meta.llama2.*": MetaLlama2ChatAdapter,
+        r"meta.llama.*": MetaLlama2ChatAdapter,
```
sjrl, line 52: Hey @tstadel I don't think we should do this. The `MetaLlama2ChatAdapter` includes with it the correct prompt template structure with the correct instruction tokens for Llama2. And the instruction tokens for Llama3 do not match. So for the ChatAdapater we'd need to make one specific to Llama3.

integrations/amazon_bedrock/src/haystack_integrations/components/generators/amazon_bedrock/adapters.py
tstadel: For this we would need to change how Adapters are initialized in general. See https://github.com/deepset-ai/haystack-core-integrations/blob/cbb51ed2c0613ee2e42a1eee303283b4d00ecb78/integrations/amazon_bedrock/src/haystack_integrations/components/generators/amazon_bedrock/generator.py#L141-L142

Even if we did, the impact would be quite limited as usually a user doesn't interact with the adapter. 
We could add a docstring to `AmazonBedrockGenerator`'s init instead. But so far we haven't added any model-specific params there. As this is more there for legacy reasons, I wouldn't bother for now... WDYT?

integrations/amazon_bedrock/src/haystack_integrations/components/generators/amazon_bedrock/chat/chat_generator.py
tstadel: Ah ok, got it. Thought it would work the same as the non-chat version. I'll revert.

integrations/amazon_bedrock/src/haystack_integrations/components/generators/amazon_bedrock/adapters.py
sjrl: All good points! I'm fine to leave as is.

integrations/amazon_bedrock/src/haystack_integrations/components/generators/amazon_bedrock/generator.py
```diff
@@ -56,9 +57,10 @@ class AmazonBedrockGenerator:
     SUPPORTED_MODEL_PATTERNS: ClassVar[Dict[str, Type[BedrockModelAdapter]]] = {
         r"amazon.titan-text.*": AmazonTitanAdapter,
         r"ai21.j2.*": AI21LabsJurassic2Adapter,
-        r"cohere.command.*": CohereCommandAdapter,
+        r"cohere.command-[^r].*": CohereCommandAdapter,
+        r"cohere.command-r.*": CohereCommandRAdapter,
         r"anthropic.claude.*": AnthropicClaudeAdapter,
-        r"meta.llama2.*": MetaLlama2ChatAdapter,
+        r"meta.llama.*": MetaLlama2ChatAdapter,
```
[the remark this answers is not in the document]
tstadel: renamed it to `MetaLlamaAdapter`

integrations/amazon_bedrock/src/haystack_integrations/components/generators/amazon_bedrock/adapters.py
```diff
@@ -108,16 +112,30 @@ def prepare_body(self, prompt: str, **inference_kwargs) -> Dict[str, Any]:
...
+            default_params = {
+                "max_tokens_to_sample": self.max_length,
+                "stop_sequences": ["\n\nHuman:"],
+                "temperature": None,
+                "top_p": None,
+                "top_k": None,
+            }
+            params = self._get_params(inference_kwargs, default_params)
+
+            body = {"prompt": f"\n\nHuman: {prompt}\n\nAssistant:", **params}
```
sjrl, line 138: I realize this would be more of a feature request, but does Claude support system prompts? In our other generators like `OpenAIGenerator` we support specifying a system prompt at init time. Also for Generators like `Llama2` or `Llama3` we can specify a system prompt by using the appropriate tokens when constructing the prompt template. It feels like Claude is treated a little differently by auto-formatting the prompt without telling the user.

tstadel: When used with messages API you can simply specify `"system"` param as in this test:
https://github.com/deepset-ai/haystack-core-integrations/blob/c16f1877ee3a9b7aa7545d31093662c4980bbbc2/integrations/amazon_bedrock/tests/test_generator.py#L262-L288

sjrl: Ahh okay I see. Sorry, but it's not immediately obvious to me, but can then I just pass a value to `system` as a kwarg to `AmazonBedRockGenerator`?

silvanocerza (approved):
Looks good to me, I have no way of testing this though.

Issue #780, opened 2024-06-03
Add `filter_policy` init parameter to all retrievers


vblagoje:
### Motivation

The introduction of the `filter_policy` parameter aims to enhance the flexibility and consistency of filter application in all retrievers. Currently, the handling of filters varies between retrievers. This inconsistency can lead to confusion and limitations for users who require a more nuanced approach to filter application.

### Proposed Solution

We propose adding an `init` parameter, `filter_policy`, to all retrievers. This parameter will control the interaction between initialization filters and runtime filters, allowing users to specify their preferred filter application method. The `filter_policy` parameter can take the following values:

- `replace`: Gives precedence to the filters specified at runtime. This is the current behavior and ensures backward compatibility.
- `merge`: Merges both runtime and initialization filters, with runtime filters overwriting initialization values.

By default, `filter_policy` will be set to `runtime` to maintain backward compatibility. This approach will standardize filter handling across all retrievers, providing users with a consistent and flexible filtering mechanism.
```[tasklist]
### Tasks
- [x] Add `filter_policy` parameter in OpenSearchEmbeddingRetriever and OpenSearchBM25Retriever
- [x] Add `filter_policy` in AstraEmbeddingRetriever
- [x] Add `filter_policy` in ChromaQueryTextRetriever and ChromaEmbeddingRetriever
- [x] Add `filter_policy` in ElasticsearchBM25Retriever and ElasticsearchEmbeddingRetriever
- [x] Add `filter_policy` in MongoDBAtlasEmbeddingRetriever
- [x] Add `filter_policy` in PgvectorKeywordRetriever and PgvectorEmbeddingRetriever
- [x] Add `filter_policy` in PineconeEmbeddingRetriever
- [x] Add `filter_policy` in QdrantEmbeddingRetriever, QdrantSparseEmbeddingRetriever and QdrantHybridRetriever
- [ ] https://github.com/deepset-ai/haystack-core-integrations/issues/834
- [x] Create new releases of all document stores after adding `filter_policy` parameter
- [x] Add `filter_policy` parameter to 2.3 documentation pages
```

vblagoje:
Note: @dfokina , a reminded to you and myself not to forget to add `FilterPolicy` to docs. The question is where? Good candidates are:
- https://docs.haystack.deepset.ai/v2.3-unstable/docs/metadata-filtering
- https://docs.haystack.deepset.ai/v2.3-unstable/docs/retrievers

[...]

PR #781, opened 2024-06-03
feat: Add filter_policy init parameter to all retrievers

vblagoje:
### Why:
The changes across various integrations (Astra, Chroma, Elasticsearch, MongoDB Atlas, OpenSearch, pgvector, Pinecone, Qdrant, and Weaviate) address a common need for more flexible filtering options within the retrieval process. By introducing a filter policy option (`replace` or `merge`), developers can now control how runtime filters are applied relative to initialization time filters. 

- fixes https://github.com/deepset-ai/haystack-core-integrations/issues/780

### What:
- Added `filter_policy` parameter with options `replace` and `merge` across multiple retrievers to control filter behavior dynamically.
- Adjustments made to initialize, serialize (to_dict), and run methods to support the new `filter conservatively` policy behavior.
- Unit tests were updated or added to validate the new functionality.

### How can it be used:

- **Dynamic Filter Behavior Adjustment:**  Users can decide whether to completely override the initial filters set during the retriever's initialization (`replace`) or merge them with runtime filters, with the latter taking precedence (`merge`).
    ```python
    # Example: Using the filter_policy parameter
    retriever = SomeRetriever(document_store=my_doc_store, filters={"status": "active"}, filter_policy="merge")
    ```

- **Complex Search Scenarios:**  
    - In cases where the context of a query might dictate altering pre-set filters without discarding them, the `merge` option allows for an additive approach.
    - For strict query contexts that require ignoring initial filters, the `replace` option offers a clean slate for filters at runtime.

### How did you test it:
- Unit tests were enhanced or newly created to cover both `replace` and `merge` scenarios for the `filter_policy` parameter. 
- Tests ensure that filter logic is correctly applied based on the policy setting, whether it merges runtime filters with initial filters or replaces them entirely.
- Additional test cases should be considered for complex filter merge scenarios to ensure priority and override mechanisms function as expected.

### Notes for the reviewer:
- Make sure all retrievers in the project are accounted and appropriately adjusted 
- Pay special attention all retrievers have the same new init param + proper pydoc
- Verify unit tests were adjusted appropriately
- Pay special attention to Qdrant  retrievers, there we needed (or not, I'm not 100% sure) to keep attribute filters as None
- It's essential to ensure backward compatibility; existing implementations should default to `replace` to maintain current behavior unless explicitly set to `merge`.

PR #825, opened 2024-06-19
feat: Add filter_policy to elasticsearch integration

vblagoje:
### Why:
Adds flexible filtering options to elasticsearch integration. By introducing a filter policy option (`replace` or `merge`), developers can now control how runtime filters are applied relative to initialization time filters. 

- partially fixes https://github.com/deepset-ai/haystack-core-integrations/issues/780
- supersedes https://github.com/deepset-ai/haystack-core-integrations/pull/781

### What:
- Added `filter_policy` parameter with options `replace` and `merge` across multiple retrievers to control filter behavior dynamically.
- Unit tests were updated or added to validate the new functionality.

### How can it be used:

- **Dynamic Filter Behavior Adjustment:**  Users can decide whether to completely override the initial filters set during the retriever's initialization (`replace`) or merge them with runtime filters, with the latter taking precedence (`merge`).

- **Complex Search Scenarios:**  
    - In cases where the context of a query might dictate altering pre-set filters without discarding them, the `merge` option allows for an additive approach.
    - For strict query contexts that require ignoring initial filters, the `replace` option offers a clean slate for filters at runtime.

### How did you test it:
- Unit tests were enhanced or newly created to cover both `replace` and `merge` scenarios for the `filter_policy` parameter. 
- Tests ensure that filter logic is correctly applied based on the policy setting, whether it merges runtime filters with initial filters or replaces them entirely.
- Additional test cases should be considered for complex filter merge scenarios to ensure priority and override mechanisms function as expected.


vblagoje:
Please delay your review until a new patch release of 2.2.x is released. We need these [two PRs](https://github.com/deepset-ai/haystack/commits/main/haystack/document_stores/types/filter_policy.py) to be included in a patch release first.

vblagoje:
Ready for review now @Amnah199  

- Since you last looked at this PR we introduced `apply_filter_policy` function that is used by all Retrievers in this [PR](https://github.com/deepset-ai/haystack/pull/7902) and we thoroughly tested that function in that PR. We then just applied it here in run method
- Please note if FilterPolicy type is used consistently both in init and in run, and that it is documented well

vblagoje:
Stefano recommended one more small improvement. Please hold your review until a new commit with it is added here

vblagoje:
@Amnah199  please have a look at https://github.com/deepset-ai/haystack-core-integrations/pull/827 @anakin87 has already gone through a through review of Astra and should help you validate this PR is well-aligned as well

integrations/elasticsearch/src/haystack_integrations/components/retrievers/elasticsearch/bm25_retriever.py
```diff
@@ -60,6 +63,7 @@ def __init__(
             for more details.
         :param top_k: Maximum number of Documents to return.
         :param scale_score: If `True` scales the Document`s scores between 0 and 1.
+        :param filter_policy: Policy to determine how filters are applied.
```
Amnah199, line 66: Would it help to give possible values for this, like REPLACE or MERGE.

vblagoje: That's a good idea but it gets outdated fast because `FilterPolicy` changes and we have to do more maintenance work. As it is defined type users can simply look up `FilterPolicy` enum and see values available.

integrations/elasticsearch/src/haystack_integrations/components/retrievers/elasticsearch/embedding_retriever.py
```diff
@@ -71,6 +75,7 @@ def __init__(
         self._filters = filters or {}
         self._top_k = top_k
         self._num_candidates = num_candidates
+        self._filter_policy = FilterPolicy.from_str(filter_policy) if isinstance(filter_policy, str) else filter_policy
```
Amnah199, line 78: Looking at edge cases, would it be helpful here to raise a type error if lets say in integer is passed? Else it would break anyways later in the code.  e.g.
raise TypeError("filter_policy must be either a str or an instance of FilterPolicy")

integrations/elasticsearch/src/haystack_integrations/components/retrievers/elasticsearch/bm25_retriever.py
Amnah199: Makes sense.

Amnah199 (approved):
One small comment, otherwise good to go!

integrations/elasticsearch/src/haystack_integrations/components/retrievers/elasticsearch/embedding_retriever.py
vblagoje: Yeah, we don't check for those typing errors, the types are defined in the signature and are assumed to be followed by the user

PR #833, opened 2024-06-19
feat: made truncation optional for BedrockGenerator

### LINKED ISSUE #783: Amazon Bedrock Generators missing `kwargs` init params when serializing
**Describe the bug**
The Amazon Bedrock Generators init method allows for `kwargs`
https://github.com/deepset-ai/haystack-core-integrations/blob/6aaf523636d923ce158c46ea39a057f5529b5a17/integrations/amazon_bedrock/src/haystack_integrations/components/generators/amazon_bedrock/generator.py#L65-L77

but in the `to_dict` method ignores these kwargs

https://github.com/deepset-ai/haystack-core-integrations/blob/6aaf523636d923ce158c46ea39a057f5529b5a17/integrations/amazon_bedrock/src/haystack_integrations/components/generators/amazon_bedrock/generator.py#L251-L267

which means serializing and then deserializing will result in a different Generator with different behavior (e.g. if a user sets the `temperature` as a `kwarg` then this value will be "forgotten").

**To Reproduce**
Serialize a Bedrock generator using a `kwarg` and then deserialize.

 - OS: [e.g. iOS] MacOS
 - Haystack version: 2.2
 - Integration version: 0.8.0

### LINKED ISSUE #785: Make truncation optional for AmazonBedrockGenerators
We've found that in some situations we'd prefer the model provider (e.g. Bedrock) to return with an error message if the prompt is too long rather than truncate the end of the prompt which can lead to non-sensical answers depending on how the prompt template is constructed. So I propose adding a parameter to toggle this behavior with the default behavior being unchanged (i.e. truncation=True is the same as the previous default behavior).

Similar solution as was implemented in this [PR](https://github.com/deepset-ai/haystack/pull/7431) for Haystack v1.

Amnah199:
### Related Issues

- fixes #785 
- fixes #783 

### Proposed Changes:

- Added 'truncate' parameter to initialization method. 
- Added kwargs in serialization method

### How did you test it?

Updated unit tests
 Integration tests

### Notes for the reviewer

NA
### Checklist

### DIFF
--- a/integrations/amazon_bedrock/src/haystack_integrations/components/generators/amazon_bedrock/generator.py
+++ b/integrations/amazon_bedrock/src/haystack_integrations/components/generators/amazon_bedrock/generator.py
@@ -103,6 +103,7 @@ class AmazonBedrockGenerator:
         self.aws_session_token = aws_session_token
         self.aws_region_name = aws_region_name
         self.aws_profile_name = aws_profile_name
+        self.kwargs = kwargs
 
         def resolve_secret(secret: Optional[Secret]) -> Optional[str]:
             return secret.resolve_value() if secret else None
@@ -271,6 +272,8 @@ class AmazonBedrockGenerator:
             aws_profile_name=self.aws_profile_name.to_dict() if self.aws_profile_name else None,
             model=self.model,
             max_length=self.max_length,
+            truncate=self.truncate,
+            **self.kwargs,
         )
 
     @classmethod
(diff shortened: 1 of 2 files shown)


Amnah199:
Should we also make truncation optional for BedrockChatGenerators?

sjrl:
> Should we also make truncation optional for BedrockChatGenerators?

I would also be in support of that :)

julian-risch (changes requested):
Looks quite good to me already. Could you please add a small test that checks the behavior when truncate is set to False. Basically a mix of the two existing test cases `test_short_prompt_is_not_truncated(mock_boto3_session)` and `test_long_prompt_is_truncated(mock_boto3_session)` where you have a long prompt and `truncate=False` and then you check that the prompt wasn't truncated.

julian-risch:
Regarding BedrockChatGenerators please create a new issue for that and then we can assign it in one of the next sprints.

julian-risch (approved):
LGTM! 🚀

PR #836, opened 2024-06-21
ci: retry tests to reduce flakyness

### LINKED ISSUE #799: Mark integration tests as flaky
We have frequent spurious CI failures due I/O and connectivity issues, leading to recurrent notifications.

Mark all tests that have side-effects with [flaky](https://docs.pytest.org/en/7.1.x/explanation/flaky.html) and turn on timed retries.

masci:
### Related Issues

- fixes #799 

### Proposed Changes:

- Introduce `pytest-rerunfailures` to retry test execution and possibly reduce failures due to network connectivity
- To minimize the impact of this PR, add the retry logic to any `hatch run test` script. The plugin supports a more granular configuration but in this PR I want to avoid a case-by-case strategy
- To avoid timing out, we now run the tests with `-x`, so we exit immediately after the first retry. Several tests might fail for the same reason, so we avoid piling up too many retries before exiting.
- (unrelated) reformatted the toml files

### How did you test it?

CI

### Notes for the reviewer

- Please let me know if you agree with the `-x` strategy
- I set a 30 seconds delay in-between retries, I'm open to change that number
- Also let me know if 3 attempts is ok

### Checklist

### DIFF
--- a/integrations/amazon_bedrock/pyproject.toml
+++ b/integrations/amazon_bedrock/pyproject.toml
@@ -10,9 +10,7 @@ readme = "README.md"
 requires-python = ">=3.8"
 license = "Apache-2.0"
 keywords = []
-authors = [
-  { name = "deepset GmbH", email = "info@deepset.ai" },
-]
+authors = [{ name = "deepset GmbH", email = "info@deepset.ai" }]
 classifiers = [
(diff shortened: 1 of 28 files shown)


anakin87 (approved):
Looks good!

If necessary, we will introduce more granular control in the future.

PR #838, opened 2024-06-21
feat: Update Cohere default LLMs, add examples and update unit tests

vblagoje:
- Update default model to `command-r`
- Add examples

### DIFF
--- a/integrations/cohere/examples/cohere_generation.py
+++ b/integrations/cohere/examples/cohere_generation.py
@@ -1,3 +1,18 @@
+# This example demonstrates a corrective Haystack pipeline with Cohere LLM integration that runs until the
+# generated output satisfies a strict JSON schema.
+#
+# The pipeline includes the following components:
+#  - BranchJoiner: https://docs.haystack.deepset.ai/reference/joiners-api#branchjoiner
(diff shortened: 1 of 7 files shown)


integrations/cohere/examples/cohere_generation.py
```diff
@@ -0,0 +1,36 @@
+from typing import List
```
anakin87, line 1: I think adding a brief comment (and/or some docs links) explaining this example would be beneficial for users unfamiliar with `BranchJoiner` and `JsonSchemaValidator`.

anakin87 (approved):
Looks good.

Please just add an explanatory comment in the generation example.

integrations/cohere/examples/cohere_generation.py
vblagoje: Deal, will do

vblagoje: Have a brief look at [4b1bee8](https://github.com/deepset-ai/haystack-core-integrations/pull/838/commits/4b1bee8df99d01d5fc1620d2c8a18391a33efda5) @anakin87

anakin87: Nice!

PR #840, opened 2024-06-21
feat: add create_index option to OpenSearchDocumentStore

tstadel:
### Related Issues

- fixes having left-over indexes due to unintended index creations

### Proposed Changes:
- add `create_index` init param to disable automatic index creation
- add `create_index()` method for more control over (manually) creating indexes

### How did you test it?
- added integration test

### Checklist

### DIFF
--- a/integrations/opensearch/src/haystack_integrations/document_stores/opensearch/document_store.py
+++ b/integrations/opensearch/src/haystack_integrations/document_stores/opensearch/document_store.py
@@ -119,10 +119,15 @@ class OpenSearchDocumentStore:
         elif self._create_index:
             # Create the index if it doesn't exist
             body = {"mappings": self._mappings, "settings": self._settings}
-            self._client.indices.create(index=self._index, body=body)
+            self._client.indices.create(index=self._index, body=body)  # type:ignore
         return self._client
-    
-    def create_index(self, index: Optional[str] = None, mappings: Optional[Dict[str, Any]] = None, settings: Optional[Dict[str, Any]] = None) -> None:
+
+    def create_index(
+        self,
+        index: Optional[str] = None,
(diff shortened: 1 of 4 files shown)


integrations/opensearch/src/haystack_integrations/document_stores/opensearch/document_store.py
```diff
@@ -113,13 +116,39 @@ def client(self) -> OpenSearch:
...
             )
-        else:
+        elif self._create_index:
             # Create the index if it doesn't exist
             body = {"mappings": self._mappings, "settings": self._settings}
-
             self._client.indices.create(index=self._index, body=body)  # type:ignore
         return self._client
 
+    def create_index(
```
vblagoje, line 125: Do we need this one to be public?

vblagoje (commented):
Hey @tstadel looks good. Can you provide a bit more context why do we need to have create_index function to be public? Who would invoke it?

integrations/opensearch/src/haystack_integrations/document_stores/opensearch/document_store.py
tstadel: Use-case for us is the following:
- whenever a pipeline is created, we want to create the index, but only in this situation.
- if the pipeline should actually be down but is still running (e.g. because embedding batches can take a couple of seconds), but the index has already been deleted async, the index shouldn't be created again automatically

To support this:
- we can set `create_index=False` init param for `OpenSearchDocumentStore`
- call `create_index()` on the document store of the pipeline whenever we load the pipeline

I'm not opinionated here if it has to be public. But I can imagine this as valuable in general whenever you don't want the index to be created automatically. For this I would see it as public, but just let me know if I should change it.

PR #841, opened 2024-06-21
feat: add custom_query param to OpenSearch retrievers

tstadel:
### Related Issues

- fixes support for custom OpenSearch queries

### Proposed Changes:
- add `custom_query` param
- we support ~`str`~ and `dict` as custom_query: the former is easier to use with YAML, the latter is easier to use with ~python~JSON

### How did you test it?
- added tests

### Notes for the reviewer
- What do I mean with easier with YAML? See how a already properly formated OpenSearch query in JSON can be simply copied into yaml under https://docs.cloud.deepset.ai/docs/opensearch-queries-for-boosting-retrieval#used-in-a-pipeline 

### Checklist

### DIFF
--- a/integrations/opensearch/src/haystack_integrations/components/retrievers/opensearch/bm25_retriever.py
+++ b/integrations/opensearch/src/haystack_integrations/components/retrievers/opensearch/bm25_retriever.py
@@ -1,7 +1,7 @@
 # SPDX-FileCopyrightText: 2023-present deepset GmbH <info@deepset.ai>
 #
 # SPDX-License-Identifier: Apache-2.0
-from typing import Any, Dict, List, Optional
+from typing import Any, Dict, List, Optional, Union
 
 from haystack import component, default_from_dict, default_to_dict
 from haystack.dataclasses import Document
@@ -19,7 +19,7 @@ class OpenSearchBM25Retriever:
         top_k: int = 10,
         scale_score: bool = False,
         all_terms_must_match: bool = False,
-        custom_query: Optional[str] = None,
+        custom_query: Optional[Union[str, Dict[str, Any]]] = None,
     ):
         """
         Create the OpenSearchBM25Retriever component.
@@ -32,7 +32,7 @@ class OpenSearchBM25Retriever:
             This is useful when comparing documents across different indexes. Defaults to False.
         :param all_terms_must_match: If True, all terms in the query string must be present in the retrieved documents.
             This is useful when searching for short text where even one term can make a difference. Defaults to False.
-        :param custom_query: The query string containing a mandatory `${query}` and an optional `${filters}` placeholder
+        :param custom_query: The query containing a mandatory `$query` and an optional `$filters` placeholder
 
             **An example custom_query:**
 
@@ -41,10 +41,10 @@ class OpenSearchBM25Retriever:
                 "query": {
                     "bool": {
                         "should": [{"multi_match": {
-                            "query": ${query},                 // mandatory query placeholder
+                            "query": $query,                 // mandatory query placeholder
                             "type": "most_fields",
                             "fields": ["content", "title"]}}],
-                        "filter": ${filters}                  // optional filter placeholder
+                        "filter": $filters                  // optional filter placeholder
                     }
                 }
             }
@@ -113,7 +113,7 @@ class OpenSearchBM25Retriever:
         top_k: Optional[int] = None,
         fuzziness: Optional[str] = None,
         scale_score: Optional[bool] = None,
-        custom_query: Optional[str] = None,
+        custom_query: Optional[Union[str, Dict[str, Any]]] = None,
     ):
         """
         Retrieve documents using BM25 retrieval.
@@ -125,7 +125,7 @@ class OpenSearchBM25Retriever:
         :param fuzziness: Fuzziness parameter for full-text queries.
         :param scale_score: Whether to scale the score of retrieved documents between 0 and 1.
             This is useful when comparing documents across different indexes.
-        :param custom_query: The query string containing a mandatory `${query}` and an optional `${filters}` placeholder
+        :param custom_query: The query containing a mandatory `$query` and an optional `$filters` placeholder
 
             **An example custom_query:**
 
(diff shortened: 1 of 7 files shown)


silvanocerza (changes requested):
Let's just use `Dict[str, Any]` for `custom_query` and scrap support for `str`.

integrations/opensearch/src/haystack_integrations/components/retrievers/opensearch/bm25_retriever.py
```diff
@@ -113,6 +166,8 @@ def run(
             fuzziness = self._fuzziness
         if scale_score is None:
             scale_score = self._scale_score
+        if custom_query is None:
+            custom_query = self._custom_query
```
davidsbatista, line 170:
```
        self._document_store = document_store
        self._filters = filters or {}
        self._fuzziness = fuzziness
        self._top_k = top_k
        self._scale_score = scale_score
        self._all_terms_must_match = all_terms_must_match
        self._custom_query = custom_query
```

is this really needed? wouldn't be enough remove this and instead do:

```
docs = self._document_store._bm25_retrieval(
            query=query,
            filters=self._filters,
            fuzziness=self_.fuzziness,
            top_k=self_.top_k,
            scale_score=self._scale_score,
            all_terms_must_match=self._all_terms_must_match,
            custom_query=self._custom_query,
        )
```

integrations/opensearch/src/haystack_integrations/components/retrievers/opensearch/embedding_retriever.py
```diff
@@ -86,10 +156,13 @@ def run(self, query_embedding: List[float], filters: Optional[Dict[str, Any]] =
...
         if top_k is None:
             top_k = self._top_k
+        if custom_query is None:
+            custom_query = self._custom_query
 
         docs = self._document_store._embedding_retrieval(
             query_embedding=query_embedding,
             filters=filters,
             top_k=top_k,
+            custom_query=custom_query,
```
davidsbatista, line 166: same as above, why not pass directly the parameter attached to the `self`?

integrations/opensearch/src/haystack_integrations/document_stores/opensearch/document_store.py
```diff
@@ -300,35 +303,68 @@ def _bm25_retrieval(
...
-            raise ValueError(msg)
+            body: Dict[str, Any] = {"query": {"bool": {"must": {"match_all": {}}}}}
+            if filters:
+                body["query"]["bool"]["filter"] = normalize_filters(filters)
+
+        if isinstance(custom_query, str):
+            template = Template(custom_query)
+            # substitute placeholder for query and filters for the custom_query template string
+            substitutions = {
+                "query": json.dumps(query),
```
davidsbatista, line 337: what if `query` is None here?

integrations/opensearch/src/haystack_integrations/document_stores/opensearch/document_store.py
```diff
@@ -300,35 +303,68 @@ def _bm25_retrieval(
...
+                "filters": json.dumps(normalize_filters(filters)),
+            }
+            custom_query_json = template.substitute(**substitutions)
+            body = json.loads(custom_query_json)
+        elif isinstance(custom_query, dict):
+            body = self._render_custom_query(custom_query, {"$query": query, "$filters": normalize_filters(filters)})
```
davidsbatista, line 343: same here, `query` can be `None` - will that be OK?

integrations/opensearch/src/haystack_integrations/document_stores/opensearch/document_store.py
tstadel: Good catch: `json.dumps(None)` evaluates to `"null"`. So probably not what we want. I'll make a proper `None` out of it and let OpenSearch handle the error.

integrations/opensearch/src/haystack_integrations/document_stores/opensearch/document_store.py
tstadel: Here it's ok. We let OpenSearch handle it.

tstadel:
> Let's just use `Dict[str, Any]` for `custom_query` and scrap support for `str`.

@silvanocerza I'd normally also like the idea to keep it simple, but you usually interact with OpenSearch using JSON, e.g. via [OpenSearch dev tools](https://opensearch.org/docs/2.14/dashboards/dev-tools/run-queries/). Don't you think this makes it quite cumbersome to set custom queries as you always have to convert them to however you want to interact with the retrievers here (e.g. from JSON to python dicts, or to YAML if you use YAML marshalling)?

tstadel:
> > Let's just use `Dict[str, Any]` for `custom_query` and scrap support for `str`.
> 
> @silvanocerza I'd normally also like the idea to keep it simple, but you usually interact with OpenSearch using JSON, e.g. via [OpenSearch dev tools](https://opensearch.org/docs/2.14/dashboards/dev-tools/run-queries/) or postman. Don't you think this makes it quite cumbersome to set custom queries as they can get quite complex and you always have to convert them to however you want to interact with the retrievers here (e.g. from JSON to python dicts, or to YAML if you use YAML marshalling)?

Ah anywas, we can readd it if we see it's not working.

integrations/opensearch/src/haystack_integrations/document_stores/opensearch/document_store.py
tstadel: deleted

integrations/opensearch/src/haystack_integrations/components/retrievers/opensearch/embedding_retriever.py
tstadel: see above

silvanocerza:
@tstadel correct JSON is almost correct Python too, the only difference being `null` instead of `None`. Copy and paste wouldn't be a great issue.

I agree that it might be annoying handling converting that to YAML for a user, though you'd get a validation that you wouldn't get if it's a string.

My main concern is that using a string could be error prone and hard to notice. The fact that most of the docstrings are not valid Python nor JSON is pretty telling. :)

Also, as you said, we can add this later if we see the need.

silvanocerza (approved):

davidsbatista (approved):
LGTM

PR #852, opened 2024-06-27
feat: add raise_on_failure param to OpenSearch retrievers

tstadel:
### Related Issues

In a hybrid retrieval setting, it can occurr that one retriever raises an error (e.g. due to a too long query for BM25), but the other is fine. In such cases it's better to return the results of the working retriever and ignore the failed ones instead of getting no results at all.

### Proposed Changes:
- add `raise_on_failure` param to OpenSearch retrievers

### How did you test it?
- added tests

### Checklist


julian-risch:
Hi @shadeMe can we get your opinion on this too? To me it makes sense to return an empty list of documents if there are errors but it's not specific to OpenSearch retrievers only. It's even similar to the `raise_on_failure` we have for evaluators.

shadeMe (approved):

PR #853, opened 2024-06-28
chore: Update ruff invocation to include check parameter

vblagoje:
- Updates ruff invocation across integration to include check parameter. 
- langfuse and astra were already including check parameter and were not updated here


anakin87 (approved):
Unstructured and Google errors are unrelated and should be fixed separately.

Issue #676, opened 2024-04-22
Qdrant - refactor and release a major version


anakin87:
### important
- the `embedding_field` init parameter of `QdrantDocumentStore` will break things if not set to the default value ("embedding") -> should be removed
- ~the `return_embedding` init parameter of `QdrantDocumentStore` is not used and is confusing -> should be removed~
  EDIT: I was wrong, it is used

### nice to have
- set the default value for `scale_score` in Retrievers to `False`, as we do for all other retrievers
- separate well the tests for Document Store retrieval methods and Retrievers (non breaking)

anakin87:
also the init parameter `duplicate_documents` is not used and should be removed.

anakin87:
done in #861 and #862

released a new major version of the package: https://pypi.org/project/qdrant-haystack/4.0.0/

PR #862, opened 2024-07-01
refactor!: Qdrant - set `scale_score` default value to `False`

anakin87:
### Related Issues

part of #676

### Proposed Changes:

Set `scale_score` default value to `False`, as we do for every Document Store in 2.x.
The motivation of the original decision can be read in https://github.com/deepset-ai/haystack/issues/6271

### How did you test it?
CI

### Checklist


integrations/qdrant/src/haystack_integrations/document_stores/qdrant/document_store.py
```diff
@@ -496,7 +496,7 @@ def _query_by_sparse(
         query_sparse_embedding: SparseEmbedding,
         filters: Optional[Union[Dict[str, Any], rest.Filter]] = None,
         top_k: int = 10,
-        scale_score: bool = True,
+        scale_score: bool = False,
```
anakin87, line 499: the changes in this file are not needed, since the only interface to these methods is the Retriever,
but I think they make the overall behavior clearer.

shadeMe (approved):

PR #874, opened 2024-07-05
Normalize logical filter conditions

### LINKED ISSUE #865: Hande Nested Logical Operators for Astra Document Store Filters
**Describe the bug**

It seems like the code right now is unable to handle nested logical filters. Let's say we have a need for a filter that looks like this.
```
{
    "operator": "AND",
    "conditions":
    [
        {
            "field": "meta.field1",
            "operator": "==",
            "value": "something"
        },
        {
            "operator": "OR",
            "conditions":
            [
                {
                    "field": "meta.field2",
                    "operator": "==",
                    "value": "something else"
                },
                {
                    "field": "meta.field2",
                    "operator": "==",
                    "value": "something else else"
                }
            ]
        }
    ]
}
```
When running `_normalize_filters` on this from https://github.com/deepset-ai/haystack-core-integrations/blob/main/integrations/astra/src/haystack_integrations/document_stores/astra/filters.py#L7. We get the following error because the code is expecting these to only be condition filters.
```
"'field' key missing in ..."
```
From the [haystack page](https://docs.haystack.deepset.ai/docs/metadata-filtering#logic), the conditions in the logic filters can be either logic filters or condition filters.

I think we can fix this by using `_normalize_filters` [here](https://github.com/deepset-ai/haystack-core-integrations/blob/main/integrations/astra/src/haystack_integrations/document_stores/astra/filters.py#L76) instead of `_parse_conditions`. Let me know what y'all think!

**To Reproduce**
If you take the above filter and run the `_normalize_filters` on it, it should error out.

Sample code looks like this:
```
filters = {
    "operator": "AND",
    "conditions":
    [
        {
            "field": "meta.field1",
            "operator": "==",
            "value": "something"
        },
        {
            "operator": "OR",
            "conditions":
            [
                {
                    "field": "meta.field2",
                    "operator": "==",
                    "value": "something else"
                },
                {
                    "field": "meta.field2",
                    "operator": "==",
                    "value": "something else else"
                }
            ]
        }
    ]
}

_normalize_filters(filters)
```

 - OS: [e.g. iOS] Sonoma 14.5
 - Haystack version: 2.2.3? (haystack-ai==2.2.3)
 - Integration version: 0.5.0? (astra-haystack==0.5.0)

ajtran:
### Related Issues

- fixes #865 

### Proposed Changes:

 
 

This is a bug. We aren't properly handling conditions for logical filters. We are evaluating them only as comparison filters when they could be both logical or comparison filters. By using the `_normalize_filters` method again we are evaluating for both logical and comparison filters.

### How did you test it?

Manual verification by testing with this case on my python environment.
```
{
    "operator": "AND",
    "conditions":
    [
        {
            "field": "meta.field1",
            "operator": "==",
            "value": "something"
        },
        {
            "operator": "OR",
            "conditions":
            [
                {
                    "field": "meta.field2",
                    "operator": "==",
                    "value": "something else"
                },
                {
                    "field": "meta.field2",
                    "operator": "==",
                    "value": "something else else"
                }
            ]
        }
    ]
}
```

### Checklist


Amnah199:
@ajtran thanks for the contribution. It would be nice if we can add new tests to check nested filters. Something like below could be a start:

```
 def test_nested_filters(self, document_store, filterable_docs):

        filter_criteria = {
            "operator": "AND",
            "conditions":
            [
                {
                    "field": "meta.field1",
                    "operator": "==",
                    "value": "Value 1."
                },
                {
                    "operator": "OR",
                    "conditions":
                    [
                        {
                            "field": "meta.field2",
                            "operator": "==",
                            "value": "Value 2."
                        },
                        {
                            "field": "meta.field2",
                            "operator": "==",
                            "value": "Value 3"
                        }
                    ]
                }
            ]
        }

        document_store.write_documents(filterable_docs)
        result = document_store.filter_documents(filters=filter_criteria)
       
        self.assert_documents_are_equal(result, [
            d for d in filterable_docs 
            if d.meta.get("field1") == "Value 1" and 
            (d.meta.get("field2") == "Value 2" or d.meta.get("field2") == "Value 3")
        ])
```

ajtran:
@Amnah199 please take another look when you get a chance. thank you!

integrations/astra/tests/test_document_store.py
```diff
@@ -172,6 +172,40 @@ def test_delete_documents_more_than_twenty_delete_ids(self, document_store: Astr
...
 
+    def test_filter_documents_nested_filters(self, document_store, filterable_docs):
+        """
+        Test filter_documents() with nested filters. Two documents from the filterable_docs mixin
+        should be found. If those documents change, this test has to change as well.
+        """
+        filter_criteria = {
+            "operator": "AND",
+            "conditions": [
+                {"field": "meta.page", "operator": ">=", "value": "100"},
```
Amnah199, line 183: ">=" operator is not applicable on strings so this will fail.
```suggestion
                {"field": "meta.page", "operator": "==", "value": "100"},
```

integrations/astra/tests/test_document_store.py
```diff
@@ -172,6 +172,40 @@ def test_delete_documents_more_than_twenty_delete_ids(self, document_store: Astr
...
+            ],
+        }
+
+        document_store.write_documents(filterable_docs)
+        result = document_store.filter_documents(filters=filter_criteria)
+
+        expected_docs = [
+            d
+            for d in filterable_docs
+            if d.meta.get("page") >= "100" and (d.meta.get("chapter") == "abstract" or d.meta.get("chapter") == "intro")
```
Amnah199, line 200:
```suggestion
            if d.meta.get("page") == "100" and (d.meta.get("chapter") == "abstract" or d.meta.get("chapter") == "intro")
```

integrations/astra/tests/test_document_store.py
```diff
@@ -172,6 +172,40 @@ def test_delete_documents_more_than_twenty_delete_ids(self, document_store: Astr
...
+        ]
+        assert len(expected_docs) == 2
```
Amnah199, line 202: `assert_documents_are_equal ` is comparing the results. So this is not needed.
```suggestion
```

integrations/astra/tests/test_document_store.py
```diff
@@ -172,6 +172,40 @@ def test_delete_documents_more_than_twenty_delete_ids(self, document_store: Astr
...
+        }
+
+        document_store.write_documents(filterable_docs)
+        result = document_store.filter_documents(filters=filter_criteria)
+
+        expected_docs = [
+            d
+            for d in filterable_docs
+            if d.meta.get("page") >= "100" and (d.meta.get("chapter") == "abstract" or d.meta.get("chapter") == "intro")
+        ]
```
Amnah199, line 201: Won't be needed as the calculation is done below
```suggestion
```

Amnah199 (changes requested):
Added the suggestions for improving the test.

integrations/astra/tests/test_document_store.py
ajtran: Right. Good catch here.

integrations/astra/tests/test_document_store.py
ajtran: ah i see! i thought it might be a more useful test to ensure that we actually get documents from the filters instead of just comparing the results since it's possible that the filters can get no documents and the comparison would still pass. what do you think?

Amnah199: That wouldn't be necessary as we are comparing two methods of retrieving the documents (using astra filters, and through simple operators). Passing these to `assert_documents_are_equal `  verifies both the number and content of the retrieved documents are equal.

Amnah199 (approved):
Thanks for your work @ajtran. Looks good for the merge.

PR #888, opened 2024-07-09
MonsterAPI Intergration

rishh76:
Monster API LLM models into Haystack.


anakin87:
Since this PR merely introduces a wrapper around OpenAI Chat Generator, I would be inclined to close it to reduce maintenance efforts.

You can instead focus on the related PR for the integration page, showing how MonsterAPI can be used in Haystack via OpenAI Generators. I will provide more details there.
 https://github.com/deepset-ai/haystack-integrations/pull/244

Issue #894, opened 2024-07-15
bug: Existing pipelines missing filter_policy in yaml files will break when loading


sjrl:
**Describe the bug**
Recently `FilterPolicy` was added in a series of PRs to Retrievers of each Document Store.

However, it was added in such a way that is required to specify a value in the yaml file even though the init methods of the Retrievers  have a default value for `FilterPolicy`. 

```python
data["init_parameters"]["filter_policy"] = FilterPolicy.from_str(data["init_parameters"]["filter_policy"])
```

**To Reproduce**
Try loading a pipeline from a yaml file with a `OpenSearchEmbeddingRetriever` without specifying `filter_policy`. This is an issue with all Retrievers all doc stores. 

**Describe your environment (please complete the following information):**
 - Haystack version: 2.3
 - Integration version: opensearch-haystack 0.8.0

wochinge:
Big thanks for the catch @sjrl !

PR #822, opened 2024-06-19
feat: Add filter_policy to opensearch integration

vblagoje:
### Why:
Adds flexible filtering options to opensearch integration. By introducing a filter policy option (`replace` or `merge`), developers can now control how runtime filters are applied relative to initialization time filters. 

- partially fixes https://github.com/deepset-ai/haystack-core-integrations/issues/780
- supersedes https://github.com/deepset-ai/haystack-core-integrations/pull/781

### What:
- Added `filter_policy` parameter with options `replace` and `merge` across multiple retrievers to control filter behavior dynamically.
- Unit tests were updated or added to validate the new functionality.

### How can it be used:

- **Dynamic Filter Behavior Adjustment:**  Users can decide whether to completely override the initial filters set during the retriever's initialization (`replace`) or merge them with runtime filters, with the latter taking precedence (`merge`).

- **Complex Search Scenarios:**  
    - In cases where the context of a query might dictate altering pre-set filters without discarding them, the `merge` option allows for an additive approach.
    - For strict query contexts that require ignoring initial filters, the `replace` option offers a clean slate for filters at runtime.

### How did you test it:
- Unit tests were enhanced or newly created to cover both `replace` and `merge` scenarios for the `filter_policy` parameter. 
- Tests ensure that filter logic is correctly applied based on the policy setting, whether it merges runtime filters with initial filters or replaces them entirely.
- Additional test cases should be considered for complex filter merge scenarios to ensure priority and override mechanisms function as expected.

PR #895, opened 2024-07-15
fix: `OpenSearch` - Fallback to default filter policy when deserializing retrievers without the init parameter

shadeMe:
### Related Issues

### Proposed Changes:

 
 
Fixes an oversight introduced in #822 that broke the deserialization of pipelines with older versions of the OpenSearch retriever components.

### How did you test it?

Unit test

### Checklist


sjrl (approved):
Thanks!

PR #899, opened 2024-07-15
fix: `Mongo` - Fallback to default filter policy when deserializing retrievers without the init parameter

vblagoje:


Amnah199 (approved):
Looks good!

PR #900, opened 2024-07-15
fix: `PgVector` - Fallback to default filter policy when deserializing retrievers without the init parameter

vblagoje:


integrations/pgvector/tests/test_retrievers.py
```diff
@@ -241,6 +241,54 @@ def test_from_dict(self, monkeypatch):
         assert retriever.filters == {"field": "value"}
         assert retriever.top_k == 5
 
+    @pytest.mark.usefixtures("patches_for_unit_tests")
+    def test_from_dict_with_filter_policy(self, monkeypatch):
```
davidsbatista, line 245:
```suggestion
    def test_from_dict_without_filter_policy(self, monkeypatch):
```

PR #920, opened 2024-07-23
feat: support aws authentication with OpenSearchDocumentStore

tstadel:
### Related Issues

Currently we can't make use of aws authentication when deserializing `OpenSearchDocumentStore` (e.g. from YAML). In code you would do the following for AWS IAM authentication:
```python
from opensearchpy import Urllib3AWSV4SignerAuth
import boto3
from haystack_integrations.document_stores.opensearch.document_store import OpenSearchDocumentStore

endpoint = 'my-endpoint'
region = 'eu-central-1'
session = boto3.Session(profile_name="my-profile", region_name=region)
credentials = session.get_credentials()
opensearch_auth = Urllib3AWSV4SignerAuth(credentials, region, service="es")

ds = OpenSearchDocumentStore(
    hosts=[{'host': endpoint, 'port': 443}],
    http_auth=opensearch_auth,
    index="my-index",
    use_ssl=True,
    verify_certs=True,
    timeout=300,
)
```
By design there is no way to serialize `Urllib3AWSV4SignerAuth`, boto3 credentials or `boto3.Session`. So we need to run this code during init.

### Proposed Changes:
- `http_auth` param accepts `AWSAuth` object
- `AWSAuth` configures additional `aws_` params for `boto3`
- if `http_auth=AWSAuth()`, load boto3 session based on `aws_` params

Above code snippet now looks like this assuming that all relevant aws environment variables are set (e.g. `AWS_DEFAULT_REGION`):
```python
from haystack_integrations.document_stores.opensearch.document_store import OpenSearchDocumentStore

ds = OpenSearchDocumentStore(
    hosts=[{'host': endpoint, 'port': 443}],
    http_auth=AWSAuth(),
    index="my-index",
    use_ssl=True,
    verify_certs=True,
    timeout=300,
)
```

### How did you test it?
- added unit tests
- tested locally with my AWS setup

### Notes for the reviewer
- this is required to enable custom components in deepset Cloud

### Checklist


tstadel:
@silvanocerza I just recalled that there are `DeserializationCallbacks`. Anyways, I think this PR is useful: due to the tight integration of OpenSearch and AWS, it would make lifes a lot easier for AWS OpenSearch users. WDYT?

silvanocerza:
I would structure this a bit differently though, it might be confusing and cause issues having alls those extra arguments just for AWS authentication.

Would be better to have a single `AWSAuth` class similar to how we approached different auth methods in [the Weaviate integration](https://github.com/deepset-ai/haystack-core-integrations/blob/main/integrations/weaviate/src/haystack_integrations/document_stores/weaviate/auth.py). That way you the `aws_auth` argument can be just `Optional[AWSAuth] = None` and it would simplify the amount of args needed.

That way we can also make the `boto3` and `botocore` dependency optional with a lazy import.

tstadel:
> I would structure this a bit differently though, it might be confusing and cause issues having alls those extra arguments just for AWS authentication.
> 
> Would be better to have a single `AWSAuth` class similar to how we approached different auth methods in [the Weaviate integration](https://github.com/deepset-ai/haystack-core-integrations/blob/main/integrations/weaviate/src/haystack_integrations/document_stores/weaviate/auth.py). That way the `aws_auth` argument can be just `Optional[AWSAuth] = None` and it would simplify the amount of args needed.
> 
> That way we can also make the `boto3` and `botocore` dependency optional with a lazy import.

@silvanocerza I applied the changes, looks way better now! :-) 
For easier serialization I added boolean options for `aws_auth`.

tstadel:
@silvanocerza I've made a couple of simplifications, which allows to pass `AWSAuth` to the already existing `http_auth` param. So changes in `OpenSearchDocumentStore` are reduced to (de)serializing `AWSAuth` objects mainly.

integrations/opensearch/src/haystack_integrations/common/opensearch/utils.py
```diff
@@ -0,0 +1,53 @@
...
+AWS_CONFIGURATION_KEYS = [
+    "aws_access_key_id",
+    "aws_secret_access_key",
+    "aws_session_token",
+    "aws_region_name",
+    "aws_profile_name",
+]
+
+
+def get_aws_session(
```
silvanocerza, line 19: I would move this in `auth.py` and make it private. It's used only there in any case.

integrations/opensearch/src/haystack_integrations/document_stores/opensearch/document_store.py
```diff
@@ -106,9 +125,14 @@ def _get_default_mappings(self) -> Dict[str, Any]:
     @property
     def client(self) -> OpenSearch:
         if not self._client:
-            self._client = OpenSearch(self._hosts, **self._kwargs)
-            # Check client connection, this will raise if not connected
-            self._client.info()  # type:ignore
```
silvanocerza, line 111: Any reason to remove this? 🤔

integrations/opensearch/tests/test_document_store.py
```diff
@@ -96,6 +112,153 @@ def test_get_default_mappings(_mock_opensearch_client):
...
 
+class TestAuth:
+    @pytest.fixture(autouse=True)
+    def mock_boto3_session(self):
+        with patch("boto3.Session") as mock_client:
+            yield mock_client
+
+    @patch("haystack_integrations.document_stores.opensearch.document_store.OpenSearch")
+    def test_init_with_basic_auth(self, _mock_opensearch_client):
+        OpenSearchDocumentStore(hosts="testhost", http_auth=("user", "pw")).client
```
silvanocerza, line 123: The access to the `client` property is a bit hidden, I would change this and the other tests to something similar to this:

```
document_store = OpenSearchDocumentStore(hosts="testhost", http_auth=("user", "pw"))
assert document_store.client
```

At that point you can also remove `# ruff: noqa: B018` from above I think.

integrations/opensearch/src/haystack_integrations/document_stores/opensearch/document_store.py
```diff
@@ -44,6 +45,10 @@ def __init__(
         mappings: Optional[Dict[str, Any]] = None,
         settings: Optional[Dict[str, Any]] = DEFAULT_SETTINGS,
         create_index: bool = True,
+        http_auth: Any = None,
```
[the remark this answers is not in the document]
tstadel: I'm fine with that. Just want to mention technically it could be anything that whatever the connection class specified takes as `http_auth` param. That's why opensearch-py also typed it as `Any` I guess. E.g. https://github.com/opensearch-project/opensearch-py/blob/a68e8b7daff4312e8c220cd531ab9a2e5f2ccf9e/opensearchpy/connection/http_urllib3.py#L108
This connection class checks for Callable, tuple and list for example:
https://github.com/opensearch-project/opensearch-py/blob/a68e8b7daff4312e8c220cd531ab9a2e5f2ccf9e/opensearchpy/connection/http_urllib3.py#L141-L149

integrations/opensearch/src/haystack_integrations/document_stores/opensearch/document_store.py
tstadel: Yes, if you scope OpenSearch calls to certain index patterns using AWS IAM policies, this would break as the resulting request is a "GET /" request against the cluster without any index pattern. Also this breaks AWS OpenSearch Serverless as "GET /" requests are not allowed. 
We don't need it anyways. The following exists() call would raise in a similar way if the cluster is not reachable.

integrations/opensearch/src/haystack_integrations/document_stores/opensearch/document_store.py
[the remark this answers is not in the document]
silvanocerza: Ok, cool. Let's leave it as `Any` then.

integrations/opensearch/src/haystack_integrations/document_stores/opensearch/document_store.py
silvanocerza: Ok, cool. 👍

silvanocerza (approved):

PR #921, opened 2024-07-24
fix: Weaviate: Use collections.list_all instead of collections._get_all

vblagoje:
- Weaviate released weaviate-client 4.7.0 library on 23.7.24 - this version started failing our CI tests.
- Seems like weaviate client removed internal method _get_all and introduced list_all. 
- This fix uses list_all


vblagoje:
@silvanocerza git blame says you are familiar with this code. Would you please also give it a nod? 🙏

silvanocerza (approved):

PR #927, opened 2024-07-29
fix: support streaming_callback param in amazon bedrock generators

tstadel:
### Related Issues

- use same streaming design/API as haystack main package to have only one way of streaming throughout haystack (incl. core-integrations)
- so far v1-like StreamingHandler was used

### Proposed Changes:
- adjust `AmazonBedrockGenerator` and `AmazonBedrockChatGenerator` to use the same streaming design as generators of haystack main package

### How did you test it?
- adjusted tests

### Notes for the reviewer
- it's safe to make this "breaking change" as the current streaming design/API resembled the v1 pattern which is not what users expect. But I'm almost 100% sure nobody is using streaming in the current form. Migration would be straight forward and you can stick to the previous release of really necessary.

### Checklist


elizaup:
Hey @silvanocerza hope you are doing well :) Would you have chance to look into this PR today at all please? We are demoing deepset cloud at a conference and would like streaming to be supported for haystack 2.0 pipelines. Thanks a lot!

silvanocerza (approved):

PR #930, opened 2024-07-30
[Test/WIP] feat: check compatibility

ArzelaAscoIi:
### Related Issues

- fixes #issue-number

### Checklist

### DIFF
--- a/.github/workflows/compatibility.yml
+++ b/.github/workflows/compatibility.yml
@@ -6,7 +6,6 @@ jobs:
   test-integrations:
     name: Install all packages
     runs-on: ubuntu-latest
-
     steps:
       - name: Checkout repository
         uses: actions/checkout@v2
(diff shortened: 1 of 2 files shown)


silvanocerza:
We shouldn't do this.

It's a mess that we decided we wanted to avoid completely with Haystack 2 and that's why we have the integrations completely separate.

This is a huge amount of work and creates a huge amount of conflicts and problems that we can't tackle in any way.

wochinge:
> It's a mess that we decided we wanted to avoid completely with Haystack 2 and that's why we have the integrations completely separate.

Fair point. However, when we talked to @shadeMe suggested otherwise. The test seems like a pretty simple way to at least be aware of conflicts.

wochinge:
@ArzelaAscoIi Talked to Madeesh about this. Let's try it out in our pipeline-images repo. Depending on whether there is any problem in the first place, we can reevaluate what the best way forward is (e.g. if there is a simple maintainable fix or if we need to change our approach for building the images). 

Based on the experience in our repository we can then probably also give better input on how much maintenance effort a similar check in this repository would be.

ArzelaAscoIi:
Sure! Sounds good

PR #937, opened 2024-08-01
fix: remove exposure of connection_string

Amnah199:
### Related Issues

Unnecessary exposure of resolved Mongodb connection string, making it prone to leaks.

### Proposed Changes:

Avoid storing the connection string value.

### How did you test it?

Ran the existing tests
### Checklist


shadeMe (approved):
Let's merge this one but a quick clarification - Just like with the component instance, anyone can still introspect the contents of the `MongoClient` instance to get the resolved connection string if they wanted to. So we're are not really moving the needle here 😅 

The important thing is to ensure that we don't serialize secrets to disk.

Amnah199:
I understand. Anyways, an unnecessary parameter removed. 😄

PR #969, opened 2024-08-12
chore: Update Qdrant tests for the new `apply_filter_policy` usage

vblagoje:
- Update qdrant `test_run_filters` test to match the results of the common `apply_filter_policy` algorithm applied across all db stores
- In this test case we need to: 
    - combine init filter and run filter (i.e MERGE filters)
    - we use AND logical operator by default
    - the result is 1 document matching both filters

**NOTE:**

- We need to wait for new haystack-ai to be released and then open this PR


davidsbatista (approved):

vblagoje:
@davidsbatista we'll wait for the new `haystack-ai` release to merge this PR, otherwise we'll break main CI

davidsbatista:
yeap, I'm aware of that, just did not want to block you when the release is done

PR #971, opened 2024-08-12
chore: Update mongodb test for the new `apply_filter_policy` usage

vblagoje:
- Update filter unit tests to use new filter syntax 
- Update MERGE unit test with new filter syntax and logic

**NOTE:**

- ~We need to wait for new haystack-ai to be released and then open this PR~


vblagoje:
@julian-risch a lot of the changes are related to simply removing old filter syntax

shadeMe (approved):

Issue #975, opened 2024-08-13
Bedrock: "Parameter 'tools' is not allowed and will be ignored."


lambda-science:
**Describe the bug**
When trying to use function calling with AWS Bedrock, the current integration ignore the tools keyword argument.
```python
generator_json = AmazonBedrockChatGenerator(
    model="anthropic.claude-3-haiku-20240307-v1:0",
    aws_region_name=Secret.from_token("eu-west-3"),
    generation_kwargs={"temperature": 0, "max_tokens": 4096}, aws_profile_name=Secret.from_token("sandbox")
)
response = generator_json.run(
    messages=messages,
    generation_kwargs={"tools":tools}
)
```
`08/13/2024 01:25:03 PM Parameter 'tools' is not allowed and will be ignored.`
Lines responsible of this error:
-  https://github.com/deepset-ai/haystack-core-integrations/blob/main/integrations/amazon_bedrock/src/haystack_integrations/components/generators/amazon_bedrock/chat/adapters.py#L76 
- https://github.com/deepset-ai/haystack-core-integrations/blob/main/integrations/amazon_bedrock/src/haystack_integrations/components/generators/amazon_bedrock/chat/adapters.py#L159

**Describe your environment (please complete the following information):**
 - OS: [e.g. iOS]: Linux Docker
 - Haystack version: 2.3.1
 - Integration version: `amazon-bedrock-haystack==0.9.3`

**Additional infos:**
Also this integration could be modernized with the new "Converse" API from Bedrock, that allow cross-model integration (same API for all providers), thus making easier tools calling. 
https://docs.aws.amazon.com/bedrock/latest/userguide/tool-use.html
> To use tools with a model you can use the Converse API ([Converse](https://docs.aws.amazon.com/bedrock/latest/APIReference/API_runtime_Converse.html) or [ConverseStream](https://docs.aws.amazon.com/bedrock/latest/APIReference/API_runtime_ConverseStream.html)). The example code in this topic uses the Converse API to show how to use a tool that gets the most popular song for a radio station. For general information about calling the Converse API, see [Use the Converse API](https://docs.aws.amazon.com/bedrock/latest/userguide/conversation-inference.html).

[...]

PR #976, opened 2024-08-13
fix(Bedrock): allow tools kwargs for AWS Bedrock Claude model

lambda-science:
### Related Issues

- fixes https://github.com/deepset-ai/haystack-core-integrations/issues/975

### Proposed Changes:
Allow the tools kwargs for AWS Bedrock Anthropic models because it is listed as valid here: https://docs.aws.amazon.com/bedrock/latest/userguide/model-parameters-anthropic-claude-messages.html

### How did you test it?

I did not, as it's a tiny fix and I didn't really have time to setup the whole testing process.

### Checklist

### DIFF
--- a/integrations/amazon_bedrock/src/haystack_integrations/components/generators/amazon_bedrock/chat/adapters.py
+++ b/integrations/amazon_bedrock/src/haystack_integrations/components/generators/amazon_bedrock/chat/adapters.py
@@ -166,6 +166,7 @@ class AnthropicClaudeChatAdapter(BedrockModelChatAdapter):
         "top_k",
         "system",
         "tools",
+        "tool_choice",
     ]
 
     def __init__(self, truncate: Optional[bool], generation_kwargs: Dict[str, Any]):
--- a/integrations/amazon_bedrock/src/haystack_integrations/components/generators/amazon_bedrock/chat/adapters.py
+++ b/integrations/amazon_bedrock/src/haystack_integrations/components/generators/amazon_bedrock/chat/adapters.py
@@ -164,6 +164,7 @@ class AnthropicClaudeChatAdapter(BedrockModelChatAdapter):
         "top_p",
         "top_k",
         "system",
+        "tools",
     ]
 
     def __init__(self, generation_kwargs: Dict[str, Any]):
--- a/integrations/amazon_bedrock/src/haystack_integrations/components/generators/amazon_bedrock/chat/adapters.py
+++ b/integrations/amazon_bedrock/src/haystack_integrations/components/generators/amazon_bedrock/chat/adapters.py
@@ -250,6 +250,10 @@ class AnthropicClaudeChatAdapter(BedrockModelChatAdapter):
                 if content.get("type") == "text":
                     meta = {k: v for k, v in response_body.items() if k not in ["type", "content", "role"]}
                     messages.append(ChatMessage.from_assistant(content["text"], meta=meta))
+                if content.get("type") == "tool_use":
+                    meta = {k: v for k, v in response_body.items() if k not in ["type", "content", "role"]}
+                    json_answer = json.dumps(content)
+                    messages.append(ChatMessage.from_assistant(json_answer, meta=meta))
         return messages
 
     def _build_streaming_chunk(self, chunk: Dict[str, Any]) -> StreamingChunk:
--- a/integrations/amazon_bedrock/src/haystack_integrations/components/generators/amazon_bedrock/chat/adapters.py
+++ b/integrations/amazon_bedrock/src/haystack_integrations/components/generators/amazon_bedrock/chat/adapters.py
@@ -253,14 +253,18 @@ class AnthropicClaudeChatAdapter(BedrockModelChatAdapter):
         """
         messages: List[ChatMessage] = []
         if response_body.get("type") == "message":
-            for content in response_body["content"]:
-                if content.get("type") == "text":
-                    meta = {k: v for k, v in response_body.items() if k not in ["type", "content", "role"]}
-                    messages.append(ChatMessage.from_assistant(content["text"], meta=meta))
-                if content.get("type") == "tool_use":
-                    meta = {k: v for k, v in response_body.items() if k not in ["type", "content", "role"]}
-                    json_answer = json.dumps(content)
-                    messages.append(ChatMessage.from_assistant(json_answer, meta=meta))
+            if response_body.get("stop_reason") == "tool_use":  # If `tool_use` we only keep the tool_use content
+                for content in response_body["content"]:
+                    if content.get("type") == "tool_use":
+                        meta = {k: v for k, v in response_body.items() if k not in ["type", "content", "role"]}
+                        json_answer = json.dumps(content)
+                        messages.append(ChatMessage.from_assistant(json_answer, meta=meta))
+            else:  # For other stop_reason, return all text content
+                for content in response_body["content"]:
+                    if content.get("type") == "text":
+                        meta = {k: v for k, v in response_body.items() if k not in ["type", "content", "role"]}
+                        messages.append(ChatMessage.from_assistant(content["text"], meta=meta))
+
         return messages
 
     def _build_streaming_chunk(self, chunk: Dict[str, Any]) -> StreamingChunk:
--- a/integrations/amazon_bedrock/CHANGELOG.md
+++ b/integrations/amazon_bedrock/CHANGELOG.md
@@ -1,5 +1,15 @@
 # Changelog
 
+## [integrations/amazon_bedrock-v1.0.0] - 2024-08-12
+
+### 🚜 Refactor
+
+- Change meta data fields (#911)
+
+### 🧪 Testing
+
+- Do not retry tests in `hatch run test` command (#954)
+
 ## [integrations/amazon_bedrock-v0.10.0] - 2024-08-12
 
 ### 🐛 Bug Fixes
--- a/integrations/amazon_bedrock/tests/test_chat_generator.py
+++ b/integrations/amazon_bedrock/tests/test_chat_generator.py
@@ -1,5 +1,6 @@
 import logging
 import os
+import json
 from typing import Optional, Type
 from unittest.mock import patch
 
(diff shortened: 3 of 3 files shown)


vblagoje:
LGTM but you have a look as well @davidsbatista but the context is in this Discord [thread](https://discord.com/channels/993534733298450452/1265258968142315561)

vblagoje:
@lambda-science would you please add a unit test, not as much to add some deep rigours unit test case here, but to have that test fail if tools was removed by whatever reason 🙏

lambda-science:
> @lambda-science would you please add a unit test, not as much to add some deep rigours unit test case here, but to have that test fail if tools was removed by whatever reason 🙏

I will look into it tomorrow :)

davidsbatista:
@lambda-science looks good to me, please as @vblagoje mentioned just add a unit test

lambda-science:
Hello,

I worked on adding a test for tool use using Bedrock. And multiple problem appeared (so it was a good idea).
The current Bedrock implementation was processing the replies too strictly and removing answers of type `tool_use`. So I had to modify the adapter to add `tool_use` replies type to the ChatMessages by dumping the `tool_use` as JSON content.

The issue is that the test is flacky, either it pass or it crash with:

```python
    @component.output_types(replies=List[ChatMessage])
    def run(
        self,
        messages: List[ChatMessage],
    ):
        """
        Generates a list of `ChatMessage` response to the given messages using the Amazon Bedrock LLM.

        :param messages: The messages to generate a response to.
        :param streaming_callback:
            A callback function that is called when a new token is received from the stream.
        :param generation_kwargs: Additional generation keyword arguments passed to the model.
        :returns: A dictionary with the following keys:
            - `replies`: The generated List of `ChatMessage` objects.
        """
        generation_kwargs = generation_kwargs or {}
        generation_kwargs = generation_kwargs.copy()

        streaming_callback = streaming_callback or self.streaming_callback
        generation_kwargs["stream"] = streaming_callback is not None

        # check if the prompt is a list of ChatMessage objects
        if not (
            isinstance(messages, list)
            and len(messages) > 0
            and all(isinstance(message, ChatMessage) for message in messages)
        ):
            msg = f"The model {self.model} requires a list of ChatMessage objects as a prompt."
            raise ValueError(msg)

        body = self.model_adapter.prepare_body(
            messages=messages, **{"stop_words": self.stop_words, **generation_kwargs}
        )
        try:
            if streaming_callback:
                response = self.client.invoke_model_with_response_stream(
                    body=json.dumps(body), modelId=self.model, accept="application/json", contentType="application/json"
                )
                response_stream = response["body"]
                replies = self.model_adapter.get_stream_responses(
                    stream=response_stream, streaming_callback=streaming_callback
                )
            else:
                response = self.client.invoke_model(
                    body=json.dumps(body), modelId=self.model, accept="application/json", contentType="application/json"
                )
                response_body = json.loads(response.get("body").read().decode("utf-8"))
                replies = self.model_adapter.get_responses(response_body=response_body)
        except ClientError as exception:
            msg = f"Could not inference Amazon Bedrock model {self.model} due: {exception}"
            raise AmazonBedrockInferenceError(msg) from exception

        # rename the meta key to be inline with OpenAI meta output keys
        for response in replies:
            if response.meta is not None and "usage" in response.meta:
>               response.meta["usage"]["prompt_tokens"] = response.meta["usage"].pop("input_tokens")
E               KeyError: 'input_tokens'

src\haystack_integrations\components\generators\amazon_bedrock\chat\chat_generator.py:209: KeyError

FAILED tests/test_chat_generator.py::TestAnthropicClaudeAdapter::test_tools_use[anthropic.claude-3-haiku-20240307-v1:0] - KeyError: 'input_tokens'
```

The reason is that this test produce two types of answers:
1. Either the model answer only ONE reply that is the tool_use type like `{'type': 'tool_use', 'id': 'toolu_bdrk_01NcAB4KKhQzdV6eX1ji1Hpk', 'name': 'top_song', 'input': {'sign': 'WZPZ'}}`
2. Or the model produce two replies: 
* One `text` type answer such as `{'type': 'text', 'text': 'Here is the most popular song played on radio station WZPZ:'}`
* One `tool_use` type answer as before: `{'type': 'tool_use', 'id': 'toolu_bdrk_01NcAB4KKhQzdV6eX1ji1Hpk', 'name': 'top_song', 'input': {'sign': 'WZPZ'}}`
But in this case only one instance of meta data is present (prompt/completion/input/ouput tokens). I'm not entirely sure how to handle this. Your input would be valuable. @vblagoje @davidsbatista 

Also I'm not sure what is the standard for `tool_use` output (in ChatMessage), I simply dumped the whole reply as json, maybe it's not the right way to do, but test is passing (C/C from Anthropic implementation -- non-bedrock).

Also this is 100% with the `InvokeModel` way of calling tools, the new `Converse` API from Bedrock use another tool spec input...

lambda-science:
Maybe we need a way to force the model to discard the first text answer if there is a tool_use answer right after ?

davidsbatista:
@lambda-science please add it as a new test function

lambda-science:
> @lambda-science please add it as a new test function

Hello, I'm not sure what you meant by this, as I added the test as a new function ?

On another note I think I solved the flacky test. What I did is that when the `stop_reason` is a `tool_use` we only return the content that has the `tool_use` type (effectively discarding the `text` type content). Else we return chat message as usual.

The flacky test was because when normalizing `meta` keys, we mutate a dictionnary that was not a deep copy (meta). So if we have a two replies, on the second repliy the dictionary keys are already modified ! My modification solve it in a way that, if it's a `tool_use`, there will be only one reply.

davidsbatista:
> Hello, I'm not sure what you meant by this, as I added the test as a new function ?

sorry, I oversee that you did it

lambda-science:
Ah, as possible further improvement for this PR we could/should also add the `tool_choice` option as well (with test, in the scope of function calling PR improvement). I will work on this in the futur day  https://docs.aws.amazon.com/bedrock/latest/userguide/model-parameters-anthropic-claude-messages.html

vblagoje:
Yes @lambda-science add it, your intuition is the best guideline here. I'll take a look tomorrow 🙏

lambda-science:
> Yes @lambda-science add it, your intuition is the best guideline here. I'll take a look tomorrow 🙏

Everything should be good now, ready for review :) params `tools` and `tool_choice` added with an implmentation test (you should run it in the CI/CD, I guess it doesn't by default)

integrations/amazon_bedrock/src/haystack_integrations/components/generators/amazon_bedrock/chat/adapters.py
```diff
@@ -252,10 +254,18 @@ def _extract_messages_from_response(self, response_body: Dict[str, Any]) -> List
...
+                    if content.get("type") == "tool_use":
+                        meta = {k: v for k, v in response_body.items() if k not in ["type", "content", "role"]}
+                        json_answer = json.dumps(content)
+                        messages.append(ChatMessage.from_assistant(json_answer, meta=meta))
+            else:  # For other stop_reason, return all text content
+                for content in response_body["content"]:
+                    if content.get("type") == "text":
+                        meta = {k: v for k, v in response_body.items() if k not in ["type", "content", "role"]}
+                        messages.append(ChatMessage.from_assistant(content["text"], meta=meta))
+
```
vblagoje, line 268: Seems a bit complex, no?

What about smth like:
```
if response_body.get("type") == "message":
    meta = {k: v for k, v in response_body.items() if k not in ["type", "content", "role"]}
    for content in response_body["content"]:
        text = content.get("text") or json.dumps(content.get("tool_use"))
        if text:
            messages.append(ChatMessage.from_assistant(text, meta=meta))

return messages
```

vblagoje (changes requested):
Looks good, but please try out this simplification I suggested

integrations/amazon_bedrock/src/haystack_integrations/components/generators/amazon_bedrock/chat/adapters.py
lambda-science: The problem of this is that as I explained when doing a tool use sometimes the LLM answer two replies leading to issues.

> Either the model answer only ONE reply that is the tool_use type like {'type': 'tool_use', 'id': 'toolu_bdrk_01NcAB4KKhQzdV6eX1ji1Hpk', 'name': 'top_song', 'input': {'sign': 'WZPZ'}} => TEST IS PASSING

> Or the model produce two replies:

> One text type answer such as {'type': 'text', 'text': 'Here is the most popular song played on radio station WZPZ:'}
> One tool_use type answer as before: {'type': 'tool_use', 'id': 'toolu_bdrk_01NcAB4KKhQzdV6eX1ji1Hpk', 'name': 'top_song', 'input': {'sign': 'WZPZ'}}

> But in this case only one instance of meta data is present (prompt/completion/input/ouput tokens).

The two answer situation leads to failing test (flacky) and also poor developper experience because now for two replies you only have one set of meta-data and you can't predict if the answer is valid JSON or just a 'pre-message' before the valid JSON for tool-use.

That's why my original code, in the case of a tool-use stop-sequence allow to discard the text answer before the actual tool-use JSON.

lambda-science: > Thanks for your patience. I think I know what is going on. You are getting the "thinking" [message](https://docs.anthropic.com/en/docs/build-with-claude/tool-use#chain-of-thought). We dealt with this by having a boolean switch in Anthropic to account for this thinking message. See #790 for more details.
> 
> What do you think we add the same init parameter here? Open to suggestions...

I think it's exactly what it is
 I guess it will need a bit more time for me to finish this PR then to add this new feature ^^

lambda-science: I'm not sure however where to put this parameters, because the `AmazonBedrockChatGenerator` is a generalist generator for a lot of models, so it's not great to put a specific init parameters for Anthropic model in its init method

vblagoje: Hmm, right. So we leave it as-is is (i.e. the current state of PR) is the best option, no?

lambda-science: > Hmm, right. So we leave it as-is is (i.e. the current state of PR) is the best option, no?


I guess yes, I don't see in which case somebody would need these thinking messages

vblagoje (approved):

lambda-science:
I would be happy to get pinged when release in pypi✌️

julian-risch:
@lambda-science ping 🙂

PR #978, opened 2024-08-14
chore: Update Anthropic example, use ChatPromptBuilder properly

vblagoje:
- Properly use ChatPromptBuilder init with documents variable
- Inject document content into prompt not just str of the document
- Verify component connections
- Verify output is based on fetched document


vblagoje:
@Amnah199 David and I talked about this example yesterday and how it needs an update. This is that update 😄

davidsbatista (approved):
I see now, `ChatPrompBuilder` changed

davidsbatista:
@vblagoje the docs also need to be updated: https://haystack.deepset.ai/integrations/anthropic#haystack-20

vblagoje:
Thank you for this @davidsbatista https://github.com/deepset-ai/haystack-integrations/pull/254 
All other docs are ok, merging this one

Issue #1004, opened 2024-08-19
Support prompt caching in Anthropic generators


julian-risch:
**Is your feature request related to a problem? Please describe.**
Anthropic, in particular the models Claude 3.5 Sonnet and Claude 3 Haiku support prompt caching or context caching. Support for Claude 3 Opus is coming soon. We should enable users to use that feature through Haystack to reduce costs and latency. 
https://docs.anthropic.com/en/docs/build-with-claude/prompt-caching
https://www.anthropic.com/news/prompt-caching

**Describe the solution you'd like**
We need to implement support for Anthropic's `cache_control` parameter in messages.

vblagoje:
To have a full benefit of prompt caching in Haystack pipelines we need to resolve https://github.com/deepset-ai/haystack-core-integrations/issues/1011

vblagoje:
Support for prompt caching added in [https://pypi.org/project/anthropic-haystack/ 1.1.0](https://pypi.org/project/anthropic-haystack/1.1.0/) release

PR #1006, opened 2024-08-19
feat: Add Anthropic prompt caching support, add example

vblagoje:
### Why:
Introduces [prompt caching](https://docs.anthropic.com/en/docs/build-with-claude/prompt-caching) for AnthropicChatGenerator. As prompt caching will be enabled by default in the near future we don't add a new init parameter for it.

- fixes https://github.com/deepset-ai/haystack-core-integrations/issues/1004

### What:
- **Added a new feature for prompt caching:** Enables caching of prompts for Anthropic LLMs to avoid repeated data fetches, reducing processing time and improving efficiency.
- **Implemented conditional usage of prompt caching based on configuration:** Allows users to enable or disable prompt caching through a configuration flag, giving them control over this feature based on their requirements.
- **Refined message handling for system and chat interactions:** Modifies how system and user chat messages are formatted and processed.
- **Updated project configurations:** Allow print statements in examples

### How can it be used:
- cache_control ChatMessage meta field
- In init of AnthropicChatGenerator we enable caching via extra_headers in generation_kwargs.

See `integrations/anthropic/example/prompt_caching.py` example for detailed usage

### How did you test it:
- Added unit tests
- Using `integrations/anthropic/example/prompt_caching.py` example and additional manual tests.


vblagoje:
cc Emil-io

TuanaCelik:
Hey @vblagoje  - here is my feedback that you asked for:
- Although I understand that you're saying that prompt caching will be enabled by default in near future, this suggests that users would also be able to turn it off if needed/wanted. So my intuition is that it would still be beneficial to add an init parameter for this, or something similar, which in future may indeed be used to _disable_
- I don't quite see how the example is showcasing the benefit/need for prompt caching. Can you explain a bit further please? Would help me review.

vblagoje:
> Hey @vblagoje - here is my feedback that you asked for:
> 
> * Although I understand that you're saying that prompt caching will be enabled by default in near future, this suggests that users would also be able to turn it off if needed/wanted. So my intuition is that it would still be beneficial to add an init parameter for this, or something similar, which in future may indeed be used to _disable_
> * I don't quite see how the example is showcasing the benefit/need for prompt caching. Can you explain a bit further please? Would help me review.

Thanks for feedback:
- prompt caching is turned on at ChatMessage level, see how Anthropic examples add `cache_control` to message [here](https://docs.anthropic.com/en/docs/build-with-claude/prompt-caching) there will be no need to set something on chat generator level 
- in the example, the benefit is that fetched doc is cached and used in subsequent inferencing with questions (everything is done on Anthropic side so it is not that visible)
  - the only visible effect is the speed. Run the example and you'll notice how questions are visibly answered faster (2nd and subsequent questions if we add them).

Emil-io:
Hey @vblagoje - first of all, this looks very interesting! Here my feedback, feel free to correct me if I made some false assumptions.

**1. How this fits into Haystack Pipelines**
I am trying to figure out the bigger picture and how this fits into Haystack. Prompt Caching works fine when the LLM is used as a stand alone component, but Haystack is not built for that. Moving to something with retrieval, I assume it makes sense to cache a long system prompt, since this is the only part that stays constant? But this would not perfectly align with the way the prompt builder is designed, as it also allows for dynamic changes in the jinja prompt template at any place. But maybe it would be nice to have an example in that way, as I assume most people would not use it as a stand alone component for haystack, but inside some more complex pipeline logic.

**2. Specifying the Caching**
To correctly specify the caching, the user definitely has to see this example (as this is also not explained in the documentation of the Anthropic Chat Component). Is it intended this way?

vblagoje:
Thanks for feedback Emil. 

1. Yes, it should work with pipelines, I discovered one bug/oversight in `ChatPromptBuilder` where we don't copy meta from messages - hence, right now, prompt caching won't work with `ChatPromptBuilder`. Having said that, as long as we set this cache_control somewhere, even in custom component just before the LLM - Anthropic caching should work in pipelines. 

2. Everything revolves around cache_control meta field of the `ChatMessage`, prompt caching is rather unobtrusive and that's how the authors of these APIs intended them to be.

vblagoje:
I've added data about prompt caching to be printed to stdout, confirming prompt caching

vblagoje:
@julian-risch please run the example yourself to see the prompt caching effect.
See the discussions above between Tuana, Emil, and me regarding this particular solution. 
I didn't add tests until we agree on this approach but it should be trivial to verify, the example prints proofs of caching
Note the discovered [bug](https://github.com/deepset-ai/haystack-core-integrations/issues/1011) in the process of testing

vblagoje:
@Emil-io have you tried the prompt caching example? @TuanaCelik can you take a look once again, run the example!

vblagoje:
@Amnah199 please have a look and I'll ask @julian-risch to have one as well. Running the example is a must. The speedup with prompt caching is visible but I expected it to be more prominent. Having said that - still important to have this feature as users will ask for it.

Amnah199:
@vblagoje, I tried the example, but the printed usage for all questions returned `'cache_creation_input_tokens': 0, 'cache_read_input_tokens': 0`. From the comments in example, I think it shouldn't be the case.
Additionally, response generation time did not significantly improve. Can we tweak the example to make the benefits of caching more obvious?

Emil-io:
> @Emil-io have you tried the prompt caching example? @TuanaCelik can you take a look once again, run the example!

HI @vblagoje 
sorry, I somehow overlooked this. Lmk if I should still try it out and run the example.

vblagoje:
Have you installed the branch version of Anthropic integration before running the example ?

vblagoje:
@Amnah199 @Emil-io the example should be easier to follow now, please try again 🙏

Amnah199:
@vblagoje explained the example in more detail and I have tested it. I think this use of prompt caching would make sense in certain use cases. Tagging @julian-risch for reference.

vblagoje:
@julian-risch let's integrate this, I can help @dfokina write a paragraph in AnthropicChatGenerator about it

julian-risch:
@vblagoje I am testing the example code right now. Still getting the "Cache not used" message with prompt_caching.py. It works with my own test code so could there be a small issue in prompt_caching.py?

vblagoje:
For some reason Anthropic caching doesn't seem to work on small messages (i.e. a short instruction). Perhaps there is a minima length they require cached content to be. I could recreate prompt_caching example in integration test? cc @julian-risch this is to test when prompt caching gets disabled as beta - perhaps we'll get some warning but I doubt an exception. Perhaps we can monitor the Anthropic prompt caching devs and eventually when prompt caching becomes default - adjust our code base at that time....

julian-risch:
@vblagoje Ah, true. Found the minimum cacheable length in their docs: https://docs.anthropic.com/en/docs/build-with-claude/prompt-caching#cache-limitations In that case, let's use 1024 tokens in the integration test? $3.75 / MTok is the cost for cache write so it's still cheap.

vblagoje:
Amazing, will do 🙏

vblagoje:
@Amnah199 and @julian-risch - this one should be ready now, lmk if you see any additional opportunities for improvement

julian-risch (approved):
LGTM! 👍 Please don't forget to write a paragraph in AnthropicChatGenerator about it.

vblagoje:
> LGTM! 👍 Please don't forget to write a paragraph in AnthropicChatGenerator about it.

Will do 🙏 - keeping this one open until prompt caching docs is integrated and a new release made

vblagoje:
Docs updated https://docs.haystack.deepset.ai/docs/anthropicchatgenerator 
@dfokina please move around and adjust the prompt caching section in the [docs](https://docs.haystack.deepset.ai/docs/anthropicchatgenerator) as you see fit.

vblagoje:
Prompt caching available in anthropic-haystack [integration](https://pypi.org/project/anthropic-haystack/) v1.1.0 onward.

PR #1023, opened 2024-08-26
feat: add NVIDIA NIM ranker support

mattf:
### Proposed Changes:

NVIDIA NIMs (hosted on https://build.nvidia.com and downloadable) support reranking models. This adds an NvidiaRanker component to access those models.

### How did you test it?

- [x] unit tests
- [x] integration tests of hosted NIM
- [x] integration tests of downloaded NIM

### Notes for the reviewer

1. should the input documents be mutated to add a score?
2. the ranker endpoints are not standardized, so i introduced an endpoint arg to rank(). an alternative is to refactor all components to pass the full inference endpoint to the backend.

### Checklist

### DIFF
--- a/integrations/nvidia/src/haystack_integrations/components/rankers/nvidia/ranker.py
+++ b/integrations/nvidia/src/haystack_integrations/components/rankers/nvidia/ranker.py
@@ -175,7 +175,7 @@ class NvidiaRanker:
         if not all(isinstance(doc, Document) for doc in documents):
             msg = "Ranker expects the `documents` parameter to be a list of Document objects."
             raise TypeError(msg)
-        if top_k is not None and (not isinstance(top_k, int) or isinstance(top_k, bool)):
+        if top_k is not None and not isinstance(top_k, int):
             msg = "Ranker expects the `top_k` parameter to be an integer."
             raise TypeError(msg)
 
--- a/integrations/nvidia/src/haystack_integrations/components/rankers/nvidia/ranker.py
+++ b/integrations/nvidia/src/haystack_integrations/components/rankers/nvidia/ranker.py
@@ -48,7 +48,7 @@ class NvidiaRanker:
     def __init__(
         self,
         model: Optional[str] = None,
-        truncate: Optional[Union[RankerTruncateMode, Literal["NONE", "END"]]] = None,
+        truncate: Optional[Union[RankerTruncateMode, str]] = None,
         api_url: Optional[str] = None,
         api_key: Optional[Secret] = None,
         top_k: int = 5,
@@ -74,7 +74,7 @@ class NvidiaRanker:
             msg = "Ranker expects the `api_url` parameter to be a string."
             raise TypeError(msg)
         if truncate is not None and not isinstance(truncate, RankerTruncateMode):
-            truncate = RankerTruncateMode(truncate)
+            truncate = RankerTruncateMode.from_str(truncate)
         if not isinstance(top_k, int):
             msg = "Ranker expects the `top_k` parameter to be an integer."
             raise TypeError(msg)
--- a/integrations/nvidia/src/haystack_integrations/components/rankers/nvidia/ranker.py
+++ b/integrations/nvidia/src/haystack_integrations/components/rankers/nvidia/ranker.py
@@ -174,14 +174,15 @@ class NvidiaRanker:
         if not isinstance(documents, list):
             msg = "Ranker expects the `documents` parameter to be a list."
             raise TypeError(msg)
-        if len(documents) == 0:
-            return {"documents": []}
         if not all(isinstance(doc, Document) for doc in documents):
             msg = "Ranker expects the `documents` parameter to be a list of Document objects."
             raise TypeError(msg)
         if top_k is not None and not isinstance(top_k, int):
             msg = "Ranker expects the `top_k` parameter to be an integer."
             raise TypeError(msg)
+
+        if len(documents) == 0:
+            return {"documents": []}
 
         top_k = top_k if top_k is not None else self._top_k
         if top_k < 1:
--- a/integrations/nvidia/src/haystack_integrations/components/rankers/nvidia/ranker.py
+++ b/integrations/nvidia/src/haystack_integrations/components/rankers/nvidia/ranker.py
@@ -22,10 +22,12 @@ class NvidiaRanker:
     Usage example:
     ```python
     from haystack_integrations.components.rankers.nvidia import NvidiaRanker
+    from haystack import Document
+    from haystack.utils import Secret
 
     ranker = NvidiaRanker(
         model="nvidia/nv-rerankqa-mistral-4b-v3",
-        api_key="your-api-key",
+        api_key=Secret.from_env_var("NVIDIA_API_KEY"),
     )
     ranker.warm_up()
 
--- a/integrations/nvidia/src/haystack_integrations/components/rankers/nvidia/ranker.py
+++ b/integrations/nvidia/src/haystack_integrations/components/rankers/nvidia/ranker.py
@@ -1,5 +1,5 @@
 import warnings
-from typing import Any, Dict, List, Literal, Optional, Union
+from typing import Any, Dict, List, Optional, Union
 
 from haystack import Document, component, default_from_dict, default_to_dict
 from haystack.utils import Secret, deserialize_secrets_inplace
--- a/integrations/nvidia/src/haystack_integrations/components/rankers/nvidia/ranker.py
+++ b/integrations/nvidia/src/haystack_integrations/components/rankers/nvidia/ranker.py
@@ -74,7 +74,7 @@ class NvidiaRanker:
         if truncate is not None and truncate not in ["NONE", "END"]:
             msg = "Invalid truncate value. Must be 'NONE' or 'END'."
             raise ValueError(msg)
-        if not isinstance(top_k, int) or isinstance(top_k, bool):
+        if not isinstance(top_k, int):
             msg = "Ranker expects the `top_k` parameter to be an integer."
             raise TypeError(msg)
 
--- a/integrations/nvidia/src/haystack_integrations/components/rankers/nvidia/ranker.py
+++ b/integrations/nvidia/src/haystack_integrations/components/rankers/nvidia/ranker.py
@@ -132,18 +132,19 @@ class NvidiaRanker:
 
         :raises ValueError: If the API key is required for hosted NVIDIA NIMs.
         """
-        model_kwargs = {}
-        if self._truncate is not None:
-            model_kwargs.update(truncate=str(self._truncate))
-        self._backend = NimBackend(
-            self._model,
-            api_url=self._api_url,
-            api_key=self._api_key,
-            model_kwargs=model_kwargs,
-        )
-        if not self._model:
-            self._model = _DEFAULT_MODEL
-        self._initialized = True
+        if not self._initialized:
+            model_kwargs = {}
+            if self._truncate is not None:
+                model_kwargs.update(truncate=str(self._truncate))
+            self._backend = NimBackend(
+                self._model,
+                api_url=self._api_url,
+                api_key=self._api_key,
+                model_kwargs=model_kwargs,
+            )
+            if not self._model:
+                self._model = _DEFAULT_MODEL
+            self._initialized = True
 
     @component.output_types(documents=List[Document])
     def run(
--- a/integrations/nvidia/src/haystack_integrations/components/rankers/nvidia/ranker.py
+++ b/integrations/nvidia/src/haystack_integrations/components/rankers/nvidia/ranker.py
@@ -1,10 +1,12 @@
 import warnings
-from typing import Any, Dict, List, Literal, Optional
+from typing import Any, Dict, List, Literal, Optional, Union
 
 from haystack import Document, component, default_from_dict, default_to_dict
 from haystack.utils import Secret, deserialize_secrets_inplace
 
 from haystack_integrations.utils.nvidia import NimBackend, url_validation
+
+from .truncate import RankerTruncateMode
 
 _DEFAULT_MODEL = "nvidia/nv-rerankqa-mistral-4b-v3"
 
@@ -46,7 +48,7 @@ class NvidiaRanker:
     def __init__(
         self,
         model: Optional[str] = None,
-        truncate: Optional[Literal["NONE", "END"]] = None,
+        truncate: Optional[Union[RankerTruncateMode, Literal["NONE", "END"]]] = None,
         api_url: Optional[str] = None,
         api_key: Optional[Secret] = None,
         top_k: int = 5,
@@ -57,7 +59,7 @@ class NvidiaRanker:
         :param model:
             Ranking model to use.
         :param truncate:
-            Truncation strategy to use. Can be "NONE" or "END". Defaults to NIM's default.
+            Truncation strategy to use. Can be "NONE", "END", or RankerTruncateMode. Defaults to NIM's default.
         :param api_key:
             API key for the NVIDIA NIM.
         :param api_url:
@@ -71,9 +73,8 @@ class NvidiaRanker:
         if not isinstance(api_url, (str, type(None))):
             msg = "Ranker expects the `api_url` parameter to be a string."
             raise TypeError(msg)
-        if truncate is not None and truncate not in ["NONE", "END"]:
-            msg = "Invalid truncate value. Must be 'NONE' or 'END'."
-            raise ValueError(msg)
+        if truncate is not None and not isinstance(truncate, RankerTruncateMode):
+            truncate = RankerTruncateMode(truncate)
         if not isinstance(top_k, int):
             msg = "Ranker expects the `top_k` parameter to be an integer."
             raise TypeError(msg)
@@ -133,7 +134,7 @@ class NvidiaRanker:
         """
         model_kwargs = {}
         if self._truncate is not None:
-            model_kwargs.update(truncate=self._truncate)
+            model_kwargs.update(truncate=str(self._truncate))
         self._backend = NimBackend(
             self._model,
             api_url=self._api_url,
--- a/integrations/nvidia/src/haystack_integrations/components/rankers/nvidia/ranker.py
+++ b/integrations/nvidia/src/haystack_integrations/components/rankers/nvidia/ranker.py
@@ -92,7 +92,7 @@ class NvidiaRanker:
         model: Optional[str] = _DEFAULT_MODEL,
         truncate: Optional[Literal["NONE", "END"]] = None,
         api_url: Optional[str] = None,
-        api_key: Optional[Secret] = Secret.from_env_var("NVIDIA_API_KEY"),
+        api_key: Optional[Secret] = None,
         top_k: int = 5,
     ):
         """
@@ -124,6 +124,7 @@ class NvidiaRanker:
 
         self._model = model
         self._truncate = truncate
+        self._api_key = api_key
         # if no api_url is provided, we're using a hosted model and can
         #  - assume the default url will work, because there's only one model
         #  - assume we won't call backend.models()
@@ -136,7 +137,8 @@ class NvidiaRanker:
                 raise ValueError(msg)
             self._api_url = None  # we handle the endpoint
             self._endpoint = _MODEL_ENDPOINT_MAP[self._model]
-        self._api_key = api_key
+            if api_key is None:
+                self._api_key = Secret.from_env_var("NVIDIA_API_KEY")
         self._top_k = top_k
         self._initialized = False
 
--- a/integrations/nvidia/src/haystack_integrations/components/rankers/nvidia/ranker.py
+++ b/integrations/nvidia/src/haystack_integrations/components/rankers/nvidia/ranker.py
@@ -1,0 +1,245 @@
+import warnings
+from typing import Any, Dict, List, Literal, Optional
+
+from haystack import Document, component, default_from_dict, default_to_dict
+from haystack.utils import Secret, deserialize_secrets_inplace
+
+from haystack_integrations.utils.nvidia import NimBackend, url_validation
+
+#
+# Ranker spec
+# -----------
+#
+# Ranker is a class that ranks a list of items (documents) based on a given text (query).
+#
+# The class should have the following methods:
+#
+# - `__init__(self,
+#             model: Optional[str],
+#             top_k: int = 5,
+#             truncate: Optional[int],
+#             api_url: Optional[str],
+#             api_key: Optional[str])`
+#   Constructor method that initializes the ranker.
+#
+# - `warm_up(self)`
+#   Initializes the ranker.
+#
+# - `to_dict(self) -> Dict[str, Any]`
+#   Serializes the ranker to a dictionary.
+#
+# - `from_dict(cls, data: Dict[str, Any]) -> "Ranker"`
+#   Deserializes the ranker from a dictionary.
+#
+# - `run(self,
+#        query: str,
+#        documents: List[Document],
+#        top_k: Optional[int]) -> List[Document]`
+#   Ranks a list of documents based on a given query.
+#
+# The class should have the following attributes:
+#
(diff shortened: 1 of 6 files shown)


integrations/nvidia/src/haystack_integrations/components/rankers/nvidia/ranker.py
```diff
@@ -0,0 +1,247 @@
...
+            warnings.warn("top_k should be at least 1, returning nothing", stacklevel=2)
+            return {"documents": []}
+
+        assert self._backend is not None
+        # rank result is list[{index: int, logit: float}] sorted by logit
+        sorted_indexes_and_scores = self._backend.rank(
+            query,
+            documents,
+            endpoint=self._endpoint,
+        )[:top_k]
```
davidsbatista, line 239:
```suggestion
        )
```

integrations/nvidia/src/haystack_integrations/components/rankers/nvidia/ranker.py
```diff
@@ -0,0 +1,247 @@
...
+        sorted_documents = []
+        for item in sorted_indexes_and_scores:
```
davidsbatista, line 241: Let's rather add it here for better readability

```suggestion
        for item in sorted_indexes_and_scores[:top_k]:
```

integrations/nvidia/src/haystack_integrations/components/rankers/nvidia/ranker.py
```diff
@@ -0,0 +1,247 @@
...
+            # todo: should we copy or mutate?
```
davidsbatista, line 242: Since we are only updating the `doc.score` field I would say there's no need to copy

integrations/nvidia/src/haystack_integrations/components/rankers/nvidia/ranker.py
```diff
@@ -0,0 +1,247 @@
...
+            raise TypeError(msg)
+        if not isinstance(documents, list):
+            msg = "Ranker expects the `documents` parameter to be a list."
+            raise TypeError(msg)
+        if len(documents) == 0:
+            return {"documents": []}
+        if not all(isinstance(doc, Document) for doc in documents):
+            msg = "Ranker expects the `documents` parameter to be a list of Document objects."
+            raise TypeError(msg)
+        if top_k is not None and (not isinstance(top_k, int) or isinstance(top_k, bool)):
```
davidsbatista, line 224: Why a `bool` ?

integrations/nvidia/src/haystack_integrations/components/rankers/nvidia/ranker.py
```diff
@@ -0,0 +1,247 @@
+import warnings
+from typing import Any, Dict, List, Literal, Optional
+
+from haystack import Document, component, default_from_dict, default_to_dict
+from haystack.utils import Secret, deserialize_secrets_inplace
+
+from haystack_integrations.utils.nvidia import NimBackend, url_validation
+
+#
```
davidsbatista, line 9: Let's remove this commented spec

integrations/nvidia/src/haystack_integrations/components/rankers/nvidia/ranker.py
```diff
@@ -0,0 +1,247 @@
...
+        if truncate is not None and truncate not in ["NONE", "END"]:
+            msg = "Invalid truncate value. Must be 'NONE' or 'END'."
+            raise ValueError(msg)
+        if not isinstance(top_k, int) or isinstance(top_k, bool):
+            msg = "Ranker expects the `top_k` parameter to be an integer."
+            raise TypeError(msg)
+
+        self._model = model
+        self._truncate = truncate
+        self._api_key = api_key
```
davidsbatista, line 127: we dropped the use of plain text API, keys, you can simply to this, and the user sets it with an environment variable  or pass with this parameter

```suggestion
        self._api_key = Secret.from_env_var("NVIDIA_API_KEY")
```

integrations/nvidia/src/haystack_integrations/components/rankers/nvidia/ranker.py
```diff
@@ -0,0 +1,247 @@
...
+        if api_url is not None:
+            self._api_url = url_validation(api_url, None, ["v1/ranking"])
+            self._endpoint = None  # we let backend.rank() handle the endpoint
+        else:
+            if model not in _MODEL_ENDPOINT_MAP:
+                msg = f"Model '{model}' is unknown. Please provide an api_url to access it."
+                raise ValueError(msg)
+            self._api_url = None  # we handle the endpoint
+            self._endpoint = _MODEL_ENDPOINT_MAP[self._model]
+            if api_key is None:
```
davidsbatista, line 140: You can remove this and initialised like in my suggestions

davidsbatista (commented):
Hi @mattf! Many thanks for your contribution, I've left a few comments

davidsbatista:
@anakin87 can you also have a look on this PR?

integrations/nvidia/src/haystack_integrations/components/rankers/nvidia/ranker.py
mattf: `isinstance(True/False, int)` -> `True`

integrations/nvidia/src/haystack_integrations/components/rankers/nvidia/ranker.py
mattf: should i entirely remove the `api_key` param to `__init__`?

integrations/nvidia/src/haystack_integrations/components/rankers/nvidia/ranker.py
mattf: there are two modes of operation -

 1. hosted - calls to api.nvidia.com, these require an api_key
 2. local - calls to some user known api_url, these do not require an api_key

i would like to only require the api_key in the first case. ok to keep this?

integrations/nvidia/src/haystack_integrations/components/rankers/nvidia/ranker.py
```diff
@@ -0,0 +1,201 @@
...
+
+        :raises ValueError: If the API key is required for hosted NVIDIA NIMs.
+        """
+        model_kwargs = {}
+        if self._truncate is not None:
+            model_kwargs.update(truncate=self._truncate)
+        self._backend = NimBackend(
+            self._model,
+            api_url=self._api_url,
+            api_key=self._api_key,
```
davidsbatista, line 137:
```suggestion
            api_key=self._api_key.resolve_value(),
```

integrations/nvidia/src/haystack_integrations/components/rankers/nvidia/ranker.py
davidsbatista: ah ok, I see, then here it's enough to to `api_key = Secret.from_env_var("NVIDIA_API_KEY")` and then call the `.resolve_value()` in the run() method

mattf: is it sufficient that resolve_value is called in the NimBackend init?

see https://github.com/deepset-ai/haystack-core-integrations/pull/1023/files#diff-4f9b82acfb7e3f7ebe10dc8f69f7f3e26a01b24e34c7b8a74cfd55c372cda5aaR41

integrations/nvidia/src/haystack_integrations/components/rankers/nvidia/ranker.py
mattf: if we allow True/False for top_k, that's what will happen.

if it's ok to accept True/False as 1/0, i can remove the bool portion of the condition.

imho, a user who pass in a bool is on the path to confusion. i'll defer to your judgement.

integrations/nvidia/src/haystack_integrations/components/rankers/nvidia/ranker.py
davidsbatista: ok, it's called there, I missed that

integrations/nvidia/src/haystack_integrations/components/rankers/nvidia/ranker.py
```diff
@@ -0,0 +1,201 @@
...
+    A component for ranking documents using ranking models provided by
+    [NVIDIA NIMs](https://ai.nvidia.com).
+
+    Usage example:
+    ```python
+    from haystack_integrations.components.rankers.nvidia import NvidiaRanker
+
+    ranker = NvidiaRanker(
+        model="nvidia/nv-rerankqa-mistral-4b-v3",
+        api_key="your-api-key",
```
silvanocerza, line 28: This wrong, `api_key` must be a `Secret`. I would change this like so:

```suggestion
        api_key=Secret.from_token("your-api-key"),
```

You also need to add its import.

integrations/nvidia/src/haystack_integrations/components/rankers/nvidia/ranker.py
```diff
@@ -0,0 +1,201 @@
...
+    ]
+
+    result = ranker.run(query, documents, top_k=2)
+    print(result["documents"])
+    ```
+    """
+
+    def __init__(
+        self,
+        model: Optional[str] = _DEFAULT_MODEL,
```
silvanocerza, line 46: Why is this `Optional` if you set a default and enforce it to be set down below?

integrations/nvidia/src/haystack_integrations/components/rankers/nvidia/ranker.py
```diff
@@ -0,0 +1,201 @@
...
+            self._model,
+            api_url=self._api_url,
+            api_key=self._api_key,
+            model_kwargs=model_kwargs,
+        )
+        if not self._model:
+            self._model = _DEFAULT_MODEL
+        self._initialized = True
+
+    @component.output_types(replies=List[Document])
```
silvanocerza, line 144:
```suggestion
    @component.output_types(documents=List[Document])
```

integrations/nvidia/src/haystack_integrations/components/rankers/nvidia/ranker.py
silvanocerza: The `bool` check is not necessary, you're already checking if it's not an `int`. Also the `run` type hint clearly says that it mus be an `int`.

```
In [1]: top_k = 1

In [2]: not isinstance(top_k, int) or isinstance(top_k, bool)
Out[2]: False

In [3]: top_k = True

In [4]: not isinstance(top_k, int) or isinstance(top_k, bool)
Out[4]: True

In [5]: top_k = 0.1

In [6]: not isinstance(top_k, int) or isinstance(top_k, bool)
Out[6]: True

In [7]: top_k = "1"

In [8]: not isinstance(top_k, int) or isinstance(top_k, bool)
Out[8]: True
```

integrations/nvidia/src/haystack_integrations/components/rankers/nvidia/ranker.py
```diff
@@ -0,0 +1,201 @@
...
+        if not isinstance(model, str):
+            msg = "Ranker expects the `model` parameter to be a string."
+            raise TypeError(msg)
+        if not isinstance(api_url, (str, type(None))):
+            msg = "Ranker expects the `api_url` parameter to be a string."
+            raise TypeError(msg)
+        if truncate is not None and truncate not in ["NONE", "END"]:
+            msg = "Invalid truncate value. Must be 'NONE' or 'END'."
+            raise ValueError(msg)
+        if not isinstance(top_k, int) or isinstance(top_k, bool):
```
silvanocerza, line 75: No need to check it's a `bool`, it must be an `int` in any case as the type hint above implies too.

integrations/nvidia/src/haystack_integrations/components/rankers/nvidia/ranker.py
```diff
@@ -0,0 +1,201 @@
...
+        if len(documents) == 0:
+            return {"documents": []}
+        if not all(isinstance(doc, Document) for doc in documents):
+            msg = "Ranker expects the `documents` parameter to be a list of Document objects."
+            raise TypeError(msg)
+        if top_k is not None and (not isinstance(top_k, int) or isinstance(top_k, bool)):
+            msg = "Ranker expects the `top_k` parameter to be an integer."
+            raise TypeError(msg)
+
+        top_k = top_k if top_k is not None else self._top_k
```
silvanocerza, line 182: Can be simplified.

```suggestion
        top_k = top_k or self._top_k
```

integrations/nvidia/src/haystack_integrations/utils/nvidia/nim_backend.py
```diff
@@ -129,3 +130,28 @@ def models(self) -> List[Model]:
...
             raise ValueError(msg)
         return models
+
+    def rank(
+        self,
+        query: str,
+        documents: List[Document],
+        endpoint: Optional[str] = None,
+    ) -> List[Dict[str, Any]]:
+        url = endpoint if endpoint else f"{self.api_url}/ranking"
```
silvanocerza, line 140:
```suggestion
        url = endpoint or f"{self.api_url}/ranking"
```

integrations/nvidia/tests/test_ranker.py
silvanocerza, line 1: Would be great if we could name tests in a more descriptive way so we can easily understand what they're testing. `test_documents_typeerror2` is really not a great name for a test.

integrations/nvidia/tests/test_base_url.py
```diff
@@ -60,6 +61,16 @@ def test_base_url_valid_generator(base_url: str) -> None:
         "http://localhost:8888/chat/completions",
     ],
 )
+# todo: add this to test_base_url_invalid_not_hosted
```
silvanocerza, line 64: What is this about?

silvanocerza (changes requested):
Left some comments, there are some imperfectiosn to fix.

integrations/nvidia/src/haystack_integrations/components/rankers/nvidia/ranker.py
mattf: my thinking is `run(..., top_k=...)` should always override `NvidiaRanker(..., top_k=...)` including when `run(..., top_k=0)`

`top_k or self._top_k` -> `self._top_k` when `top_k=0`

silvanocerza: Ah right, didn't consider `top_k` could be `0`. 🤦 

Disregard this.

integrations/nvidia/tests/test_base_url.py
mattf: a note to myself, which i decided to skip because it was a refactor. i'll remove it.

integrations/nvidia/src/haystack_integrations/components/rankers/nvidia/ranker.py
mattf: i changed it to be None and set default within the constructor

integrations/nvidia/tests/test_ranker.py
mattf: i combined the two tests into one

shadeMe:
@mattf FYI - we avoid force-pushing to a branch once a review's been published. Otherwise, the review comments lose their anchors and it's difficult for the reviewer to diff changes since their previous review.

integrations/nvidia/src/haystack_integrations/components/rankers/nvidia/ranker.py
```diff
@@ -0,0 +1,204 @@
...
+    ]
+
+    result = ranker.run(query, documents, top_k=2)
+    print(result["documents"])
+    ```
+    """
+
+    def __init__(
+        self,
+        model: Optional[str] = None,
```
silvanocerza, line 48: I still don't get the benefit of this being `Optional[str]`, can't we just set it to `str` and a have proper default?

mattf:
> @mattf FYI - we avoid force-pushing to a branch once a review's been published. Otherwise, the review comments lose their anchors and it's difficult for the reviewer to diff changes since their previous review.

understood. it was one of the options the github presented to catch up to main when the pr was out of date. it may be possible to disable this in the project settings.

integrations/nvidia/src/haystack_integrations/components/rankers/nvidia/ranker.py
mattf: this is future proofing. in the case of a downloaded NIM, the default can be detected by querying /v1/models on the api_url.

an alternative is to `model: str = _DEFAULT_MODEL` and in the future check `api_url is not None and model == _DEFAULT_MODEL` before getting a new default based on `{api_url}/models`. a corner case being if the user actually specified a `model=...` that happened to match `_DEFAULT_MODEL`. my inclination is to keep it `None` so we can tell if the user passed `model` or not (sans the case where they pass `model=None`).

integrations/nvidia/src/haystack_integrations/components/rankers/nvidia/ranker.py
```diff
@@ -0,0 +1,204 @@
...
+            msg = "Ranker expects the `api_url` parameter to be a string."
+            raise TypeError(msg)
+        if truncate is not None and truncate not in ["NONE", "END"]:
+            msg = "Invalid truncate value. Must be 'NONE' or 'END'."
+            raise ValueError(msg)
+        if not isinstance(top_k, int):
+            msg = "Ranker expects the `top_k` parameter to be an integer."
+            raise TypeError(msg)
+
+        # todo: detect default in non-hosted case (when api_url is provided)
```
shadeMe, line 81: Is this not something we can handle in this PR?

integrations/nvidia/src/haystack_integrations/components/rankers/nvidia/ranker.py
```diff
@@ -0,0 +1,204 @@
...
+
+    result = ranker.run(query, documents, top_k=2)
+    print(result["documents"])
+    ```
+    """
+
+    def __init__(
+        self,
+        model: Optional[str] = None,
+        truncate: Optional[Literal["NONE", "END"]] = None,
```
shadeMe, line 49: Can we replace this with an enum like we've done in the [embedders](https://github.com/deepset-ai/haystack-core-integrations/blob/36c5401f2c8a3ae4465afcc00702e0d0f5cacdc7/integrations/nvidia/src/haystack_integrations/components/embedders/nvidia/truncate.py#L4)?

integrations/nvidia/src/haystack_integrations/components/rankers/nvidia/ranker.py
```diff
@@ -0,0 +1,204 @@
...
+            if api_key is None:
+                self._api_key = Secret.from_env_var("NVIDIA_API_KEY")
+        self._top_k = top_k
+        self._initialized = False
+
+    def to_dict(self) -> Dict[str, Any]:
+        """
+        Serialize the ranker to a dictionary.
+
+        :return: A dictionary containing the ranker's attributes.
```
shadeMe, line 106:
```suggestion
        :returns: A dictionary containing the ranker's attributes.
```

integrations/nvidia/src/haystack_integrations/components/rankers/nvidia/ranker.py
```diff
@@ -0,0 +1,204 @@
...
+            api_key=self._api_key,
+        )
+
+    @classmethod
+    def from_dict(cls, data: Dict[str, Any]) -> "NvidiaRanker":
+        """
+        Deserialize the ranker from a dictionary.
+
+        :param data: A dictionary containing the ranker's attributes.
+        :return: The deserialized ranker.
```
shadeMe, line 123:
```suggestion
        :returns: The deserialized ranker.
```

integrations/nvidia/src/haystack_integrations/components/rankers/nvidia/ranker.py
```diff
@@ -0,0 +1,204 @@
...
+            model_kwargs.update(truncate=self._truncate)
+        self._backend = NimBackend(
+            self._model,
+            api_url=self._api_url,
+            api_key=self._api_key,
+            model_kwargs=model_kwargs,
+        )
+        if not self._model:
+            self._model = _DEFAULT_MODEL
+        self._initialized = True
```
shadeMe, line 145: Let's exit early if `self._initialized` is already `True`.

integrations/nvidia/src/haystack_integrations/components/rankers/nvidia/ranker.py
```diff
@@ -0,0 +1,204 @@
...
+
+        :param query: The query to rank the documents against.
+        :param documents: The list of documents to rank.
+        :param top_k: The number of documents to return.
+        :return: A list of ranked documents
+
+        :raises RuntimeError: If the ranker has not been loaded.
+        :raises TypeError: If the arguments are of the wrong type.
+
+        :return: A dictionary containing the ranked documents.
```
shadeMe, line 165:
```suggestion
        :returns: A dictionary containing the ranked documents.
```

integrations/nvidia/src/haystack_integrations/components/rankers/nvidia/ranker.py
```diff
@@ -0,0 +1,204 @@
...
+            msg = "Ranker expects the `documents` parameter to be a list."
+            raise TypeError(msg)
+        if len(documents) == 0:
+            return {"documents": []}
+        if not all(isinstance(doc, Document) for doc in documents):
+            msg = "Ranker expects the `documents` parameter to be a list of Document objects."
+            raise TypeError(msg)
+        if top_k is not None and not isinstance(top_k, int):
+            msg = "Ranker expects the `top_k` parameter to be an integer."
+            raise TypeError(msg)
```
shadeMe, line 183:
```suggestion
        if not all(isinstance(doc, Document) for doc in documents):
            msg = "Ranker expects the `documents` parameter to be a list of Document objects."
            raise TypeError(msg)
        if top_k is not None and not isinstance(top_k, int):
            msg = "Ranker expects the `top_k` parameter to be an integer."
            raise TypeError(msg)

        if len(documents) == 0:
            return {"documents": []}            
```
To group the errors.

shadeMe (changes requested):
Just a few small changes before we can merge it.

integrations/nvidia/src/haystack_integrations/components/rankers/nvidia/ranker.py
mattf: i added `RankerTruncateMode(str, Enum)`

i did not use `from_str`, instead relying on `Enum` constructor. let me know if i should implement `from_str`.

integrations/nvidia/src/haystack_integrations/components/rankers/nvidia/ranker.py
mattf: i would prefer to do this in a follow up feat, if that's ok with you

integrations/nvidia/src/haystack_integrations/components/rankers/nvidia/ranker.py
mattf: updated to skip second warm_ups

integrations/nvidia/src/haystack_integrations/components/rankers/nvidia/ranker.py
shadeMe: We'd prefer to keep it consistent across the codebase (it's a common pattern that we use), so it'd be great if we could add `from_str`, use `truncate: Optional[Union[RankerTruncateMode, str]] = None` as the init param and call `RankerTruncateMode.from_str` if the passed value is a string. You can basically use the same code as the one in the embedder truncate enum.

integrations/nvidia/src/haystack_integrations/components/rankers/nvidia/ranker.py
shadeMe: Works for me.

shadeMe (approved):
LGMT! Thanks for the quick turnaround!

PR #1025, opened 2024-08-27
refactor: Use Qdrant's Query API

Anush008:
## Description

This PR updates the Qdrant integration to use the new [Query API](https://qdrant.tech/blog/qdrant-1.10.x/) that supersedes the search API.
It also support server-side fusion of the hybrid results.


anakin87 (approved):
That looks like a nice simplification.

Thank you for the PR!

PR #1034, opened 2024-08-28
fix: converting `Pinecone` metadata fields from float back to int

davidsbatista:
### Proposed Changes:

Pinecone converts all meta numbers in the meta field to float (https://docs.pinecone.io/guides/data/filter-with-metadata). This causes the `SentenceWindowRetriever` to crash completely.

This PR checks if the metadata values are floats and convert them back to integers making `Pinecone` supported by the `SentenceWindowRetriever`

### How did you test it?

- manual verification, unit tests and integration tests

### Checklist


integrations/pinecone/src/haystack_integrations/document_stores/pinecone/document_store.py
```diff
@@ -267,6 +267,19 @@ def _embedding_retrieval(
 
         return self._convert_query_result_to_documents(result)
 
+    @staticmethod
+    def _convert_to_int(metadata: Dict[str, Any]) -> Dict[str, Any]:
+        """
+        Convert metadata original int values back to integers again, since PineCone store numbers as float.
```
anakin87, line 273: Let's explain we only do this for specific metadata fields used by some retrievers.
nit: "Pinecone" instead of "PineCone"

integrations/pinecone/tests/test_document_store.py
```diff
@@ -257,3 +260,59 @@ def test_embedding_retrieval(self, document_store: PineconeDocumentStore):
         assert len(results) == 2
         assert results[0].content == "Most similar document"
         assert results[1].content == "2nd best document"
+
+    @pytest.mark.skipif("PINECONE_API_KEY" not in os.environ, reason="PINECONE_API_KEY not set")
```
anakin87, line 264:
```suggestion
```
this marker is inherited from `TestDocumentStore` test class, so we don't need it.

integrations/pinecone/tests/test_document_store.py
```diff
@@ -257,3 +260,59 @@ def test_embedding_retrieval(self, document_store: PineconeDocumentStore):
...
+    def test_sentence_window_retriever(self, sleep_time):
```
anakin87, line 265: I believe this test could be made much shorter if we use the `document_store` fixture, as done for example in `test_write_documents_duplicate_overwrite`.

The fixture should take care of creating and deleting the document store, automatically sleep after writing docs, etc...

anakin87 (commented):
The logic looks good.

I found some small opportunities for improvement.
Let's also add a small unit test for the new class method.

davidsbatista:
Do you think an extra unit test is necessary? It's a private function, and testing its functionality is already covered by other test

anakin87:
> Do you think an extra unit test is necessary? It's a private function, and testing its functionality is already covered by other test

I'll leave it up to you; I'd like to have it.

integrations/pinecone/tests/test_document_store.py
davidsbatista: good point 👍🏽 thanks

davidsbatista:
> > Do you think an extra unit test is necessary? It's a private function, and testing its functionality is already covered by other test
> 
> I'll leave it up to you; I'd like to have it.

integrations/pinecone/tests/test_document_store.py
```diff
@@ -257,3 +260,59 @@ def test_embedding_retrieval(self, document_store: PineconeDocumentStore):
         assert len(results) == 2
         assert results[0].content == "Most similar document"
         assert results[1].content == "2nd best document"
+
+    def test_sentence_window_retriever(self, sleep_time, document_store: PineconeDocumentStore):
```
anakin87, line 264: You don't need `sleep_time`. It should be automatically applied when using the `PineconeDocumentStore` fixture.
https://github.com/deepset-ai/haystack-core-integrations/pull/1034#discussion_r1735809950

integrations/pinecone/src/haystack_integrations/document_stores/pinecone/document_store.py
```diff
@@ -267,6 +267,19 @@ def _embedding_retrieval(
 
         return self._convert_query_result_to_documents(result)
 
+    @staticmethod
+    def _convert_to_int(metadata: Dict[str, Any]) -> Dict[str, Any]:
+        """
+        Convert metadata original int values back to integers again, since Pinecone store metadata numbers as float.
```
anakin87, line 273: Let's explain we only do this for specific metadata fields used by some retrievers.

davidsbatista: I need your help in understanding what's missing, it's explicitly on the docstring why and how is used, even being a private function.

Please add here, in the comment, the exact wording you would like to see, and I change it.

anakin87 (approved):
I pushed some small changes.
I'll merge if the tests pass.

PR #1049, opened 2024-09-04
fix: #1047 Remove count_documents from delete_documents

### LINKED ISSUE #1047: AstraDB delete_documents performs an unnecessary count
https://github.com/deepset-ai/haystack-core-integrations/blob/main/integrations/astra/src/haystack_integrations/document_stores/astra/document_store.py#L406-L427

In delete_documents, an unnecessary check is being done, which is performing a full count of the collection as a naive check to ensure the collection is not empty. This is an extremely expensive operation and not one that is necessary, as performing a delete on nonexisting documents is not harmful. In reality, the existing check does not actually check if the document exists or not anyway.

The count_documents check should be removed.

brendancicchi:
### Related Issues

- fixes #1047

### Proposed Changes:

Removed the expensive full-count check to see if the collection is non-empty. Instead finding any single document as a check for non-emptiness.

### How did you test it?

Ran:
- `hatch run lint:all`
- `hatch run test`

### Notes for the reviewer

This [test](https://github.com/deepset-ai/haystack/blob/main/haystack/testing/document_store.py#L161C5-L163C61) ensures that the behavior for an empty collection is not the same as the behavior when all document_ids to be deleted are not found.

### Checklist


vblagoje (approved):
@brendancicchi looks good. I somehow overlooked notification about this PR 🙏

Issue #977, opened 2024-08-13
Bedrock Converse API integration


FloRul:
# Integrate AWS Bedrock Converse API into Haystack

## Background
AWS Bedrock released earlier this year their new converse API that aims to standardize their api calls to their models.

## Objective
Integrate the AWS Bedrock Converse API into the Haystack integration library to allow users to leverage AWS Bedrock converse within their Haystack pipelines, paving the way potentially for easier agent development and multimodal integration.

## Proposed Features
1. Add a new `BedrockConversationGenerator` node in Haystack integrations.
2. Support easy tool calling with some new classes to ease the process of tools definition and calling.
3. Support streaming responses from the Bedrock Converse API.
4. Implement error handling and retries for API calls.

## Implementation Details
- Use the AWS SDK for Python (Boto3) to interact with the Bedrock Converse API.
- Follow Haystack's node implementation guidelines for consistency.

## Resources
- [AWS Bedrock Converse Documentation](https://docs.aws.amazon.com/bedrock/latest/APIReference/API_runtime_Converse.html)
- [Model/Features compatibility](https://docs.aws.amazon.com/bedrock/latest/userguide/conversation-inference.html)

### Alternatives
This could also be a feature for the existing AmazonBedrock generators instead of a new one.

FloRul:
Currently working on it

lambda-science:
> Currently working on it

Following on this, it will help a lot with tool use across all AWS Bedrock models with this unified API

julian-risch:
@vblagoje Related issue: https://github.com/deepset-ai/haystack-core-integrations/issues/1229

vblagoje:
Completed with https://github.com/deepset-ai/haystack-core-integrations/pull/1219
https://pypi.org/project/amazon-bedrock-haystack/ v2.0.0 released

PR #1061, opened 2024-09-07
feat: bedrock converse

FloRul:
### Related Issues
[#977](https://github.com/deepset-ai/haystack-core-integrations/issues/977)

### Proposed Changes:
Added a new converse generator, some other utility classes have been added to facilitate its use such as ConverseMessage or ToolConfig

### How did you test it?

unit test and manual testing mostly

### Checklist


FloRul:
@silvanocerza  It seems like the AWS region is not assigned in the CI. I had the same issues in my previous PR.

silvanocerza:
@FloRul PR from forks can't have access to repository secrets. 
Those tests should be skipped if the necessary secrets are not found, see [this as an example](https://github.com/deepset-ai/haystack-core-integrations/blob/7cd71f17a3f86a22f74e85db2825f993c597e5b8/integrations/anthropic/tests/test_chat_generator.py#L165-L168).

silvanocerza:
Why do we need a new converse generator? Isn't it akin to a chat generator that we already have? Shouldn't we update that?

It seems like all the abstractions added are extremely similar to existing ones for chat generators. 🤔

FloRul:
> Why do we need a new converse generator? Isn't it akin to a chat generator that we already have? Shouldn't we update that?
> 
> It seems like all the abstractions added are extremely similar to existing ones for chat generators. 🤔

That is a fair point, the thing with the current bedrock chat is that it relies on adpaters, would it be more logic to consider converse its own adapter ? Keeping in mind that aws seems to aim at converse being the main point of entry for bedrock inference api.

silvanocerza:
> That is a fair point, the thing with the current bedrock chat is that it relies on adpaters, would it be more logic to consider converse its own adapter ? Keeping in mind that aws seems to aim at converse being the main point of entry for bedrock inference api.

I would consider converse its own adapter. If AWS goes all in with this kind of common API we can ditch completely the adapter concepts too. That's less code to maintain. :)

Maybe the best approach would be to ditch the adapters for those models that support the new Converse API, that should work right?

FloRul:
The idea behind adding a new generator was because of the standardization, which results in a richer output from the generator as well as facilitating the multimodal integration.
I think trying to put the implementation in the former ChatGenerator would either not be retrocompatible or making it lose all the benefits this implementation tries to provide: easier tool calling, multimodal integration and richer output.

silvanocerza:
We already have a standard and that is the `ChatMessage`. Adding all the classes from `utils.py` goes against that.

This Component will be extremely hard to use in any existing `Pipeline`s in place of other generators also. The `ConverseMessage` is a completely new type that no other Component returns as output nor expects as input. It would be completely isolated.

That's why I'm advocating to use the existing interfaces and types.

FloRul:
Ok @silvanocerza I'll look into that.

anakin87:
Superseded by #1219

PR #1068, opened 2024-09-09
chore: PgVector - remove legacy filter support

vblagoje:
- Removes legacy filter support in PgVector (i.e. use of `convert` from from `haystack-ai`)
- See https://github.com/deepset-ai/haystack/pull/8342 for more details


vblagoje:
@Amnah199 can you please have a look as @anakin87 is likely travelling today

Amnah199 (approved):

PR #1083, opened 2024-09-12
chore: Update changelog after removing legacy filters

vblagoje:
I ran git-cliff with parameters from github CI runs that failed. Note that most changelogs have this unreleased section while elastic has the actual tag. I wasn't sure what's right here, perhaps you know - I'd guess the actual tag?


vblagoje:
Yeah, I think we should replace unreleased with actual tag....

vblagoje:
Please have a look now @silvanocerza

silvanocerza (approved):
Looks good 👍

PR #1086, opened 2024-09-16
feat: Cohere LLM - adjust token counting meta to match OpenAI format

vblagoje:
- we forgot to update Cohere LLM so that the meta token counting format matches OpenAI
- we need this format to match for general portability of generators
- we need this to turn on token counting and generation span in Langfuse integration


davidsbatista (approved):
LGTM - don't forget to update the integration docs/code example - now the output is different

vblagoje:
> LGTM - don't forget to update the integration docs/code example - now the output is different
Docs don't mention these details - which is ok.  👍

Issue #746, opened 2024-05-23
Avoid database connections in document stores constructors


masci:
**Describe the bug**
Not a but per-se, but currently the behaviour of some document stores prevents pipeline's validation and serialisation.

The way Haystack pipelines validate and serialise requires all the components to be able to initialise fast, hence any operation loading or downloading data is demanded to the `warm_up` method of the component classes (when needed).

This should be the case for document stores too, for example:
- We don't necessarily want to reach the network and connect to a database only in order to create (and validate) a pipeline
- If we're building a pipeline in order to serialise it and send it over the wire, the database might not be available at "pipeline-build time"

Contrary to regular components, a `warm_up` strategy wouldn't be effective, because we don't have a single point in the codebase where the pipeline can call `document_store.warm_up()`. In fact, a document store can be passed to a writer, to a reader, or even used standalone.

For the time being, we decided to roll out a better init strategy for document stores without introducing new abstractions nor breaking changes. This means that every document store, for convention and where it makes sense, will defer the database connection until really needed.

This is the list of document stores affected that we need to change:

```[tasklist]
### Tasks
- [ ] https://github.com/deepset-ai/haystack-core-integrations/issues/767
- [ ] https://github.com/deepset-ai/haystack-core-integrations/issues/765
- [ ] https://github.com/deepset-ai/haystack-core-integrations/issues/768
- [ ] https://github.com/deepset-ai/haystack-core-integrations/issues/749
- [ ] https://github.com/deepset-ai/haystack-core-integrations/issues/771
- [ ] https://github.com/deepset-ai/haystack-core-integrations/issues/776
- [ ] https://github.com/deepset-ai/haystack-core-integrations/issues/747
- [ ] https://github.com/deepset-ai/haystack-core-integrations/issues/762
```

PR #1094, opened 2024-09-19
feat: Chroma - allow remote HTTP connection

### LINKED ISSUE #1071: Chroma Remote HttpClient Integration
It'd be great to be able to use the chroma integration with a remote chroma database using the httpclient, rather than having to use an embedded database.

A simple option when creating the chroma component in a pipeline to point at a remote chroma server.

The only other alternatives would potentially be to use a different vector db with an existing integration that could be used remotely.

**Additional context**
Unsure if any additional context, other than to say adding this would give a lot more flexibility when using haystack with chroma.

alperkaya:
### Related Issues

- fixes #1071
### Proposed Changes:

A remote http client is added
 

### How did you test it?

Manual test

### Checklist

### DIFF
--- a/integrations/chroma/src/haystack_integrations/document_stores/chroma/document_store.py
+++ b/integrations/chroma/src/haystack_integrations/document_stores/chroma/document_store.py
@@ -82,6 +82,11 @@ class ChromaDocumentStore:
         self._embedding_function = embedding_function
         self._embedding_function_params = embedding_function_params
         self._distance_function = distance_function
+
+        self._persist_path = persist_path
+        self._host = host
+        self._port = port
+
         # Create the client instance
         if persist_path and (host or port is not None):
             error_message = (
@@ -96,18 +101,12 @@ class ChromaDocumentStore:
                 host=host,
                 port=port,
             )
-            self._host = host
-            self._port = port
-            self._client_type = "http"
         elif persist_path is None:
             # In-memory storage
             self._chroma_client = chromadb.Client()
-            self._client_type = "in_memory"
         else:
             # Local persistent storage
             self._chroma_client = chromadb.PersistentClient(path=persist_path)
-            self._persist_path = persist_path
-            self._client_type = "persistent"
 
         embedding_func = get_embedding_function(embedding_function, **embedding_function_params)
 
@@ -375,9 +374,9 @@ class ChromaDocumentStore:
             self,
             collection_name=self._collection_name,
             embedding_function=self._embedding_function,
-            persist_path=getattr(self, "_persist_path", None),
-            host=getattr(self, "_host", None),
-            port=getattr(self, "_port", None),
+            persist_path=self._persist_path,
+            host=self._host,
+            port=self._port,
             distance_function=self._distance_function,
             **self._embedding_function_params,
         )
--- a/integrations/chroma/src/haystack_integrations/document_stores/chroma/document_store.py
+++ b/integrations/chroma/src/haystack_integrations/document_stores/chroma/document_store.py
@@ -34,6 +34,8 @@ class ChromaDocumentStore:
         collection_name: str = "documents",
         embedding_function: str = "default",
         persist_path: Optional[str] = None,
+        host: Optional[str] = None,  # For remote connection
+        port: Optional[int] = None,  # For remote connection
         distance_function: Literal["l2", "cosine", "ip"] = "l2",
         metadata: Optional[dict] = None,
         **embedding_function_params,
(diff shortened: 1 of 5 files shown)


integrations/chroma/src/haystack_integrations/document_stores/chroma/document_store.py
```diff
@@ -79,9 +83,21 @@ def __init__(
         self._persist_path = persist_path
         self._distance_function = distance_function
         # Create the client instance
-        if persist_path is None:
-            self._chroma_client = chromadb.Client()
+        # Ensure only one of persist_path or host is provided
+        if persist_path and host:
+            error_message = "You cannot provide both `persist_path` and `host`. Please choose one."
```
anakin87, line 88:
```suggestion
        if persist_path and (host or port is not None):
            error_message = "You must specify `persist_path` for local persistent storage or, alternatively, `host` and `port` for remote HTTP client connection. You cannot specify both options."
```

let's check also `port`.
probably the `error_message` should be split over multiple lines.

integrations/chroma/src/haystack_integrations/document_stores/chroma/document_store.py
```diff
@@ -79,9 +83,21 @@ def __init__(
...
+            raise ValueError(error_message)
+        if host:
+            # Remote connection via HTTP client
+            self._chroma_client = chromadb.HttpClient(
+                host=host,
+                port=port or 8000,
+            )
```
anakin87, line 95:
```suggestion
        if host and port is not None:
            # Remote connection via HTTP client
            self._chroma_client = chromadb.HttpClient(
                host=host,
                port=port,
            )
```

Let's be explicit: in case users want to connect to a remote Chroma instance, they should specify both `host` and `port`.

integrations/chroma/src/haystack_integrations/document_stores/chroma/document_store.py
```diff
@@ -79,9 +83,21 @@ def __init__(
...
+        elif persist_path is None:
+            # In-memory storage
+            self._chroma_client = chromadb.Client()  # In-memory client
```
anakin87, line 98:
```suggestion
            self._chroma_client = chromadb.Client()
```
nit: redundant comment

anakin87 (changes requested):
Thanks for the contribution!

I found some small opportunities for improvement.

Also, let's **test the new behavior**.
This can be done as follows:
- add a new step in the [Chroma testing workflow](https://github.com/deepset-ai/haystack-core-integrations/blob/main/.github/workflows/chroma.yml) before `Run tests` to run the DB server as described in [Chroma docs](https://docs.trychroma.com/guides#running-chroma-in-client-server-mode).
Something like this should work: `hatch run chroma run`

- add a new test in https://github.com/deepset-ai/haystack-core-integrations/blob/main/integrations/chroma/tests/test_document_store.py to test the different init options (one of them is the newly introduced remote Chroma).

In case you need help adding the test, LMK and I can help/do it.

Last note: the introduction of this connection option conflicts with #746.
So after this PR is merged, I can take care of adapting the Document Store to defer the DB connection.

alperkaya:
> Thanks for the contribution!
> 
> I found some small opportunities for improvement.
> 
> Also, let's **test the new behavior**. This can be done as follows:
> 
>     * add a new step in the [Chroma testing workflow](https://github.com/deepset-ai/haystack-core-integrations/blob/main/.github/workflows/chroma.yml) before `Run tests` to run the DB server as described in [Chroma docs](https://docs.trychroma.com/guides#running-chroma-in-client-server-mode).
>       Something like this should work: `hatch run chroma run`
> 
>     * add a new test in https://github.com/deepset-ai/haystack-core-integrations/blob/main/integrations/chroma/tests/test_document_store.py to test the different init options (one of them is the newly introduced remote Chroma).
> 
> 
> In case you need help adding the test, LMK and I can help/do it.
> 
> Last note: the introduction of this connection option conflicts with #746. So after this PR is merged, I can take care of adapting the Document Store to defer the DB connection.

Hi,

Thanks for feedback!

I made a slight change in __init__() to make it more testable, added some test to check different init configs, and updated to to_dict() to support http client as well.

anakin87:
I will have a look later...

In the meantime, to simplify things, I think we can test the HTTP client on Linux only (and skip the corresponding test on Windows).

anakin87 (approved):
I have taken the liberty of making some simplifications.
Thanks for the contribution!

I will merge this PR but I won't immediately release a new version of this package.

As I said before, I want to take care to make this Document Store consistent with others, by deferring the DB connection (see #746). I will do this in another PR.

PR #1107, opened 2024-09-25
feat: Chroma - defer the DB connection

### LINKED ISSUE #1101: Chroma - defer the DB connection
In #1094, we introduced the possibility of connecting via HTTP to a Chroma server.

Previously, Chroma could only be run in-memory, within the same process as the Python script utilizing it.

This new option conflicts with the work done in #746 to avoid database connections in document stores constructors.

**We should refactor Chroma as we did for other Document Stores and only then release a new version.**

alperkaya:
### Related Issues

- fixes #1101

### How did you test it?

run unit & integration tests

### Checklist


integrations/chroma/src/haystack_integrations/document_stores/chroma/document_store.py
```diff
@@ -79,61 +79,76 @@ def __init__(
...
+        self._metadata = metadata
+        self._collection = None
 
         self._persist_path = persist_path
         self._host = host
         self._port = port
+        self._chroma_client = None
+
+    @property
+    def chroma_client(self):
```
shadeMe, line 94: Given the incoming support for async clients in document stores, this pattern of lazy-initializing is no longer suitable. Please use [this pattern](https://github.com/deepset-ai/haystack-experimental/blob/90d5966edfe88a6801d7d8efba121d802af73188/haystack_experimental/document_stores/opensearch/document_store.py#L160) instead (combine both the client and the collection's initialization in it) and ensure that it's called in every public method.

shadeMe (changes requested):

integrations/chroma/src/haystack_integrations/document_stores/chroma/document_store.py
alperkaya: I used suggested pattern but because of lint errors, I had to raise valueError for every public method which looks ugly :/

integrations/chroma/src/haystack_integrations/document_stores/chroma/document_store.py
```diff
@@ -90,7 +90,14 @@ def __init__(
...
         self._chroma_client = None
 
-    @property
+        self._initialized = False
+
+    def _ensure_initialized(self):
+        if not self._initialized:
+            self._chroma_client = self.chroma_client()
+            self._collection = self.collection()
+            self._initialized = True
```
shadeMe, line 99: Let's inline these two function calls and remove the individual methods - I don't think they are used anywhere else.

integrations/chroma/src/haystack_integrations/document_stores/chroma/document_store.py
```diff
@@ -148,7 +154,11 @@ def count_documents(self) -> int:
 
         :returns: how many documents are present in the document store.
         """
-        return self.collection.count()
+        self._ensure_initialized()
+        if self._collection is None:
+            msg = "Collection is not initialized"
+            raise ValueError(msg)
```
shadeMe, line 160: Both the collection and client variables should be initialized after the call to `_ensure_initialized` , so there's no reason to raise. We can replace that with `assert self._collection is not None` instead (same with the other variable wherever it needs to be directly accessed).

shadeMe (changes requested):

integrations/chroma/src/haystack_integrations/document_stores/chroma/document_store.py
shadeMe: See below.

alperkaya:
I made the changes as you pointed out. LMK your comments.

shadeMe (approved):
@anakin87 I'll leave the final approval to you.

alperkaya:
> @anakin87 I'll leave the final approval to you (once the tests are fixed).

I've been looking into the test failures, but I'm having trouble interpreting them, as they seem to be open to different interpretations. I’d appreciate it if you could help clarify what might be causing the issue or what direction I should take to resolve this.

anakin87:
I will take a look...

anakin87 (approved):
After fixing some small things, I'm going to merge this PR.
Thanks!

PR #1115, opened 2024-09-30
chore: unpin `llama-cpp-python`

anakin87:
### Related Issues

- Starting from the release of `llama-cpp-python==0.3.0`, [integrations tests were running for > 4 hours](https://github.com/deepset-ai/haystack-core-integrations/actions/runs/11043172578/job/30676976129).
- In #1111, I pinned the package to a previous version.
- Since the issues were not solved with the release of `llama-cpp-python==0.3.1`, I spent some time debugging the problem.

### Proposed Changes:
- make the hanging test run in a reasonable amount of time
  - force the choice of the tool (as shown in the [docs](https://llama-cpp-python.readthedocs.io/en/latest/#function-calling))
  - also set `n_ctx`  to 512 (default in `llama-cpp-python`, while the default in our integration is 0 = inferred from the model)
- unpin `llama-cpp-python`

### How did you test it?
CI

### Notes for the reviewer
I think in the future we will rework function calling with `llama-cpp-python` and remove the test with this model (`functionary-small-v2.4`), which is becoming obsolete.
In the meantime, I prefer to try to be compatible with the latest versions of this library.

### Checklist


julian-risch (approved):
LGTM! 👍

Issue #905, opened 2024-07-16
Support the FastEmbed GPU implementation


aymbot:
**Is your feature request related to a problem? Please describe.**
Currently the **Fastembed....** embedders and retrievers are not utilizing the _GPUs_ which makes it so that i.e. _SPLADE_ takes a substantial amount of time for embedding vs. it's counterparts. 

**Describe the solution you'd like**
_QDrant_ supports _GPUs_ with another library, see [here](https://github.com/qdrant/fastembed?tab=readme-ov-file#%EF%B8%8F-fastembed-on-a-gpu). Utilizing that library would allow us to leverage our _GPUs_. _GPU_-mode could be enabled with a flag or another method.

**Describe alternatives you've considered**
Besides the *Fastembed..* embedders, there are no out-of-the-box, nor integration, alternatives for sparse embeddings, meaning the only alternative would be to not use _GPUs_.

**Additional context**
For now, none.

PR #1119, opened 2024-10-01
feat: Add option to select ONNXRuntime Execution Providers for fastembed

Fannovel16:
### Related Issues

- In some cases, fastembed does not use GPU despite `onnxruntime-gpu` and `fastembed-gpu` are installed correctly

### Proposed Changes:

 - Add option `onnx_providers` to set ONNXRuntime Execution Providers manually.

### How did you test it?

- WIP

### Checklist

<!-- - I added unit tests and updated the docstrings

### DIFF
--- a/integrations/fastembed/src/haystack_integrations/components/embedders/fastembed/embedding_backend/fastembed_backend.py
+++ b/integrations/fastembed/src/haystack_integrations/components/embedders/fastembed/embedding_backend/fastembed_backend.py
@@ -5,6 +5,7 @@ from tqdm import tqdm
 
 from fastembed import TextEmbedding
 from fastembed.sparse.sparse_text_embedding import SparseTextEmbedding
+from fastembed.common.types import OnnxProvider
 
 
 class _FastembedEmbeddingBackendFactory:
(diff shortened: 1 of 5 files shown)


anakin87:
Just a quick thought: I don't think this would work, because this integration explicitly depends on `fastembed`.
GPU support requires `fastembed-gpu`, which cannot be installed in the same environment as `fastembed` ([FastEmbed docs](https://qdrant.github.io/fastembed/examples/FastEmbed_GPU/)).

Related: https://github.com/deepset-ai/haystack-core-integrations/issues/905

Fannovel16:
> Just a quick thought: I don't think this would work, because this integration explicitly depends on `fastembed`. GPU support requires `fastembed-gpu`, which cannot be installed in the same environment as `fastembed` ([FastEmbed docs](https://qdrant.github.io/fastembed/examples/FastEmbed_GPU/)).
> 
> Maybe I'm wrong, but I think the whole integration needs to be reworked to support GPU. There may be several ways to do that...
> 
> Related: #905

@anakin87 On Google Colab, I installed fastembed-gpu and correct onnxruntime-gpu but it does not pick up GPU by default, even if "CUDAExecutionProvider" is available. Explicily stating providers fix that

PR #1122, opened 2024-10-02
feat: Add Azure AI Search integration

### LINKED ISSUE #950: Integration for Azure AI Search
## Summary and motivation

Motivation from this issue [#6749](https://github.com/deepset-ai/haystack/issues/6749) raised by community members 
## Detailed design

Create a new document store for Azure AI search allowing operations of vector search, hybrid search and keyword search.

## Implementation Breakdown

- [x] Create a document store with CURD operations.
- [x]  Tests for document stores.
- [ ]  Add retrievers.
- [ ] Tests for retrievers.
- [x] Add filters and search operations.
- [ ] Test for filters.

## Checklist

```[tasklist]
### Tasks
- [ ] A label named like `integration:<your integration name>` has been added to this repo
- [ ] There is an example available to demonstrate the feature
```

Amnah199:
### Related Issues

- fixes #950 

### Proposed Changes:

This PR includes the following features:

- Implement a document store for [Azure AI Search.](https://learn.microsoft.com/en-us/azure/search/search-how-to-create-search-index?tabs=portal).
- Implement an embedding retriever
- Implement filters. Note that Azure Search supports additional functionalities such as `any` and `all`. The current filter implementation includes [metadata filtering](https://docs.haystack.deepset.ai/docs/metadata-filtering) using operators in Haystack filters and content filtering using `search` function in Azure. This is how its done in other integrations. More filters can be added later.

### How did you test it?

- Added unit tests for components
- Used `haystack.testing.document_store` for testing filters and CRUD operations
- Manual testing by running examples
- 
### Notes for the reviewer
Config files need to be fixed. Also the `ReadMe.md `and log files will be added.

### Checklist


integrations/azure_ai_search/src/haystack_integrations/document_stores/azure_ai_search/document_store.py
```diff
@@ -0,0 +1,378 @@
...
+
+            embedding = azure_doc.get("embedding")
+            if embedding == self._dummy_vector:
+                embedding = None
+
+            # Filter out meta fields
+            meta = {
+                key: value
+                for key, value in azure_doc.items()
+                if key not in ["id", "content", "embedding"] and not key.startswith("@")
```
Amnah199, line 326: @ttmenezes Could you provide some input on this?

The goal is to retrieve all relevant document fields from the index—such as `id`, `content`, `embedding`, and other metadata—while avoiding internal fields like `@search_score`. I haven't found a function that only loads the relevant document fields. While the current implementation seems to work, it might have issues.

Do you have any suggestions for a better solution?

ttmenezes: Asking our API team if there is a better way to do this, but manually filtering the list as such might be the best option to get this result.

ttmenezes, line 326: Broadly, we need to make sure that the integration supports indexes that do not have explicit "id" or "content" fields.

On the AI Search side, we support arbitrary naming of fields, and the only requirement is that at least one of the fields is designated as the "key" field. In indexes that we assist the user in creating, the key field is "id" and "content" is another field, but the user can definitely configure their index with a different naming scheme, so we'll need to support that scenario as well.

integrations/azure_ai_search/src/haystack_integrations/components/azure_ai_search/retrievers/embedding_retriever.py
```diff
@@ -0,0 +1,105 @@
...
+        :returns: List of Document that are most similar to `query_embedding`
+        """
+
+        if not query_embedding:
+            msg = "query_embedding must be a non-empty list of floats"
+            raise ValueError(msg)
+
+        # embedding = get_embeddings(input=query, model=embedding_model_name, dimensions=self._embedding_dimension)
+
+        vector_query = VectorizedQuery(vector=query_embedding, k_nearest_neighbors=3, fields="embeddings")
```
ttmenezes, line 101: Should the `k_nearest_neighbors` field be set to the `top_k` function parameter?

Amnah199: Good catch! Will update this.

integrations/azure_ai_search/src/haystack_integrations/document_stores/azure_ai_search/document_store.py
Amnah199: I understand your reasoning here, but in our [DocumentStores](https://docs.haystack.deepset.ai/docs/document-store#work-with-documents), we specifically store Haystack [Documents](https://docs.haystack.deepset.ai/docs/data-classes#document). Which means normally users convert their data to `Document`, which always includes an `id` and `content`.  Based on this information, I think its okay to keep `id`, `content` and `embedding` as default fields.

ttmenezes: Ok I think that's a fair decision to make if it is expected that users will be creating their indexes and upserting documents via Haystack. The potential for a mismatch between existing AI Search indexes' columns and the expected columns here will be something to keep in mind when users begin using this integration.

I imagine that the same assumption is made for integrations with other providers too, so it seems fine to move forward.

integrations/azure_ai_search/src/haystack_integrations/document_stores/azure_ai_search/document_store.py
```diff
@@ -0,0 +1,414 @@
...
+        try:
+            if not self._index_client:
+                self._index_client = SearchIndexClient(self._azure_endpoint, credential, **self._kwargs)
+            if not self.index_exists(self._index_name):
+                # Create a new index if it does not exist
+                logger.debug(
+                    "The index '%s' does not exist. A new index will be created.",
+                    self._index_name,
+                )
+                self.create_index(self._index_name)
```
ttmenezes, line 130: I noticed that the `create_index` function signature allows kwargs to be passed in. Should they be passed in here? I was thinking about the case where the user wants to enable semantic re-ranking/search in their index, which can be toggled in the `SearchIndex()` call below.

integrations/azure_ai_search/src/haystack_integrations/document_stores/azure_ai_search/document_store.py
```diff
@@ -0,0 +1,414 @@
...
+
+        :raises ValueError: If `query_embedding` is an empty list
+        :returns: List of Document that are most similar to `query_embedding`
+        """
+
+        if not query_embedding:
+            msg = "query_embedding must be a non-empty list of floats"
+            raise ValueError(msg)
+
+        vector_query = VectorizedQuery(vector=query_embedding, k_nearest_neighbors=3, fields="embedding")
```
ttmenezes, line 411: During testing, I noticed that this value overrides the `top_k` parameter that is passed in on the next line, so the line should be:
```
vector_query = VectorizedQuery(vector=query_embedding, k_nearest_neighbors=top_k, fields="embedding")
```

integrations/azure_ai_search/src/haystack_integrations/document_stores/azure_ai_search/document_store.py
ttmenezes: Spoke with my team, and we want to prioritize allowing users to toggle on semantic re-ranking. This is a setting should be set at index creation time, so adding a toggle for this setting in the `AzureAISearchDocumentStore` seems like the way forward. At retrieval time, we can allow the user to toggle semantic ranking, which will only work if their index was initialized with it. It is also worth noting that on the free tier of the semantic ranker, the user has a quota of 1,000 semantic requests per month, and then they can upgrade beyond that.

Thoughts on implementing support for this?

integrations/azure_ai_search/src/haystack_integrations/document_stores/azure_ai_search/document_store.py
Amnah199: Yes, the same assumption applies to all integrations. If anything is out of order, it will be addressed when other engineers on our team review the PR.

integrations/azure_ai_search/src/haystack_integrations/document_stores/azure_ai_search/document_store.py
Amnah199: That's a great suggestion. I believe we can add support for semantic re-ranking during index creation. We usually allow passing `kwargs` specific to the database, but I'll need to check the documentation to understand how this will affect index creation and the remaining methods. Regardless, I'll add support for semantic re-ranking after referring to API.

integrations/azure_ai_search/src/haystack_integrations/document_stores/azure_ai_search/document_store.py
Amnah199: This is added in the next commit. 🙏

integrations/azure_ai_search/src/haystack_integrations/document_stores/azure_ai_search/filters.py
```diff
@@ -0,0 +1,104 @@
+from typing import Any, Dict
+
+from dateutil import parser
+
+from .errors import AzureAISearchDocumentStoreFilterError
+
+LOGICAL_OPERATORS = {"AND": "and", "OR": "or", "NOT": "not"}
```
vblagoje, line 7: Don't we have enum for these somewhere, please search

integrations/azure_ai_search/src/haystack_integrations/document_stores/azure_ai_search/document_store.py
```diff
@@ -0,0 +1,445 @@
...
+        """
+        return default_to_dict(
+            self,
+            azure_endpoint=self._azure_endpoint.to_dict() if self._azure_endpoint is not None else None,
+            api_key=self._api_key.to_dict() if self._api_key is not None else None,
+            index_name=self._index_name,
+            create_index=self._create_index,
+            embedding_dimension=self._embedding_dimension,
+            metadata_fields=self._metadata_fields,
+            vector_search_configuration=self._vector_search_configuration,
```
vblagoje, line 205: Is this non primitive VectorSearch serializable?

vblagoje (commented):
Left some rough first pass feedback, will revisit

integrations/azure_ai_search/example/document_store.py
```diff
@@ -0,0 +1,37 @@
+from haystack import Document
```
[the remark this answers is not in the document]
Amnah199: Done. Added the instructions.

integrations/azure_ai_search/example/embedding_retrieval.py
```diff
@@ -0,0 +1,37 @@
...
+        content="""Elephants have been observed to behave in a way that indicates a
+         high level of self-awareness, such as recognizing themselves in mirrors."""
+    ),
+    Document(
+        content="""In certain parts of the world, like the Maldives, Puerto Rico, and
+          San Diego, you can witness the phenomenon of bioluminescent waves."""
+    ),
+]
+
+document_embedder = SentenceTransformersDocumentEmbedder(model=model)
```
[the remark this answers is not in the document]
Amnah199: If you prefer it that way, I'll update the example.

integrations/azure_ai_search/src/haystack_integrations/document_stores/azure_ai_search/filters.py
Amnah199: I found them [here](https://github.com/deepset-ai/haystack/blob/78292422f00592bb0a6b5d58bbbb679f4b8718da/haystack/utils/filters.py#L48) in haystack.utils but thats not what we need. Plus other integrations also define these operators based on requirements.

integrations/azure_ai_search/src/haystack_integrations/document_stores/azure_ai_search/document_store.py
Amnah199: Decided to introduce this change as a separate PR to avoid extending the current PR further.

integrations/azure_ai_search/.gitignore
```diff
@@ -0,0 +1,163 @@
+# Byte-compiled / optimized / DLL files
```
anakin87, line 1: I would say that this file is not needed: we already have [this](https://github.com/deepset-ai/haystack-core-integrations/blob/e97cfee4c2b0868d96e4a4a335ec5cd7ee0f6cf1/.gitignore) at the root level

integrations/azure_ai_search/pyproject.toml
```diff
@@ -0,0 +1,164 @@
...
+  "Development Status :: 4 - Beta",
+  "Programming Language :: Python",
+  "Programming Language :: Python :: 3.8",
+  "Programming Language :: Python :: 3.9",
+  "Programming Language :: Python :: 3.10",
+  "Programming Language :: Python :: 3.11",
+  "Programming Language :: Python :: Implementation :: CPython",
+  "Programming Language :: Python :: Implementation :: PyPy",
+]
+dependencies = ["haystack-ai", "azure-search-documents>=11.5", "azure-identity", "torch>=1.11.0"]
```
anakin87, line 25: Why is `torch` needed?

integrations/azure_ai_search/pyproject.toml
```diff
@@ -0,0 +1,164 @@
+[build-system]
+requires = ["hatchling", "hatch-vcs"]
+build-backend = "hatchling.build"
+
+[project]
+name = "azure-ai-search-haystack"
+dynamic = ["version"]
+description = 'Haystack 2.x Document Store for Azure AI Search'
+readme = "README.md"
+requires-python = ">=3.8,<3.13"
```
anakin87, line 10: 3.8 already reached End Of Life, so for new integrations, I would drop it here and in test matrices

integrations/azure_ai_search/src/haystack_integrations/components/retrievers/azure_ai_search/embedding_retriever.py
```diff
@@ -0,0 +1,125 @@
...
+    """
+
+    def __init__(
+        self,
+        *,
+        document_store: AzureAISearchDocumentStore,
+        filters: Optional[Dict[str, Any]] = None,
+        top_k: int = 10,
+        filter_policy: Union[str, FilterPolicy] = FilterPolicy.REPLACE,
+        raise_on_failure: bool = True,
```
anakin87, line 28: Unless there is a strong reason for this, I would avoid this parameter: `raise_on_failure`
If we decide to keep it, let's document it in the docstring below.

integrations/azure_ai_search/src/haystack_integrations/document_stores/azure_ai_search/document_store.py
```diff
@@ -0,0 +1,450 @@
...
+    def __init__(
+        self,
+        *,
+        api_key: Secret = Secret.from_env_var("AZURE_SEARCH_API_KEY", strict=False),  # noqa: B008
+        azure_endpoint: Secret = Secret.from_env_var("AZURE_SEARCH_SERVICE_ENDPOINT", strict=False),  # noqa: B008
+        index_name: str = "default",
+        embedding_dimension: int = 768,
+        metadata_fields: Optional[Dict[str, type]] = None,
+        vector_search_configuration: VectorSearch = None,
+        create_index: bool = True,
```
anakin87, line 76: `create_index` is not documented in the docstring. Maybe it's not used and should be removed...

integrations/azure_ai_search/src/haystack_integrations/document_stores/azure_ai_search/document_store.py
```diff
@@ -0,0 +1,450 @@
...
+            index_fields = self._index_client.get_index(self._index_name).fields
+            self._index_fields = [field.name for field in index_fields]
+            self._client = self._index_client.get_search_client(self._index_name)
+        else:
+            msg = "Search Index Client is not initialized."
+            raise AzureAISearchDocumentStoreConfigError(msg)
+
+        return self._client
+
+    def create_index(self, index_name: str, **kwargs) -> None:
```
anakin87, line 158:
```suggestion
    def _create_index(self, index_name: str, **kwargs) -> None:
```
This seems an internal method, no?

integrations/azure_ai_search/src/haystack_integrations/document_stores/azure_ai_search/document_store.py
```diff
@@ -0,0 +1,450 @@
...
+        :returns:
+            Deserialized component.
+        """
+
+        deserialize_secrets_inplace(data["init_parameters"], keys=["api_key", "azure_endpoint"])
+        if (vector_search_configuration := data["init_parameters"].get("vector_search_configuration")) is not None:
+            data["init_parameters"]["vector_search_configuration"] = VectorSearch.from_dict(vector_search_configuration)
+        return default_from_dict(cls, data)
+
+    def count_documents(self, **kwargs: Any) -> int:
```
anakin87, line 228: The DocumentStore protocol is not respected due to the presence of kwargs
See https://github.com/deepset-ai/haystack/blob/13804293b1bb79743e5a30e980b76a0561dcfaf8/haystack/document_stores/types/protocol.py#L37

integrations/azure_ai_search/src/haystack_integrations/document_stores/azure_ai_search/document_store.py
```diff
@@ -0,0 +1,450 @@
...
+
+            # Create the document with meta only if it's non-empty
+            doc = Document(
+                id=azure_doc["id"], content=azure_doc["content"], embedding=embedding, meta=meta if meta else {}
+            )
+
+            documents.append(doc)
+        return documents
+
+    def index_exists(self, index_name: Optional[str]) -> bool:
```
anakin87, line 348: internal method, no?

integrations/azure_ai_search/src/haystack_integrations/document_stores/azure_ai_search/document_store.py
```diff
@@ -0,0 +1,450 @@
...
+        if self.count_documents == 0:
+            return
+        documents = self._get_raw_documents_by_id(document_ids)
+        if documents:
+            self.client.delete_documents(documents)
+
+    def get_documents_by_id(self, document_ids: List[str]) -> List[Document]:
+        return self._convert_search_result_to_documents(self._get_raw_documents_by_id(document_ids))
+
+    def search_documents(self, search_text: Optional[str] = "*", top_k: Optional[int] = 10) -> List[Document]:
```
anakin87, line 295: We generally don't expose such methods directly but through a retriever. Is that different?

`top_k: Optional[int] = 10` -> `top_k: int = 10`
same for `search_text`

integrations/azure_ai_search/src/haystack_integrations/document_stores/azure_ai_search/document_store.py
```diff
@@ -0,0 +1,450 @@
...
+
+    def _map_metadata_field_types(self, metadata: Dict[str, type]) -> Dict[str, str]:
+        """Map metadata field types to Azure Search field types."""
+
+        metadata_field_mapping = {}
+
+        for key, value_type in metadata.items():
+
+            # Azure Search index only allows field names starting with letters
+            field_name = next((key[i:] for i, char in enumerate(key) if char.isalpha()), key)
```
anakin87, line 407: I would probably raise an error...

integrations/azure_ai_search/tests/test_document_store.py
```diff
@@ -0,0 +1,365 @@
...
+        it should override this method. This can happen for example when the Document Store sets
+        a score to returned Documents. Since we can't know what the score will be, we can't compare
+        the Documents reliably.
+        """
+        sorted_recieved = sorted(received, key=lambda doc: doc.id)
+        sorted_expected = sorted(expected, key=lambda doc: doc.id)
+        assert sorted_recieved == sorted_expected
+
+    # Dataframes are not supported in serach index
+    def test_comparison_equal_with_dataframe(self, document_store, filterable_docs):
```
anakin87, line 228: Let's use an explicit skip mark like this
https://github.com/deepset-ai/haystack-core-integrations/blob/e97cfee4c2b0868d96e4a4a335ec5cd7ee0f6cf1/integrations/pinecone/tests/test_filters.py#L41

integrations/azure_ai_search/tests/test_document_store.py
```diff
@@ -0,0 +1,365 @@
...
+            [
+                d
+                for d in filterable_docs
+                if d.meta.get("date") is not None
+                and datetime.strptime(d.meta["date"], "%Y-%m-%dT%H:%M:%SZ").replace(tzinfo=timezone.utc)
+                > datetime.strptime("1972-12-11T19:54:58Z", "%Y-%m-%dT%H:%M:%SZ").replace(tzinfo=timezone.utc)
+            ],
+        )
+
+    def test_comparison_greater_than_equal_with_iso_date(self, document_store, filterable_docs):
```
anakin87, line 264: Why do we need to override the original test?

anakin87 (commented):
I felt free to take a look and noticed some possible improvements.

integrations/azure_ai_search/src/haystack_integrations/document_stores/azure_ai_search/document_store.py
Amnah199: For this one, I was wondering if it would be helpful to expose it? Can serve some purpose for user?

integrations/azure_ai_search/src/haystack_integrations/document_stores/azure_ai_search/document_store.py
Amnah199: Found a similar method in Astra, so I thought we can keep it.
https://github.com/deepset-ai/haystack-core-integrations/blob/e97cfee4c2b0868d96e4a4a335ec5cd7ee0f6cf1/integrations/astra/src/haystack_integrations/document_stores/astra/document_store.py#L354

integrations/azure_ai_search/tests/test_document_store.py
Amnah199: Mentioned the reason [above](https://github.com/deepset-ai/haystack-core-integrations/blob/23d1e2298ab620b421e063ec359f71ea6723b365/integrations/azure_ai_search/tests/test_document_store.py#L246). Although I think I could explain it better by specifically mentioning **ISO 8601 format**.

integrations/azure_ai_search/pyproject.toml
Amnah199: I think I added this while trying to fix lib support in python- 3.13. Will remove.

integrations/azure_ai_search/src/haystack_integrations/document_stores/azure_ai_search/document_store.py
anakin87: Sorry I was talking fof `search_documents`

integrations/azure_ai_search/src/haystack_integrations/document_stores/azure_ai_search/document_store.py
anakin87: I would avoid mentioning index, apart from initialization.
Some explanation here: https://github.com/deepset-ai/haystack-core-integrations/pull/1160#issue-2620805781

ttmenezes (approved):
Core Azure AI Search functionality looks good, and I have tested it locally after pulling the code. Semantic reranking, hybrid retrieval, and BM25 retrieval will be added in future PRs.

integrations/azure_ai_search/src/haystack_integrations/document_stores/azure_ai_search/document_store.py
Amnah199: Thanks for explaining. Agreed 👍

integrations/azure_ai_search/src/haystack_integrations/document_stores/azure_ai_search/document_store.py
Amnah199: Related issue discussed [here](https://deepset-ai.slack.com/archives/C06FBMD8LMV/p1730376446186879). For now will just an error.

anakin87: In general, I would not rename fields.
In case some names are not supported, let's emit a clear warning and drop them.

integrations/azure_ai_search/src/haystack_integrations/document_stores/azure_ai_search/document_store.py
Amnah199: In that case, would you prefer to have `search_documents` as an internal method used for content filtering. This can be called by user through `filter_documents` using `content` key?

integrations/azure_ai_search/src/haystack_integrations/document_stores/azure_ai_search/document_store.py
```diff
@@ -0,0 +1,451 @@
...
+            self.client.upload_documents(documents_to_write)
+        return len(documents_to_write)
+
+    def delete_documents(self, document_ids: List[str]) -> None:
+        """
+        Deletes all documents with a matching document_ids from the search index.
+
+        :param document_ids: ids of the documents to be deleted.
+        """
+        if self.count_documents == 0:
```
vblagoje, line 282: We need to invoke count_documents method with parentheses here. This condition is never true.

integrations/azure_ai_search/tests/test_document_store.py
```diff
@@ -0,0 +1,359 @@
...
+    assert isinstance(document_store._azure_endpoint, EnvVarSecret)
+    assert document_store._index_name == "default"
+    assert document_store._embedding_dimension == 768
+    assert document_store._metadata_fields is None
+    assert document_store._vector_search_configuration == DEFAULT_VECTOR_SEARCH
+
+
+@patch("haystack_integrations.document_stores.azure_ai_search.document_store.AzureAISearchDocumentStore")
+def test_init_is_lazy(_mock_azure_search_client):
+    AzureAISearchDocumentStore(azure_endppoint=Secret.from_token("test_endpoint"))
```
vblagoje, line 81: typo in azure_endppoint

vblagoje (changes requested):
Cursor review found two more issues to address.

integrations/azure_ai_search/src/haystack_integrations/document_stores/azure_ai_search/document_store.py
```diff
@@ -0,0 +1,451 @@
...
+    ):
+        """
+        A document store using [Azure AI Search](https://azure.microsoft.com/products/ai-services/ai-search/)
+        as the backend.
+
+        :param azure_endpoint: The URL endpoint of an Azure AI Search service.
+        :param api_key: The API key to use for authentication.
+        :param index_name: Name of index in Azure AI Search, if it doesn't exist it will be created.
+        :param embedding_dimension: Dimension of the embeddings.
+        :param metadata_fields: A dictionary of metatada keys and their types to create
```
anakin87, line 86:
```suggestion
        :param metadata_fields: A dictionary of metadata keys and their types to create
```

integrations/azure_ai_search/src/haystack_integrations/document_stores/azure_ai_search/document_store.py
```diff
@@ -0,0 +1,451 @@
...
+            additional fields in index schema. As fields in Azure SearchIndex cannot be dynamic,
+            it is necessary to specify the metadata fields in advance.
```
anakin87, line 88: As an user, it's difficult for me to understand how these metadata fields should be provided. Can ypu add a little example?

integrations/azure_ai_search/src/haystack_integrations/document_stores/azure_ai_search/document_store.py
```diff
@@ -0,0 +1,451 @@
...
+        azure_endpoint = azure_endpoint or os.environ.get("AZURE_SEARCH_SERVICE_ENDPOINT") or None
+        if not azure_endpoint:
+            msg = "Please provide an Azure endpoint or set the environment variable AZURE_OPENAI_ENDPOINT."
+            raise ValueError(msg)
+
+        api_key = api_key or os.environ.get("AZURE_SEARCH_API_KEY") or None
+
+        self._client = None
+        self._index_client = None
+        self._index_fields = []  # type: List[Any]  # stores all fields in the final schema of index
```
anakin87, line 111: why is type in comments?

Amnah199: Resolved

integrations/azure_ai_search/src/haystack_integrations/document_stores/azure_ai_search/document_store.py
```diff
@@ -0,0 +1,451 @@
...
+
+type_mapping = {
+    str: "Edm.String",
+    bool: "Edm.Boolean",
+    int: "Edm.Int32",
+    float: "Edm.Double",
+    datetime: "Edm.DateTimeOffset",
+}
+
+MAX_UPLOAD_BATCH_SIZE = 1000
```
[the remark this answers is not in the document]
Amnah199: Didnt implement batch processing. Will remove it for now.

vblagoje (approved):

PR #1120, opened 2024-10-01
modify regex to allow cross-region inference in bedrock 

1greentangerine:
more info here: https://aws.amazon.com/de/blogs/machine-learning/getting-started-with-cross-region-inference-in-amazon-bedrock/

for now this feature is only available for anthropic claude models

### Related Issues

- fixes #issue-number

### Checklist

- I have read the [contributors guidelines](https://github.com/deepset-ai/haystack-core-integrations/blob/main/CONTRIBUTING.md) and the [code of conduct](https://github.com/deepset-ai/haystack-core-integrations/blob/main/CODE_OF_CONDUCT.md)
- I have updated the related issue with new insights and changes
- I added unit tests and updated the docstrings
- I've used one of the [conventional commit types](https://www.conventionalcommits.org/en/v1.0.0/) for my PR title: `fix:`, `feat:`, `build:`, `chore:`, `ci:`, `docs:`, `style:`, `refactor:`, `perf:`, `test:`.

PR #1127, opened 2024-10-07
feat: add prefixes to supported model patterns to allow cross region …

abrahamy:
…model ids

### Related Issues

- fixes #issue-number

### Checklist

### DIFF
--- a/integrations/amazon_bedrock/src/haystack_integrations/components/generators/amazon_bedrock/generator.py
+++ b/integrations/amazon_bedrock/src/haystack_integrations/components/generators/amazon_bedrock/generator.py
@@ -65,13 +65,13 @@ class AmazonBedrockGenerator:
     """
 
     SUPPORTED_MODEL_PATTERNS: ClassVar[Dict[str, Type[BedrockModelAdapter]]] = {
-        r"amazon.titan-text.*": AmazonTitanAdapter,
-        r"ai21.j2.*": AI21LabsJurassic2Adapter,
-        r"cohere.command-[^r].*": CohereCommandAdapter,
-        r"cohere.command-r.*": CohereCommandRAdapter,
-        r"(.+\.)?anthropic.claude.*": AnthropicClaudeAdapter,
-        r"meta.llama.*": MetaLlamaAdapter,
-        r"mistral.*": MistralAdapter,
+        r"(eu\.|us\.)?amazon.titan-text.*": AmazonTitanAdapter,
+        r"(eu\.|us\.)?ai21.j2.*": AI21LabsJurassic2Adapter,
+        r"(eu\.|us\.)?cohere.command-[^r].*": CohereCommandAdapter,
+        r"(eu\.|us\.)?cohere.command-r.*": CohereCommandRAdapter,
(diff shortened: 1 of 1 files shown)


anakin87:
This recent contribution seems related: https://github.com/deepset-ai/haystack-core-integrations/pull/1120

integrations/amazon_bedrock/src/haystack_integrations/components/generators/amazon_bedrock/generator.py
```diff
@@ -65,13 +65,13 @@ class AmazonBedrockGenerator:
...
 
     SUPPORTED_MODEL_PATTERNS: ClassVar[Dict[str, Type[BedrockModelAdapter]]] = {
-        r"amazon.titan-text.*": AmazonTitanAdapter,
-        r"ai21.j2.*": AI21LabsJurassic2Adapter,
-        r"cohere.command-[^r].*": CohereCommandAdapter,
-        r"cohere.command-r.*": CohereCommandRAdapter,
-        r"(.+\.)?anthropic.claude.*": AnthropicClaudeAdapter,
-        r"meta.llama.*": MetaLlamaAdapter,
-        r"mistral.*": MistralAdapter,
+        r"(eu\.|us\.)?amazon.titan-text.*": AmazonTitanAdapter,
```
shadeMe, line 68: Let's use the reg-ex seen [here](https://github.com/deepset-ai/haystack-core-integrations/pull/1120) - This will also handle other regions that might get supported in the future.

If we know that AWS uses the [ISO country code](https://en.wikipedia.org/wiki/List_of_ISO_3166_country_codes), we should also restrict the length.

integrations/amazon_bedrock/src/haystack_integrations/components/generators/amazon_bedrock/generator.py
```diff
@@ -65,13 +65,13 @@ class AmazonBedrockGenerator:
...
+        r"(eu\.|us\.)?ai21.j2.*": AI21LabsJurassic2Adapter,
+        r"(eu\.|us\.)?cohere.command-[^r].*": CohereCommandAdapter,
+        r"(eu\.|us\.)?cohere.command-r.*": CohereCommandRAdapter,
+        r"(eu\.|us\.)?(.+\.)?anthropic.claude.*": AnthropicClaudeAdapter,
```
shadeMe, line 72: This addition seems superfluous?

shadeMe (changes requested):

integrations/amazon_bedrock/src/haystack_integrations/components/generators/amazon_bedrock/generator.py
abrahamy: Regex updated to the simple `(.+\.)?`.

integrations/amazon_bedrock/src/haystack_integrations/components/generators/amazon_bedrock/generator.py
abrahamy: Updated the regex to support ISO Alpha-2 country code prefixes like so: `([a-z]{2}\.)?`.

shadeMe (approved):

PR #1131, opened 2024-10-10
feat: add `keep_alive` parameter to Ollama Generators

### LINKED ISSUE #1129: Support full parameter set for Ollama integrations

Ollama integrations do not support the full capabilities of the generation methods. Namely `OllamaGenerator.generate` and `OllamaChatGenerator.chat` methods. I was trying to add the keep_alive option to the generation using the OllamaGenerator. But when I inspected the [code](https://github.com/deepset-ai/haystack-core-integrations/blob/9438634f2082bc8c7ebd518f319512ed4fbe9882/integrations/ollama/src/haystack_integrations/components/generators/ollama/generator.py#L175), I've seen that the method only passed down the `model`, `prompt`, `stream` and `options` parameters, even though the Ollama client is capable of utilizing more:

```python
@overload
  def generate(
    self,
    model: str = '',
    prompt: str = '',
    suffix: str = '',
    system: str = '',
    template: str = '',
    context: Optional[Sequence[int]] = None,
    stream: Literal[True] = True,
    raw: bool = False,
    format: Literal['', 'json'] = '',
    images: Optional[Sequence[AnyStr]] = None,
    options: Optional[Options] = None,
    keep_alive: Optional[Union[float, str]] = None,
  ) -> Iterator[Mapping[str, Any]]: ...
```

Adding all the top level parameters that ollama client is able to use to the `OllamaGenerator` and `OllamaChatGenerator` constructors respectively.

Using kwargs to pass down the parameters manually:

```python
class OllamaGenerator:
    def __init__(
        ...
        run_kwargs: Optional[Dict[str, Any]] = None # added line
    ) 
    ...
    def run(
        ...
        response = self._client.generate(model=self.model,
                                         prompt=prompt,
                                         stream=stream,
                                         options=generation_kwargs,
                                         **self.run_kwargs) # added line
        ...

# usage
generator = OllamaGenerator(
    run_kwargs={"keep_alive": "5m"},
)
```

**Additional context**

emso-c:
### Related Issues

- Partially resolves #1129

### Proposed Changes:

Added keep_alive parameter to `OllamaGenerator` and `OllamaChatGenerator`

### How did you test it?

Ran unit tests:
```bash
hatch run test -m"not integration"
```
Manual tests:
Set keep alive to 0 and checked process consumptions.
```python
from haystack_integrations.components.generators.ollama import OllamaGenerator, OllamaChatGenerator
from haystack.dataclasses import ChatMessage, ChatRole

generator = OllamaGenerator(keep_alive=0)
replies = generator.run("Say hello").get("replies")
assert replies and replies[0].strip() != ''

generator = OllamaChatGenerator(keep_alive=0)
message = ChatMessage(content="Hello", role=ChatRole.USER, name=None)
replies = generator.run([message]).get("replies")
assert replies and replies[0].meta['done'] == True
```

#### Expected behavior:
The processes immediately getting unloaded from memory.

#### Actual behaviour:
The processes immediately getting unloaded from memory.

### Checklist

- [X] I have read the [contributors guidelines](https://github.com/deepset-ai/haystack-core-integrations/blob/main/CONTRIBUTING.md) and the [code of conduct](https://github.com/deepset-ai/haystack-core-integrations/blob/main/CODE_OF_CONDUCT.md)
- [X] I have updated the related issue with new insights and changes
- [X] I added unit tests and updated the docstrings
- [X] I've used one of the [conventional commit types](https://www.conventionalcommits.org/en/v1.0.0/) for my PR title: `fix:`, `feat:`, `build:`, `chore:`, `ci:`, `docs:`, `style:`, `refactor:`, `perf:`, `test:`.

### DIFF
--- a/integrations/ollama/src/haystack_integrations/components/generators/ollama/chat/chat_generator.py
+++ b/integrations/ollama/src/haystack_integrations/components/generators/ollama/chat/chat_generator.py
@@ -56,7 +56,13 @@ class OllamaChatGenerator:
             A callback function that is called when a new token is received from the stream.
             The callback function accepts StreamingChunk as an argument.
         :param keep_alive:
-            The option that controls how long the model will stay loaded into memory following the request.
+            The option that controls how long the model will stay loaded into memory following the request (default: `5m`)
+            
+            The value can be set to:
+            - a duration string (such as "10m" or "24h")
+            - a number in seconds (such as 3600)
+            - any negative number which will keep the model loaded in memory (e.g. -1 or "-1m")
+            - '0' which will unload the model immediately after generating a response
         """
 
         self.timeout = timeout
--- a/integrations/ollama/src/haystack_integrations/components/generators/ollama/chat/chat_generator.py
+++ b/integrations/ollama/src/haystack_integrations/components/generators/ollama/chat/chat_generator.py
@@ -56,14 +56,13 @@ class OllamaChatGenerator:
             A callback function that is called when a new token is received from the stream.
             The callback function accepts StreamingChunk as an argument.
         :param keep_alive:
-            The option that controls how long the model will stay loaded into memory following the request
-            (default: `5m`)
-
+            The option that controls how long the model will stay loaded into memory following the request.
+            If not set, it will use the default value from the Ollama (5 minutes).
             The value can be set to:
             - a duration string (such as "10m" or "24h")
             - a number in seconds (such as 3600)
             - any negative number which will keep the model loaded in memory (e.g. -1 or "-1m")
-            - '0' which will unload the model immediately after generating a response
+            - '0' which will unload the model immediately after generating a response.
         """
 
         self.timeout = timeout
--- a/integrations/ollama/src/haystack_integrations/components/generators/ollama/chat/chat_generator.py
+++ b/integrations/ollama/src/haystack_integrations/components/generators/ollama/chat/chat_generator.py
@@ -169,7 +169,9 @@ class OllamaChatGenerator:
 
         stream = self.streaming_callback is not None
         messages = [self._message_to_dict(message) for message in messages]
-        response = self._client.chat(model=self.model, messages=messages, stream=stream, keep_alive=self.keep_alive, options=generation_kwargs)
+        response = self._client.chat(
+            model=self.model, messages=messages, stream=stream, keep_alive=self.keep_alive, options=generation_kwargs
+        )
 
         if stream:
             chunks: List[StreamingChunk] = self._handle_streaming_response(response)
--- a/integrations/ollama/src/haystack_integrations/components/generators/ollama/chat/chat_generator.py
+++ b/integrations/ollama/src/haystack_integrations/components/generators/ollama/chat/chat_generator.py
@@ -57,7 +57,7 @@ class OllamaChatGenerator:
             The callback function accepts StreamingChunk as an argument.
         :param keep_alive:
             The option that controls how long the model will stay loaded into memory following the request (default: `5m`)
-            
+
             The value can be set to:
             - a duration string (such as "10m" or "24h")
             - a number in seconds (such as 3600)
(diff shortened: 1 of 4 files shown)


integrations/ollama/src/haystack_integrations/components/generators/ollama/chat/chat_generator.py
```diff
@@ -54,12 +55,15 @@ def __init__(
         :param streaming_callback:
             A callback function that is called when a new token is received from the stream.
             The callback function accepts StreamingChunk as an argument.
+        :param keep_alive:
+            The option that controls how long the model will stay loaded into memory following the request.
```
anakin87, line 59: Let's add a comprehensive explanation of the possible values for this parameter.
https://github.com/ollama/ollama/blob/7fe3902552f762e6cc4f78eb83b2a60c6a6e61d8/docs/faq.md?plain=1#L246C1-L250C74

Let's also add that if not set, the default time is 5 minutes.

integrations/ollama/src/haystack_integrations/components/generators/ollama/generator.py
```diff
@@ -59,13 +60,16 @@ def __init__(
         :param streaming_callback:
             A callback function that is called when a new token is received from the stream.
             The callback function accepts StreamingChunk as an argument.
+        :param keep_alive:
+            The option that controls how long the model will stay loaded into memory following the request.
```
anakin87, line 64: same as above

anakin87 (changes requested):
Thanks for the PR! :blue_heart: 

I found some opportunities for improvement.

In addition to my comments, please also take care of:
- serialize this new parameter in the `to_dict` method
- update the init tests and the serialization/deserialization tests (`to_dict`/`from_dict`) to take this new parameter into account

integrations/ollama/src/haystack_integrations/components/generators/ollama/chat/chat_generator.py
emso-c: > Let's also add that if not set, the default time is 5 minutes.

Wouldn't it be more future-proof to keep the value null so that Ollama handles the default behaviour by itself? We'd have to update our default value as well if the Ollama's default value would be changed in the future.

anakin87: Yes. I agree.
I was only talking about adding this information in the docstring.

emso-c: Oh I see. But still, I believe adding such statement would require maintenance in the same sense. I think only describing the behaviour (even this can be removed for minimum coupling) and referencing the source documentation would be sufficient. Like this:

<big><pre>
:param keep_alive:
    The option that controls how long the model will stay loaded into memory following the request. It can be set to:
    - a duration string (such as "10m" or "24h")
    - a number in seconds (such as 3600)
    - any negative number which will keep the model loaded in memory (e.g. -1 or "-1m")
    - '0' which will unload the model immediately after generating a response
    For detailed information and usage examples, see the relevant sections in [Ollama API docs](https://github.com/ollama/ollama/blob/7fe3902552f762e6cc4f78eb83b2a60c6a6e61d8/docs/api.md?plain=1#L55) and [Ollama FAQ's](https://github.com/ollama/ollama/blob/7fe3902552f762e6cc4f78eb83b2a60c6a6e61d8/docs/faq.md?plain=1#L246C1-L250C74)
</pre></big>

I can still add the default time if you want.

anakin87: I prefer to add the default value to the docstring. Even if I am familiar with Ollama, it was not straightforward to find it, so I think it is a valuable information (agree with you that this will require maintaining it). We can avoid references to Ollama docs if we document this parameter comprehensively. 🙏

emso-c:
@anakin87 now that we addressed that the default value is `"5m"` in the docs, you think we should update the default value to `"5m"` in the code as well? It might be confusing for users to see that the value is  `None`, contradicting the docs.

anakin87 (approved):
I clarified better in the docstrings that, if not set, the parameter will use Ollama default.

Thanks for this contribution, which is ready to be merged!

PR #1132, opened 2024-10-10
docs: explain different connection string formats in the docstring

### LINKED ISSUE #928: Bug/Workaround/feature proposal : DBString Parsing in PgvectorDocumentStore
**Issue with Connection String Parsing in `PgvectorDocumentStore`**

**Description:**

Hi,

I encountered a problem connecting to an Azure Cosmos DB using `PgvectorDocumentStore` from the `haystack_integrations.document_stores.pgvector` module. The issue seems to stem from the inability of `psycopg2` (and the underlying C library `libpq`) to correctly parse my connection string.

**Steps to Reproduce:**

1. Using the following code:

    ```python
    from haystack_integrations.document_stores.pgvector import PgvectorDocumentStore
    import os 
    
    os.environ['PG_CONN_STR'] = "postgresql://USR:PWD@c-cosmos-datarelfrdev01.jiivzschmkljcv.postgres.cosmos.azure.com:5432/data?sslmode=require"
    
    document_store = PgvectorDocumentStore(
        table_name="docs_embeding",
        embedding_dimension=1024,
        vector_function="cosine_similarity",
        search_strategy="hnsw",
    )
    
    document_store.count_documents()
    ```

    I receive a `Name or service not known` error (which I verified is not due to a network issue).

2. I identified the issue as a problem with the connection string parsing in `psycopg2`. Testing with the following code:

    ```python
    import psycopg2
    
    def test_connection_str():
        try: 
            conn_url = "postgresql://USR:PWD@c-cosmos-datarelfrdev01.jiivzschmkljcv.postgres.cosmos.azure.com:5432/data?sslmode=require"
            connection = psycopg2.connect(conn_url)
            print("Connection string successful!")
        except: 
            print("Connection string failed!")
    
    def test_connection():
        try: 
            connection = psycopg2.connect(
                host="c-cosmos-datarelfrdev01.jiivzschmkljcv.postgres.cosmos.azure.com",
                port="5432",
                dbname="data",
                user="USR",
                password="PWD",
                sslmode="require"
            )
            print("Connection successful!")
        except: 
            print("Connection failed!")
    
    if __name__ == "__main__":
        test_connection()
        test_connection_str()
    ```

    The results are:

    ```
    Connection string failed!
    Connection successful!
    ```

**Workaround:**

I was able to bypass the issue by switching to a different connection string format:
```python
    conn_url = "host=c-cosmos-datarelfrdev01.jiivzschmkljcv.postgres.cosmos.azure.com port=5432 dbname=data user=USR password=PWD sslmode=require"
```

This code works as expected:
```python
    import os 

    os.environ['PG_CONN_STR'] = "host=c-cosmos-datarelfrdev01.jiivzschmkljcv.postgres.cosmos.azure.com port=5432 dbname=data user=USR password=PWD sslmode=require"

    document_store = PgvectorDocumentStore(
        table_name="docs_embeding",
        embedding_dimension=1024,
        vector_function="cosine_similarity",
        search_strategy="hnsw",
    )

    document_store.count_documents()
```
**Suggestion:**

I suggest implementing connection handling through individual arguments rather than relying solely on a connection string format. This would enhance compatibility and flexibility.

**Environment:**
- **Haystack Version:** 2.3.1
- **Integration Version:** 0.5.1

kanenorman:
### Related Issues

- fixes #928 
### Proposed Changes:
Allow user to provide connection parameters as a dictionary of individual arguments 

### How did you test it?

- Ran existing test
- Updated existing test
- Added unit test for `to_dict()` and `from_dict()`

### Notes for the reviewer

This PR is labeled as a feature rather than a bug because the underlying library, rather than Haystack itself, contains the bug.

### Checklist


integrations/pgvector/src/haystack_integrations/document_stores/pgvector/document_store.py
```diff
@@ -129,6 +130,12 @@ def __init__(
             `"hnsw"`. You can find more information about this parameter in the
             [pgvector documentation](https://github.com/pgvector/pgvector?tab=readme-ov-file#hnsw).
         :param keyword_index_name: Index name for the Keyword index.
+        :param connection_param_kwargs: A dictionary of parameters for the PostgreSQL connection.
+            If you prefer not to use the `connection_string`, you can specify connection parameters here.
```
davidsbatista, line 134:
```suggestion
            You can specify individual connections parameters here instead of a  `connection_string`
```

davidsbatista: I would suggest moving the `connection_param_kwargs` below the `connection_string` parameter, since they are both related and then also move the `connection_param_kwargs` docstring also right after the `connection_string`.

davidsbatista (commented):
thanks @kanenorman for your contribution! I just left a few minor comments

integrations/pgvector/src/haystack_integrations/document_stores/pgvector/document_store.py
kanenorman: Good call out @davidsbatista. Thank you for the review. I've made the changes in the latest commit.

anakin87 (changes requested):
@kanenorman I am sorry to step in after you have worked on this PR.

I think that the original issue proposes an acceptable workaround for the problem. (I tried it and it also works on default use cases)

After some discussion, we think that it is better to explain the possible connection string formats in the docstring instead of introducing a new init parameter, which might complicate the API and make future maintenance more difficult.

Therefore, my proposal is to:
- remove the `connection_param_kwargs` init parameter
- make the `connection_string` docstring more comprehensive, including an example like this: `"host=HOST port=PORT dbname=DBNAME user=USER password=PASSWORD"`.

kanenorman:
No problems @anakin87! Thank you for stepping in. I reverted the changes and enhanced the docstrings in the latest commit.

anakin87 (approved):
Thanks!

PR #1137, opened 2024-10-15
fix: make sure that streaming works with function calls - (drop python3.8)

anakin87:
### Related Issues

Tests are failing: https://github.com/deepset-ai/haystack-core-integrations/actions/runs/11337043949
The conversion of function calls into dictionaries is not working properly.

### Proposed Changes:
- Use the chunk dictionaries to access all required attributes (instead of Google Proto objects).
- Officially drop Python 3.8 in this package (`google-generativeai` never supported it)

### How did you test it?
CI

### Checklist

### DIFF
--- a/integrations/google_ai/pyproject.toml
+++ b/integrations/google_ai/pyproject.toml
@@ -7,7 +7,7 @@ name = "google-ai-haystack"
 dynamic = ["version"]
 description = 'Use models like Gemini via Makersuite'
 readme = "README.md"
-requires-python = ">=3.8"
+requires-python = ">=3.9"
 license = "Apache-2.0"
 keywords = []
 authors = [{ name = "deepset GmbH", email = "info@deepset.ai" }]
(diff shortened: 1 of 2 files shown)


Amnah199:
LG! Would you say we should also replace Proto objects approach for below non-stream responses to prevent any failures later?
https://github.com/deepset-ai/haystack-core-integrations/blob/a223e6f0705ec0d788205ae6c996919d5d0258d1/integrations/google_ai/src/haystack_integrations/components/generators/google_ai/chat/gemini.py#L325

anakin87:
> LG! Would you say we should also replace Proto objects approach for below non-stream responses to prevent any failures later?
> 
> https://github.com/deepset-ai/haystack-core-integrations/blob/a223e6f0705ec0d788205ae6c996919d5d0258d1/integrations/google_ai/src/haystack_integrations/components/generators/google_ai/chat/gemini.py#L325

Since this is working now and is covered by tests, I would leave it unchanged. But I leave the decision to you...

(Also, I envision a massive refactoring of this component, when we will add better support for tools.)

Amnah199:
In that case, makes sense to refactor it once Tools are supported. Thanks for the fix.

Amnah199 (approved):

PR #1150, opened 2024-10-22
feat: add `JinaReaderConnector`

### LINKED ISSUE #663: Add support for Reader API to convert HTMLs into Documents
There's no component to use [Jina's Reader API](https://jina.ai/reader/) with Haystack.

A new JinaHTMLtoDocument (name TBD) component to use [Jina's Reader API](https://jina.ai/reader/) to convert URLs into Haystack Documents. This component should accept a URL and output a Haystack Document.

- This component can output a markdown file and users might use MarkdownConverter to use that component in a pipeline (not Haystack intuitive but might have advantages)
- Depending on how the Reader API works, it can accept a list of URLs and return a list of Haystack Documents

**Additional context**

jlonge4:
### Related Issues

- part of #663

### Proposed Changes:

Implement the Jina AI Reader as a haystack component 

### How did you test it?

Manual Verification

### Notes for the reviewer

N/A

### Checklist


jlonge4:
Hi @anakin87, @Anitha6g and I were trying to think of the best way to handle returning the output of the tool. 
A document definitely makes sense for the read and search mode, but what about ground as it returns: Statement, Factuality, Result, Reason.

The unfortunate thing is that it is all markdown and not structured. Should we parse the content and put those things in doc metadata? Thoughts?

anakin87:
Let me have a look ⌛

anakin87:
I'm finally back, after doing a deeper investigation :detective:  and discussing this with @bilgeyucel.

- We think this component should be named `JinaReaderConnector` and placed in components/connectors

- While studying the issue with grounding, I found that you can always get a JSON response by all endpoints (reader, search, grounding), if you pass the header `'Accept': 'application/json'`. You can play with this and see the results at https://jina.ai/reader/.

- After discovering this JSON response, I would say that we should use this by default because it is more granular and informative (we can add an init parameter called `json_response` wich defaults to `True`). This implies that for search we won't have a Document but one for each search result. So I would change the general return type to `List[Document]` and do something similar to what is discussed below.
  - If `json_response==False`: return a list of only one Document with only the content, which is a markdown string.
  - If `json_response==True`, this means we should do some parsing/packing of the results depending on the mode:
    - for reader, create a list of one Document with the returned content as content and all other fields (`title`, `URL`, ...) copied into `meta`
    - for search, create a list of Documents (one for each search result) with the returned content as content and all other fields (`title`, `URL`, ...) copied into `meta`
    - for grounding (not a great solution, but OK),  create a list of one Document with the `reason` field as content and all fields (`result`, `reference`, ..., including `reason`) copied into `meta`

- By playing with the website, I discovered that there are many other options to run this API which can be set using headers. For this reason, I would add a parameter called  `headers` to the `run` method, to make these option available to people.


Let me know what you think. I am quite convinced about the plan, even if it requires some work.

jlonge4:
@anakin87 @bilgeyucel great points, will get right on it and rebase in a couple days. Thanks for the direction as always 😎

anakin87:
I will try reviewing this PR today or tomorrow...

integrations/jina/src/haystack_integrations/components/reader/jina/__init__.py
```diff
@@ -0,0 +1,6 @@
+# SPDX-FileCopyrightText: 2023-present deepset GmbH <info@deepset.ai>
+#
+# SPDX-License-Identifier: Apache-2.0
+from .reader import JinaReaderConnector
+
+__all__ = ["JinaReaderConnector"]
```
anakin87, line 6: we should put this component in  
`integrations/jina/src/haystack_integrations/components/converters/`

integrations/jina/src/haystack_integrations/components/reader/jina/reader.py
```diff
@@ -0,0 +1,130 @@
...
+        """
+        resolved_api_key = api_key.resolve_value()
+        self.api_key = api_key
+        self._session = requests.Session()
+        self.json_response = json_response
+        self._session.headers.update(
+            {
+                "Authorization": f"Bearer {resolved_api_key}",
+                "Accept-Encoding": "identity",
+                "Content-type": "application/json",
```
anakin87, line 58:
```suggestion

```

integrations/jina/src/haystack_integrations/components/reader/jina/reader.py
```diff
@@ -0,0 +1,130 @@
...
+            }
+        )
+
+        if self.json_response:
+            self._session.headers.update({"Accept": "application/json"})
+
+        if isinstance(mode, JinaReaderMode):
+            self.mode = mode.value
+        elif isinstance(mode, str):
+            self.mode = mode.upper()
```
anakin87, line 68:
```suggestion
        if isinstance(mode, str):
            self.mode = JinaReaderMode.from_str(mode)
```

integrations/jina/src/haystack_integrations/components/reader/jina/reader.py
```diff
@@ -0,0 +1,130 @@
...
+    def to_dict(self) -> Dict[str, Any]:
+        """
+        Serializes the component to a dictionary.
+        :returns:
+            Dictionary with serialized data.
+        """
+        return default_to_dict(
+            self,
+            api_key=self.api_key.to_dict(),
+            mode=self.mode,
```
anakin87, line 79:
```suggestion
            mode=str(self.mode),
```

integrations/jina/src/haystack_integrations/components/reader/jina/reader.py
```diff
@@ -0,0 +1,130 @@
...
+        :param data:
+            Dictionary to deserialize from.
+        :returns:
+            Deserialized component.
+        """
+        deserialize_secrets_inplace(data["init_parameters"], keys=["api_key"])
+        return default_from_dict(cls, data)
+
+    def parse_json_response(self, data: dict) -> Document:
+        if self.mode == "GROUND":
```
anakin87, line 96:
```suggestion
        if self.mode == JinaReaderMode.GROUND:
```

integrations/jina/src/haystack_integrations/components/reader/jina/reader.py
```diff
@@ -0,0 +1,130 @@
...
+    def run(self, query: str, headers: Optional[Dict[str, str]] = None):
+        """
+        Process the query using the Jina AI reader service.
+
+        :param query: The query string or URL to process.
+        :returns: A list containing a single Document object with the processed content and metadata.
+        """
+        if headers:
+            self._session.headers.update(headers)
+        mode_map = {"READ": "r", "SEARCH": "s", "GROUND": "g"}
```
anakin87, line 114: this can be exposed at the module level.
Something like:
`mode_map = {JinaReaderMode.READ: "r", JinaReaderMode.SEARCH: "s", JinaReaderMode.GROUND: "g"}`

anakin87 (commented):
I thought I was late with the review, then I realized it is still work in progress. :-)

in any case, I hope it is useful.

jlonge4:
@anakin87 very useful as always! Should be sorted now :)

integrations/jina/src/haystack_integrations/components/converters/jina/reader.py
```diff
@@ -0,0 +1,127 @@
...
+
+    def parse_json_response(self, data: dict) -> Document:
+        if self.mode == JinaReaderMode.GROUND:
+            content = data.pop("reason")
+        else:
+            content = data.pop("content")
+        document = Document(content=content, meta=data)
+        return document
+
+    # TODO add headers param
```
anakin87, line 100:
```suggestion
```

anakin87 (commented):
I took another look.

Could you add tests?

jlonge4:
@vblagoje thanks very much! Still have to whip up those tests but should be done(ish) today 🤞🏼

jlonge4:
@vblagoje @anakin87 what stupid mistake am I making with these test failures and import structure?

anakin87:
@jlonge4 the circular import error should be fixed now

jlonge4:
@anakin87  how are we looking?

anakin87:
@jlonge4 in the coming days, I will take a proper look, so as to push this to the finish line.

PR #1152, opened 2024-10-22
refactor: avoid downloading tokenizer if `truncate` is `False`

### LINKED ISSUE #1125: AmazonBedrockGenerator shouldn't load tokenizer if truncate is set to False.
DefaultPromptHandler downloads unnecessary GPT2 tokenizer to a read-only filesystem (eg. AWS Lambda), when `truncate` argument is set to `False`.

Check https://github.com/deepset-ai/haystack-core-integrations/blob/cf52ce94c9a1ae7f33bbde14c8639e2058262707/integrations/amazon_bedrock/src/haystack_integrations/components/generators/amazon_bedrock/generator.py#L148-L156

If truncate is set to false, GPT2 tokenizer for token estimation shouldn't be loaded.

Simply place a condition to only load the tokenizer if truncate is set to true.
 
```python
        # Truncate prompt if prompt tokens > model_max_length-max_length
        # (max_length is the length of the generated text)
        # we use GPT2 tokenizer which will likely provide good token count approximation
--->    if self.truncate:
            self.prompt_handler = DefaultPromptHandler(
                tokenizer="gpt2",
                model_max_length=model_max_length,
                max_length=self.max_length or 100,
            )
```

anakin87:
### Related Issues

- fixes #1125

### Proposed Changes:
If `truncate` is `False`, do not initialize the `PromptHandler` (which internally downloads the tokenizer).
In the Generator, the `PromptHandler` is only used to truncate the text if needed.

### How did you test it?
CI + a new unit test.

### Notes for the reviewer
This change only applies to `AmazonBedrockGenerator`.

I would have liked to do the same for the corresponding Chat Generator.
Unfortunately, in the Chat Generator, the `PromptHandler` (and its internal tokenizer) have 2 different purposes:
- transform messages into text with special tokens (via `tokenizer.apply_chat_template`)
- truncate the text, like in the Generator

Therefore, we cannot easily get rid of the tokenizer.

(I hope we can prioritize #977 soon: this would greatly simplify the Chat Generator.)

### Checklist

### DIFF
--- a/integrations/amazon_bedrock/src/haystack_integrations/components/generators/amazon_bedrock/generator.py
+++ b/integrations/amazon_bedrock/src/haystack_integrations/components/generators/amazon_bedrock/generator.py
@@ -153,7 +153,7 @@ class AmazonBedrockGenerator:
         model_max_length = kwargs.get("model_max_length", 4096)
 
         # we initialize the prompt handler only if truncate is True: we avoid unnecessarily downloading the tokenizer
-        if truncate:
+        if self.truncate:
             # Truncate prompt if prompt tokens > model_max_length-max_length
             # (max_length is the length of the generated text)
             # we use GPT2 tokenizer which will likely provide good token count approximation
--- a/integrations/amazon_bedrock/src/haystack_integrations/components/generators/amazon_bedrock/generator.py
+++ b/integrations/amazon_bedrock/src/haystack_integrations/components/generators/amazon_bedrock/generator.py
@@ -152,15 +152,16 @@ class AmazonBedrockGenerator:
         # We pop the model_max_length as it is not sent to the model but used to truncate the prompt if needed
         model_max_length = kwargs.get("model_max_length", 4096)
 
-        # Truncate prompt if prompt tokens > model_max_length-max_length
-        # (max_length is the length of the generated text)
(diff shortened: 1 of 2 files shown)


integrations/amazon_bedrock/src/haystack_integrations/components/generators/amazon_bedrock/generator.py
```diff
@@ -152,15 +152,16 @@ def resolve_secret(secret: Optional[Secret]) -> Optional[str]:
...
-        # (max_length is the length of the generated text)
-        # we use GPT2 tokenizer which will likely provide good token count approximation
-
-        self.prompt_handler = DefaultPromptHandler(
-            tokenizer="gpt2",
-            model_max_length=model_max_length,
-            max_length=self.max_length or 100,
-        )
+        # we initialize the prompt handler only if truncate is True: we avoid unnecessarily downloading the tokenizer
+        if truncate:
```
Amnah199, line 156: Not a big suggestion, but we can use `self.truncate` here for consistency? As we are using that in `run` method to check whether `prompt_handler` is called.

anakin87: done!

Amnah199 (approved):
LG!

PR #1165, opened 2024-11-04
fix: adapt our implementation to breaking changes in Chroma 0.5.17 

anakin87:
### Related Issues
- failing tests: https://github.com/deepset-ai/haystack-core-integrations/actions/runs/11655758403
- in 0.5.17, Chroma introduced some breaking changes:
  - writing documents that do not contain a valid string (`None`) causes an error
  - they changed filter default: from `{}` to `None`

### Proposed Changes:
- incorporate their changes
- pin `chroma>=0.5.17`

### How did you test it?
CI

### Checklist

### DIFF
--- a/integrations/chroma/src/haystack_integrations/document_stores/chroma/document_store.py
+++ b/integrations/chroma/src/haystack_integrations/document_stores/chroma/document_store.py
@@ -248,7 +248,7 @@ class ChromaDocumentStore:
 
             if doc.content is None:
                 logger.warning(
-                    "ChromaDocumentStore cannot store documents without `content`. "
+                    "ChromaDocumentStore cannot store documents with `content=None`. "
                     "`array`, `dataframe` and `blob` are not supported. "
                     "Document with id %s will be skipped.",
                     doc.id,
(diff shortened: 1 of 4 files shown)


integrations/chroma/src/haystack_integrations/document_stores/chroma/document_store.py
```diff
@@ -248,9 +248,11 @@ def write_documents(self, documents: List[Document], policy: DuplicatePolicy = D
 
             if doc.content is None:
                 logger.warning(
-                    "ChromaDocumentStore can only store the text field of Documents: "
-                    "'array', 'dataframe' and 'blob' will be dropped."
+                    "ChromaDocumentStore cannot store documents without `content`. "
+                    "Document with id %s will be skipped.",
```
Amnah199, line 252: Did something change about storing arrays and dataframes etc. Otherwise we should still include that in the warning?

anakin87: i improved the message now.

davidsbatista: nit: I would replace the "without" with something like: `ChromaDocumentStore cannot index empty Document(s) (i.e.: content=None)` but just a nit - all looks good to me as it

davidsbatista (approved):
LGTM

PR #1178, opened 2024-11-12
feat: Fastembed - add FastembedRanker

paulmartrencharpro:
### Proposed Changes:
New feature: integrate with the new ranker from fastembed. (https://qdrant.tech/articles/cross-encoder-integration-gsoc/, https://github.com/qdrant/fastembed/tree/main?tab=readme-ov-file#-rerankers)

I based my code on CohereRanker and Fastembed existing code

### How did you test it?

I wrote all the tests in test_fastembed_ranker.py, based on the other tests for fastembed.

### Checklist


integrations/fastembed/src/haystack_integrations/components/rankers/fastembed/ranker.py
```diff
@@ -0,0 +1,194 @@
...
+    ```python
+    from haystack import Document
+    from haystack_integrations.components.rankers.fastembed import FastembedRanker
+
+    ranker = FastembedRanker(model_name="Xenova/ms-marco-MiniLM-L-6-v2", top_k=2)
+
+    docs = [Document(content="Paris"), Document(content="Berlin")]
+    query = "What is the capital of germany?"
+    output = ranker.run(query=query, documents=docs)
+    docs = output["documents"]
```
anakin87, line 27:
```suggestion
   print(output["documents"][0].content)
   
   # Berlin
```

I think adding an example of the output can help.

integrations/fastembed/src/haystack_integrations/components/rankers/fastembed/ranker.py
```diff
@@ -0,0 +1,194 @@
...
+                If > 1, data-parallel encoding will be used, recommended for offline encoding of large datasets.
+                If 0, use all available cores.
+                If None, don't use data-parallel processing, use default onnxruntime threading instead.
+        :param local_files_only: If `True`, only use the model files in the `cache_dir`.
+        :param meta_fields_to_embed: List of meta fields that should be concatenated
+            with the document content for reranking.
+        :param meta_data_separator: Separator used to concatenate the meta fields
+            to the Document content.
+        """
+        self.model_name = model_name
```
anakin87, line 63:
```suggestion
        if top_k <= 0:
            raise ValueError(f"top_k must be > 0, but got {top_k}")
            
        self.model_name = model_name
```

let's add this check

integrations/fastembed/src/haystack_integrations/components/rankers/fastembed/ranker.py
```diff
@@ -0,0 +1,194 @@
...
+
+        :param query:
+            Query string.
+        :param documents:
+            List of Documents.
+        :param top_k:
+            The maximum number of Documents you want the Ranker to return.
+        :returns:
+            A dictionary with the following keys:
+            - `documents`: List of Documents most similar to the given query in descending order of similarity.
```
anakin87, line 147:
```suggestion
        Returns a list of documents ranked by their similarity to the given query, using FastEmbed.

        :param query:
            The input query to compare the documents to.
        :param documents:
            A list of documents to be ranked.
        :param top_k:
            The maximum number of documents to return.

        :returns:
            A dictionary with the following keys:
            - `documents`: A list of documents closest to the query, sorted from most similar to least similar.
```

integrations/fastembed/src/haystack_integrations/components/rankers/fastembed/ranker.py
```diff
@@ -0,0 +1,194 @@
...
+        if not isinstance(documents, list) or documents and not isinstance(documents[0], Document):
+            msg = "FastembedRanker expects a list of Documents as input. "
+            raise TypeError(msg)
+        if query == "":
+            msg = "No query provided"
+            raise ValueError(msg)
+
+        if documents.count == 0:
+            msg = "No documents provided"
+            raise ValueError(msg)
```
anakin87, line 160:
```suggestion
        if not documents:
            return {"documents": []}
```

we usually don't raise an error in this case

integrations/fastembed/src/haystack_integrations/components/rankers/fastembed/ranker.py
```diff
@@ -0,0 +1,194 @@
...
+        batch_size: int = 64,
+        parallel: Optional[int] = None,
+        local_files_only: bool = False,
+        meta_fields_to_embed: Optional[List[str]] = None,
+        meta_data_separator: str = "\n",
+    ):
+        """
+        Creates an instance of the 'FastembedRanker'.
+
+        :param model_name: Fastembed model name. Check the list of supported models in the [Fastembed documentation](https://docs.Fastembed.com/docs/models).
```
anakin87, line 46: wrong URL

anakin87 (changes requested):
Hey! Thank you for this PR.

I've had a first look and it seems good.

I have an idea: in general, we have an embedding backend for local Embedders because we expect users to use both Text Embedder and Doc Embedder based on the same model; this saves resources (the model is loaded only once).

For rankers, we don't have this problem.
So I would ask you to **remove the backend and do everything in the component** (Cohere can be taken as example).

Let me know and thanks again!

integrations/fastembed/src/haystack_integrations/components/rankers/fastembed/ranker.py
paulmartrencharpro: Indeed. I changed it
Note: the models are not there yet. I made a PR for this on the fastembed part (https://github.com/qdrant/fastembed/pull/393). It's merged, but I think it needs a release to be pushed on the website

integrations/fastembed/tests/test_fastembed_ranker.py
```diff
@@ -0,0 +1,235 @@
...
+        assert ranker.batch_size == 50
+        assert ranker.parallel == 1
+        assert ranker.local_files_only
+        assert ranker.meta_fields_to_embed == ["test_field"]
+        assert ranker.meta_data_separator == " | "
+
+    def test_embed_incorrect_input_format(self):
+        """
+        Test for checking incorrect input format when creating embedding.
+        """
```
anakin87, line 165:
```suggestion
    def test_run_incorrect_input_format(self):
        """
        Test for checking incorrect input format.
        """
```

integrations/fastembed/tests/test_fastembed_ranker.py
```diff
@@ -0,0 +1,235 @@
...
+        with pytest.raises(
+            TypeError,
+            match="FastembedRanker expects a list of Documents as input.",
+        ):
+            ranker.run(query=query, documents=list_integers_input)
+
+    def test_embed_metadata(self):
+        """
+        Test for checking output dimensions and embedding dimensions for documents
+        with a custom instruction and metadata.
```
anakin87, line 188:
```suggestion
        Tests the embedding of metadata fields in document content for ranking.
```

integrations/fastembed/tests/test_fastembed_ranker.py
```diff
@@ -0,0 +1,235 @@
...
+        ]
+
+        result = ranker.run(query=query, documents=documents)
+
+        assert len(result["documents"]) == 2
+        first_document = result["documents"][0]
+
+        assert isinstance(first_document, Document)
+        assert first_document.content == "fastembed is supported by and maintained by Qdrant."
+        assert first_document.score == 5.472434997558594
```
anakin87, line 235: Let's only check that first document score > second document score.
Otherwise, this test can become flaky due to temporary circumstances and external changes.

paulmartrencharpro: Done

paulmartrencharpro:
> In general, this PR is almost ready to be merged.
> 
> TODO:
> 
>     * modify this line in README to mention ranker: https://github.com/deepset-ai/haystack-core-integrations/blob/c3437763e7c80e11e9ed8678f1b235353a316781/README.md?plain=1#L37
> 
>     * add more tests if you can/want (uncovered lines: 66-67, 109, 115->exit, 162-163, 166, 170-171, 174-175)
> 
> 
> Then, if you also want to open a PR on haystack-integrations to update this page ( https://github.com/deepset-ai/haystack-integrations/blob/main/integrations/fastembed.md), it would be appreciated - otherwise, I'll take care of it.

I updated the Readme and added tests.
The only test I cannot write is the one that check that we don't create 2 TextCrossEncoder when we call warm_up multiple times. I don't understand the patch/mockup code to replicate it. I wrote this but it's not working:

    @patch('fastembed.rerank.cross_encoder.TextCrossEncoder')
    def test_warmup_twice(self, mock_encoder):
        """
        Test for checking that warm up only creates one TextCrossEncoder
        """
        ranker = FastembedRanker(model_name="Xenova/ms-marco-MiniLM-L-12-v2")
        ranker.warm_up()
        ranker.warm_up()

        # Assert that TextCrossEncoder is called only once
        mock_encoder.assert_called_once()
        

I'll make the PR on the other repo

anakin87 (approved):
Everything looks good now!

Thanks for the good work. :heart_decoration: 

(I'll track the related stuff in https://github.com/deepset-ai/haystack-core-integrations/issues/1188)

anakin87 (approved):

Issue #1154, opened 2024-10-24
Connect the generators to prompts in langfuse 


alex-stoica:
**Is your feature request related to a problem? Please describe.**
Hello! Currently, pipeline runs (equivalent to traces in Langfuse) do not have their corresponding prompts tied with the "generation" blocks. This can be seen in the example from the [Haystack Langfuse Integration blog post](https://haystack.deepset.ai/blog/langfuse-integration).
![image](https://github.com/user-attachments/assets/51ba6221-a99d-4c87-90de-d4e8d4689b5e)

**Describe the solution you'd like** 
When prompts exist, they should be added to the tracer file, such as in the following example:

```python
m = meta[0]
try:
    prompt = self._tracer.get_prompt(m.get("prompt_name")) # assumes that in model's output there is an additional parameter "prompt_name"
    if prompt:
        span._span.update(prompt=prompt)
except Exception as e:
    print(f"Prompt not found or error occurred: {e}")
span._span.update(usage=m.get("usage") or None, model=m.get("model"))
```
This would ensure that prompts are tied to the corresponding generation blocks, giving better traceability.

**Describe alternatives you've considered**
N/A

**Additional context**
Desired output: 
![image](https://github.com/user-attachments/assets/a1b7f1fb-45a0-4910-a628-3c1887d019f3)

One challenge is that the prompt name is not available in the generator by default. This could complicate the integration, as it may require accessing the prompt metadata manually. In some cases, creating a custom generator that explicitly handles the prompt name and metadata might be necessary for an integration with Langfuse + allowing that custom generator to be traced (see https://github.com/deepset-ai/haystack-core-integrations/issues/1153)

vblagoje:
Hey @alex-stoica let me see if I understand everything correctly here. It seems the assumption that `prompt_name` is, by default, added to ChatMessage meta seems a bit over the top and rather langfuse specific. Am I understanding this correctly? Isn't this prompt_name info available somewhere in langfuse meta by any chance?

alex-stoica:
Hello @vblagoje,
To add context: tracking users, prompts, and generation scores is valuable—especially for analyzing prompts. Currently, Haystack's Langfuse integration:

- Creates a trace per run (good, 1:1 correspondence)
- Logs a generation per generator run in the pipeline (also useful)
---
[...]

PR #1184, opened 2024-11-12
Fixed TypeError in LangfuseTrace

### LINKED ISSUE #1174: TypeError: LangfuseTracer.trace() got an unexpected keyword argument 'parent_span'
**Describe the bug**
I had a running Haystack pipeline that was using langfuse for tracing. However, all of the sudden I received an error "TypeError: LangfuseTracer.trace() got an unexpected keyword argument 'parent_span'". @marcklingen stated, that "Parent_span does not exist on a trace, which is correct". Here is a simple example of the code I am using:

**To Reproduce**
Example code:

```from haystack.document_stores.in_memory import InMemoryDocumentStore
from haystack.components.converters import PyPDFToDocument
from haystack.components.preprocessors import DocumentSplitter, DocumentCleaner
from haystack.components.writers import DocumentWriter
from haystack.components.embedders import SentenceTransformersDocumentEmbedder, SentenceTransformersTextEmbedder
from haystack.components.generators import HuggingFaceAPIGenerator
from haystack import Pipeline
from haystack_integrations.components.connectors.langfuse import LangfuseConnector

embedding_model = "sentence-transformers/all-MiniLM-L6-v2"

document_embedder = SentenceTransformersDocumentEmbedder(model=embedding_model)
query_embedder = SentenceTransformersTextEmbedder(model=embedding_model)

llm_generator = HuggingFaceAPIGenerator(
    api_type="serverless_inference_api",
    api_params={"model": "mistralai/Mistral-7B-Instruct-v0.3"},
    generation_kwargs={
        "temperature": 1,
    }
)

prompt_template = """Based on the following documents, answer the question:

{% for doc in documents %}
    {{ doc.content }}
{% endfor %}

Question: {{ query }}
Answer:"""

document_store = InMemoryDocumentStore()

pdf_converter = PyPDFToDocument()
document_cleaner = DocumentCleaner()
document_splitter = DocumentSplitter(split_by="sentence", split_length=5, split_overlap=0)
document_writer = DocumentWriter(document_store=document_store)

indexing_pipeline = Pipeline()
indexing_pipeline.add_component("tracer", LangfuseConnector("Simple Haystack Example"))
indexing_pipeline.add_component("pdf_converter", pdf_converter)
indexing_pipeline.add_component("document_cleaner", document_cleaner)
indexing_pipeline.add_component("document_splitter", document_splitter)
indexing_pipeline.add_component("embedder", document_embedder)
indexing_pipeline.add_component("document_writer", document_writer)

indexing_pipeline.connect("pdf_converter", "document_cleaner")
indexing_pipeline.connect("document_cleaner", "document_splitter")
indexing_pipeline.connect("document_splitter", "embedder")
indexing_pipeline.connect("embedder", "document_writer")

document_path = ["ADD_PDF_HERE.pdf"]
indexing_pipeline.run({"pdf_converter": {"sources": document_path}})
```

 - Pythen 3.11
 - Haystack version: 2.7.0
 - Integration (Langfuse) version: 0.5.0

alex-stoica:
### Related Issues

- fixes #1174 and another issue related to it (missing / empty `span._data`)

### Proposed Changes:

The primary issue was caused by the missing `parent_span` parameter in `LangfuseTracer.trace`, as explained in the original issue discussion.

Additionally, after updating the function signature to accept `parent_span`, `meta` was not found anymore as the code didn’t rely on it directly. The solution involves creating a root span when on creating a root span when `parent_span` does not exist or a child span when it does.

### How did you test it?

Locally with the following setup:

```python
pipe.add_component("llm", AzureOpenAIGenerator(...))
pipe.add_component("pb", PromptBuilder(...))
pipe.connect("pb.prompt", "llm.prompt")
```

And with a custom generator

### Notes for the reviewer
I’ve left some commented code for a potential future implementation related to https://github.com/deepset-ai/haystack-core-integrations/issues/1154


vblagoje:
@silvanocerza do you want to take over this PR review?

integrations/langfuse/src/haystack_integrations/tracing/langfuse/tracer.py
```diff
@@ -115,86 +117,105 @@ def __init__(self, tracer: "langfuse.Langfuse", name: str = "Haystack", public:
...
+            yield span
+        finally:
             self._context.pop()
-
+            if created_root_span:
+                self._context.pop()
+
+            # Update span metadata based on component type
+            if tags.get(COMPONENT_TYPE_KEY) in _SUPPORTED_GENERATORS:
+                meta = span._data.get(COMPONENT_TYPE_KEY, {}).get("meta")
```
vblagoje, line 177: Did you mean?
```suggestion
                meta = span._data.get(COMPONENT_OUTPUT_KEY, {}).get("meta")
```

integrations/langfuse/src/haystack_integrations/tracing/langfuse/tracer.py
```diff
@@ -115,86 +117,105 @@ def __init__(self, tracer: "langfuse.Langfuse", name: str = "Haystack", public:
...
+                if meta:
+                    # Haystack returns one meta dict for each message, but the 'usage' value
+                    # is always the same, let's just pick the first item
+                    m = meta[0]
+                    span._span.update(usage=m.get("usage") or None, model=m.get("model"))
+                    # add prompt object to generator #1154
```
vblagoje, line 183: Can you please remove this commented code here 🙏

vblagoje (changes requested):
Some minor issues found, please have another look @alex-stoica  🙏

alex-stoica:
Nice catches, @vblagoje. Solved them

alex-stoica:
@vblagoje Indeed, I had overlooked the `end()` method call, and it turned out to be a bit trickier than I initially expected. At first, I tried updating the span with `start_time` and `end_time`, but after taking a closer look, I realized that `end()` effectively addressed the issue 
image

I've also tested locally with `chat.py` with `AzureOpenAIChatGenerator` and it worked as expected. 

Besides this, I see some future ideas for improving the Langfuse integration:
1. A test with loops / cycles 
2. Adding a trace name instead of leaving the default "Haystack"

LastRemote:
@alex-stoica Thanks for making this change! I added the parent_span parameter in Haystack 2.7 with the intention to fix a bug where Langfuse traces overlap with each if there are two or more pipeline.run() running concurrently (which is very common in deployed pipelinements), but I haven't had a chance to update the Langfuse integration part since then.

Your changes look good, although I'd recommend we _always_ create a new root span if `parent_span` is None to prevent the concurrency issue (there would be cases that the 2nd pipeline.run has `parent_span` is None with a non-empty `current_span`, resulting the follow-up trace be registered as a `span` instead of `trace`.

Here is something I have in my draft:

```python
tracing_context: ContextVar[Dict[Any, Any]] = ContextVar("tracing_context", default={})
"""
A context variable containing information related to tracing. This can be used to store trace, user, and session IDs,
custom tags and pipeline version information.
"""

class LangfuseTracer(Tracer):
    @contextlib.contextmanager
    def trace(
        self, operation_name: str, tags: Optional[Dict[str, Any]] = None, parent_span: Optional[Span] = None
    ) -> Iterator[Span]:
        """
        Start and manage a new trace span.

        :param operation_name: The name of the operation.
        :param tags: A dictionary of tags to attach to the span.
        :param parent_span: The parent span to use for the newly created span.
        :return: A context manager yielding the span.
        """
        # pylint: disable=protected-access
        # This is to make it consistent with the original implementation
        tags = tags or {}
        span_name = tags.get("haystack.component.name", operation_name)

        if not parent_span:
            if operation_name != "haystack.pipeline.run":
                logger.warning(
                    "Creating a new trace without a parent span is not recommended for operation '%s'.", operation_name
                )
            # Create a new trace if no parent span is provided
            span = LangfuseSpan(
                self._tracer.trace(
                    name=self._name,
                    public=self._public,
                    id=tracing_context.get().get("trace_id"),
                    user_id=tracing_context.get().get("user_id"),
                    session_id=tracing_context.get().get("session_id"),
                    tags=tracing_context.get().get("tags"),
                    version=tracing_context.get().get("version"),
                )
            )
        else:
            if tags.get("haystack.component.type") in self.ALL_SUPPORTED_GENERATORS:
                span = LangfuseSpan(parent_span.raw_span().generation(name=span_name))
            else:
                span = LangfuseSpan(parent_span.raw_span().span(name=span_name))

        self._context.append(span)
        span.set_tags(tags)

        yield span

        ...  # The rest of the code is left unchanged
```

I hope this helps.

alex-stoica:
@LastRemote I see where you’re coming from. I also envisioned some "extra-storage" in the tracer (https://github.com/deepset-ai/haystack-core-integrations/issues/1154#issuecomment-2472993625), however I was unaware of the concurrency issues as  the fix that allowed hayhooks to run concurrently different `run()` for the same pipeline was very recently merged https://github.com/deepset-ai/hayhooks/pull/27 

Unfortunately, I will be unavailable for the next 3-4 days, so I recommend either integrating the fix as it stands or extending it with the context variable proposal you mentioned, provided @vblagoje or someone else from Haystack team approves. In this case I think you can add a commit to extend this PR. 

IMO, the people involved in the Langfuse integration (both from Haystack and external contributors like us) should be on the same page on multiple issues. In this case, I believe the context var approach could also be an opportunity to explore how it might assist in linking generators with their corresponding prompts, but I need to think a little bit more about it

silvanocerza (approved):
Thanks everyone for the contribution.
I integrated the changes from @LastRemote and made some other changes.

I'll merge and release so the integration is again usable.

LastRemote:
> I think the people involved in the Langfuse integration (both from Haystack and external contributors like us) should be on the same page on multiple issues. In this case, I believe the context var approach could also be an opportunity to explore how it might assist in linking generators with their corresponding prompts, but I need to think a little bit more about it

I agree. I hope there is a mailing list or a well-maintained GitHub label or maybe a Discord channel to track all ongoing issues and feature requests (personally I did a bad job maintaining issues though). We can bring this up to the community managers and see how this can be done.

PR #1192, opened 2024-11-14
Add AnthropicVertexChatGenerator component

### LINKED ISSUE #1106: Extend Anthropic integration to AnthropicVertex
The current Anthropic integration calls the general endpoint provided by Anthropic but cannot call the models through GCP vertex. The google vertex ai integration does not allow to select anthopic model (seems to be a limitation of the google client)

Alow to pass an alternative client to the `AnthropicChatGenerator` using `AnthropicVertex` instead of `Anthropic` (both from the anthropic lib) so we can use the chat generator but calling a model through GCP vertex.

Update the google vertex integration but it seems more complicated since the google client seems to be focus on Gemini.

**Additional context**
You're doing a great job, love your lib!

Amnah199:
### Related Issues

- fixes #1106 

### Proposed Changes:

- Create `AnthropicVertexChatGenerator` class that inherits from `AnthropicChatGenerator`
- Replace `api_key` with parameters needed for `AnthropicVertex`

### How did you test it?

- Added new tests for the component
- Locally tested some examples

### Notes for the reviewer

A documentation will be created for this component
### Checklist


integrations/anthropic/src/haystack_integrations/components/generators/anthropic/adapter.py
```diff
@@ -0,0 +1,98 @@
...
+            else:
+                _fields[_field.name] = getattr(self, _field.name)
+
+        return {"type": self.TYPE, "init_parameters": _fields}
+
+    @staticmethod
+    def from_dict(data: Dict[str, Any]) -> "BaseAdapter":
+        """
+        Converts a dictionary representation to an adapter object.
+        """
```
Amnah199, line 46: This method is needed only if we want serialize the adapter. Not sure if thats needed.

integrations/anthropic/src/haystack_integrations/components/generators/anthropic/generator.py
```diff
@@ -52,7 +55,9 @@ class AnthropicGenerator:
 
     def __init__(
         self,
-        api_key: Secret = Secret.from_env_var("ANTHROPIC_API_KEY"),  # noqa: B008
+        api_key: Optional[Secret] = Secret.from_env_var("ANTHROPIC_API_KEY"),  # noqa: B008
+        region: Optional[str] = None,
+        project_id: Optional[str] = None,
```
Amnah199, line 60: All parameters are kept optional, allowing users to select which ones to provide based on the targeted model. However, without an explicit option for users to choose between Vertex and Anthropic, it may lead to confusion. WDYT?

integrations/anthropic/src/haystack_integrations/components/generators/anthropic/generator.py
```diff
@@ -126,6 +147,12 @@ def run(self, prompt: str, generation_kwargs: Optional[Dict[str, Any]] = None):
          - `replies`: A list of generated replies.
          - `meta`: A list of metadata dictionaries for each reply.
         """
+
+        # Select the model adapter based on the provided env variables
+        model_adapter_cls = self.get_model_adapter()
+        client = model_adapter_cls.client()
+        model_name = model_adapter_cls.set_model(self.model)  # set the model name in the format required by the API
```
Amnah199, line 154: I would prefer to initialize the client locally if it doesnt have any implications.

integrations/anthropic/src/haystack_integrations/components/generators/anthropic/chat/vertex_chat_generator.py
```diff
@@ -0,0 +1,135 @@
...
+            - `top_k`: The top_k value to use for top-k sampling.
+            - `extra_headers`: A dictionary of extra headers to be passed to the model (i.e. for beta features).
+        :param ignore_tools_thinking_messages: Anthropic's approach to tools (function calling) resolution involves a
+            "chain of thought" messages before returning the actual function names and parameters in a message. If
+            `ignore_tools_thinking_messages` is `True`, the generator will drop so-called thinking messages when tool
+            use is detected. See the Anthropic [tools](https://docs.anthropic.com/en/docs/tool-use#chain-of-thought-tool-use)
+            for more details.
+        """
+        self.region = region or os.environ.get("REGION")
+        self.project_id = project_id or os.environ.get("PROJECT_ID")
```
silvanocerza, line 97: These should be `Secret`s I think. If you do it like this you can leak them when serializing the component.

Amnah199: In this case `project_id` and `region` do not grant access to GCP resources on their own, as proper authentication and permissions are required. So I thought these values are not sensitive data; but more of configuration parameters. Wdyt?

silvanocerza: Good point, we treated them as secrets on some other components though. Good for me to change this.

In any case I would make them mandatory arguments. If the user wants to pass them as env vars that's up to them.

silvanocerza (approved):
Great job! 👏

PR #1193, opened 2024-11-14
feat: Improvements to NvidiaRanker and adding user input timeout

sjrl:
### Related Issues

- fixes #issue-number

### Proposed Changes:

 
 
- Adds user specifiable timeout
- Improves the NvidiaRanker to have the same feature set as other rankers in Haystack

### How did you test it?

- added tests

### Checklist

### DIFF
--- a/integrations/nvidia/src/haystack_integrations/components/embedders/nvidia/document_embedder.py
+++ b/integrations/nvidia/src/haystack_integrations/components/embedders/nvidia/document_embedder.py
@@ -2,15 +2,18 @@
 #
 # SPDX-License-Identifier: Apache-2.0
 
+import os
 import warnings
 from typing import Any, Dict, List, Optional, Tuple, Union
 
-from haystack import Document, component, default_from_dict, default_to_dict
+from haystack import Document, component, default_from_dict, default_to_dict, logging
 from haystack.utils import Secret, deserialize_secrets_inplace
 from tqdm import tqdm
 
 from haystack_integrations.components.embedders.nvidia.truncate import EmbeddingTruncateMode
 from haystack_integrations.utils.nvidia import NimBackend, is_hosted, url_validation
+
+logger = logging.getLogger(__name__)
 
 _DEFAULT_API_URL = "https://ai.api.nvidia.com/v1/retrieval/nvidia"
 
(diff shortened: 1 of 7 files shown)


integrations/nvidia/src/haystack_integrations/components/embedders/nvidia/text_embedder.py
```diff
@@ -44,6 +47,7 @@ def __init__(
         prefix: str = "",
         suffix: str = "",
         truncate: Optional[Union[EmbeddingTruncateMode, str]] = None,
+        timeout: Optional[float] = None,
```
davidsbatista, line 50: This new parameter is missing in the `to_dict()`/serialization function

integrations/nvidia/src/haystack_integrations/components/generators/nvidia/generator.py
```diff
@@ -49,6 +50,7 @@ def __init__(
         api_url: str = _DEFAULT_API_URL,
         api_key: Optional[Secret] = Secret.from_env_var("NVIDIA_API_KEY"),
         model_arguments: Optional[Dict[str, Any]] = None,
+        timeout: Optional[float] = None,
```
davidsbatista, line 53: also here, same thing as above

integrations/nvidia/src/haystack_integrations/components/embedders/nvidia/text_embedder.py
sjrl: Yeah good point, I was following the pattern we have for the OpenAIGenerator where we also don't save the timeout in the to_dict. But maybe that was a bug and we should do that here and there? What do you think?

davidsbatista: I see - I would serialize it as well in both places.

sjrl: okay sounds good!

integrations/nvidia/src/haystack_integrations/components/embedders/nvidia/document_embedder.py
```diff
@@ -76,6 +80,9 @@ def __init__(
         :param truncate:
             Specifies how inputs longer that the maximum token length should be truncated.
```
davidsbatista, line 81:
```suggestion
            Specifies how inputs longer than the maximum token length should be truncated.
```

davidsbatista (approved):
LGTM

PR #1202, opened 2024-11-20
feat: Pgvector - recreate the connection if it is no longer valid

### LINKED ISSUE #847: Pgvector: preserve the connection
Yes, i'm currently using haystack pipeline & pgvector with FastAPI to serve embedding search but seems the current implementation of psycopg doesn't handle the reconnection after a long idle time Eg. overnight without any calls to API
the exception that thrown is `psycopg.OperationalError: the connection is lost`

i'd like to implement psycopg_pool to make it better in handling lost connection also for anyone who serve it via API this 
is more scaleable to use a connection pool.

Increasing DB Connection max lifetime but seems not a good practice, make a retry call to PG also doesn't work with current implementation, re-initiate the connection are impossible because of the connection var never be None once it initialized
https://github.com/deepset-ai/haystack-core-integrations/blob/60c666624cfb20a3600a28be074697fdf4cf2553/integrations/pgvector/src/haystack_integrations/document_stores/pgvector/document_store.py#L169-L173

**Additional context**
I'd be happy to make a PR for this myself, just want to make sure that I'm not missing anything before working on this.
open for discussion & also correct me if i'm wrong 🙇

anakin87:
### Related Issues

- fixes #847

### Proposed Changes:
- Every time we use the `connection`, we check whether it is still valid (implementation inspired by psycopg pool).
- If not, we try to close the existing connection and a new connection is created.

### How did you test it?
CI, new test

manual test:
- I created a Pgvector container
- in a Python shell, I created a `PgvectorDocumentStore` and called `count_documents`
- I stopped and restarted the container
- in the same Python shell, I called `count_documents` again
- everything works but internally the connection object has been recreated (it has a different `id()`)

### Notes for the reviewer
This might not be the best way to solve this problem, but it only requires a little refactoring and does not bring new dependencies (such as `psycopg_pool`).

### Checklist


integrations/pgvector/src/haystack_integrations/document_stores/pgvector/document_store.py
```diff
@@ -189,16 +201,40 @@ def _create_connection(self):
...
         self._create_keyword_index_if_not_exists()
 
         if self.search_strategy == "hnsw":
             self._handle_hnsw()
 
-        return self._connection
+        self._table_initialized = True
+
+    @staticmethod
+    def _check_connection(connection):
```
silvanocerza, line 225: I would rename this method to something else, it's not clear what check actually means at first glance.

Something like this probably:

```suggestion
    def _connection_is_valid(connection):
```

anakin87: right!

Done.

silvanocerza (approved):

PR #1213, opened 2024-11-22
feat: add `create_extension` parameter to control vector extension creation

wuqunfei:
Related Issues
Normal Postgresql User failed to connect PG host, because missing the admin role.
Create Extension need HIGH Permission in many cloud environment, like Azure https://learn.microsoft.com/en-us/azure/postgresql/extensions/how-to-allow-extensions?tabs=portal#create-extension

Proposed Changes:
Recommend to remove to check extension if exists pgvector in every connection.

Operational Level like DBA can create Vector extension once
Developer can connect it without considering this precondition.
How did you test it?
manual tests

Notes for the reviewer
Checklist
I have read the [contributors guidelines](https://github.com/deepset-ai/haystack-core-integrations/blob/main/CONTRIBUTING.md) and the [code of conduct](https://github.com/deepset-ai/haystack-core-integrations/blob/main/CODE_OF_CONDUCT.md)
I have updated the related issue with new insights and changes
I added unit tests and updated the docstrings
I've used one of the [conventional commit types](https://www.conventionalcommits.org/en/v1.0.0/) for my PR title: fix:, feat:, build:, chore:, ci:, docs:, style:, refactor:, perf:, test:.

### DIFF
--- a/.github/workflows/pgvector.yml
+++ b/.github/workflows/pgvector.yml
@@ -33,7 +33,7 @@ jobs:
         python-version: ["3.9", "3.10", "3.11"]
     services:
       pgvector:
-        image: ankane/pgvector:latest
+        image: pgvector/pgvector:pg17
         env:
           POSTGRES_USER: postgres
           POSTGRES_PASSWORD: postgres
--- a/integrations/pgvector/src/haystack_integrations/document_stores/pgvector/document_store.py
+++ b/integrations/pgvector/src/haystack_integrations/document_stores/pgvector/document_store.py
@@ -90,6 +90,7 @@ class PgvectorDocumentStore:
         hnsw_index_name: str = "haystack_hnsw_index",
         hnsw_ef_search: Optional[int] = None,
         keyword_index_name: str = "haystack_keyword_index",
+        create_extension: bool = True,
     ):
         """
         Creates a new PgvectorDocumentStore instance.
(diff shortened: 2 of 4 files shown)


wuqunfei:
Linked to the https://github.com/deepset-ai/haystack-core-integrations/pull/1203

wuqunfei:
check points all passed.

anakin87:
thanks I will have a look soon

.github/workflows/pgvector.yml
```diff
@@ -33,7 +33,7 @@ jobs:
         python-version: ["3.9", "3.10", "3.11"]
     services:
       pgvector:
-        image: ankane/pgvector:latest
+        image: pgvector/pgvector:pg17
```
anakin87, line 36: unrelated: I noticed that we were using a very old Docker image for testing and I updated it

anakin87 (approved):
I like this approach. Thanks for contributing! :blue_heart: 

(I pushed some small refinements)

anakin87:
Just released a new version of the package with this change: https://pypi.org/project/pgvector-haystack/1.2.0/

wuqunfei:
Amazing @anakin87, you are the best. !!! thanks for your hard work in the Friday night, thanks for the unit test and clear documentation. 

One Side, I agreed that the parameter need to order by alphabet. In the meanwhile, I prefer put new parameter in the end, because it could keep the old version class migrate into new version easily when developer initialized the class without parameter key name, for example: 
```
# Old Version 
store = PgvectorDocumentStore('postgresql://user:pwd@db:5432', ....)

# New Version 
store = PgvectorDocumentStore(True,'postgresql://user:pwd@db:5432', ....)

```

anakin87:
Thanks!

Since we force the parameter to be a keyword-only argument (not positional), I think it is ok to change the order.
https://realpython.com/python-asterisk-and-slash-special-parameters/

```python
    def __init__(
        self,
        *,
        connection_string: Secret = Secret.from_env_var("PG_CONN_STR"),
        create_extension: bool = True, ...
```

PR #1231, opened 2024-12-08
feat: warn if LangfuseTracer initialized without tracing enabled

### LINKED ISSUE #1230: Warn if LangfuseTracer initialized without HAYSTACK_CONTENT_TRACING_ENABLED
If LangfuseConnector component is initialized without HAYSTACK_CONTENT_TRACING_ENABLED=true and added to a pipeline, the pipeline will run successfully without any warnings, but the traces added to Languse will not have relevant information (only metadata).

Warn if LangfuseTracer initialized without HAYSTACK_CONTENT_TRACING_ENABLED.

**Additional context**

isfuku:
### Related Issues

- fixes #1230 

### Proposed Changes:

Add a warning during `LangfuseTracer` initialization in case tracing is disabled.

### How did you test it?

Tested locally by running the tracer with `LangufuseConnector` in the following cases:
- importing `haystack.Pipeline` before setting `HAYSTACK_CONTENT_TRACING_ENABLED` (warning raised)
- importing `haystack.Pipeline` after setting `HAYSTACK_CONTENT_TRACING_ENABLED` (no warning)

### Notes for the reviewer
I have not added unit tests because it requires to clear haystack modules and re-import it while having the appropriate HAYSTACK_CONTENT_TRACING_ENABLED env variable set. I though about adding it by removing haystack imports via sys.modules, and re-importing having HAYSTACK_CONTENT_TRACING_ENABLED=false for raising the warning, but apparently by doing so other tests fails, I think I would have to re-import the module while having HAYSTACK_CONTENT_TRACING_ENABLED=true. Let me know if adding a test for it is necessary. 

### Checklist

### DIFF
--- a/integrations/langfuse/src/haystack_integrations/tracing/langfuse/tracer.py
+++ b/integrations/langfuse/src/haystack_integrations/tracing/langfuse/tracer.py
@@ -7,7 +7,8 @@ from typing import Any, Dict, Iterator, List, Optional, Union
 from haystack import logging
 from haystack.components.generators.openai_utils import _convert_message_to_openai_format
 from haystack.dataclasses import ChatMessage
-from haystack.tracing import Span, Tracer, tracer
+from haystack.tracing import Span, Tracer
+from haystack.tracing import tracer as proxy_tracer
 from haystack.tracing import utils as tracing_utils
 
 import langfuse
@@ -78,7 +79,7 @@ class LangfuseSpan(Span):
         :param key: The content tag key.
         :param value: The content tag value.
         """
-        if not tracer.is_content_tracing_enabled:
+        if not proxy_tracer.is_content_tracing_enabled:
             return
         if key.endswith(".input"):
             if "messages" in value:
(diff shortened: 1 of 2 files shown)


julian-risch (changes requested):
@isfuku Thank you for contributing to Haystack! The changes look very good to me. Could you please just add a test before we merge? That test should check if the warning is triggered of we initialize the tracer while the env variable is not set to true. Should be something similar to 
```python
import logging

import pytest

...

def test_init_with_tracing_disabled(self, monkeypatch, caplog):
    monkeypatch.setenv("HAYSTACK_CONTENT_TRACING_ENABLED", “false”)
    with caplog.at_level(logging.WARNING):
        assert "tracing is disabled" in caplog.text
```
in `integrations/langfuse/tests/test_tracer.py`

julian-risch (approved):
LGTM! 👍 We could further extend the test with `@pytest.mark.parametrize` and also check that no warning is emitted if tracing is enabled. And we could create a helper function with the code that clears imported haystack modules so that it can be reused. Still, the code in its current shape is ready to be merged in my opinion.
Thank you and congratulations on your first contribution to Haystack @isfuku !

julian-risch:
I'll release a new version of the integration tomorrow.

Issue #1236, opened 2024-12-09
Update all components with `ChatMessage` to use `text` instead of `content`


anakin87:
part of https://github.com/deepset-ai/haystack/issues/8583

Since `content` will be removed in future in favor of `text` (and other types of content),
we are already transitioning to `text` to smooth the transition process.

PR #1242, opened 2024-12-10
fix: Google Vertex - fix the content type of `ChatMessage` `content` from function

anakin87:
### Related Issues

- part of #1236 (accessing `text` instead of `content`)
- twin PR of #1241 (see it for details and review it first)
  - I noticed that in messages from function, the `content` is a `dict` instead of `str` (see [Haystack `ChatMessage` definition](https://github.com/deepset-ai/haystack/blob/248dccbdd341941b988b622c51c26d1b9af66876/haystack/dataclasses/chat_message.py#L31)). This happened also because in integrations we skipping mypy checks for Haystack (this is not easy to fix)

### Proposed Changes:
- fix the content type
- use `text` instead of `content` to access the text

### How did you test it?
CI

### Notes for the reviewer
- This is a bug fix but also a breaking change. To make things clearer, I think that we can release a major version
- In the future, when new `ChatMessage` + `Tool` are adopted, we should refactor the tool calling section of this Chat Generator and I'm confident that we can handle it better

### Checklist

### DIFF
--- a/integrations/google_vertex/src/haystack_integrations/components/generators/google_vertex/chat/gemini.py
+++ b/integrations/google_vertex/src/haystack_integrations/components/generators/google_vertex/chat/gemini.py
@@ -1,3 +1,4 @@
+import json
 import logging
 from typing import Any, Callable, Dict, Iterable, List, Optional, Union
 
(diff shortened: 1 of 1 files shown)


vblagoje (approved):
Looks good! One change is that we previously, for some reason, dumped dict to content directly (which was a mistake) but now are properly json string serializing it. Isn't this going to break some existing clients? Even if it does - it's for the better, right?

anakin87:
Yes, this can potentially break the existing behavior. But it is actually a bug fix.
That is why I will release a new major version.

PR #1247, opened 2024-12-13
ci: delete all azure_ai_search indexes

Amnah199:
### Related Issues

- fixes [failing nightly tests](https://github.com/deepset-ai/haystack-core-integrations/actions/workflows/azure_ai_search.yml)  for azure_ai_search 

### Proposed Changes:

There may be an internal bug in Azure AI causing some index deletions to fail silently. These missed deletions during nightly tests can lead to a full index quota, causing subsequent tests to fail. 

I've added a `fixture` to clear all indexes from the search service at the end of each test session, preventing leftover indexes after nightly runs.
### How did you test it?

Ran all the tests
### Checklist


anakin87 (approved):
I was probably too optimistic.
I definitely see a risk of deleting indexes in use if tests for different operating systems and python versions run in parallel.

But let's try to see if it fixes the problem or creates other bugs.
I would only ask you to specify it clearly in the test docstring.

For the title, I would prefer something like
"ci: delete Azure AI test indexes"

PR #1277, opened 2025-01-06
fix: PgvectorDocumentStore - use appropriate schema name if dropping index

### LINKED ISSUE #1275: pgvector - hnsw index table drop is not using schema name
**Describe the bug**
If using pgvector, hnsw and multiple schema names (not the default one) the "hnsw index drop" does not drop the table, even if 
hnsw_recreate_index_if_exists=True.

It seems to me that there is missing schema_name before table_name in drop index SQL query in:
[pgvector/document_store.py](https://github.com/deepset-ai/haystack-core-integrations/blob/67e4ab6d4dd8fcbed33280688073c177f6f7f01d/integrations/pgvector/src/haystack_integrations/document_stores/pgvector/document_store.py) (row 392)

Current code:
`sql_drop_index = SQL("DROP INDEX IF EXISTS {index_name}").format(index_name=Identifier(self.hnsw_index_name))`

Should be:
`sql_drop_index = SQL("DROP INDEX IF EXISTS {schema_name}.{index_name}").format(schema_name=Identifier(self.schema_name),index_name=Identifier(self.hnsw_index_name))`

**To Reproduce**
1. Create PostgreSQL with pgvector using docker ([pgvector/pgvector:pg17](https://hub.docker.com/r/pgvector/pgvector))
2. Try to write documents 2 times with same table name (hnsw index should be dropped but it is not.
Minimal example code: [https://gist.github.com/mabartcz/4faa3e1ae7a56c8a9536b1ab8e7d4572](https://gist.github.com/mabartcz/4faa3e1ae7a56c8a9536b1ab8e7d4572)
3. Get error: "psycopg.errors.DuplicateTable: relation "test_table_index" already exists"

 - OS: Windows 11
 - pgvector-haystack: 1.2.0
 - haystack-ai: 2.8.0

If I am not missing something, I can create PR to fix the issue :)

mabartcz:
### Related Issues

- fixes #1275

### Proposed Changes:

In pgvector if HNSW index should drop, only index_name was used, but not schema_name. So if there were multiple schemas, the index was not dropped and could not be recreated.

### How did you test it?

The check was added directly to the code, so if the index existed and should be dropped, the Error is raised, because I consider it as critical Error.

### Notes for the reviewer

I tested this on multiple machines with similar setups, and the behavior was not consistent. On one PC this fix was needed, but on the other PC this fix was not needed, and it was working even without the fix, but I was not able to discover the difference...
But I believe that the fix is correct, and should be applied.

### Checklist

- [x]  I have read the [contributors guidelines](https://github.com/deepset-ai/haystack-core-integrations/blob/main/CONTRIBUTING.md) and the [code of conduct](https://github.com/deepset-ai/haystack-core-integrations/blob/main/CODE_OF_CONDUCT.md)
- [x]  I have updated the related issue with new insights and changes
- [x]  I added unit tests and updated the docstrings
- [x]  I've used one of the [conventional commit types](https://www.conventionalcommits.org/en/v1.0.0/) for my PR title: `fix:`, `feat:`, `build:`, `chore:`, `ci:`, `docs:`, `style:`, `refactor:`, `perf:`, `test:`.

### DIFF
--- a/integrations/pgvector/src/haystack_integrations/document_stores/pgvector/document_store.py
+++ b/integrations/pgvector/src/haystack_integrations/document_stores/pgvector/document_store.py
@@ -394,18 +394,6 @@ class PgvectorDocumentStore:
         )
         self._execute_sql(sql_drop_index, error_msg="Could not drop HNSW index")
 
-        index_exists_after_deletion = bool(
-            self._execute_sql(
-                "SELECT 1 FROM pg_indexes WHERE schemaname = %s AND tablename = %s AND indexname = %s",
-                (self.schema_name, self.table_name, self.hnsw_index_name),
-                "Could not check if HNSW index exists",
-            ).fetchone()
-        )
-
-        if index_exists and index_exists_after_deletion:
-            error_message = "The HNSW was not dropped"
-            raise DocumentStoreError(error_message)
-
         self._create_hnsw_index()
 
     def _create_hnsw_index(self):
--- a/integrations/pgvector/src/haystack_integrations/document_stores/pgvector/document_store.py
+++ b/integrations/pgvector/src/haystack_integrations/document_stores/pgvector/document_store.py
@@ -389,8 +389,22 @@ class PgvectorDocumentStore:
             )
             return
 
-        sql_drop_index = SQL("DROP INDEX IF EXISTS {index_name}").format(index_name=Identifier(self.hnsw_index_name))
+        sql_drop_index = SQL("DROP INDEX IF EXISTS {schema_name}.{index_name}").format(
+            schema_name=Identifier(self.schema_name), index_name=Identifier(self.hnsw_index_name)
+        )
         self._execute_sql(sql_drop_index, error_msg="Could not drop HNSW index")
+
+        index_exists_after_deletion = bool(
+            self._execute_sql(
+                "SELECT 1 FROM pg_indexes WHERE schemaname = %s AND tablename = %s AND indexname = %s",
+                (self.schema_name, self.table_name, self.hnsw_index_name),
+                "Could not check if HNSW index exists",
+            ).fetchone()
+        )
+
+        if index_exists and index_exists_after_deletion:
+            error_message = "The HNSW was not dropped"
+            raise DocumentStoreError(error_message)
 
         self._create_hnsw_index()
 
--- a/integrations/pgvector/tests/test_document_store.py
+++ b/integrations/pgvector/tests/test_document_store.py
@@ -2,9 +2,6 @@
 #
 # SPDX-License-Identifier: Apache-2.0
 
-import os
-import random
-import string
 from unittest.mock import patch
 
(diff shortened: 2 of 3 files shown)


integrations/pgvector/src/haystack_integrations/document_stores/pgvector/document_store.py
```diff
@@ -389,9 +389,23 @@ def _handle_hnsw(self):
...
+            self._execute_sql(
+                "SELECT 1 FROM pg_indexes WHERE schemaname = %s AND tablename = %s AND indexname = %s",
+                (self.schema_name, self.table_name, self.hnsw_index_name),
+                "Could not check if HNSW index exists",
+            ).fetchone()
+        )
+
+        if index_exists and index_exists_after_deletion:
+            error_message = "The HNSW was not dropped"
+            raise DocumentStoreError(error_message)
```
anakin87, line 407: I would remove this check.
If the HNSW index is not dropped, the following invocation `self._create_hnsw_index()` will fail. Right?

integrations/pgvector/tests/test_document_store.py
```diff
@@ -259,3 +263,41 @@ def test_from_pg_to_haystack_documents():
     assert haystack_docs[2].meta == {"meta_key": "meta_value"}
     assert haystack_docs[2].embedding == [0.7, 0.8, 0.9]
     assert haystack_docs[2].score is None
+
+
+@pytest.mark.integration
+def test_hnsw_index_recreation_in_new_schema():
```
anakin87, line 269: I think the test can be improved/changed.

I would probably first create a Document Store instance,
then create another one with `hnsw_recreate_index_if_exists=True` and verify in some way that the HNSW index is first dropped and then recreated.

I can help with the implementation, if needed.

anakin87 (commented):
Thanks for the PR!
I agree with the change, but I found some opportunities for improvement.

integrations/pgvector/src/haystack_integrations/document_stores/pgvector/document_store.py
mabartcz: You are right; I overthought that. The only reason it might fail during creation is if it wasn't deleted beforehand. However, you wouldn't know that was the cause. That said, this will be tested moving forward.

integrations/pgvector/tests/test_document_store.py
mabartcz: Your help would be greatly appreciated, as I’ve been trying to find a way to check if the index was recreated, but I haven’t found a solution yet. From this "high-level" scope, I can’t verify if the deletion was successful. Because it happened inside the "process".

anakin87: I find a way to test this, based on OIDs. Take a look if you want.

anakin87 (approved):
Thanks, @mabartcz.
Merging now!

Issue #1262, opened 2024-12-24
Langfuse Integration: start/end time of generators are not logged


marcklingen:
**Describe the bug**

While this works for non-LLM/generator nodes, generator nodes do not log start/end times to Langfuse.

Example trace: https://cloud.langfuse.com/project/cloramnkj0002jz088vzn1ja4/traces/42cb3022-b114-4e33-b8c7-774229d451b6?observation=9326286a-fdd4-4f12-8fe0-b22e480df767&display=details

**To Reproduce**

https://github.com/langfuse/langfuse-docs/blob/main/cookbook/integration_haystack.ipynb

**Describe your environment (please complete the following information):**

Colab with latest packages, see above.

vblagoje:
Thanks for logging this bug, @marcklingen. Just to confirm: once this bug is fixed, the Langfuse web UI will show how long it took to generate the response—similar to the other components in the screenshot, right?
<img width="448" alt="Image" src="https://github.com/user-attachments/assets/bbdf2f4b-0817-4d08-b86d-8c501157c763" />

marcklingen:
That's correct @vblagoje! This currently works for all components except for LLM/generator nodes

vblagoje:
Fixed and included in https://pypi.org/project/langfuse-haystack/0.7.0/ release

PR #1301, opened 2025-01-17
fix: End langfuse generation spans properly

vblagoje:
### Why:
Refines the tracing logic to handle different types of spans and their proper end()ing. 

- fixes https://github.com/deepset-ai/haystack-core-integrations/issues/1262

### What:
- Added support for finalizing both regular and generation spans by extending type checks.
- Introduced comments to clarify the span ending logic and lifecycle management.

### How can it be used:
Internal change not visible to clients

### How did you test it:
- Relying on existing unit tests
- Verified manually via `chat.py` example that generated this [trace](https://cloud.langfuse.com/trace/5c85a1ff-9b01-40fd-891d-042bb860a085)
- Contrast that trace to [this one](https://cloud.langfuse.com/project/clvdk4bcb000099hzwhts3g66/traces/4fdfd20a-fc29-4e61-a6ba-1e7062d78483) that doesn't have end() called for generation. The end result is that the latency for generation is not calculated and thus the entire trace doesn't have latency calculated as well. The difference is very subtle to notice. Below are images of both:
1) With generation.end() invoked:
Screenshot 2025-01-20 at 11 09 44

2) generation.end() NOT invoked: 
Screenshot 2025-01-20 at 11 10 09

  

### Notes for the reviewer:
NA

### DIFF
--- a/integrations/langfuse/src/haystack_integrations/tracing/langfuse/tracer.py
+++ b/integrations/langfuse/src/haystack_integrations/tracing/langfuse/tracer.py
@@ -202,7 +202,7 @@ class LangfuseTracer(Tracer):
 
         raw_span = span.raw_span()
 
-        # In this section, we finalize both regular spans and generation spans that were created using the LangfuseSpan class.
+        # In this section, we finalize both regular spans and generation spans created using the LangfuseSpan class.
         # It's important to end() these spans to ensure they are properly closed and all relevant data is recorded.
         # Note that we do not call end() on the main trace span itself, as its lifecycle is managed differently.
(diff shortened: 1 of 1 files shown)


vblagoje:
@julian-risch I think this is important fix but the review doesn't need to be rushed as we planned this item for the next week anyway.
 
cc @mpangrazzi @marcklingen

vblagoje:
@mpangrazzi can you please have a look - I wanted to combine this PR and https://github.com/deepset-ai/haystack-core-integrations/pull/1308 to release a new langfuse integration

mpangrazzi (approved):

Issue #1151, opened 2024-10-22
Add httpx_client parameter in LangfuseConnector


alex-stoica:
The current environment variables `LANGFUSE_SECRET_KEY` / `LANGFUSE_API_KEY`, `LANGFUSE_PUBLIC_KEY`, and `LANGFUSE_HOST` may not be sufficient for tracing in Langfuse.

For example, the following code:
```
tracer = Langfuse()
tracer.create_prompt(
    name="greeting-prompt",
    prompt="..."
)
```
might return an error:
```Unexpected error occurred. Please check your request and contact support: https://langfuse.com/support.
...
ConnectError: [SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed
```

However, this works correctly if the env variables are set correctly: 
```
Langfuse(
    httpx_client=httpx.Client(verify=os.getenv('LANGFUSE_PATH_TO_CERTIFICATE'))
)
```
To address this, please update the wrapper in `haystack_integrations.components.connectors.langfuse.LangfuseConnector` to accept `httpx_client` as a parameter. 
 The code fix would only affect the lines 
- https://github.com/deepset-ai/haystack-core-integrations/blob/61ac2f457c53697e8f906d9f44c39995a008f140/integrations/langfuse/src/haystack_integrations/components/connectors/langfuse/langfuse_connector.py#L108       
```self.tracer = LangfuseTracer(tracer=Langfuse(), name=name, public=public)``` 
->  ```tracer=Langfuse(httpx_client=httpx_client or httpx.Client(verify=os.getenv('LANGFUSE_PATH_TO_CERTIFICATE')))```
-  https://github.com/deepset-ai/haystack-core-integrations/blob/61ac2f457c53697e8f906d9f44c39995a008f140/integrations/langfuse/src/haystack_integrations/components/connectors/langfuse/langfuse_connector.py#L108` ->  `def __init__(self, name: str, public: bool = False, httpx_client: Optional[httpx.Client] = None)`

Thanks!

alex-stoica:
Tried locally to overwrite the LangfuseConnector in this way:

```@component
class LangfuseConnector:
    def __init__(self, name: str, public: bool = False, httpx_client: Optional[httpx.Client] = None):
        self.name = name
        self.tracer = LangfuseTracer(
            tracer=Langfuse(httpx_client=httpx_client or httpx.Client(verify=os.getenv('LANGFUSE_PATH_TO_CERTIFICATE')))                     
[...]

PR #1308, opened 2025-01-21
feat: LangfuseConnector - add httpx.Client init param

vblagoje:
### Why:
Enhances the LangfuseConnector by adding support for custom HTTP clients, allowing for greater flexibility and control over API communication.

- fixes https://github.com/deepset-ai/haystack-core-integrations/issues/1151

### What:
- Introduced an optional `httpx.Client` parameter to the `LangfuseConnector` to allow custom HTTP client usage.
- Updated constructor and documentation to include the new `httpx_client` parameter.
- Ensured that the `httpx_client` is integrated into the Langfuse API call handling.

### How can it be used:
With these changes, users can now pass a custom HTTP client to the LangfuseConnector as shown in the code snippet below:

```python
from httpx import Client
from haystack_integrations.components.connectors.langfuse import LangfuseConnector

custom_client = Client(timeout=10.0)
connector = LangfuseConnector(httpx_client=custom_client)
```

### How did you test it:
Manual testing was conducted to validate the integration of the custom `httpx.Client` with the LangfuseConnector's API calls, ensuring proper function and no regressions in behavior.

### Notes for the reviewer:
Attention should be given to ensure that the `httpx_client` integration does not introduce side effects or unintended behavior with existing configurations. Consider edge cases where no custom client is provided.

### DIFF
--- a/integrations/langfuse/pyproject.toml
+++ b/integrations/langfuse/pyproject.toml
@@ -22,7 +22,7 @@ classifiers = [
   "Programming Language :: Python :: Implementation :: CPython",
   "Programming Language :: Python :: Implementation :: PyPy",
 ]
-dependencies = ["haystack-ai>=2.9.0", "langfuse", "httpx"]
+dependencies = ["haystack-ai>=2.9.0", "langfuse"]
 
 [project.urls]
 Documentation = "https://github.com/deepset-ai/haystack-core-integrations/tree/main/integrations/langfuse#readme"
@@ -72,7 +72,6 @@ dependencies = [
     "black>=23.1.0",
     "mypy>=1.0.0",
     "ruff>=0.0.243",
-    "httpx",  # needed for types
 ]
 
 [tool.hatch.envs.lint.scripts]
@@ -173,6 +172,7 @@ module = [
   "haystack_integrations.*",
   "pytest.*",
   "numpy.*",
+  "httpx.*",
 ]
 ignore_missing_imports = true
 
--- a/integrations/langfuse/pyproject.toml
+++ b/integrations/langfuse/pyproject.toml
@@ -22,7 +22,7 @@ classifiers = [
   "Programming Language :: Python :: Implementation :: CPython",
   "Programming Language :: Python :: Implementation :: PyPy",
 ]
-dependencies = ["haystack-ai>=2.9.0", "langfuse"]
+dependencies = ["haystack-ai>=2.9.0", "langfuse", "httpx"]
 
 [project.urls]
 Documentation = "https://github.com/deepset-ai/haystack-core-integrations/tree/main/integrations/langfuse#readme"
@@ -67,7 +67,13 @@ python = ["3.8", "3.9", "3.10", "3.11", "3.12"]
 [tool.hatch.envs.lint]
 installer = "uv"
 detached = true
-dependencies = ["pip", "black>=23.1.0", "mypy>=1.0.0", "ruff>=0.0.243"]
+dependencies = [
+    "pip",
+    "black>=23.1.0",
+    "mypy>=1.0.0",
+    "ruff>=0.0.243",
+    "httpx",  # needed for types
+]
 
 [tool.hatch.envs.lint.scripts]
 typing = "mypy --install-types --non-interactive --explicit-package-bases {args:src/ tests}"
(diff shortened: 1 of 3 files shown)


integrations/langfuse/pyproject.toml
```diff
@@ -22,7 +22,7 @@ classifiers = [
   "Programming Language :: Python :: Implementation :: CPython",
   "Programming Language :: Python :: Implementation :: PyPy",
 ]
-dependencies = ["haystack-ai>=2.9.0", "langfuse"]
+dependencies = ["haystack-ai>=2.9.0", "langfuse", "httpx"]
```
anakin87, line 25: I see that `httpx` is a dependency of  `langfuse` (https://github.com/langfuse/langfuse-python/blob/d4ac21403c2de9cc490997355fb8249de1bd70c4/pyproject.toml#L11), so I would say that we don't need to explicitly depend on it

If there are issues related to mypy, adding a line here may help:
https://github.com/deepset-ai/haystack-core-integrations/blob/0678b7d13493569d0bc3ae59247180c0814a7bb2/integrations/langfuse/pyproject.toml#L163

vblagoje: Great points 🙏

integrations/langfuse/src/haystack_integrations/components/connectors/langfuse/langfuse_connector.py
```diff
@@ -112,6 +114,7 @@ def __init__(
             only accessible to the Langfuse account owner. The default is `False`.
         :param public_key: The Langfuse public key. Defaults to reading from LANGFUSE_PUBLIC_KEY environment variable.
         :param secret_key: The Langfuse secret key. Defaults to reading from LANGFUSE_SECRET_KEY environment variable.
+        :param httpx_client: Optional custom httpx.Client instance to use for Langfuse API calls.
```
anakin87, line 117: Perhaps we should mention the serialization limitation in the docstring.

anakin87 (approved):
Please improve the docstring, then feel free to merge.

integrations/langfuse/src/haystack_integrations/components/connectors/langfuse/langfuse_connector.py
vblagoje: Very good point. Have one last quick look again 🙏

integrations/langfuse/src/haystack_integrations/components/connectors/langfuse/langfuse_connector.py
```diff
@@ -112,6 +114,9 @@ def __init__(
             only accessible to the Langfuse account owner. The default is `False`.
         :param public_key: The Langfuse public key. Defaults to reading from LANGFUSE_PUBLIC_KEY environment variable.
         :param secret_key: The Langfuse secret key. Defaults to reading from LANGFUSE_SECRET_KEY environment variable.
+        :param httpx_client: Optional custom httpx.Client instance to use for Langfuse API calls. Note that this client
+            is not serializable, so when recreating a pipeline from YAML, any custom client passed here will not be
+            preserved - a new default client will be used instead.
```
anakin87, line 119: I'm not sure if this is what actually happens: my impression is that since the client is not serialized, the recreated Pipeline won't have this parameter. Could you verify?

anakin87 (changes requested):

integrations/langfuse/src/haystack_integrations/components/connectors/langfuse/langfuse_connector.py
vblagoje: But that's what the doc says, no?

vblagoje: Maybe put NOT in capital...

anakin87: ah you are referring to the Langfuse default client.
What I want to make clear is that in serde the HTTPX client is destroyed and never recreated...

vblagoje: Got it, how about https://github.com/deepset-ai/haystack-core-integrations/pull/1308/commits/357e5b2d05fec863f166a8299cd042fbcbd005b6 ?

anakin87 (approved):

PR #1312, opened 2025-01-21
feat: Update Ragas integration

sahusiddharth:
**Added `RagasEvaluator` component for Ragas version 0.2.* support**

This update introduces the new `RagasEvaluator` component, designed to be fully compatible with Ragas version 0.2.*. The component integrates with the latest changes and enhancements in the Ragas framework.


sahusiddharth:
Hi @davidsbatista, 

The test cases are failing in Python 3.9 due to a regex mismatch, although they are passing in Python 3.10. Could you please advise on the best way to proceed?

integrations/ragas/src/haystack_integrations/components/evaluators/ragas/evaluator.py
```diff
@@ -1,171 +1,299 @@
...
         Deserializes the component from a dictionary.
 
-        :param data:
-            Dictionary to deserialize from.
-        :returns:
-            Deserialized component.
+        :param data: Dictionary to deserialize from.
+        :returns: Deserialized component.
         """
         return default_from_dict(cls, data)
```
julian-risch, line 299: We can leave out the `from_dict` method entirely as it doesn't differ from the default implementation. Users can still call 
```python
from haystack_integrations.components.evaluators.ragas import RagasEvaluator
from haystack.core.serialization import component_from_dict, component_to_dict
evaluator = component_from_dict(RagasEvaluator, data=data, name="evaluator")
```

integrations/ragas/src/haystack_integrations/components/evaluators/ragas/evaluator.py
```diff
@@ -1,171 +1,299 @@
...
+            "multi_responses": "['Response 1', 'Response 2']",
+            "reference": "'A reference string'",
+            "rubrics": "{'score1': 'high_similarity'}",
+        }
+        return examples.get(field, "An appropriate value based on the field's type")
 
-        if not check_serializable(self.metric_params):
-            msg = "Ragas evaluator cannot serialize the metric parameters"
-            raise DeserializationError(msg)
+    def to_dict(self) -> Dict[str, Any]:
```
julian-risch, line 277: We can leave out the `to_dict` method entirely if it doesn't differ from the default implementation. Users can still call 
```python
from haystack_integrations.components.evaluators.ragas import RagasEvaluator
from haystack.core.serialization import component_from_dict, component_to_dict
evaluator = RagasEvaluator(...)
data = component_to_dict(evaluator, name="evaluator")
```

integrations/ragas/src/haystack_integrations/components/evaluators/ragas/evaluator.py
```diff
@@ -1,171 +1,299 @@
...
-from typing import Any, Callable, Dict, List, Optional, Union
+import re
+from typing import Any, Dict, List, Optional, Union, get_args, get_origin
 
-from datasets import Dataset
-from haystack import DeserializationError, component, default_from_dict, default_to_dict
+from haystack import Document, component, default_from_dict, default_to_dict
+from haystack.dataclasses import ChatMessage
+from langchain_core.embeddings import Embeddings as LangchainEmbeddings  # type: ignore
+from langchain_core.language_models import BaseLanguageModel as LangchainLLM  # type: ignore
```
julian-risch, line 7: Just out of curiosity, could you please briefly explain why the `# type: ignore` is needed here?

julian-risch (changes requested):
Hello @sahusiddharth and thank you for opening this pull request! I left a few comments, just smaller change requests. Regarding the failing test, I have an idea what's happening here. The regex and the output don't match for Python 3.9 because type(expected_type) outputs typing.Dict, which is not the same as `Dict`. In Python 3.10+, type(expected_type) outputs dict. `typing.Dict[str, str]` is just making it even more complex. (see https://peps.python.org/pep-0585/)

> Regex: "The 'rubrics' field expected 'one of Dict, NoneType', but got 'list'."
> E            Input: "Validation error occured while running RagasEvaluator Component:\nThe 'rubrics' field expected 'one of typing.Dict[str, str], NoneType', but got 'list'.\nHint: Provide {'score1': 'high_similarity'}"

This regex matching of the message is anyway error-prone and we should make the test more robust. Let's not try to match the exact error message with a regex.

integrations/ragas/src/haystack_integrations/components/evaluators/ragas/evaluator.py
sahusiddharth: Hi Julian,

The `# type: ignore` comment is needed in this case because when I ran the `hatch run lint:typing` command, I encountered the following errors:

```
src/haystack_integrations/components/evaluators/ragas/evaluator.py:6: error: Cannot find implementation or library stub for module named "langchain_core.embeddings" [import-not-found]
src/haystack_integrations/components/evaluators/ragas/evaluator.py:7: error: Cannot find implementation or library stub for module named "langchain_core.language_models" [import-not-found]
src/haystack_integrations/components/evaluators/ragas/evaluator.py:8: error: Cannot find implementation or library stub for module named "pydantic" [import-not-found]
src/haystack_integrations/components/evaluators/ragas/evaluator.py:10: error: Module "ragas" has no attribute "evaluate" [attr-defined]
```

To resolve this, I referred to a previous implementation where a similar issue was addressed with `# type: ignore`. You can see how it was solved in this [link](https://github.com/deepset-ai/haystack-core-integrations/blob/main/integrations/ragas/src/haystack_integrations/components/evaluators/ragas/evaluator.py#L7).

sahusiddharth: If I missed something or if there’s something else I should be doing, please feel free to let me know.

integrations/ragas/example/evaluation_from_pipeline_example.py
```diff
@@ -0,0 +1,140 @@
...
+    haystack_documents = response["answer_builder"]["answers"][0].documents
+    # extracting context from haystack documents
+    single_turn['retrieved_contexts'] = [doc.content for doc in haystack_documents]
+
+    evals_list.append(single_turn)
+
+evaluation_dataset = EvaluationDataset.from_list(evals_list)
+
+llm = ChatOpenAI(model="gpt-4o-mini")
+evaluator_llm = LangchainLLMWrapper(llm)
```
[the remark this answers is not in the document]
sahusiddharth: We use Langchain wrappers underneath to handle the evaluations, so if we move forward with Haystack, we might need to implement similar functionality ourselves or at least ensure the integration aligns with the existing workflow.

But we can definitely explore this further if you think it’s worth the effort!

julian-risch (approved):
Looks good to me! Thank you for addressing the review comments so quickly. I will take care of releasing a new version of the integration now. Feel free to open a PR that edits the example code in this [file](https://github.com/deepset-ai/haystack-integrations/blob/main/integrations/ragas.md) to update this overview page: https://haystack.deepset.ai/integrations/ragas

And yes, let's explore a HaystackLLMWrapper!

sahusiddharth:
I’m currently working on the code example and will raise a PR for it shortly. I’ll keep you updated once it’s ready!

sahusiddharth:
Hi @julian-risch,

I’ve raised the PR for the code example. Could you please take a look when you have a moment and let me know your thoughts?
- https://github.com/deepset-ai/haystack-integrations/pull/297

PR #1304, opened 2025-01-20
feat: AmazonBedrockChatGenerator - add tools support

vblagoje:
### Why:
Adds Haystack tool support to `AmazonBedrockChatGenerator`

- fixes https://github.com/deepset-ai/haystack/issues/8732

### What:
- Added method to convert Haystack tools to Amazon Bedrock toolConfig format, facilitating tool integration.
- Modified `AmazonBedrockChatGenerator` to accept, serialize, and handle tool specifications as part of the chat messages.
- Updated testing suite to cover tool-related functionalities, ensuring robustness and reliability.

### How can it be used:
The changes enable users to pass tools to `AmazonBedrockChatGenerator`, enhancing the model's capabilities by allowing it to use external tools within chat interactions. Here is a simplified example:
```python
from haystack.dataclasses import ChatMessage
from haystack_integrations.components.generators.amazon_bedrock import AmazonBedrockChatGenerator
from haystack.tools import Tool

def weather(city: str):
        """Get weather for a given city."""
        return f"The weather in {city} is sunny and 32°C"

tool_parameters = {"type": "object", "properties": {"city": {"type": "string"}}, "required": ["city"]}
tool = Tool(
    name="weather",
    description="useful to determine the weather in a given location",
    parameters=tool_parameters,
    function=weather,
)

messages = [ChatMessage.from_user("What's the weather like in Paris?")]
generator = AmazonBedrockChatGenerator(model="model_name", tools=[tool])
response = generator.run(messages=messages)
```

### How did you test it:
Implemented thorough unit and integration tests to validate the addition of tool-related functionalities. Tests included:
- Serialization and deserialization of tool configurations.
- Ensuring tool calls within conversations behave as expected.
- Integration tests with mock and live scenarios for tool usage within chat interactions.

### Notes for the reviewer:
Please pay special attention to the serialization and deserialization processes for tool objects and the handling of tool calls within the chat messages. Consider reviewing the edge cases in the test suite that handle different orderings of tool messages in replies.

PR #1314, opened 2025-01-22
refactor!: AmazonBedrockGenerator - remove truncation 

vblagoje:
### Why:
Remove truncation in AmazonBedrockGenerator. 
- Models have super long context nowadays and truncation is rarely needed
- We don't have truncation uniformly across generators
- We have extra dependency on transformers to calculate tokens
- Truncation can be done in a seperate component that feeds final prompt to generator

### What:
- Removed the `transformers` dependency from the `pyproject.toml`.
- Deprecated `max_length` and `truncate` parameters and associated warning logic from the generator.
- Deleted the `DefaultPromptHandler` class along with associated prompt truncation logic.

### How can it be used:
See above

### How did you test it:
- Updated tests to not test truncation
- Relying on exiting tests to make sure no regressions were made

### Notes for the reviewer:
Please focus on the removal of the prompt handler logic and verify that all necessary tests continue to pass. Special attention may be needed to confirm that `max_length` and `truncate` parameters' deprecation does not affect downstream functionality or produce any warnings at runtime.


vblagoje:
@mpangrazzi not to distract you with something that @anakin87 is already familiar with and has complete context. @anakin87 I know we are not following deprecation policy here but with generators being slowly removed anyways I thought it would be an overkill to follow the policy here - so I just eviscerated the truncation support altogether.

integrations/amazon_bedrock/src/haystack_integrations/components/generators/amazon_bedrock/generator.py
```diff
@@ -140,6 +138,15 @@ def __init__(
...
         self.max_length = max_length
         self.truncate = truncate
+
+        if max_length is not None or truncate is not None:
+            warnings.warn(
+                "The 'max_length' and 'truncate' parameters have been removed and no longer have any effect. "
+                "No truncation will be performed.",
+                DeprecationWarning,
+                stacklevel=2,
+            )
```
anakin87, line 148:
```suggestion
        if max_length is not None or truncate is not None:
            warnings.warn(
                "The 'max_length' and 'truncate' parameters have been removed and no longer have any effect. "
                "No truncation will be performed.",
            )
```

Technically, this is not just a deprecation, because the parameter no longer has effect. This is an (acceptable) breaking change. So let's modify the warning accordingly to make it more prominent.

anakin87 (changes requested):
In general, I like this change.

I left a comment.

I would also make this change more evident in the changelog with a PR title like:
"refactor!: AmazonBedrockGenerator - remove truncation "

anakin87 (approved):
Feel free to merge once linting errors are fixed (restore stacklevel).

vblagoje:
@anakin87 Let's batch it with https://github.com/deepset-ai/haystack-core-integrations/pull/1304 once it becomes ready so we can issue a new release at that time

PR #1317, opened 2025-01-23
refactor: GoogleAIGeminiGenerator - make some attributes public

sjrl:
### Related Issues

- fixes an issue in dC when we want to use streaming

### Proposed Changes:

 
 

Change class variables to be public instead of private to be consistent with our other generators. 

### How did you test it?

Existing tests

### Checklist


tstadel (approved):
LGTM!

anakin87:
@sjrl could you please better explain what the underlying issue is?

In any case, we don't enforce Generators to adhere to a specific protocol (they are just components),
so it's very likely that there are inconsistencies.

If this represents an issue in this particular case, I would prefer that the change be as small as possible. For example, I don't think that `_model` -which is basically a client- should be public.

sjrl:
Re your point here 

> For example, I don't think that _model -which is basically a client- should be public.

we do make the clients in our other generators public (ie OpenAI, Azure OpenAI, Cohere, Mistral, Bedrock, etc.) so if anything I'd say that Google AI is the odd one out here.

integrations/google_ai/src/haystack_integrations/components/generators/google_ai/gemini.py
```diff
@@ -100,12 +100,12 @@ def __init__(
...
-        self._generation_config = generation_config
-        self._safety_settings = safety_settings
-        self._model = GenerativeModel(self._model_name)
-        self._streaming_callback = streaming_callback
+        self.api_key = api_key
+        self.model_name = model
+        self.generation_config = generation_config
+        self.safety_settings = safety_settings
+        self.model = GenerativeModel(self.model_name)
+        self.streaming_callback = streaming_callback
```
anakin87: I would make only the following public: `generation_config`, `safety_settings` and `streaming_callback`.
If I'm not mistaken, changing these attributes at runtime actually modifies the behavior of the component.
(this is not true for `model_name`, for example)

anakin87 (commented):
I left a comment...

Please also change the PR title to something more expressive like
`refactor: GoogleAIGeminiGenerator - make some attributes public`.
In core integrations, the PR title is used to generate the CHANGELOG.md file.

integrations/google_ai/src/haystack_integrations/components/generators/google_ai/gemini.py
sjrl: I agree with that, but I also may want as a user to easily inspect the name of the model I'm using in a component when working with these in python and it would feel odd that I would need to access a private variable to do so, right?

anakin87: ok let's leave `model_name` public. What about `api_key`?

sjrl: Maybe in the spirit of minimizing changes let's leave it internal. But it is public in most of our other components. Perhaps a discussion on standardizing what should and shouldn't be public would be a good idea?

sjrl:
> > Re your point here
> > > For example, I don't think that _model -which is basically a client- should be public.
> > 
> > 
> > we do make the clients in our other generators public (ie OpenAI, Azure OpenAI, Cohere, Mistral, Bedrock, etc.) so if anything I'd say that Google AI is the odd one out here.
> 
> IMHO, the client should be internal (as we do for example in the HF API), but I can only agree with you that most of the time we made it public.

I definitely see your point! I'd be happy to discuss and make this change towards more private variables in the future :) and fine to leave this one private now as well.

anakin87 (approved):
Feel free to merge and, if you release a new version of the package (maybe 5.0.1), do it right after the merge (before other PRs are merged).

Otherwise, the CHANGELOG will be wrong.

edit: test is failing for unrelated reasons (usage quota), so feel free to merge.

PR #1318, opened 2025-01-23
feat: CohereChatGenerator - add tools support 

vblagoje:
### Why:
Upgrades the Cohere chat API integration from v1 to v2 and adds modern Haystack tools support

### What:
- Migrated the `CohereChatGenerator` to use the v2 Cohere chat API, incorporating tools as a key parameter 
- Improved message and response formatting to comply with the v2 Cohere chat API specifications
- Added tool handling logic to enable seamless integration of predefined tools
- Updated tests to cover new functionality, including tool integration and live streaming scenarios.

### How can it be used:
The changes enable the chatbot to leverage predefined tools dynamically in conversation, using the v2 Cohere chat API. For example, you can define a tool for weather information:
```python
from haystack import Pipeline
from haystack.dataclasses import ChatMessage
from haystack.components.tools import ToolInvoker
from haystack.tools import Tool
from haystack_integrations.components.generators.cohere import CohereChatGenerator

# Create a weather tool
def weather(city: str) -> str:
    return f"The weather in {city} is sunny and 32°C"

weather_tool = Tool(
    name="weather",
    description="useful to determine the weather in a given location",
    parameters={
        "type": "object",
        "properties": {
            "city": {
                "type": "string",
                "description": "The name of the city to get weather for, e.g. Paris, London",
            }
        },
        "required": ["city"],
    },
    function=weather,
)

# Create and set up the pipeline
pipeline = Pipeline()
pipeline.add_component("generator", CohereChatGenerator(model="command-r", tools=[weather_tool]))
pipeline.add_component("tool_invoker", ToolInvoker(tools=[weather_tool]))
pipeline.connect("generator", "tool_invoker")

# Run the pipeline with a weather query
results = pipeline.run(
    data={"generator": {"messages": [ChatMessage.from_user("What's the weather like in Paris?")]}}
)

# The tool result will be available in the pipeline output
print(results["tool_invoker"]["tool_messages"][0].tool_call_result.result)
# Output: "The weather in Paris is sunny and 32°C"
```

### How did you test it:
Test cases were expanded to validate new functionalities, including:
- Unit tests for message formatting and tool usage within the `CohereChatGenerator`.
- Integration tests for tool invocation scenarios ensuring proper execution and results.
- Live tests to verify interaction with the v2 Cohere chat API, confirming that real-time data is appropriately processed.

### Notes for the reviewer:
[...]

PR #1321, opened 2025-01-24
refactor: Migrate Cohere to V2

vblagoje:
### Why:
- Migrates our Cohere integration to V2 (embedder(s) and ranker).
- Embedders can now use `embedding_types` to choose the size of the embeddings, providing more control over memory usage and performance.
- (Chat)Generator migration is done in a separate [PR](https://github.com/deepset-ai/haystack-core-integrations/pull/1318).

### What:
- Introduced `EmbeddingTypes` enum to define and manage supported embedding types:
  - **Allowed values**:
    - `float`: Default float embeddings (valid for all models).
    - `int8`: Signed int8 embeddings (valid for v3 models only).
    - `uint8`: Unsigned int8 embeddings (valid for v3 models only).
    - `binary`: Signed binary embeddings (valid for v3 models only).
    - `ubinary`: Unsigned binary embeddings (valid for v3 models only).
- Updated `CohereDocumentEmbedder` and `CohereTextEmbedder` to handle multiple embedding types.
- Updated `CohereRanker` to use Cohere V2 API, added `max_tokens_per_doc` for token-level control, and deprecated the use of `max_chunks_per_doc` (now ignored).

### How can it be used:
Users can specify the desired embedding type and initialize the embedders or ranker:

```python
from haystack_integrations.components.embedders.cohere import CohereDocumentEmbedder
from haystack_integrations.components.embedders.cohere.embedding_types import EmbeddingTypes
from haystack_integrations.components.rankers.cohere import CohereRanker

embedder = CohereDocumentEmbedder(
    api_key="your_api_key",
    model="embed-english-v3.0",
    embedding_type=EmbeddingTypes.INT8,
)

ranker = CohereRanker(
    api_key="your_api_key",
    model="rerank-english-v2.0",
    max_tokens_per_doc=500,
)
```

### How did you test it:
- Updated and added unit tests for `CohereDocumentEmbedder` and `CohereTextEmbedder` to validate functionality with multiple embedding types.
- Verified correct handling of embedding type combinations (e.g., `int8`, `binary`).
- Updated  `CohereRanker` tests

### Notes for the reviewer:
- Check the handling of `embedding_types` in the embedders to ensure that only valid types are accepted for v3 models.
- Verify that `max_tokens_per_doc` in `CohereRanker` works as intended and aligns with the Cohere V2 API.
- Ensure the integration supports seamless switching between embedding types and document configurations.
- To be merged in tandem with [PR](https://github.com/deepset-ai/haystack-core-integrations/pull/1318) and then a new major version of cohere-haystack integration should be released.

### DIFF
--- a/integrations/cohere/src/haystack_integrations/components/rankers/cohere/ranker.py
+++ b/integrations/cohere/src/haystack_integrations/components/rankers/cohere/ranker.py
@@ -40,7 +40,7 @@ class CohereRanker:
         max_chunks_per_doc: Optional[int] = None,
         meta_fields_to_embed: Optional[List[str]] = None,
         meta_data_separator: str = "\n",
-        max_tokens_per_doc: int = 4096
+        max_tokens_per_doc: int = 4096,
     ):
         """
         Creates an instance of the 'CohereRanker'.
--- a/integrations/cohere/src/haystack_integrations/components/rankers/cohere/ranker.py
+++ b/integrations/cohere/src/haystack_integrations/components/rankers/cohere/ranker.py
@@ -40,6 +40,7 @@ class CohereRanker:
         max_chunks_per_doc: Optional[int] = None,
         meta_fields_to_embed: Optional[List[str]] = None,
         meta_data_separator: str = "\n",
+        max_tokens_per_doc: Optional[int] = None,
     ):
         """
         Creates an instance of the 'CohereRanker'.
@@ -57,6 +58,7 @@ class CohereRanker:
             with the document content for reranking.
         :param meta_data_separator: Separator used to concatenate the meta fields
             to the Document content.
+        :param max_tokens_per_doc: The maximum number of tokens to embed for each document.
         """
         self.model_name = model
         self.api_key = api_key
@@ -65,7 +67,18 @@ class CohereRanker:
         self.max_chunks_per_doc = max_chunks_per_doc
         self.meta_fields_to_embed = meta_fields_to_embed or []
         self.meta_data_separator = meta_data_separator
-        self._cohere_client = cohere.Client(
+        self.max_tokens_per_doc = max_tokens_per_doc
+        if max_chunks_per_doc is not None:
+            # Note: max_chunks_per_doc is currently not supported by the Cohere V2 API
+            # See: https://docs.cohere.com/reference/rerank
+            import warnings
+
+            warnings.warn(
+                "The max_chunks_per_doc parameter currently has no effect as it is not supported by the Cohere V2 API.",
+                UserWarning,
+                stacklevel=2,
+            )
+        self._cohere_client = cohere.ClientV2(
             api_key=self.api_key.resolve_value(), base_url=self.api_base_url, client_name="haystack"
         )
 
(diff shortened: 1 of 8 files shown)


vblagoje:
@davidsbatista please review ranker changes. This is for the next week @anakin87 @davidsbatista

anakin87 (commented):
I left some comments, but in general I agree with the direction of this PR.

Just one question: should we use the opportunity of this refactoring to move the client definition to `__init__`?

integrations/cohere/src/haystack_integrations/components/rankers/cohere/ranker.py
```diff
@@ -152,7 +166,7 @@ def run(self, query: str, documents: List[Document], top_k: Optional[int] = None
             model=self.model_name,
             query=query,
             documents=cohere_input_docs,
-            max_chunks_per_doc=self.max_chunks_per_doc,
+            max_tokens_per_doc=self.max_tokens_per_doc,
```
davidsbatista, line 169: I would suggest to set at the same default value as the API sets if nothing is given.

```suggestion
            max_tokens_per_doc=self.max_tokens_per_doc or 4096,
```

integrations/cohere/src/haystack_integrations/components/rankers/cohere/ranker.py
```diff
@@ -57,6 +58,7 @@ def __init__(
             with the document content for reranking.
         :param meta_data_separator: Separator used to concatenate the meta fields
             to the Document content.
+        :param max_tokens_per_doc: The maximum number of tokens to embed for each document.
```
davidsbatista, line 61:
```suggestion
        :param max_tokens_per_doc: The maximum number of tokens to embed for each document defaults to 4096.
```

integrations/cohere/tests/test_cohere_ranker.py
```diff
@@ -48,6 +48,7 @@ def test_init_default(self, monkeypatch):
         assert component.max_chunks_per_doc is None
         assert component.meta_fields_to_embed == []
         assert component.meta_data_separator == "\n"
+        assert component.max_tokens_per_doc is None
```
davidsbatista, line 51:
```suggestion
        assert component.max_tokens_per_doc = 4096
```

davidsbatista (commented):
LGTM, I would just suggest to replace the `None` by 4096 as this is the default value from Chohere API V2 (https://docs.cohere.com/v2/reference/rerank)

davidsbatista (approved):
LGTM from my side

anakin87:
Looks good, but please answer this question:
> should we use the opportunity of this refactoring to move the client definition to `__init__`?

davidsbatista:
@anakin87 so that we can then simply do, for instance: 

`from haystack_integrations.amazon_bedrock import AmazonBedrockDocumentEmbedder` 

or any other component?

I would vote for this simplification in importing

anakin87:
@davidsbatista no no, sorry if I did not explain myself better (we have discussed this with Vladimir in the past).

Currently, the Cohere client is recreated every time the `run` method is called.
In other components, we usually create the client at initialization and reuse it, so I propose to consider this refactoring as an opportunity to introduce this improvement.

vblagoje:
Ah right, @anakin87 apologies - too many context switches. Yes, let's fix that one in a separate PR to ease the cognitive load - as IIRC we agreed privately?

anakin87 (approved):
>  Yes, let's fix that one in a separate PR to ease the cognitive load - as IIRC we agreed privately?

OK

PR #1333, opened 2025-01-29
feat: Google AI Chat Generator - add tool config + improvements

anakin87:
### Related Issues
While working on the Vertex PR (#1327), I thought of several improvements for Google AI.

### Proposed Changes:
- add support for [`ToolConfig`](https://ai.google.dev/api/caching#ToolConfig)
- simplify conversion of Generation Config to dictionary
- better streaming support, in line with Vertex: stream also function calls and handle usage metadata

### How did you test it?
CI, new test

### Checklist


integrations/google_ai/src/haystack_integrations/components/generators/google_ai/chat/gemini.py
```diff
@@ -353,25 +348,42 @@ def _stream_response_and_convert_to_messages(
...
+                    content_to_stream += new_text
+                    text += new_text
+                elif new_function_call := part.get("function_call"):
+                    content_to_stream += json.dumps(dict(new_function_call))
                     tool_calls.append(
                         ToolCall(
-                            tool_name=part["function_call"]["name"],
-                            arguments=dict(part["function_call"]["args"]),
+                            tool_name=new_function_call["name"],
+                            arguments=dict(new_function_call["args"]),
```
vblagoje, line 369: do we need these two dict wrappers?

vblagoje (commented):
Just one question out of curiosity!

integrations/google_ai/src/haystack_integrations/components/generators/google_ai/chat/gemini.py
anakin87: thanks for asking...

I've verified that in the `_stream_response_and_convert_to_messages` method we don't need `dict` because we are already using a dictionary (`chunk.to_dict()`).

It was probably a copy-paste from `_convert_response_to_messages`, where we need `dict`, otherwise `part.function_call.args` is a `proto.marshal.collections.maps.MapComposite` object.

vblagoje (approved):
Let's go 🚀

Issue #1340, opened 2025-02-01
Add LIKE and NOT_LIKE operator for pgvector document store


rhajou:
Currently the metadata used for filtering are limited to : 
- `==`
- `!=`
- `>`
- `>=`
- `<`
- `<=`
- `in`
- `not in`

According to this haystack page: [here](https://docs.haystack.deepset.ai/docs/metadata-filtering)

I came upon a problem that I wanted to check if a specific metadata include a specific string (this is similar to str contains in python). 

I resolved this by manually adding the LIKE and NOT_LIKE operator to haystack:

```
from haystack_integrations.document_stores.pgvector.filters import COMPARISON_OPERATORS
from typing import Any, Tuple

def _like(field: str, value: Any) -> Tuple[str, Any]:
    if not isinstance(value, str):
        msg = f"{field}'s value must be a str when using 'LIKE' "
        raise FilterError(msg)
    return f"{field} LIKE %s", value


def _not_like(field: str, value: Any) -> Tuple[str, Any]:
    if not isinstance(value, str):
        msg = f"{field}'s value must be a str when using 'LIKE' "
        raise FilterError(msg)
    return f"{field} NOT LIKE %s", value


COMPARISON_OPERATORS["LIKE"] = _like
COMPARISON_OPERATORS["NOT_LIKE"] = _not_like

```
Working well! I suggest to integrate this with the library

PR #1341, opened 2025-02-01
feat: Pgvector - add like and not like filters

rhajou:
### Related Issues

- fixes: #1340
### Proposed Changes:
Currently the metadata used for filtering are limited to : 
- `==`
- `!=`
- `>`
- `>=`
- `<`
- `<=`
- `in`
- `not in`

According to this haystack page: [here](https://docs.haystack.deepset.ai/docs/metadata-filtering)

I came upon a problem that I wanted to check if a specific metadata include a specific string (this is similar to str contains in python). 

I resolved this by manually adding the LIKE and NOT_LIKE operator to haystack pg vector.

Maybe we need to update it for all other databases as welll?

### DIFF
--- a/integrations/pgvector/src/haystack_integrations/document_stores/pgvector/filters.py
+++ b/integrations/pgvector/src/haystack_integrations/document_stores/pgvector/filters.py
@@ -238,11 +238,13 @@ def _like(field: str, value: Any) -> Tuple[str, Any]:
         raise FilterError(msg)
     return f"{field} LIKE %s", value
 
+
 def _not_like(field: str, value: Any) -> Tuple[str, Any]:
     if not isinstance(value, str):
         msg = f"{field}'s value must be a str when using 'LIKE' "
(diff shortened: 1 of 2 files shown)


anakin87:
Hey, thanks for the contribution!

For the moment, let's only add these operators to Pgvector.

Could you please add some tests to test the behavior of these new operators here in `TestFilters`?
https://github.com/deepset-ai/haystack-core-integrations/blob/958aace78165985964297a0298310ca8a18f8665/integrations/pgvector/tests/test_filters.py#L20C19-L20C38

The tests should be similar to this: https://github.com/deepset-ai/haystack/blob/379711f63ee8e58181194eaf58beea052723a3f4/haystack/testing/document_store.py#L279

rhajou:
> Looks good.
> 
> Thanks for also adding the tests!

Thanks! will work later on adding these operators if needed to other vector databases

PR #1347, opened 2025-02-03
ci: skip tests when only CHANGELOG.md is modified

anakin87:
### Related Issues

After #1330, every time we release a new integration, a PR with the proposed CHANGELOG change is opened and tests run on it, wasting time and computational resources.

### Proposed Changes:
- for each integration, do not run tests if only CHANGELOG.md is modified

### How did you test it?
Not easy to test before merging.
Some tests failures are unrelated to this PR.

### Checklist


anakin87:
> (nit: It may also be worth to add a similar rule for README.md?)

Good idea! I will exclude all .md files in the root folder of each integration.

PR #1358, opened 2025-02-06
adding tools to MistralChatGenerator - based on OpenAIChatGenerator

davidsbatista:
### Related Issues

- fixes https://github.com/deepset-ai/haystack/issues/8736

### Proposed Changes:

- mostly based on OpenAIChatGenerator code
 
### Checklist

### DIFF
--- a/integrations/mistral/src/haystack_integrations/components/generators/mistral/chat/chat_generator.py
+++ b/integrations/mistral/src/haystack_integrations/components/generators/mistral/chat/chat_generator.py
@@ -1,14 +1,17 @@
 # SPDX-FileCopyrightText: 2023-present deepset GmbH <info@deepset.ai>
 #
 # SPDX-License-Identifier: Apache-2.0
+
+import json
 from typing import Any, Callable, Dict, List, Optional
 
-from haystack import component
+from haystack import component, logging
 from haystack.components.generators.chat import OpenAIChatGenerator
-from haystack.dataclasses import StreamingChunk
+from haystack.dataclasses import StreamingChunk, ChatMessage, ToolCall
 from haystack.tools import Tool
 from haystack.utils.auth import Secret
 
+logger = logging.getLogger(__name__)
 
 @component
 class MistralChatGenerator(OpenAIChatGenerator):
@@ -55,7 +58,7 @@ class MistralChatGenerator(OpenAIChatGenerator):
     def __init__(
         self,
         api_key: Secret = Secret.from_env_var("MISTRAL_API_KEY"),
-        model: str = "mistral-tiny",
+        model: str = "mistral-small-latest",
         streaming_callback: Optional[Callable[[StreamingChunk], None]] = None,
         api_base_url: Optional[str] = "https://api.mistral.ai/v1",
         generation_kwargs: Optional[Dict[str, Any]] = None,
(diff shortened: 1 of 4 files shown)


integrations/mistral/tests/test_mistral_chat_generator.py
```diff
@@ -267,3 +281,17 @@ def __call__(self, chunk: StreamingChunk) -> None:
 
         assert callback.counter > 1
         assert "Paris" in callback.responses
+
+    def test_live_run_with_tools(self, tools):
```
anakin87, line 285: A round trip test like [this](https://github.com/deepset-ai/haystack-core-integrations/blob/fd5def89af2a25041ae3117549d5ce410ec2c8af/integrations/google_ai/tests/generators/chat/test_chat_gemini.py#L362) would be more comprehensive: it tests tool call generation but also how the model behaves when you pass a tool call result.

integrations/mistral/tests/test_mistral_chat_generator.py
```diff
@@ -267,3 +281,17 @@ def __call__(self, chunk: StreamingChunk) -> None:
...
+        chat_messages = [ChatMessage.from_user("What's the weather like in Paris?")]
+        component = MistralChatGenerator(tools=tools)
+        results = component.run(chat_messages)
+        assert len(results["replies"]) == 1
+        message = results["replies"][0]
+
+        assert message.tool_calls
+        tool_call = message.tool_call
```
anakin87, line 293: we discussed about the removal of `assert not message.text`...

Which kind of text does the model produce along with a tool call?
Is it expected? I can't find it mentioned in [Mistral docs], but I haven't studied the topic in depth.

integrations/mistral/src/haystack_integrations/components/generators/mistral/chat/chat_generator.py
```diff
@@ -95,4 +104,80 @@ def __init__(
             api_base_url=api_base_url,
             organization=None,
             generation_kwargs=generation_kwargs,
+            tools=tools,
+        )
+
+    def to_dict(self) -> Dict[str, Any]:
```
anakin87, line 110: could you explain why this method and the following one need to be overridden?

integrations/mistral/src/haystack_integrations/components/generators/mistral/chat/chat_generator.py
```diff
@@ -54,10 +60,11 @@ class MistralChatGenerator(OpenAIChatGenerator):
     def __init__(
         self,
         api_key: Secret = Secret.from_env_var("MISTRAL_API_KEY"),
-        model: str = "mistral-tiny",
+        model: str = "mistral-small-latest",
```
anakin87, line 63: if there is a good reason to change the default model, let's mark this PR as breaking ("!" in the title).
In any case, the title needs to be aligned with the common format.

integrations/mistral/src/haystack_integrations/components/generators/mistral/chat/chat_generator.py
vblagoje: yeah, of course. We have a mismatch with super class when it comes to init params. So it seems like we have two options:
1) Match the params from OpenAIChatGenerator
2) Adjust serialization of only selected parameters

In case of 2, if we don't do this, the serde tests were failing because we were init-ing params like max_retries from super class. 
We can also do option 1) and remove to_dict()

wdyt?

integrations/mistral/src/haystack_integrations/components/generators/mistral/chat/chat_generator.py
vblagoje: ok yeah, will do mark it as breaking, couldn't find this model listed any more nor supporting FC, see https://docs.mistral.ai/capabilities/function_calling/

integrations/mistral/src/haystack_integrations/components/generators/mistral/chat/chat_generator.py
anakin87: ok I think that your strategy makes sense. Please add a short comment in the code explaining this, so we remember in future.

integrations/mistral/src/haystack_integrations/components/generators/mistral/chat/chat_generator.py
```diff
@@ -95,4 +104,80 @@ def __init__(
...
+            self,
+            model=self.model,
+            streaming_callback=callback_name,
+            api_base_url=self.api_base_url,
+            generation_kwargs=self.generation_kwargs,
+            api_key=self.api_key.to_dict(),
+            tools=[tool.to_dict() for tool in self.tools] if self.tools else None,
         )
+
+    def _convert_streaming_chunks_to_chat_message(self, chunk: Any, chunks: List[StreamingChunk]) -> ChatMessage:
```
anakin87, line 128: please also explain this

anakin87 (approved):
- Please address the 2 small comments I've left.
- Change the title to reflect our conventions.

Then feel free to merge!

PR #1361, opened 2025-02-10
fix: Chroma - make `filter_documents` method return embeddings

### LINKED ISSUE #1286: `filter_documents()` Method Missing Embeddings in Retrieval with filter_documents() in ChromaDocumentStore class
**Description:**  
I encountered an issue while trying to search for persisted embeddings using the `filter_documents()` method. The method fetches all documents, but the embeddings appear as `NoneType`. The embeddings are indeed stored in the persisted ChromaDB; however, the `filter_documents()` method retrieves the data using:  

```python
result = self._collection.get()
```

This approach seems to omit the embeddings. When modified to explicitly include embeddings like so:

```python
result = self._collection.get(including=['documents', 'embeddings', 'metadatas'])
```

(as done in the retriever), it works as expected.  

**Proposed Fix:**  
Update the `filter_documents()` method to include the `including=['documents', 'embeddings', 'metadatas']` parameter when calling `self._collection.get()` to ensure embeddings are returned correctly.

Thank you for looking into this issue! Let me know if you need further details or a code sample.

https://github.com/deepset-ai/haystack-core-integrations/blob/e2a008aac4aa37becf325f51bc521ee326b6cb4e/integrations/chroma/src/haystack_integrations/document_stores/chroma/document_store.py#L222

anakin87:
### Related Issues

- fixes #1286

### Proposed Changes:
- make `filter_documents` method return embeddings using the `include` kwarg

### How did you test it?
CI; new test

### Checklist


integrations/chroma/src/haystack_integrations/document_stores/chroma/document_store.py
```diff
@@ -416,8 +417,11 @@ def _get_result_to_documents(result: GetResult) -> List[Document]:
                 document_dict["meta"] = result_metadata[i]
 
             result_embeddings = result.get("embeddings")
-            if result_embeddings:
-                document_dict["embedding"] = list(result_embeddings[i])
+            if result_embeddings is not None:
+                if isinstance(result_embeddings[i], ndarray):
```
Amnah199, line 421: I looked into the API and I believe the embeddings are always converted to a np.array. [Source here](https://github.com/chroma-core/chroma/blob/8ca12208aa3d3b33192a4a0f8b37501fd8293bfb/chromadb/api/models/CollectionCommon.py#L463).
If thats the case, I think we wouldn't need this check.

anakin87: Mypy was complaining, so I looked at [`GetResult`](https://github.com/chroma-core/chroma/blob/8ca12208aa3d3b33192a4a0f8b37501fd8293bfb/chromadb/api/types.py#L376) and it seems that they also accept a `PyVector`: https://github.com/chroma-core/chroma/blob/8ca12208aa3d3b33192a4a0f8b37501fd8293bfb/chromadb/types.py#L203

Amnah199: Got it!

Amnah199 (approved):

PR #1370, opened 2025-02-12
chore: pgvector - remove support for dataframe

anakin87:
### Related Issues

- part of #1371

### Proposed Changes:

This is the first PR to remove support for `dataframe` in document stores.
I'm doing that for Pgvector.
This is a breaking change -> I will release a major version of this integration.

### How did you test it?
CI;
I also tested locally that a table created with the current version of the Document Store still works with the new version (but does not support writing `dataframe`).

### Checklist

### DIFF
--- a/integrations/pgvector/pyproject.toml
+++ b/integrations/pgvector/pyproject.toml
@@ -163,6 +163,7 @@ markers = ["integration: integration tests"]
 module = [
   "haystack.*",
   "haystack_integrations.*",
+  "numpy.*",
   "pgvector.*",
   "psycopg.*",
   "pytest.*",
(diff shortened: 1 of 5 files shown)


integrations/pgvector/pyproject.toml
```diff
@@ -163,6 +163,7 @@ markers = ["integration: integration tests"]
 module = [
   "haystack.*",
   "haystack_integrations.*",
+  "numpy.*",
```
anakin87, line 166: this is unrelated, but probably needed by new mypy 1.15.0

davidsbatista (approved):
LGTM - I was looking for a test with a Document containing a Dataframe, but Dataframe support in Document will be dropped, doesn't make sense to have it.

Issue #1371, opened 2025-02-12
Remove `dataframe` support from Document Stores in core-integrations


anakin87:

anakin87:
done in the linked PRs

PR #1377, opened 2025-02-14
chore!: Elasticsearch - remove dataframe support

anakin87:
### Related Issues

- part of #1371

### Proposed Changes:
- Remove dataframe support from Elasticsearch
- This is a breaking change -> I will release a major version of this integration.

### How did you test it?
CI; new tests

### Checklist

### DIFF
--- a/integrations/elasticsearch/tests/test_document_store.py
+++ b/integrations/elasticsearch/tests/test_document_store.py
@@ -147,11 +147,12 @@ class TestDocumentStore(DocumentStoreBaseTests):
 
     def test_deserialize_document_dataframe_ignored(self, document_store: ElasticsearchDocumentStore):
         hit = {
-            "_source": {"id": "1", "content": "test", "dataframe": {"a": [1, 2, 3]}},
+            "_source": {"id": "1", "content": "test", "dataframe": {"a": [1, 2, 3]}, "_score": 1.0},
         }
         doc = document_store._deserialize_document(hit)
         assert doc.id == "1"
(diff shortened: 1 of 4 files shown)


integrations/elasticsearch/tests/test_document_store.py
```diff
@@ -130,6 +131,31 @@ def test_write_documents(self, document_store: ElasticsearchDocumentStore):
         with pytest.raises(DuplicateDocumentError):
             document_store.write_documents(docs, DuplicatePolicy.FAIL)
 
+    def test_write_documents_dataframe_ignored(self, document_store: ElasticsearchDocumentStore):
+        doc = Document(id="1", content="test")
+        doc.dataframe = DataFrame({"a": [1, 2, 3]})
```
davidsbatista, line 136: this test will fail when we remove dataframe from documents

davidsbatista (approved):
LGTM - just a simple comment regarding a test

integrations/elasticsearch/tests/test_document_store.py
anakin87: I've assigned the dataframe separately to avoid the failure.

This works, for example:
```python
from haystack import Document
doc = Document(content="test")
doc.my_attribute = "test"

print(doc.my_attribute)
# test
```

correct?

davidsbatista:
gotcha you - you are just "injecting" my force a new attribute in an instance of the class 👍🏽

PR #1406, opened 2025-02-17
chore!: Weaviate - remove dataframe support

anakin87:
### Related Issues

- part of #1371

### Proposed Changes:
- Remove dataframe support from Weaviate
- This is a breaking change -> I will release a major version of this integration.

### How did you test it?
CI; added specific tests;
I also tested locally that a collection created with the current version of the Document Store still works with the new version (but skips dataframe during writing and reading).

### Checklist

### DIFF
--- a/integrations/weaviate/src/haystack_integrations/document_stores/weaviate/document_store.py
+++ b/integrations/weaviate/src/haystack_integrations/document_stores/weaviate/document_store.py
@@ -297,20 +297,20 @@ class WeaviateDocumentStore:
             dataframe = data.pop("dataframe")
             if dataframe:
                 logger.warning(
-                    "Document %s has the `dataframe` field set. "
+                    "Document {id} has the `dataframe` field set. "
                     "WeaviateDocumentStore no longer supports dataframes and this field will be ignored. "
                     "The `dataframe` field will soon be removed from Haystack Document.",
-                    data["_original_id"],
+                    id=data["_original_id"],
                 )
 
         if "sparse_embedding" in data:
             sparse_embedding = data.pop("sparse_embedding", None)
             if sparse_embedding:
                 logger.warning(
-                    "Document %s has the `sparse_embedding` field set,"
+                    "Document {id} has the `sparse_embedding` field set,"
                     "but storing sparse embeddings in Weaviate is not currently supported."
                     "The `sparse_embedding` field will be ignored.",
-                    data["_original_id"],
+                    id=data["_original_id"],
                 )
 
         return data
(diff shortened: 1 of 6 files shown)


integrations/weaviate/src/haystack_integrations/document_stores/weaviate/document_store.py
```diff
@@ -294,6 +292,16 @@ def _to_data_object(self, document: Document) -> Dict[str, Any]:
                 "Document %s has the unsupported `_split_overlap` meta field. It will be ignored.", data["_original_id"]
             )
 
+        if "dataframe" in data:
+            dataframe = data.pop("dataframe")
+            if dataframe:
+                logger.warning(
```
sjrl, line 298: Hey @anakin87 we should be using the logger from Haystack right? So `from haystack import logging`

sjrl:
This looks good! One additional question would this also break already existing Weaviate document stores that have ingested a Document with a dataframe?

integrations/weaviate/src/haystack_integrations/document_stores/weaviate/document_store.py
anakin87: Good catch. Done in https://github.com/deepset-ai/haystack-core-integrations/pull/1406/commits/8f7012d0e2f671997201ec7df59f490d7ad8ab2d

I've also noticed that we are not consistent across integrations. I'll create an issue for that.

anakin87:
> This looks good! One additional question would this also break already existing Weaviate document stores that have ingested a Document with a dataframe?

In general, I think this won't break existing Weaviate Document Stores where Documents with dataframes were ingested.
I've tested this locally, saving a Document with a dataframe with the old version and retrieving it with the new version.

Of course, with the new version, the dataframe is not returned. (This qualifies this change as breaking)

integrations/weaviate/src/haystack_integrations/document_stores/weaviate/document_store.py
```diff
@@ -294,6 +292,16 @@ def _to_data_object(self, document: Document) -> Dict[str, Any]:
...
+                    "Document %s has the `dataframe` field set. "
+                    "WeaviateDocumentStore no longer supports dataframes and this field will be ignored. "
+                    "The `dataframe` field will soon be removed from Haystack Document.",
+                    data["_original_id"],
+                )
```
sjrl, line 303: I'd double check that this warning message is formatted correctly. I believe we have to format the string like 
```
        logger.debug(
            "Connecting '{sender_component}.{sender_socket_name}' to '{receiver_component}.{receiver_socket_name}'",
            sender_component=sender_component_name,
            sender_socket_name=sender_socket.name,
            receiver_component=receiver_component_name,
            receiver_socket_name=receiver_socket.name,
        )
```
I'm unsure if the `%s` syntax also works

anakin87: yes, fixed now.

sjrl (approved):
Looks good!

PR #1412, opened 2025-02-17
feat: integration with Weave the tracer from Weight and Bias 

davidsbatista:
### Related Issues

- part of [103](https://github.com/deepset-ai/haystack-private/issues/103)

### Proposed Changes:

- This integration is a wrapper around  Weave python client so that haystack tracing tool can send traces to the Weave framework for component runs.

### How did you test it?

- unit tests, integration tests and manual verification

### Notes for the reviewer

- I'm still trying to add an example that runs with the docker release of weaver
- I'm soon opening a PR on https://github.com/deepset-ai/haystack-integrations

### Checklist


integrations/weights_bias/src/haystack_integrations/tracing/weave/tracer.py
```diff
@@ -0,0 +1,158 @@
...
+                op=operation_name,
+                inputs=tags,
+                parent=parent_span.raw_span() if parent_span else None,
+            )
+            span = WeaveSpan(call=call)
+        self._current_span = span
+
+        if tags:
+            span.set_tags(tags)
+
```
anakin87, line 113: The following part seems to me a bit hard to read and understand, mostly because of nested logic.
It would be great if it could be simplified a bit.

integrations/weights_bias/src/haystack_integrations/tracing/weave/tracer.py
```diff
@@ -0,0 +1,158 @@
...
+        parent=parent_span.raw_span() if parent_span else None,
+    )
+
+    return call
+
+
+class WeaveTracer(Tracer):
+    """A simple bridge between Haystack's Tracer interface and Weave."""
+
+    def __init__(self, project_name: str) -> None:
```
anakin87, line 82: missing docstring

integrations/weights_bias/src/haystack_integrations/components/connectors/weave_connector.py
```diff
@@ -0,0 +1,60 @@
...
+        """
+        Initialize WeaveConnector.
+
+        :param pipeline_name: The name of the pipeline you want to trace.
+        """
+        self.pipeline_name = pipeline_name
+
+        # TODO: This is a hack because content tracing enabled seems to rely on import order which is hard to debug.
+        # As a workaround, we assume that users adding the WeaveConnector to a pipeline always want to trace content.
+        haystack_tracer.is_content_tracing_enabled = True
```
anakin87, line 28: we should not do this: if users want to do content tracing, they should explicitly opt in 

[An example of how this is handled in Langfuse Tracer](https://github.com/deepset-ai/haystack-core-integrations/blob/65c46c09fbb26a9a1fb19776042fa16ed2dc5c61/integrations/langfuse/src/haystack_integrations/tracing/langfuse/tracer.py#L311)

integrations/weights_bias/src/haystack_integrations/components/connectors/weave_connector.py
```diff
@@ -0,0 +1,60 @@
...
+
+    def warm_up(self) -> None:
+        """Initialize the WeaveTracer."""
+        if self.tracer is None:
+            self.tracer = WeaveTracer(project_name=self.pipeline_name)
+            tracing.enable_tracing(self.tracer)
+
+    @component.output_types(no_op=str)
+    def run(self, no_op: str = "no_op") -> Dict[str, str]:
+        return {"no_op": no_op}
```
anakin87, line 40: This seems like a placeholder.
In [Langfuse](https://github.com/deepset-ai/haystack-core-integrations/blob/65c46c09fbb26a9a1fb19776042fa16ed2dc5c61/integrations/langfuse/src/haystack_integrations/components/connectors/langfuse/langfuse_connector.py#L163) this is done differently and makes more sense to me.

integrations/weights_bias/src/haystack_integrations/components/connectors/weave_connector.py
```diff
@@ -0,0 +1,60 @@
...
+
+    def to_dict(self) -> Dict[str, Any]:
+        """
+        Serializes the component to a dictionary.
+
+        :returns:
+            Dictionary with all the necessary information to recreate this component.
+        """
+        return default_to_dict(self, pipeline_name=self.pipeline_name)  # type: ignore
```
anakin87, line 49: why `type:ignore` ?

anakin87 (commented):
I left some initial comments.
I would also like @vblagoje to review the tracing logic since he is more familiar with it.

integrations/weights_bias/pyproject.toml
```diff
@@ -0,0 +1,161 @@
...
+test = "pytest {args:tests}"
+test-cov = "coverage run -m pytest {args:tests}"
+test-cov-retry = "test-cov --reruns 3 --reruns-delay 30 -x"
+cov-report = ["- coverage combine", "coverage report"]
+cov = ["test-cov", "cov-report"]
+cov-retry = ["test-cov-retry", "cov-report"]
+docs = ["pydoc-markdown pydoc/config.yml"]
+
+[[tool.hatch.envs.all.matrix]]
+python = ["3.8", "3.9", "3.10", "3.11", "3.12"]
```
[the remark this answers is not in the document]
davidsbatista: good catch - I will also change the default styles for black and ruff to `py39`

integrations/weights_bias/src/haystack_integrations/components/connectors/weave_connector.py
davidsbatista: This was a bit confusing to me - I will adapt the langfuse tracer strategy

integrations/weights_bias/src/haystack_integrations/tracing/weave/tracer.py
davidsbatista: I agree. I'm trying to understand this code a bit more in detail and see how to improve it - although I'm opening this PR the code was made by someone else

integrations/weights_bias/src/haystack_integrations/components/connectors/weave_connector.py
davidsbatista: I'm reading into the Weave API and trying to understand what else would make sense to pass here

anakin87: Langfuse does this
`return {"name": self.name, "trace_url": self.tracer.get_trace_url()}`

Would it make sense in this case?

@abrahamy any insights to share?

integrations/weights_bias/src/haystack_integrations/tracing/weave/tracer.py
davidsbatista: I did some refactoring for the trace method, following also on an extended docstring, explaining better what's happening in the trace method and why the two approaches depending on the type of operation we want to trace

davidsbatista:
@anakin87 I'm also adding @mathislucka as a reviewer since he was the original author of the code

integrations/weights_bias/src/haystack_integrations/components/connectors/weave_connector.py
```diff
@@ -0,0 +1,122 @@
...
+        self.pipeline_name = pipeline_name
+        content_tracing_enabled = os.getenv("HAYSTACK_CONTENT_TRACING_ENABLED", "false").lower()
+        self.enable_tracing = content_tracing_enabled == "true"
+        haystack_tracer.is_content_tracing_enabled = self.enable_tracing
+        if not haystack_tracer.is_content_tracing_enabled:
+            logger.warning(
+                "Traces will not be logged to Weave because Haystack tracing is disabled. "
+                "To enable, set the HAYSTACK_CONTENT_TRACING_ENABLED environment variable to true "
+                "before importing Haystack."
+            )
```
anakin87, line 90: Maybe it's better to do this check directly in the Tracer (or also in the Tracer),
otherwise we can have the same problem there.

integrations/weights_bias/src/haystack_integrations/components/connectors/weave_connector.py
```diff
@@ -0,0 +1,122 @@
...
+
+    def warm_up(self) -> None:
+        """Initialize the WeaveTracer."""
+        if self.tracer is None:
+            self.tracer = WeaveTracer(project_name=self.pipeline_name)
+            tracing.enable_tracing(self.tracer)
+
+    @component.output_types(no_op=str)
+    def run(self) -> dict[str, str]:
+        return {}  # this is a no-op component - it doesn't do anything, just sends traces to Weights & Biases
```
anakin87, line 102: While the `no_op` argument was confusing, I would say that now we are declaring an output key (`no_op`) which is always missing.

Perhaps we can return the `name`, as Langfuse does:
`return {"name": self.name, "trace_url": self.tracer.get_trace_url()}`

anakin87 (commented):
- I left two small comments
- Can we also test the Tracer ([see Langfuse](https://github.com/deepset-ai/haystack-core-integrations/blob/main/integrations/langfuse/tests/test_tracer.py))?
- (It would be great if also @mathislucka can review this PR)

integrations/weights_bias/src/haystack_integrations/components/connectors/weave_connector.py
davidsbatista: there's no name in this class, the only attribute is the `pipeline_name` - if you prefer we can send the `pipeline_name`

anakin87: yes... it is similar

https://github.com/deepset-ai/haystack-core-integrations/blob/911d3b1b8bb14e7afb6ed19880a947923db317a6/integrations/langfuse/src/haystack_integrations/components/connectors/langfuse/langfuse_connector.py#L130

integrations/weights_bias/src/haystack_integrations/components/connectors/weave_connector.py
davidsbatista: we can replicate the check in the Tracer for the case of not being called with the connector

anakin87: yes but since the Tracer is used in the Connector, we can also do it directly in the Tracer

davidsbatista: I completely remove it from the connector, and it's only being done in the init of the Tracer

anakin87 (commented):
For my part, at this point, I would like to see some tests for the tracer, such as these: https://github.com/deepset-ai/haystack-core-integrations/blob/main/integrations/langfuse/tests/test_tracer.py

Otherwise, based on my limited experience with this topic, everything seems fine.

davidsbatista:
> For my part, at this point, I would like to see some tests for the tracer, such as these: https://github.com/deepset-ai/haystack-core-integrations/blob/main/integrations/langfuse/tests/test_tracer.py
> 
> Otherwise, based on my limited experience with this topic, everything seems fine.

sorry - tracer tests failed to make to the commit, my mistake

anakin87 (approved):
Looks good.

As I mentioned, I'm not an expert on this topic.
It would be great if you could get another review from @mathislucka or @vblagoje.

integrations/weights_bias/src/haystack_integrations/tracing/weave/tracer.py
```diff
@@ -0,0 +1,222 @@
...
+        """
+        Initialize the WeaveTracer.
+
+        :param project_name: The name of the project to trace, this is will be the name appearing in Weave project.
+        """
+
+        content_tracing_enabled = os.getenv("HAYSTACK_CONTENT_TRACING_ENABLED", "false").lower()
+        haystack_tracer.is_content_tracing_enabled = content_tracing_enabled == "true"
```
mathislucka, line 79: @davidsbatista does this solve the import order problem that we talked about yesterday? If so, we should keep that in mind for the langfuse connector.

integrations/weights_bias/src/haystack_integrations/tracing/weave/tracer.py
```diff
@@ -0,0 +1,222 @@
...
+        if not haystack_tracer.is_content_tracing_enabled:
+            logger.warning(
+                "Traces will not be logged to Weave because Haystack tracing is disabled. "
```
mathislucka, line 82: Only content will not be traced (inputs and outputs of components). We still get basic traces like which component's ran in what order.

integrations/weights_bias/src/haystack_integrations/tracing/weave/tracer.py
```diff
@@ -0,0 +1,222 @@
...
+                "To enable, set the HAYSTACK_CONTENT_TRACING_ENABLED environment variable to true "
+                "before importing Haystack."
```
mathislucka, line 84: Before importing Haystack? I think the example doesn't show that.

integrations/weights_bias/src/haystack_integrations/tracing/weave/tracer.py
```diff
@@ -0,0 +1,222 @@
...
+            attributes=attributes,
+            client=self._client,
+            parent_span=parent_span,
+            operation_name=operation_name,
+        )
+        span.set_call(call)
+        comp_output = attributes.pop("haystack.component.output", {})
+        self._client.finish_call(
+            call,
+            output={key: coerce_tag_value(value) for key, value in comp_output.items()},
```
mathislucka, line 139: We could move the coerce_tag_value to `set_tag` so that we don't need to do it in two places (for inputs and outputs) but not absolutely necessary for me.

mathislucka (approved):
Some minor questions and comments but overall looks good!

integrations/weights_bias/src/haystack_integrations/tracing/weave/tracer.py
davidsbatista: thanks I will updated the log message accordingly

integrations/weights_bias/src/haystack_integrations/tracing/weave/tracer.py
davidsbatista: not 100& sure, but I think regardless of the import order if this variable is not set traces will not be generated

davidsbatista: by variable, I mean ªhaystack_tracer.is_content_tracing_enabled`

integrations/weights_bias/src/haystack_integrations/tracing/weave/tracer.py
anakin87: Since this is a common issue encountered by users, I think that we should show this in the example (setting the env var before importing Haystack).

davidsbatista: This is a leftover from adapting `langfuse` and shouldn't be here - in practise, the variable is set on OS before running the script

integrations/weights_bias/src/haystack_integrations/tracing/weave/tracer.py
davidsbatista: I did a test and I think something is odd, here, run a script to send traces and both versions work, regardless of the env variable being set or not.

```bash
WANDB_API_KEY="..." python weights_bias/example/hybrid_retrieval.py
```

```bash
HAYSTACK_CONTENT_TRACING_ENABLED=true WANDB_API_KEY="..." python weights_bias/example/hybrid_retrieval.py
```

mathislucka: So you are sending content in both cases? Because basic traces are sent either way, just the content (input/output of components) should be missing.

davidsbatista: Yes, I see the content, in Weaver, for both cases, by content I mean for instance, the question/string and the retrieved documents.

integrations/weights_bias/src/haystack_integrations/tracing/weave/tracer.py
davidsbatista: I'm not sure how set tag is supposed to work, and if we need to call `coerce_tag_value` for all input in `set_Tag`- I would rather leave it as is for now - we can change it later.

PR #1421, opened 2025-02-20
chore: remove Python 3.8 support

davidsbatista:
### Related Issues

- Removing support for Python 3.8

### Checklist


anakin87 (commented):
If we want to really drop python 3.8 support we should also change all instances of 
`requires-python = ">=3.8"`
with
`requires-python = ">=3.9"`

PR #1429, opened 2025-02-24
feat: adding async support to ElasticSearch retrievers and document store

davidsbatista:
### Related Issues

- fixes [#1388](https://github.com/deepset-ai/haystack-core-integrations/issues/1388)

### Proposed Changes:

- Adds async method to both retrievers (BM25 + Embeddings) and the ElasticSearch document store

### How did you test it?

- unit tests, integration tests, manual verification

### Checklist


integrations/elasticsearch/tests/test_document_store.py
```diff
@@ -242,16 +241,16 @@ def test_bm25_not_all_terms_must_match(self, document_store: ElasticsearchDocume
         Test that not all terms must mandatorily match for BM25 retrieval to return a result.
         """
         documents = [
-            Document(id=1, content="There are over 7,000 languages spoken around the world today."),
+            Document(id="1", content="There are over 7,000 languages spoken around the world today."),
```
mpangrazzi, line 244: Did you switched to strings here for consistency with other tests?

mpangrazzi (approved):
Looks good! Left only a small comment

integrations/elasticsearch/tests/test_document_store.py
davidsbatista: The `Document.id` should be a str and not an int

https://github.com/deepset-ai/haystack/blob/main/haystack/dataclasses/document.py#L71

mpangrazzi: Got it. It's just weird that the test still passes using ints.

mpangrazzi:
@davidsbatista one last thing: since we're closing the client on async tests, shall we do the same with sync ones?

So adding

```python
store.client.close()
```

to `document_store` sync fixture?

PR #1442, opened 2025-02-26
fix: avoid thinking end tag on first text block

tstadel:
### Related Issues

- fixes non-thinking requests starting with `</thinking>`
- as a workaround `include_thinking` can be set to `False` but we should not break non-thinking requests

### Proposed Changes:
- don't add `</thinking>` tag if text content block is at index 0

### How did you test it?
- improved tests

### Checklist


davidsbatista (approved):
LGTM

PR #1445, opened 2025-02-26
feat: adding async to `AmazonChatGenerator`

### LINKED ISSUE #1381: add run_async for AmazonBedrockChatGenerator

davidsbatista:
### Related Issues

- fixes #1381 

## pyproject.toml`
- Added new dependency: `aioboto3>=14.0.0`
- Added new test dependency: `pytest-asyncio`
- Formatting changes in tool configurations

## `utils.py`
- Added import for `aioboto3`
-  Enhanced `get_aws_session()` function: new parameter `async_client: bool = False`, for returning async AWS session using `aioboto3` when `async_client=True`

### refactoring
- Moved utility functions to separate module:
  - `_format_tools`
  - `_format_messages`
  - `_parse_completion_response`
  - `_parse_streaming_response`
  - Added new `_parse_streaming_response_async`

### adding async
- `_get_async_client()`: method to create async Bedrock client
- `_prepare_request_params()`: prepare request parameters used by both run versions
- `run_async()`: async version of the run method

### NOTES:

- Running the `hatch run style:fmt` applied black code formatting to some files unrelated with the changes on this PR

### How did you test it?

- unit tests, integration tests, manual verification, CI tests

### Checklist


integrations/amazon_bedrock/src/haystack_integrations/common/amazon_bedrock/utils.py
```diff
@@ -20,6 +21,7 @@ def get_aws_session(
     aws_session_token: Optional[str] = None,
     aws_region_name: Optional[str] = None,
     aws_profile_name: Optional[str] = None,
+    async_client: bool = False,
```
mpangrazzi, line 24: I would call this param only `async` since this method returns a `Session` (and then you get the client later using `Session.client(...)`. WDYT?

integrations/amazon_bedrock/src/haystack_integrations/components/generators/amazon_bedrock/chat/chat_generator.py
```diff
@@ -529,3 +379,44 @@ def run(
...
+        :param generation_kwargs: Optional dictionary of generation parameters.
+        :param tools: Optional list of Tool objects that the model can use.
+        :return: Dictionary containing the model's replies as a list of ChatMessage objects.
+        """
+        params, callback = self._prepare_request_params(
+            messages=messages, streaming_callback=streaming_callback, generation_kwargs=generation_kwargs, tools=tools
+        )
+
+        try:
+            async_client_coroutine = await self._get_async_client()
```
mpangrazzi, line 405: Hmm this seems to create a new async client each time `run_async` is called. 

We may want to assign the async client to `self.async_client` on `_get_async_client` and then add an initial check. So like:

```python
async def _get_async_client(self):

        if self._async_client is not None:
            return self._async_client

        # ...

        self._async_client = async_session.client( ... )
```

mpangrazzi (changes requested):

integrations/amazon_bedrock/src/haystack_integrations/common/amazon_bedrock/utils.py
davidsbatista: I can change to some other name, `async` is a reserved keyword

integrations/amazon_bedrock/src/haystack_integrations/components/generators/amazon_bedrock/chat/chat_generator.py
mpangrazzi: or we can initialize the client ini `__init__` like on sync version, your choice ;)

integrations/amazon_bedrock/src/haystack_integrations/common/amazon_bedrock/utils.py
mpangrazzi: yeah you're right... what about `use_async_session` then?

integrations/amazon_bedrock/src/haystack_integrations/components/generators/amazon_bedrock/chat/chat_generator.py
davidsbatista: solved in on-line session/pair programming

mpangrazzi (approved):

Issue #1454, opened 2025-02-28
ddtrace breaks AnthropicChatGenerator streaming


mathislucka:
**Describe the bug**

See this issue on the ddtrace repo for context: https://github.com/DataDog/dd-trace-py/issues/12562

ddtrace wraps Anthropic's `Stream` in a way that breaks class inheritance which is why our check for `isinstance(response, Stream)` fails which leads to an error for anyone who wants to use AnthropicChatgenerator, datadog and streaming together. We ran into this in deepset Studio.

I know that they should fix it, but if they don't provide a timely fix we could consider changing our implementation to check for `if streaming_callback is not None` instead, which also signals that the response should come back as a stream.

**To Reproduce**
Steps to reproduce the behavior. Feel free to link a Colab we can run to investigate the issue.

**Describe your environment (please complete the following information):**
 - Haystack version:
 - Integration version:

PR #1439, opened 2025-02-26
feat: support thinking parameter for Claude 3.7

tstadel:
### Related Issues

- fixes #issue-number

### Checklist

- I have read the [contributors guidelines](https://github.com/deepset-ai/haystack-core-integrations/blob/main/CONTRIBUTING.md) and the [code of conduct](https://github.com/deepset-ai/haystack-core-integrations/blob/main/CODE_OF_CONDUCT.md)
- I have updated the related issue with new insights and changes
- I added unit tests and updated the docstrings
- I've used one of the [conventional commit types](https://www.conventionalcommits.org/en/v1.0.0/) for my PR title: `fix:`, `feat:`, `build:`, `chore:`, `ci:`, `docs:`, `style:`, `refactor:`, `perf:`, `test:`.

PR #1455, opened 2025-02-28
feat: support extended thinking mode with AnthropicGenerator

tstadel:
### Related Issues

With [Claude 3.7](https://docs.anthropic.com/en/docs/build-with-claude/extended-thinking#implementing-extended-thinking) the API supports the `thinking` parameter to enable extended thinking mode, e.g.:
```
"thinking": {
    "type": "enabled",
    "budget_tokens": 1024
}
```

Example:
https://docs.anthropic.com/en/docs/build-with-claude/extended-thinking#implementing-extended-thinking

Note that thinking output is separated from normal text output, other than for example for Deepseek-R1. So we have to decide how to incorporate thinking output into the reply. Per default we include it using <thinking></thinking> tags, whereas this behavior is fully controllable via parameters:
- `include_thinking`: if `False` no thinking output is incorporated into the reply (default: `True`)
- `thinking_tag`: the name of the thinking tag, defaults to `thinking`. If `None`, the thinking part of the reply won't be encapsulated within tags.

This is the same approach as in https://github.com/deepset-ai/haystack-core-integrations/pull/1439

This also contains a fix for https://github.com/deepset-ai/haystack-core-integrations/issues/1454 in `AnthropicGenerator`

### Proposed Changes:
- support parameter `thinking`
- add options for how to incorporate the thinking response into the reply
- add `streaming_callback` run param

### How did you test it?
- tested against Anthropic API

### Checklist

### DIFF
--- a/integrations/anthropic/src/haystack_integrations/components/generators/anthropic/generator.py
+++ b/integrations/anthropic/src/haystack_integrations/components/generators/anthropic/generator.py
@@ -7,11 +7,17 @@ from haystack.utils import Secret, deserialize_callable, deserialize_secrets_inp
 from anthropic import Anthropic, Stream
 from anthropic.types import (
     ContentBlockDeltaEvent,
+    ContentBlockStartEvent,
+    ContentBlockStopEvent,
     Message,
     MessageDeltaEvent,
     MessageParam,
     MessageStartEvent,
     MessageStreamEvent,
+    TextBlock,
+    TextDelta,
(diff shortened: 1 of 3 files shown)


mpangrazzi (approved):
LGTM - BTW there is a minor conflict to solve yet ;)

mathislucka:
@tstadel is off this week. I'll take over, resolve the conflict and get this merged.

Issue #1457, opened 2025-03-03
Nvidia Ranker URL issue


2-ulfhednar-2:
**Describe the bug**
The URL finalization for the NvidiaRanker in the NIM-Backend function rank is broken.
It doesnt fully create the link. 

**To Reproduce**
Just run the example code provided in the ranker.py for nvidia

**Describe your environment (please complete the following information):**
 - OS: ubuntu 22.04
 - Haystack version: Haystack 2.10.0
 - Integration version: Nvidia-Haystack 0.1.6

To Fix:
Change lines 173/174 in the nim_backend.py to properly complete the URL
from:
    def rank(self, query_text: str, document_texts: List[str]) -> List[Dict[str, Any]]:
        url = self.api_url

To:
    def rank(self, query_text: str, document_texts: List[str]) -> List[Dict[str, Any]]:
        url = f"{self.api_url}/ranking"

right now the ranker.py takes the provided URL adds v1 to it and then passes it to the def rank. To properly access the endpoint usually /ranking is needed.

julian-risch:
Hello @2-ulfhednar-2 thanks for opening this issue. I wasn't able to reproduce it with the latest version of the integration. Here is the code I used:

```python
from haystack_integrations.components.rankers.nvidia import NvidiaRanker
from haystack import Document
from haystack.utils import Secret

ranker = NvidiaRanker(
    model="nvidia/nv-rerankqa-mistral-4b-v3",
    api_key=Secret.from_env_var("NVIDIA_API_KEY"),
)
ranker.warm_up()

query = "What is the capital of Germany?"
documents = [
    Document(content="Berlin is the capital of Germany."),
    Document(content="The capital of Germany is Berlin."),
    Document(content="Germany's capital is Berlin."),
]

result = ranker.run(query, documents, top_k=2)
print(result["documents"])
```

When I run this code after installing `nvidia-haystack 0.1.6` with `pip install nvidia-haystack`, I get the expected response
`[Document(id=ded27a743b8980b58b13311c6f640dfbdccf0b2e523055c5845ee772ef473869, content: 'The capital of Germany is Berlin.', score: 27.171875), Document(id=f780ec15ed0011c1af18da28519f9205d5144de3ee84d7c613cf6bc0ad3f5da9, content: 'Germany's capital is Berlin.', score: 24.46875)]`

[...]

PR #1458, opened 2025-03-03
test: add test cases for all utils methods for Nvidia integration

julian-risch:
### Related Issues

- Related to https://github.com/deepset-ai/haystack-core-integrations/issues/1457

When I investigated the above issue, I noticed that methods in utils are not tested at all. Before changing core parts of the integration in order to fix the issue, such as removing url validation, we should cover the current behavior with tests.

### Proposed Changes:

- Add test cases for all methods in utils
- Change logic of `validate_hosted_model`
- Fix and extend `validate` function of `Model` class

### How did you test it?

- New unit tests for utils but not for `validate` function of `Model` class

### Checklist

### DIFF
--- a/integrations/nvidia/tests/test_utils.py
+++ b/integrations/nvidia/tests/test_utils.py
@@ -1,0 +1,92 @@
+# SPDX-FileCopyrightText: 2024-present deepset GmbH <info@deepset.ai>
+#
+# SPDX-License-Identifier: Apache-2.0
+
+from typing import Any
+
+import pytest
+
+from haystack_integrations.utils.nvidia import is_hosted
+from haystack_integrations.utils.nvidia.models import EMBEDDING_MODEL_TABLE, CHAT_MODEL_TABLE, RANKING_MODEL_TABLE
+from haystack_integrations.utils.nvidia.utils import lookup_model, determine_model, validate_hosted_model
+
+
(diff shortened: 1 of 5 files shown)


integrations/nvidia/tests/test_utils.py
```diff
@@ -0,0 +1,105 @@
...
+
+def test_validate_hosted_model_is_unknown() -> None:
+    with pytest.raises(ValueError, match="is unknown"):
+        assert validate_hosted_model("not-a-model", "NvidiaGenerator")
+    with pytest.raises(ValueError, match="is unknown"):
+        assert validate_hosted_model("not-a-model")
+
+
+def test_validate_hosted_model() -> None:
+    assert validate_hosted_model("snowflake/arctic-embed-l", "NvidiaTextEmbedder")
```
Amnah199, line 105: Another test suggestion if you think it adds more coverage:

```
def test_validate_hosted_model_with_client() -> None:
    """Test when model's client matches the provided client."""
    model = validate_hosted_model("meta/codellama-70b", "NvidiaGenerator")
    assert model is not None
    assert model.client == "NvidiaGenerator"
```

Amnah199 (approved):
Left an optional comment otherwise looks good!

PR #1450, opened 2025-02-27
feat: Use Haystack logging across integrations

mdrazak2001:
### Related Issues

- fixes #1409

### Proposed Changes:

Use Haystack logging across integrations

- Replace standard logging with Haystack logging module
- Update log messages to use keyword arguments instead of positional args

### How did you test it?

- CI

### Checklist

- I have read the [contributors guidelines](https://github.com/deepset-ai/haystack-core-integrations/blob/main/CONTRIBUTING.md) and the [code of conduct](https://github.com/deepset-ai/haystack-core-integrations/blob/main/CODE_OF_CONDUCT.md)
- I have updated the related issue with new insights and changes
- I added unit tests and updated the docstrings
- I've used one of the [conventional commit types](https://www.conventionalcommits.org/en/v1.0.0/) for my PR title: `fix:`, `feat:`, `build:`, `chore:`, `ci:`, `docs:`, `style:`, `refactor:`, `perf:`, `test:`.

PR #1484, opened 2025-03-07
chore: Use Haystack logging across integrations

### LINKED ISSUE #1409: Use Haystack logging across integrations
For each integration, we should check whether we are using the Haystack logging or not (`from haystack import logging`).

Changing from the standard logging to Haystack patched version also requires using keyword arguments.

An example:
```python
import logging

my_id="my custom id"

logger.warning("Document %s has the `dataframe` field set. ",
                my_id)
```

becomes

```python
from haystack import logging

my_id="my custom id"


logger.warning("Document {id} has the `dataframe` field set. ",
               id=my_id)
```

anakin87:
### Related Issues

- fixes #1409
- recovered from #1450 (some tests were failing due to permission issues with PRs coming from forks)

### Proposed Changes:

Use Haystack logging across integrations

- Replace standard logging with Haystack logging module
- Update log messages to use keyword arguments instead of positional args

### How did you test it?

- CI
- llama.cpp Windows tests failures are unrelated

### Checklist


mpangrazzi (approved):
Looks good! 👍  Do you think we should add a note on contributing about how to use logging here or not?

anakin87:
> Looks good! 👍 Do you think we should add a note on contributing about how to use logging here or not?

I would avoid it for the time being, hoping that the examples in all the integrations are sufficient.

PR #1487, opened 2025-03-10
feat: Add MCP tooling integration

vblagoje:
### Why:
Introduces an initial integration of the Model Context Protocol (MCP) tool support

### What:
- Added MCP integration providing support for HTTP and stdio transport.
- Introduced the `MCPTool` tool to ease communication with MCP servers.
- Implemented new examples and test suites to demonstrate usage and validity.

### How can it be used:
An example of using the `MCPTool` class to connect with an HTTP server:
```python
from haystack_integrations.tools.mcp import MCPTool, HttpMCPServerInfo

server_info = HttpMCPServerInfo(base_url="http://localhost:8000")
result = tool.invoke(a=5, b=3)
```

### How did you test it:
Integrated tests were written, including both unit and integration tests, using pytest to validate tool operations and connectivity. MCP clients were mocked to simulate real server interactions, along with actual server-client integration tests using subprocesses.

### Notes for the reviewer:
Note that we don't use typing package as we are basing this package on 3.10+
Please focus on reviewing the `MCPTool` initialization and error handling logic, ensuring compatibility with various server types. Also, verify the completeness of the transport protocol handling in `mcp_tool.py`, and pay attention to the test coverage for potential integration pitfalls.

### DIFF
--- a/integrations/mcp/examples/mcp_http_client.py
+++ b/integrations/mcp/examples/mcp_http_client.py
@@ -1,0 +1,41 @@
+# SPDX-FileCopyrightText: 2022-present deepset GmbH <info@deepset.ai>
+#
+# SPDX-License-Identifier: Apache-2.0
+
+from haystack_integrations.tools.mcp import MCPTool, HttpMCPServerInfo
+
+# run this client after running the server mcp_http_server.py
+# it shows how easy it is to use the MCPTool with HTTP transport
+
+
+def main():
+    """Example of synchronous usage of MCPTool with HTTP transport."""
+
+    server_info = HttpMCPServerInfo(
+        base_url="http://localhost:8000",
+    )
+
+    tool = MCPTool(name="add", server_info=server_info)
+
+    tool_subtract = MCPTool(name="subtract", server_info=server_info)
+
+    try:
+        print(f"Tool spec: {tool.tool_spec}")
+
+        result = tool.invoke(a=7, b=3)
+        print(f"7 + 3 = {result}")
+
+        result = tool_subtract.invoke(a=5, b=3)
+        print(f"5 - 3 = {result}")
+
+        result = tool.invoke(a=10, b=20)
+        print(f"10 + 20 = {result}")
+
+    except Exception as e:
+        print(f"Error in sync example: {e}")
+    finally:
+        pass
+
+
+if __name__ == "__main__":
+    main()
--- a/integrations/mcp/pyproject.toml
+++ b/integrations/mcp/pyproject.toml
@@ -1,0 +1,181 @@
+[build-system]
+requires = ["hatchling", "hatch-vcs"]
+build-backend = "hatchling.build"
+
+[project]
+name = "mcp-haystack"
+dynamic = ["version"]
+description = "Haystack integration for Model Context Protocol (MCP)"
+readme = "README.md"
+requires-python = ">=3.10"
+license = "Apache-2.0"
+keywords = [
+    "MCP",
+    "Haystack",
+    "Model Context Protocol",
+]
+authors = [
+    { name = "deepset GmbH", email = "info@deepset.ai" },
+]
+classifiers = [
+    "License :: OSI Approved :: Apache Software License",
+    "Development Status :: 4 - Beta",
+    "Programming Language :: Python",
+    "Programming Language :: Python :: 3.10",
+    "Programming Language :: Python :: 3.11",
+    "Programming Language :: Python :: 3.12",
+    "Programming Language :: Python :: Implementation :: CPython",
+]
+dependencies = ["mcp", "haystack-ai>=2.9.0"]
+
+[project.urls]
+Documentation = "https://github.com/deepset-ai/haystack-core-integrations/tree/main/integrations/mcp#readme"
+Issues = "https://github.com/deepset-ai/haystack-core-integrations/issues"
+Source = "https://github.com/deepset-ai/haystack-core-integrations/tree/main/integrations/mcp"
+
+[tool.hatch.build.targets.wheel]
+packages = ["src/haystack_integrations"]
+
+[tool.hatch.version]
+source = "vcs"
+tag-pattern = 'integrations\/mcp-v(?P<version>.*)'
+
+[tool.hatch.version.raw-options]
+root = "../.."
+git_describe_command = 'git describe --tags --match="integrations/mcp-v[0-9]*"'
+
+[tool.hatch.envs.default]
+installer = "uv"
+dependencies = [
+    "coverage[toml]>=6.5",
+    "haystack-pydoc-tools",
+    "pytest",
+    "anyio", 
+    "pytest-asyncio",    
+    "pytest-tornasync",
+    "pytest-rerunfailures",
+    "pytest-pythonpath",
+    "mcp_server_time" ## time server for tests
+]
+
+[tool.hatch.envs.default.scripts]
+test = "pytest {args:tests}"
+test-cov = "coverage run -m pytest {args:tests}"
+test-cov-retry = "test-cov --reruns 3 --reruns-delay 30 -x"
+cov-report = ["- coverage combine", "coverage report"]
+cov = ["test-cov", "cov-report"]
+docs = ["pydoc-markdown pydoc/config.yml"]
+
+[[tool.hatch.envs.all.matrix]]
+python = ["3.10", "3.11", "3.12"]
+
+[tool.hatch.envs.lint]
+installer = "uv"
+detached = true
+dependencies = [
+    "pip",
+    "black>=23.1.0",
+    "mypy>=1.0.0",
+    "ruff>=0.0.243",
+]
+
+[tool.hatch.envs.lint.scripts]
+typing = "mypy --install-types --non-interactive --explicit-package-bases {args:src/ tests}"
+style = ["ruff check {args:.}", "black --check --diff {args:.}"]
+fmt = ["black {args:.}", "ruff check --fix {args:.}", "style"]
+all = ["style", "typing"]
+
+[tool.black]
+target-version = ["py310"]
+line-length = 120
+skip-string-normalization = true
+
+[tool.ruff]
+target-version = "py310"
+line-length = 120
+
+[tool.ruff.lint]
+select = [
+    "A",
+    "ARG",
+    "B",
+    "C",
+    "DTZ",
+    "E",
+    "EM",
+    "F",
+    "I",
+    "ICN",
+    "ISC",
+    "N",
+    "PLC",
+    "PLE",
+    "PLR",
+    "PLW",
+    "Q",
+    "RUF",
+    "S",
+    "T",
+    "TID",
+    "UP",
+    "W",
+    "YTT",
+]
+ignore = [
+    # Allow non-abstract empty methods in abstract base classes
+    "B027",
+    # Ignore checks for possible passwords
+    "S105",
+    "S106",
+    "S107",
+    # Ignore complexity
+    "C901",
+    "PLR0911",
+    "PLR0912",
+    "PLR0913",
+    "PLR0915",
+]
+unfixable = [
+    # Don't touch unused imports
+    "F401",
+]
+
+[tool.ruff.lint.isort]
+known-first-party = ["haystack_integrations"]
+
+[tool.ruff.lint.flake8-tidy-imports]
+ban-relative-imports = "parents"
+
+[tool.ruff.lint.per-file-ignores]
+# Tests can use magic values, assertions, and relative imports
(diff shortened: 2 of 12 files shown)


vblagoje:
@julian-risch one of the integration tests involves interacting with [mcp-brave-search](https://github.com/modelcontextprotocol/servers/tree/main/src/brave-search) through docker via stdio. See the link to create account and obtain the brave api key. Once you do simply add it to CI under BRAVE_API_KEY and the failing test will start to pass.

vblagoje:
@davidsbatista this one landed on you but please allow @anakin87 to have a look as well

julian-risch:
> @julian-risch one of the integration tests involves interacting with [mcp-brave-search](https://github.com/modelcontextprotocol/servers/tree/main/src/brave-search) through docker via stdio. See the link to create account and obtain the brave api key. Once you do simply add it to CI under `BRAVE_API_KEY` and the failing test will start to pass.

I added a `BRAVE_API_KEY` to the CI now.

julian-risch:
We should adjust the tests in a way that they don't fail if the API key is not available though.

vblagoje:
> We should adjust the tests in a way that they don't fail if the API key is not available though.

Yes it was one of those cases where, I believe, where env value is a str " " and that's how it didn't skip it because if was not falsy.

README.md
```diff
@@ -41,6 +41,7 @@ Please check out our [Contribution Guidelines](CONTRIBUTING.md) for all the deta
...
 | [langfuse-haystack](integrations/langfuse/)                                                                    | Tracer                      | [![PyPI - Version](https://img.shields.io/pypi/v/langfuse-haystack.svg?color=orange)](https://pypi.org/project/langfuse-haystack)                        | [![Test / langfuse](https://github.com/deepset-ai/haystack-core-integrations/actions/workflows/langfuse.yml/badge.svg)](https://github.com/deepset-ai/haystack-core-integrations/actions/workflows/langfuse.yml)                                     |
+| [mcp-haystack](integrations/mcp/)                                                                    | Tracer                      | [![PyPI - Version](https://img.shields.io/pypi/v/mcp-haystack.svg?color=orange)](https://pypi.org/project/mcp-haystack)                        | [![Test / mcp](https://github.com/deepset-ai/haystack-core-integrations/actions/workflows/mcp.yml/badge.svg)](https://github.com/deepset-ai/haystack-core-integrations/actions/workflows/mcp.yml)                                     |
```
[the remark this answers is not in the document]
vblagoje: But it is alphabetic, no?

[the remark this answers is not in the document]
sjrl: It should come after llama-cpp-haystack which is one line below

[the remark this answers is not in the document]
sjrl: Also this isn't a Tracer. Should we call it Tool instead?
```suggestion
| [mcp-haystack](integrations/mcp/)                                                                    | Tool                      | [![PyPI - Version](https://img.shields.io/pypi/v/mcp-haystack.svg?color=orange)](https://pypi.org/project/mcp-haystack)                        | [![Test / mcp](https://github.com/deepset-ai/haystack-core-integrations/actions/workflows/mcp.yml/badge.svg)](https://github.com/deepset-ai/haystack-core-integrations/actions/workflows/mcp.yml)                                     |
```

integrations/mcp/pyproject.toml
```diff
@@ -0,0 +1,182 @@
...
+authors = [
+    { name = "deepset GmbH", email = "info@deepset.ai" },
+]
+classifiers = [
+    "License :: OSI Approved :: Apache Software License",
+    "Development Status :: 4 - Beta",
+    "Programming Language :: Python",
+    "Programming Language :: Python :: 3.10",
+    "Programming Language :: Python :: 3.11",
+    "Programming Language :: Python :: 3.12",
```
sjrl, line 26: After adding python 3.13 to the CI I would also add it as a supported here

integrations/mcp/src/haystack_integrations/tools/mcp/mcp_tool.py
```diff
@@ -0,0 +1,625 @@
...
+            # Process and validate the response content
+            processed_result = self._process_response_content(result, tool_name)
+            return processed_result
+
+        except Exception as e:
+            if isinstance(e, MCPError):
+                # Re-raise specific MCP errors
+                raise
+            message = f"Failed to invoke tool '{tool_name}': {e!s}"
+            raise MCPInvocationError(message) from e
```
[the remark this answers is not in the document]
sjrl: Yeah I'm a little confused as to why we need to have specific MCPErrors defined by us?

If we think the additional context of `message = f"Failed to invoke tool '{tool_name}': {e!s}"` is useful we should just always include it no matter what the Exception is

integrations/mcp/src/haystack_integrations/tools/mcp/mcp_tool.py
```diff
@@ -0,0 +1,625 @@
...
+
+    def _process_response_content(self, result, tool_name: str) -> Any:
```
sjrl, line 108: It seems like we could add a type for `result` which is `CallToolResult`. Could we go ahead and add it?

integrations/mcp/src/haystack_integrations/tools/mcp/mcp_tool.py
```diff
@@ -0,0 +1,625 @@
...
+
+        :param result: CallToolResult from MCP tool call
+        :param tool_name: Name of the called tool (for error messages)
+        :returns: The original CallToolResult object
+        :raises MCPResponseTypeError: If content type is not TextContent
+        :raises MCPInvocationError: If the tool call resulted in an error
+        """
+        # Return the result directly if it's not a CallToolResult structure
+        if not hasattr(result, "content"):
+            return result
```
sjrl, line 120: Why do we return a result if it doesn't abide by the `CallToolResult` structure?

integrations/mcp/src/haystack_integrations/tools/mcp/mcp_tool.py
```diff
@@ -0,0 +1,625 @@
...
+        if hasattr(result, "isError") and result.isError:
+            error_msg = "Tool execution failed"
+            if result.content and isinstance(result.content, list) and len(result.content) > 0:
+                first_item = result.content[0]
+                if hasattr(first_item, "text"):
+                    error_msg = first_item.text
+                else:
+                    error_msg = str(first_item)
+            message = f"Tool '{tool_name}' returned an error: {error_msg}"
+            raise MCPInvocationError(message)
```
sjrl, line 132: Could we just use a built in python Error type like `RuntimeError`?

integrations/mcp/src/haystack_integrations/tools/mcp/mcp_tool.py
```diff
@@ -0,0 +1,625 @@
...
+        # Validate content types - only allow TextContent
+        if result.content and isinstance(result.content, list):
+            for item in result.content:
+                if not (hasattr(item, "type") and item.type == "text" and hasattr(item, "text")):
+                    # Reject any non-TextContent
+                    message = (
+                        f"Unsupported content type in response from tool '{tool_name}'. "
+                        f"Only TextContent is currently supported."
+                    )
+                    raise MCPResponseTypeError(message)
```
sjrl, line 143: Same here why not just raise a native `TypeError`?

integrations/mcp/src/haystack_integrations/tools/mcp/mcp_tool.py
```diff
@@ -0,0 +1,625 @@
...
+        except Exception as e:
+            if isinstance(e, MCPError):
+                # Re-raise specific MCP errors
+                raise
+            message = f"Failed to invoke tool '{tool_name}': {e!s}"
+            raise MCPInvocationError(message) from e
+
+    def _process_response_content(self, result, tool_name: str) -> Any:
+        """
+        Process response content from an MCP tool call, accepting only TextContent.
```
sjrl, line 110: I think the name of the function is a bit broad. I originally thought `_process` meant you'd do something to the result, but here we are just validating it meets a certain format. So could we rename the function something like `_validate` or `_check` instead?

integrations/mcp/src/haystack_integrations/tools/mcp/mcp_tool.py
vblagoje: Mistake 🙏

LastRemote:
Copying the comment from the other PR: This feels so much smoother than configuring Claude Desktop. One stupid question though, does `StdioMCPServerInfo` respect the conda or venv runtime environment to run local MCP server? I am not exactly sure how stdio should be used.

vblagoje:
Thanks @LastRemote - iirc it does, see examples. Although I notice pros seems to prefer npx and uvx to run local programs to hook stdio into. I'm still learning myself 😎

sjrl:
> @sjrl it seems that you are arguing for removal of MCPError, MCPConnectionError, MCPToolNotFoundError etc exceptions, is that correct? What is your line of thinking there? I find that in these relatively complex async/sync setups it helps to have nice exception hierarchy to quickly understand what is going on, even though users won't catch these specific exceptions and act upon them. I think it is also suggested to do that in Zen of Python. Perhaps we can narrow down to few most important ones, without going overboard - but anyways, regardless - would love to hear your thoughts - why.

After thinking it over I believe you're right that having specific MCP errors like this could be helpful rather than using the generic errors from python. I think what we are missing is to take advantage of them and save additional info about each type of error to make it easier to debug. For example if you take a look at the errors defined in [OpenAI's SDK](https://github.com/openai/openai-python/blob/a6b493071b843bec3db807637e441c1768b695f8/src/openai/_exceptions.py#L34) you can see that they store additional class variables like `message`, `request`, etc. which I think can be quite powerful for debugging for users. 

For example, your MCPResponseTypeError could probably follow a similar structure as OpenAI's [APIResponseValidationError](https://github.com/openai/openai-python/blob/a6b493071b843bec3db807637e441c1768b695f8/src/openai/_exceptions.py#L69)

integrations/mcp/examples/mcp_http_client.py
```diff
@@ -0,0 +1,41 @@
+# SPDX-FileCopyrightText: 2022-present deepset GmbH <info@deepset.ai>
```
anakin87, line 1: Since there are several useful examples, would it make sense to create a README.md file in integrations/mcp/examples, which briefly introduces and links the different examples?

mpangrazzi (commented):
I've left two comments - IMHO in general I think we should refer to SSE directly and not to HTTP, since MCP servers are mainly using `stdio` and `SSE` as transports (see [here](https://modelcontextprotocol.io/docs/concepts/transports#built-in-transport-types)).

Of course, I know that SSE is over HTTP :) but they're actually quite different things - HTTP is request-response while SSE is server-push.

However, it's reasonable to expect a HTTP transport implementation in the future ;) But I guess it will be different from the currently support SSE one.

Last note: also Cursor (MCP client) refers to `SSE` and `stdio` as transport when you add an MCP server:
Screenshot 2025-03-11 at 16 10 56

integrations/mcp/src/haystack_integrations/tools/mcp/mcp_tool.py
```diff
@@ -0,0 +1,615 @@
...
+        # Remove the type field as it's not a constructor parameter
+        data_copy = data.copy()
+        data_copy.pop("type", None)
+
+        # Create an instance of the class with the remaining fields
+        return cls(**data_copy)
+
+
+@dataclass
+class HttpMCPServerInfo(MCPServerInfo):
```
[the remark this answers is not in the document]
vblagoje: Yes, I agree. Good point @mpangrazzi - I'll adjust the PR for this name change. 🙏

vblagoje:
As @mpangrazzi correctly HTTP -> SEE renaming. I think I'll rename the `HttpMCPServerInfo` to `SSEServerInfo` and `StdioMCPServerInfo` to `StdioServerInfo` as MCP is really implied here and superflous imho. LMK your thoughts as well @sjrl @anakin87 @davidsbatista - this is super important tool and every detail does matter.

vblagoje:
Amazing feedback everyone!!! @anakin87 @sjrl @mpangrazzi @davidsbatista lmk if I missed something to address.

anakin87 (approved):
I'm a bit concerned about having multiple nested try/except blocks, but I think that as a first version, it can be OK.

Please @mpangrazzi take a final look too...

@vblagoje I would also create an issue with [this template](https://github.com/deepset-ai/haystack-core-integrations/blob/main/.github/ISSUE_TEMPLATE/new-integration-proposal.md) to make sure that you are not forgetting something on this new integration.

PR #1489, opened 2025-03-10
test: llama.cpp - do not test on Windows

anakin87:
### Related Issues
Windows test have been failing for several days.
The issue is on llama-cpp-python, even if I cannot find a single GitHub issue that tracks this problem; there are several ones related to compilation problems.
I tried several approaches in #1466 (pinning to a previous version, using pre-built wheels...), but none of them worked.

### Proposed Changes:
Do not test on Windows. (We can change that in the future)

### How did you test it?
CI

### Checklist

### DIFF
--- a/.github/workflows/llama_cpp.yml
+++ b/.github/workflows/llama_cpp.yml
@@ -30,8 +30,8 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [ubuntu-latest, windows-latest, macos-latest]
-        python-version: ["3.9", "3.10"]
+        os: [ubuntu-latest, macos-latest] # we don't test on windows becaus of issues on llama-cpp-python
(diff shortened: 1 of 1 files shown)


.github/workflows/llama_cpp.yml
```diff
@@ -30,8 +30,8 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [ubuntu-latest, windows-latest, macos-latest]
-        python-version: ["3.9", "3.10"]
+        os: [ubuntu-latest, macos-latest] # we don't test on windows becaus of issues on llama-cpp-python
```
davidsbatista, line 33:
```suggestion
        os: [ubuntu-latest, macos-latest] # we don't test on windows because of issues on llama-cpp-python
```

davidsbatista (approved):
LGTM

PR #1490, opened 2025-03-10
fix: pgvector - improve `_create_table_if_not_exists` to be used without admin rights

### LINKED ISSUE #1354: Do not require admin rights to use the pgvector integration
At present, the Pgvector integration starts by checking whether a table for the index exists using `CREATE TABLE ... IF NOT EXISTS`. Unfortunately in Postgres a process executing this requires CREATE TABLE privileges even if the table already exists. This means we have to give our technical user DDL rights even though we know it should only ever read data.

It would be better to attempt a `SELECT` and go on to create the table if the `SELECT` fails.

**Additional context**
https://github.com/deepset-ai/haystack-core-integrations/blob/63738bddfd795edc7b3608ef6290ddfe24be6e38/integrations/pgvector/src/haystack_integrations/document_stores/pgvector/document_store.py#L311

anakin87:
### Related Issues

- fixes #1354

### Proposed Changes:
- modify the `_create_table_if_not_exists` method to be usable by non-admin users when the table already exists

### How did you test it?
CI; added a new test for the method

### Checklist

### DIFF
--- a/integrations/pgvector/src/haystack_integrations/document_stores/pgvector/document_store.py
+++ b/integrations/pgvector/src/haystack_integrations/document_stores/pgvector/document_store.py
@@ -23,7 +23,7 @@ from .filters import _convert_filters_to_where_clause_and_params
 logger = logging.getLogger(__name__)
 
 CREATE_TABLE_STATEMENT = """
-CREATE TABLE IF NOT EXISTS {schema_name}.{table_name} (
+CREATE TABLE {schema_name}.{table_name} (
 id VARCHAR(128) PRIMARY KEY,
 embedding VECTOR({embedding_dimension}),
 content TEXT,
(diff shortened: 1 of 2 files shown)


integrations/pgvector/src/haystack_integrations/document_stores/pgvector/document_store.py
```diff
@@ -23,7 +23,7 @@
 logger = logging.getLogger(__name__)
 
 CREATE_TABLE_STATEMENT = """
-CREATE TABLE IF NOT EXISTS {schema_name}.{table_name} (
```
Amnah199, line 26: Whats the reason of removing this?

Amnah199 (approved):
LG!

PR #1087, opened 2024-09-16
feat: Langfuse - support generation span for more LLMs

vblagoje:
- Adds GENERATION span (tag) Langfuse support for HuggingFace, Anthropic and Cohere LLMs
- Adds token counting (except HuggingFaceAPIChatGenerator and HuggingFaceAPIGenerator)
- Cohere support  is contingent to https://github.com/deepset-ai/haystack-core-integrations/pull/1086
- All tests were done manually in addition to updated automated integration tests.

- fixes https://github.com/deepset-ai/haystack-core-integrations/issues/887

- Screenshot below show Anthropic sonnet, other LLMs look similar
Screenshot 2024-09-16 at 15 19 01

PR #1577, opened 2025-03-30
feat: adapt Ollama metadata to OpenAI format; support Ollama in Langfuse

### LINKED ISSUE #1463: haystack langfuse integration: SUPPORTED_GENERATORS meta tracing question.
Related issue: #1153 #1310 #1475 #1423 

How should we handle the meta data from the generators?

The current setting is:

https://github.com/henryclw/deepset-ai-haystack-core-integrations/blob/00c43748edac70fbcd4efbca1a467138446c7edf/integrations/langfuse/src/haystack_integrations/tracing/langfuse/tracer.py#L264-L285

```python
                span._span.update(
                    usage=meta.get("usage") or None,
                    model=meta.get("model"),
                    completion_start_time=completion_start_time,
                )
```

But this only works with the selected generators.

```python
_SUPPORTED_GENERATORS = [
    "AzureOpenAIGenerator",
    "OpenAIGenerator",
    "AnthropicGenerator",
    "HuggingFaceAPIGenerator",
    "HuggingFaceLocalGenerator",
    "CohereGenerator",
]
_SUPPORTED_CHAT_GENERATORS = [
    "AzureOpenAIChatGenerator",
    "OpenAIChatGenerator",
    "AnthropicChatGenerator",
    "HuggingFaceAPIChatGenerator",
    "HuggingFaceLocalChatGenerator",
    "CohereChatGenerator",
]
_ALL_SUPPORTED_GENERATORS = _SUPPORTED_GENERATORS + _SUPPORTED_CHAT_GENERATORS
```

For example, the `OllamaChatGenerator` could also return the token counts information in response, but it's not in the `usage` key under the `meta` dict. Should we manually add a convergence to form a `usage` dict? If so, should this code be placed under langfuse code folder or ollama code folder? And should we address this issue in the docs / faq? Cause lots of people are unaware of this. The LLM tracing actually only works with certain generators. I think this might need to bring up in a discussion, of how the code should work and how we should deal with future situations. (Say more different generators are integrated)

Currently I have worked around and place my code in ollama folder. If this is the preferred way after discussion, I'm more than happy to open a PR.

~~My question is: the current tracer.py~~

~~https://github.com/deepset-ai/haystack-core-integrations/blob/main/integrations/langfuse/src/haystack_integrations/tracing/langfuse/tracer.py#L34~~

~~What is the point of setting `SUPPORTED_GENERATORS`? According to haystack documents, there are many more generators except for these few. <https://docs.haystack.deepset.ai/docs/generators>~~
~~For example, personally I use [OllamaChatGenerator](https://docs.haystack.deepset.ai/docs/ollamachatgenerator), why is the [OpenAIChatGenerator](https://docs.haystack.deepset.ai/docs/openaichatgenerator) in the `SUPPORTED_GENERATORS` list while the OllamaChatGenerator is not? They both have the compatible input output type for tracing I assume.~~

~~Suggestion:~~
~~- Clarify the meaning of `SUPPORTED_GENERATORS`~~
~~- Remove the `SUPPORTED_GENERATORS` or add every official implemented generator in the list.~~

henryclw:
### Related Issues

- fixes #1463
- 
### Proposed Changes:

Add support for OllamaChatGenerator in Langfuse tracer, log the meta data in ollama (e.g. model name)

### How did you test it?

Manually tested on my local machine with ollama set up.


anakin87 (commented):
Hello and thanks for the contribution!

The goal of this PR makes sense to me.

Please sign the CLA and fix the linting error.

henryclw:
@anakin87 Thank you for the kind remind. I just finished up everything. Hopefully this could be merge. Have a good one!

vblagoje:
Perhaps you already implemented this @henryclw but please also make sure you have added support for token counting, see details in https://github.com/deepset-ai/haystack-core-integrations/pull/1087 but also note that the codebase has been somewhat refactored when this PR was integrated.

anakin87:
> make sure you have added support for token counting
@vblagoje what do you mean specifically?

anakin87 (approved):
I've pushed some fixes + a test.

This PR is good to go. Thanks @henryclw!

PR #1591, opened 2025-04-03
fix: Ensure common package names are treated as namespace packages

sjrl:
### Related Issues

- fixes #issue-number

### Proposed Changes:

 
 
Remove inits in folders that should be treated as namespace packages instead of regular packages. 

More info [here](https://packaging.python.org/en/latest/guides/packaging-namespace-packages/#native-namespace-packages) but basically for package names that are shared across integrations e.g. `haystack_integrations`, `components`, `retrievers`, etc. we want to make sure these are treated as namespace packages so the namespaces are shareable when installing multiple integrations at once. 

### Checklist

### DIFF
--- a/integrations/amazon_bedrock/src/haystack_integrations/__init__.py
+++ b/integrations/amazon_bedrock/src/haystack_integrations/__init__.py
@@ -1,3 +1,0 @@
-# SPDX-FileCopyrightText: 2023-present deepset GmbH <info@deepset.ai>
-#
-# SPDX-License-Identifier: Apache-2.0
--- a/integrations/amazon_bedrock/src/haystack_integrations/common/__init__.py
+++ b/integrations/amazon_bedrock/src/haystack_integrations/common/__init__.py
@@ -1,3 +1,0 @@
-# SPDX-FileCopyrightText: 2023-present deepset GmbH <info@deepset.ai>
-#
-# SPDX-License-Identifier: Apache-2.0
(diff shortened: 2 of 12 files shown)


anakin87 (commented):
This seems to make sense to me.

- Could you describe what's the current error we get?
- (in general, for PRs that target several integrations to be released, it would be better to do different PRs, in order for changelog generation to work well)

sjrl:
> This seems to make sense to me.
> 
> * Could you describe what's the current error we get?
> * (in general, for PRs that target several integrations to be released, it would be better to do different PRs, in order for changelog generation to work well)

Ahh okay I can split this up into separate PRs so we can easily make releases. 

I ran into this error when I tried adding __init__.py files to all package levels in langfuse and then hatch run test stopped working due to an import error. When I removed the __init__.py files from folders that shared the same name between packages the import errors disappeared. Then I learned more about namespace packages and realized that we should remain as consistent as possible across our integrations. 

This error doesn't always appear (e.g. installing amazon-bedrock-haystack and cohere-haystack via pip install) still seems to work and not lead to import errors, but from what I've read on the topic this can be fragile (E.g. why the hatch run test fails).

anakin87 (commented):
Ah, it's clearer now...

I would say that in this case, a single PR may still be fine, but it is possible that the changelog will become messy and you may have to fix it manually.
However, the decision is up to you...

anakin87 (approved):

sjrl:
Closing in favor of separate PRs for each integration to make releases easier

PR #1603, opened 2025-04-03
fix: perform case-insensitive check on changelog generation

anakin87:
### Related Issues
We recently improved changelog generation to exclude commits that only perform changelog updates
However, some of these commits are still included: see https://github.com/deepset-ai/haystack-core-integrations/pull/1599 for an example.

After investigation, I discovered that in this case, there have been manual edits and the commit message is different: see https://github.com/deepset-ai/haystack-core-integrations/commit/63738bddfd795edc7b3608ef6290ddfe24be6e38. It includes "update changelog for" (lowercase).

### Proposed Changes:
- update cliff.toml to perform a case-insensitive check using [`lower` filter](https://keats.github.io/tera/docs/#lower).

### Checklist

### DIFF
--- a/cliff.toml
+++ b/cliff.toml
@@ -25,9 +25,9 @@ body = """
         If we don't do this we get an empty section since we don't show
         commits that update the changelog
     #}\
-    {% if commits | length == 1 and 'Update changelog for' in commits[0].message | lower %}\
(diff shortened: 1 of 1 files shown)


sjrl (approved):
Looks good!

PR #1618, opened 2025-04-04
fix: Update test and unpin chroma

sjrl:
### Related Issues

- fixes #1609 

### Proposed Changes:

 
 

Chroma updated what is returned when calling list_collections() making an if statement fail. Updated the code to accommodate this.

### How did you test it?

Existing tests.

### Checklist

- I have read the [contributors guidelines](https://github.com/deepset-ai/haystack-core-integrations/blob/main/CONTRIBUTING.md) and the [code of conduct](https://github.com/deepset-ai/haystack-core-integrations/blob/main/CODE_OF_CONDUCT.md)
- I have updated the related issue with new insights and changes
- I added unit tests and updated the docstrings
- I've used one of the [conventional commit types](https://www.conventionalcommits.org/en/v1.0.0/) for my PR title: `fix:`, `feat:`, `build:`, `chore:`, `ci:`, `docs:`, `style:`, `refactor:`, `perf:`, `test:`.

PR #1619, opened 2025-04-04
fix: Make chroma backwards compatible

sjrl:
### Related Issues

Follow up of https://github.com/deepset-ai/haystack-core-integrations/pull/1618

### Proposed Changes:

 
 
Make it so the integration works with older versions of chroma

### Checklist


julian-risch (approved):
LGTM! 👍 
That's good. Other option is to limit to >= 1.0.0 here https://github.com/deepset-ai/haystack-core-integrations/blob/fix-chroma/integrations/chroma/pyproject.toml#L26
That would be fair to in my opinion.

PR #1626, opened 2025-04-08
feat: Add MCPToolset

vblagoje:
### Why:
This PR introduces the `MCPToolset`, a dynamic integration layer that connects Haystack with MCP-compliant servers. It enables automatic discovery, access, and invocation of tools hosted remotely (via SSE) or locally (via stdio), allowing users to incorporate external toolchains into their Haystack pipelines with minimal configuration.

### What:
- Added `MCPToolset` class to manage and dynamically discover tools from MCP-compliant servers.
- Implemented examples demonstrating usage of `MCPToolset` with both SSE and stdio connections.
- Modified `pyproject.toml` to adjust dependencies and update configurations.

### How can it be used:
```python
import os

from haystack import Pipeline
from haystack.components.converters import OutputAdapter
from haystack.components.generators.chat import OpenAIChatGenerator
from haystack.components.tools import ToolInvoker

from haystack_integrations.tools.mcp import MCPToolset, StdioServerInfo


def main():
    # Create server info for the time service
    server_info = StdioServerInfo(command="uvx", args=["mcp-server-time", "--local-timezone=Europe/Berlin"])

    # Create the toolset - this will automatically discover all available tools
    mcp_toolset = MCPToolset(server_info)
    pipeline = Pipeline()
    pipeline.add_component("llm", OpenAIChatGenerator(model="gpt-4o-mini", tools=mcp_toolset))
    pipeline.add_component("tool_invoker", ToolInvoker(tools=mcp_toolset))
    pipeline.add_component(
        "adapter",
        OutputAdapter(
            template="{{ initial_msg + initial_tool_messages + tool_messages }}",
            output_type=list[ChatMessage],
            unsafe=True,
        ),
    )
    pipeline.add_component("response_llm", OpenAIChatGenerator(model="gpt-4o-mini"))
    pipeline.connect("llm.replies", "tool_invoker.messages")
    pipeline.connect("llm.replies", "adapter.initial_tool_messages")
    pipeline.connect("tool_invoker.tool_messages", "adapter.tool_messages")
    pipeline.connect("adapter.output", "response_llm.messages")

    user_input = "What is the time in New York? Be brief."  # can be any city
    user_input_msg = ChatMessage.from_user(text=user_input)

    result = pipeline.run({"llm": {"messages": [user_input_msg]}, "adapter": {"initial_msg": [user_input_msg]}})

    print(result["response_llm"]["replies"][0].text)


if __name__ == "__main__":
    main()

...
```

### How did you test it:
- Used a mix of unit tests with mock objects to simulate MCPToolset behavior, and integration tests to validate interactions with a mock server implementing the MCP, ensuring tools are discovered and functional.
- See examples as well

### Notes for the reviewer:
 - Review _run_sync in detail
 - Agree with integration strategy, see `integrations/mcp/pyproject.toml` diff


anakin87:
I ask @mpangrazzi to do a review for me, since he has more experience with MCP (I had already discussed it with him).

integrations/mcp/src/haystack_integrations/tools/mcp/mcp_toolset.py
```diff
@@ -0,0 +1,201 @@
...
+
+logger = logging.getLogger(__name__)
+
+
+class MCPToolset(Toolset):
+    """
+    A Toolset that connects to an MCP (Model Context Protocol) server and provides
+    access to its tools.
+
+    MCPToolset dynamically discovers and loads tools from any MCP-compliant server,
```
bilgeyucel, line 23:
```suggestion
    MCPToolset dynamically discovers and loads all tools from any MCP-compliant server,
```

bilgeyucel (commented):
LGTM! Left one small comment

integrations/mcp/src/haystack_integrations/tools/mcp/mcp_toolset.py
```diff
@@ -0,0 +1,201 @@
...
+        self.connection_timeout = connection_timeout
+        self.invocation_timeout = invocation_timeout
+
+        # Connect and load tools
+        try:
+            # Create the appropriate client using the factory method
+            client = self.server_info.create_client()
+
+            # Connect and get available tools
+            tools = self._run_sync(client.connect(), timeout=self.connection_timeout)
```
mpangrazzi, line 105: I think this might cause the constructor to fail if the server is unavailable or connection times out. What about deferring the connection in another method which actually list tools?

integrations/mcp/src/haystack_integrations/tools/mcp/mcp_toolset.py
```diff
@@ -0,0 +1,201 @@
...
+    def _run_sync(self, coro, timeout: float | None = None):
+        """
+        Run an async coroutine synchronously.
+
+        :param coro: The coroutine to run
+        :param timeout: Optional timeout in seconds
+        :returns: The result of the coroutine
+        :raises TimeoutError: If the operation times out
+        """
+        try:
```
mpangrazzi, line 132: Why don't simply use `asyncio.run(asyncio.wait_for(coro, timeout))` here? It will handle event loop management automatically.

integrations/mcp/src/haystack_integrations/tools/mcp/mcp_toolset.py
```diff
@@ -0,0 +1,201 @@
...
+        """
+        inner_data = data["data"]
+
+        # Reconstruct the server_info object
+        server_info_dict = inner_data.get("server_info", {})
+        server_info_class = import_class_by_name(server_info_dict["type"])
+        server_info = server_info_class.from_dict(server_info_dict)
+
+        # Create a new MCPToolset instance
+        return cls(
```
mpangrazzi, line 197: Here we are creating a new instance of the class (which is ok), but as commented above, it will call `__init__` and so call the remote server, which can be unwanted. It will be solved if we defer the connection on a method which e.g. list tools. WDYT?

integrations/mcp/src/haystack_integrations/tools/mcp/mcp_toolset.py
vblagoje: I know I know but we are constrained here in Tool because we don't have lifecycle method to bootstrap the Tool like we have say in component - warm_up. So we have to do it in constructor. At least for now - but @anakin87 and a few of us can have a discussion if there is a way to enhance the Tool strapping in the future.

integrations/mcp/src/haystack_integrations/tools/mcp/mcp_toolset.py
vblagoje: Yeah, I'll try it - there was some issue about event loop running and not running, detecting if it is and so on. I'll revisit 👍

integrations/mcp/src/haystack_integrations/tools/mcp/mcp_toolset.py
vblagoje: Yeah, I wish we could. See above.

integrations/mcp/src/haystack_integrations/tools/mcp/mcp_toolset.py
mpangrazzi: For sync scenarios (like this one), [asyncio.run](https://docs.python.org/3/library/asyncio-runner.html#running-an-asyncio-program) should be enough to sync-run a coroutine. For async ones, I expect you will implement a `_run_async`?

sjrl: I think I agree with @mpangrazzi here to make this simpler. We opted not to do this in AsyncPipeline currently and to avoid the `This event loop is already running` we added 
```python
import nest_asyncio
nest_asyncio.apply()
```
to the top of the Jupyter notebook we were running. See [here](https://haystack.deepset.ai/cookbook/async_pipeline).

When are you running into the `"This event loop is already running" errors`?

sjrl: Also if we really do need to keep this it seems like we have an almost duplicate implementation of `_run_sync` in `mcp_tool.py`. Could we reuse that here especially since the method could be static in both?

integrations/mcp/src/haystack_integrations/tools/mcp/mcp_toolset.py
```diff
@@ -0,0 +1,222 @@
...
+        # Store configuration
+        self.server_info = server_info
+        self.tool_names = tool_names
+        self.connection_timeout = connection_timeout
+        self.invocation_timeout = invocation_timeout
+
+        # Connect and load tools
+        try:
+            # Create the appropriate client using the factory method
+            client = self.server_info.create_client()
```
sjrl, line 116: Sorry if this is a silly question, but is this always an async client?

integrations/mcp/src/haystack_integrations/tools/mcp/mcp_toolset.py
vblagoje: Yes, will reuse it, god point!

vblagoje: > 
> When are you running into the `"This event loop is already running" errors`?

I wish it was that simple - `asyncio.run` cannot be called when another asyncio event loop is running in the same thread. So when I use this approach in hayhooks - I'm running into "This event loop is already running" there. As we talked internally It is not easy to satisfy all the requirements users expect - having MCPTool work flawlessly in sync/async context, in standalone scripts, in hayhooks deployments, all without any additional explicit code on user's side.

vblagoje:
@sjrl @mpangrazzi I have a nice solution for `MCPToolset` and builds upon ideas from https://github.com/deepset-ai/haystack-core-integrations/pull/1643 that Michele and I agreed on - but I think we can't integrate this one yet as we need Toolset from Haystack main to be available, and it is not yet until we release next version of Haystack. Because we can't integrate this PR with dependency on the Haystack main as it is now. See [this](https://github.com/deepset-ai/haystack-core-integrations/pull/1626/files#diff-835a96f91c1c36155e5a5b53d45b4c0ae8ff46b75a57ab6827bd068291903733)

So we wait for next Haystack release? cc @julian-risch

vblagoje:
@sjrl and @mpangrazzi I wanted to point out this [commit](https://github.com/deepset-ai/haystack-core-integrations/pull/1626/commits/01eb16c8d0f8eb4a6cac3c8870c4edc01e0a3b11) as it may look unusual to use bare Tool instance rather than MCPTool - which feels more natural. The reason is that all the underlying tools in a particular Toolset are "hosted" on the same target MCP server. So it is not necessary to make x many underlying connections to the same MCP server via each MCPTool instance when we can, for the lack of a better word, - multiplex all the tool invocation via the same client connection. From all the examples I've tried this works well. And it is internal detail to MCPToolset so we can always tune it.

vblagoje:
@mpangrazzi @sjrl I really battle tested this one from dev X perspective with the upcoming agent + mcp demo that I needed to do. I've ran into several issues in multiple MCP server + agent setup and I've noticed how error messages could be improved to empower users to quickly diagnose what has gone wrong. The last few commits aim to alleviate the shortcomings and provide first-rate experience with mcp toolset. 🙏

integrations/mcp/src/haystack_integrations/tools/mcp/mcp_toolset.py
```diff
@@ -0,0 +1,232 @@
...
+
+        :param server_info: Connection information for the MCP server
+        :param tool_names: Optional list of tool names to include. If provided, only tools with
+                          matching names will be added to the toolset.
+        :param connection_timeout: Timeout in seconds for server connection
+        :param invocation_timeout: Default timeout in seconds for tool invocations
+        :raises MCPToolNotFoundError: If any of the specified tool names are not found on the server
+        """
+        # Initialize with empty tools list first
+        super().__init__(tools=[])
```
sjrl, line 109: Small nit, why not super at the end once we have all the tools set up instead of using `self.add`?

integrations/mcp/src/haystack_integrations/tools/mcp/mcp_toolset.py
```diff
@@ -0,0 +1,232 @@
...
+                    message = (
+                        f"The following tools were not found: {', '.join(missing_tools)}. "
+                        f"Available tools: {', '.join(available_tools)}"
+                    )
+                    raise MCPToolNotFoundError(
+                        message=message, tool_name=next(iter(missing_tools)), available_tools=list(available_tools)
+                    )
+
+            # This is the function will give Tool instance to invoke
+            def create_invoke_tool(client, tool_name, invocation_timeout):
```
sjrl, line 139: Please use different variable names for `client` and `invocation_timeout`. I get this warning from my IDE `Shadows name 'invocation_timeout' from outer scope `, which basically means we are doing something risky here since these variables are defined already in the outer scope of this function. 

So it's safest if we avoid this name collision to avoid unexpected behavior.

integrations/mcp/src/haystack_integrations/tools/mcp/mcp_toolset.py
```diff
@@ -0,0 +1,232 @@
...
+                def invoke_tool(**kwargs) -> Any:
+                    """Invoke a tool using the existing client and AsyncExecutor."""
+                    result = AsyncExecutor.get_instance().run(
+                        client.call_tool(tool_name, kwargs), timeout=invocation_timeout
+                    )
+                    return result
+
+                return invoke_tool
+
+            # Create Tool instances not MCPTool for each available tool and add them
```
sjrl, line 149: Maybe also a naive question but why aren't we using MCPTool here? From an outside perspecitve it does feel a little odd that a MCPToolSet is not a Set of MCPTools. Totally fine if this is how it needs to be, just curious.

integrations/mcp/src/haystack_integrations/tools/mcp/mcp_toolset.py
```diff
@@ -0,0 +1,232 @@
...
+            for tool_info in tools:
+                # Skip tools not in the tool_names list if tool_names is provided
+                if self.tool_names is not None and tool_info.name not in self.tool_names:
+                    logger.debug(f"Skipping tool '{tool_info.name}' as it's not in the requested tool_names list")
```
sjrl, line 153: Once we use the Haystack logger we will need to update all logging statements slightly. Check out this [comment](https://github.com/deepset-ai/haystack-core-integrations/issues/1409) to see how.

integrations/mcp/src/haystack_integrations/tools/mcp/mcp_toolset.py
```diff
@@ -0,0 +1,232 @@
...
+    # Use the toolset as shown in the pipeline example above
+    ```
+    """
+
+    def __init__(
+        self,
+        server_info: MCPServerInfo,
+        tool_names: list[str] | None = None,
+        connection_timeout: int = 30,
+        invocation_timeout: int = 30,
```
sjrl, line 96: The timeouts are of type float not int. E.g. Your run method of `AsyncExecutor` has this signature
```python
def run(self, coro: Coroutine[Any, Any, Any], timeout: float | None = None) -> Any:
```

integrations/mcp/src/haystack_integrations/tools/mcp/mcp_toolset.py
```diff
@@ -0,0 +1,232 @@
+import logging
+from typing import Any
+
+import httpx
+from exceptiongroup import ExceptionGroup
```
sjrl, line 5: Could we add `httpx` as explicit dependencies to the pyproject.toml? It looks like right now we are relying on this being imported by mcp or haystack.

integrations/mcp/src/haystack_integrations/tools/mcp/mcp_toolset.py
```diff
@@ -0,0 +1,232 @@
...
+class MCPToolset(Toolset):
+    """
+    A Toolset that connects to an MCP (Model Context Protocol) server and provides
+    access to its tools.
+
+    MCPToolset dynamically discovers and loads all tools from any MCP-compliant server,
+    supporting both network-based SSE connections and local process-based stdio connections.
+    This dual connectivity allows for integrating with both remote and local MCP servers.
+
+    Example using MCPToolset in a Haystack Pipeline:
```
sjrl, line 30: Could we add to this example all steps needed to get it to run? I just tried copy-pasting and ran into an issue about `uvx` not being found. 

It'd be great if we could suggest what keys and additional things you need to install to be able to use this example as comments within the python code block.

integrations/mcp/src/haystack_integrations/tools/mcp/mcp_toolset.py
```diff
@@ -0,0 +1,232 @@
+import logging
```
sjrl, line 1: Also let's add the deepset license header to this file

integrations/mcp/examples/google_maps_agent.py
```diff
@@ -0,0 +1,93 @@
...
+
+def main():
+    # tracer = LangfuseConnector("Agent google maps search")
+    # Optionally, you can use Langfuse to trace the agent's activity but it needs
+    # additional configuration.
+    # See [Langfuse integration](https://github.com/deepset-ai/haystack-core-integrations/tree/main/integrations/langfuse) for more information.
+    toolset = MCPToolset(
+        SSEServerInfo(base_url="http://localhost:8000"), tool_names=["maps_geocode", "maps_search_places"]
+    )
+    agent = Agent(chat_generator=OpenAIChatGenerator(model="gpt-4.1"), tools=toolset)
```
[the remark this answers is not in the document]
vblagoje: Good catch, we need to update all CG integrations to be like https://github.com/deepset-ai/haystack/blob/main/haystack/components/generators/chat/openai.py#L91 and also agent 🙏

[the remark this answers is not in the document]
sjrl: Okay I gotcha. Could you open an issue for this?

integrations/mcp/src/haystack_integrations/tools/mcp/mcp_toolset.py
sjrl: Thanks!

integrations/mcp/examples/google_maps_agent.py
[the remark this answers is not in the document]
sjrl: Being tracked here https://github.com/deepset-ai/haystack/issues/9283#issuecomment-2821029324

sjrl:
Hey @vblagoje I'm seeing a lot of tests failing silently https://github.com/deepset-ai/haystack-core-integrations/actions/runs/14592655087/job/40931223373#step:10:53

sjrl:
@vblagoje I have the feeling that we need to be using `@pytest.mark.asyncio` to mark tests that are using async operations. E.g. See https://github.com/deepset-ai/haystack-core-integrations/blob/ba0a310269262517bce2b702e4f5bd679b391863/integrations/opensearch/tests/test_document_store_async.py#L12

sjrl (approved):
Thanks for you work on this!

vblagoje:
Thank you @sjrl - After 2.13 gets release I'll just change the [dependency](https://github.com/deepset-ai/haystack-core-integrations/pull/1626/files#diff-835a96f91c1c36155e5a5b53d45b4c0ae8ff46b75a57ab6827bd068291903733R32) to haystack-ai

PR #1627, opened 2025-04-08
fix!: Make sure to JSON serialize objects before setting content tags

sjrl:
### Related Issues

- fixes #issue-number

### Proposed Changes:

 
 
We have a util called `coerce_tag_value` that we should be using when setting tags (especially when setting tags that consist of Haystack objects (e.g. Documents, ChatMessage, etc.) which makes sure that the tag value is JSON serializable **and** makes sure to call `to_dict` on objects like `Document`, `ChatMessage`, etc. when it's available. 

### How did you test it?

Tested locally

### Notes for the reviewer

Originally inspired by the fact that I saw ChatMessage being serialized with the private variables `_role` instead of the non-private ones. 

**Before**
Screenshot 2025-04-08 at 13 31 37

 **After**
Screenshot 2025-04-08 at 13 30 59

### Checklist


sjrl:
@anakin87 let me know if you think this warrants a breaking change label and a major version update. It would technically change the keys of objects in new traces as shown for ChatMessage in the PR description.

anakin87 (approved):
Ok for me.

Yes, this is a breaking change and it's appropriate to release a new major version.

I would change the PR title to express that this is a breaking change (`fix!:`)
and maybe choose a more explicit PR title for people reading the changelog.

PR #1637, opened 2025-04-10
feat: Add GitHub integration with agent_prompts and github_components

julian-risch:
### Related Issues

- Related to https://github.com/deepset-ai/haystack-experimental/issues/250
@sjrl brought up that one way to keep example files from the experimental Agent around is an integration https://github.com/deepset-ai/haystack-experimental/pull/263#issuecomment-2775755734

### Proposed Changes:

- Move github_components from experimental to a new integration
- Move agent_prompts from experimental to a new integration

The idea is to enable users to run the example notebook (or a version with updated imports) after having installed this new integration

### How did you test it?

Didn't test it for now. Just a draft.

### Checklist

### DIFF
--- a/integrations/github-haystack/LICENSE.txt
+++ b/integrations/github-haystack/LICENSE.txt
@@ -1,0 +1,73 @@
+Apache License
+Version 2.0, January 2004
+http://www.apache.org/licenses/
+
+TERMS AND CONDITIONS FOR USE, REPRODUCTION, AND DISTRIBUTION
+
+1. Definitions.
+
+"License" shall mean the terms and conditions for use, reproduction, and distribution as defined by Sections 1 through 9 of this document.
+
+"Licensor" shall mean the copyright owner or entity authorized by the copyright owner that is granting the License.
+
+"Legal Entity" shall mean the union of the acting entity and all other entities that control, are controlled by, or are under common control with that entity. For the purposes of this definition, "control" means (i) the power, direct or indirect, to cause the direction or management of such entity, whether by contract or otherwise, or (ii) ownership of fifty percent (50%) or more of the outstanding shares, or (iii) beneficial ownership of such entity.
+
+"You" (or "Your") shall mean an individual or Legal Entity exercising permissions granted by this License.
+
+"Source" form shall mean the preferred form for making modifications, including but not limited to software source code, documentation source, and configuration files.
+
+"Object" form shall mean any form resulting from mechanical transformation or translation of a Source form, including but not limited to compiled object code, generated documentation, and conversions to other media types.
+
+"Work" shall mean the work of authorship, whether in Source or Object form, made available under the License, as indicated by a copyright notice that is included in or attached to the work (an example is provided in the Appendix below).
+
+"Derivative Works" shall mean any work, whether in Source or Object form, that is based on (or derived from) the Work and for which the editorial revisions, annotations, elaborations, or other modifications represent, as a whole, an original work of authorship. For the purposes of this License, Derivative Works shall not include works that remain separable from, or merely link (or bind by name) to the interfaces of, the Work and Derivative Works thereof.
+
+"Contribution" shall mean any work of authorship, including the original version of the Work and any modifications or additions to that Work or Derivative Works thereof, that is intentionally submitted to Licensor for inclusion in the Work by the copyright owner or by an individual or Legal Entity authorized to submit on behalf of the copyright owner. For the purposes of this definition, "submitted" means any form of electronic, verbal, or written communication sent to the Licensor or its representatives, including but not limited to communication on electronic mailing lists, source code control systems, and issue tracking systems that are managed by, or on behalf of, the Licensor for the purpose of discussing and improving the Work, but excluding communication that is conspicuously marked or otherwise designated in writing by the copyright owner as "Not a Contribution."
+
+"Contributor" shall mean Licensor and any individual or Legal Entity on behalf of whom a Contribution has been received by Licensor and subsequently incorporated within the Work.
+
+2. Grant of Copyright License. Subject to the terms and conditions of this License, each Contributor hereby grants to You a perpetual, worldwide, non-exclusive, no-charge, royalty-free, irrevocable copyright license to reproduce, prepare Derivative Works of, publicly display, publicly perform, sublicense, and distribute the Work and such Derivative Works in Source or Object form.
+
+3. Grant of Patent License. Subject to the terms and conditions of this License, each Contributor hereby grants to You a perpetual, worldwide, non-exclusive, no-charge, royalty-free, irrevocable (except as stated in this section) patent license to make, have made, use, offer to sell, sell, import, and otherwise transfer the Work, where such license applies only to those patent claims licensable by such Contributor that are necessarily infringed by their Contribution(s) alone or by combination of their Contribution(s) with the Work to which such Contribution(s) was submitted. If You institute patent litigation against any entity (including a cross-claim or counterclaim in a lawsuit) alleging that the Work or a Contribution incorporated within the Work constitutes direct or contributory patent infringement, then any patent licenses granted to You under this License for that Work shall terminate as of the date such litigation is filed.
+
+4. Redistribution. You may reproduce and distribute copies of the Work or Derivative Works thereof in any medium, with or without modifications, and in Source or Object form, provided that You meet the following conditions:
+
+     (a) You must give any other recipients of the Work or Derivative Works a copy of this License; and
+
+     (b) You must cause any modified files to carry prominent notices stating that You changed the files; and
+
+     (c) You must retain, in the Source form of any Derivative Works that You distribute, all copyright, patent, trademark, and attribution notices from the Source form of the Work, excluding those notices that do not pertain to any part of the Derivative Works; and
+
+     (d) If the Work includes a "NOTICE" text file as part of its distribution, then any Derivative Works that You distribute must include a readable copy of the attribution notices contained within such NOTICE file, excluding those notices that do not pertain to any part of the Derivative Works, in at least one of the following places: within a NOTICE text file distributed as part of the Derivative Works; within the Source form or documentation, if provided along with the Derivative Works; or, within a display generated by the Derivative Works, if and wherever such third-party notices normally appear. The contents of the NOTICE file are for informational purposes only and do not modify the License. You may add Your own attribution notices within Derivative Works that You distribute, alongside or as an addendum to the NOTICE text from the Work, provided that such additional attribution notices cannot be construed as modifying the License.
+
+     You may add Your own copyright statement to Your modifications and may provide additional or different license terms and conditions for use, reproduction, or distribution of Your modifications, or for any such Derivative Works as a whole, provided Your use, reproduction, and distribution of the Work otherwise complies with the conditions stated in this License.
+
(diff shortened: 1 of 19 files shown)


integrations/github/src/haystack_integrations/prompts/github/comment_tool.py
```diff
@@ -0,0 +1,19 @@
+comment_prompt = """
```
sjrl, line 1: More just a question, but for for variables like these I believe it's typical to make them in all caps, so COMMENT_PROMPT. What do you think?

integrations/github/src/haystack_integrations/prompts/github/context.py
```diff
@@ -0,0 +1,175 @@
+haystack_context_prompt = """
+
+Haystack-Agent was specifically designed to help developers with the Haystack-framework and any Haystack related
```
sjrl, line 3: Remove extra new lines
```suggestion
haystack_context_prompt = """Haystack-Agent was specifically designed to help developers with the Haystack-framework and any Haystack related
```

integrations/github/src/haystack_integrations/prompts/github/file_editor_tool.py
```diff
@@ -0,0 +1,120 @@
+file_editor_prompt = """
+Use the file editor to edit an existing file in the repository.
```
sjrl, line 2: Fine to ignore this comment, but just so you are aware we are adding an extra new line at the beginning of the prompt by starting the text on the second line. 
```suggestion
file_editor_prompt = """Use the file editor to edit an existing file in the repository.
```

integrations/github/src/haystack_integrations/prompts/github/comment_tool.py
julian-risch: yes, make sense. 👍

integrations/github/README.md
```diff
@@ -0,0 +1,21 @@
+# github-haystack
```
sjrl, line 1: I think it would be good for us to add examples here in the Readme on how to use or to link to the tutorial/google colab for how to use. 

Also another relevant detail I think is that these prompts were optimized using Anthropic models. Could be a useful thing for users to know.

integrations/github/src/haystack_integrations/components/connectors/github/pr_creator.py
```diff
@@ -0,0 +1,270 @@
...
+
+import requests
+from haystack import component, default_from_dict, default_to_dict, logging
+from haystack.utils import Secret, deserialize_secrets_inplace
+
+logger = logging.getLogger(__name__)
+
+
+@component
+class GithubPRCreator:
```
sjrl, line 12: I believe github is usually capitalized as GitHub, so worth considering updating our names I think
```suggestion
class GitHubPRCreator:
```

integrations/github/README.md
julian-risch: google colab in the cookbook and some more examples in an integration page is what I imagine. The README's we currently don't fill out, for example see: https://pypi.org/project/opensearch-haystack/
Might be a good idea to change that and use a copy of the integrations page. I don't see a good reason to keep it empty but I would prefer a consistent solution. I'll talk to Bilge.

sjrl:
@julian-risch maybe a general comment on the structure here. I see that the prompts aren't being used within the library and I understand they will be used in a future tutorial/colab. 

I wonder then if it would be helpful to instead pre-assemble the tools within the repo so users could easily import the tools and immediately pass them to an Agent. What do you think?

integrations/github/src/haystack_integrations/components/connectors/github/file_editor.py
```diff
@@ -0,0 +1,287 @@
+from base64 import b64decode, b64encode
```
sjrl, line 1: We should also add the license header text to all python files right?

integrations/github/src/haystack_integrations/components/connectors/github/file_editor.py
```diff
@@ -0,0 +1,287 @@
...
+            "replacement": "def new_function():",
+            "message": "Renamed function for clarity"
+        }
+    )
+    ```
+    """
+
+    def __init__(
+        self,
+        github_token: Secret = Secret.from_env_var("GITHUB_TOKEN"),
```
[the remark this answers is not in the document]
sjrl: Let me know what you think! If you agree, I'd advocated for updating the other components to do the same.

integrations/github/src/haystack_integrations/components/connectors/github/issue_viewer.py
```diff
@@ -0,0 +1,212 @@
...
+            "Accept": "application/vnd.github.v3+json",
+            "User-Agent": "Haystack/GithubIssueViewer",
+        }
+
+    def _get_request_headers(self) -> dict:
+        """
+        Get headers with resolved token for the request.
+
+        :return: Dictionary of headers including authorization if token is present
+        """
```
sjrl, line 62: I like this pattern for an incidental reason. By not putting the resolve of the Secret in the init method we allow users to not need to provide the secret until run time which is nice especially when running pipeline validation like in the deepset platform where the secrets are not available to the pipeline in the pipeline builder. 

Could be nice to do this in the other components, basically to move the resolve_value() outside of the init method.

sjrl:
@julian-risch overall this looks really good! I mostly have minor comments and only one larger conceptual one about maybe providing users Tools directly instead of needing to compose them, themselves. 

I didn't comb through every line since there is a lot, but it's well tested so it's good to go from my perspective! We can always make quick updates to this if things arise and depending on usage.

integrations/github/src/haystack_integrations/prompts/github/comment_tool.py
julian-risch: Done

integrations/github/src/haystack_integrations/prompts/github/context.py
julian-risch: Done.

integrations/github/src/haystack_integrations/prompts/github/file_editor_tool.py
julian-risch: Done.

integrations/github/src/haystack_integrations/components/connectors/github/pr_creator.py
julian-risch: Done.

integrations/github/src/haystack_integrations/components/connectors/github/file_editor.py
julian-risch: Done.

integrations/github/src/haystack_integrations/components/connectors/github/file_editor.py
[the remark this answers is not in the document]
julian-risch: I went for the first option and we now enforce kwargs in all components/tools of the GitHub integration.

integrations/github/src/haystack_integrations/tools/github/file_editor_tool.py
```diff
@@ -0,0 +1,80 @@
...
+from haystack.utils import Secret, deserialize_secrets_inplace
+
+from haystack_integrations.components.connectors.github.file_editor import GitHubFileEditor
+from haystack_integrations.prompts.github.file_editor_tool import FILE_EDITOR_PROMPT, FILE_EDITOR_SCHEMA
+
+
+class GitHubFileEditorTool(ComponentTool):
+    """
+    A Haystack tool for editing files in GitHub repositories.
+    """
```
sjrl, line 17: Oh interesting I wasn't thinking to inherit from ComponentTool but do something like 
```python
GitHubFileEditorTool = ComponentTool(...)
```
and then people could import the pre-made `GitHubFileEditorTool` but I can see how this version would be more customizable.

integrations/github/src/haystack_integrations/tools/github/file_editor_tool.py
```diff
@@ -0,0 +1,80 @@
...
+        return default_to_dict(
+            self,
+            name=self.name,
+            description=self.description,
+            parameters=self.parameters,
+            github_token=self.github_token.to_dict(),
+            repo=self.repo,
+            branch=self.branch,
+            raise_on_failure=self.raise_on_failure,
+        )
```
sjrl, line 67: If we go this route of inheriting from ComponentTool, we won't be able to use this `to_dict` method I think. At least we use different sede methods for Tools with this dict structure 
```python
{"type": generate_qualified_class_name(type(self)), "data": serialized}
```
so we'd probably need to follow that as well right? 

Since when deserializing this in a pipeline I believe we will eventually call `deserialize_tools_or_toolset_inplace`. Do these methods work in that case?

integrations/github/src/haystack_integrations/components/connectors/github/issue_viewer.py
julian-risch: Done for all components.

julian-risch:
@sjrl I added a test called `test_pipeline_serialization`, added `_get_request_headers` to all components and here is the example notebook with updated code up until the GitHub token is required. I commented out `"message": {"source": "documents", "handler": message_handler},` because it didn't work for me and need to ask @mathislucka for advice.

https://colab.research.google.com/drive/1ktlwQ-CDLGDs2uZXvzgG8XspfjPidYqZ?usp=sharing

@sjrl If GitHubFileEditorTool looks good to you, I will add tools for all other components and probably update the directory structure a bit.

integrations/github/src/haystack_integrations/tools/github/file_editor_tool.py
sjrl: @julian-risch okay interesting that this works since you added the pipeline serialization test.

I still wonder to be consistent if we should not rename `init_parameters` to `data` in the serialized dict since that appears to be the pattern we use in our other tools and expect in `deserialize_tools_or_toolset_inplace`

integrations/github/src/haystack_integrations/tools/github/issue_commenter_tool.py
```diff
@@ -1,19 +1,46 @@
...
 from haystack_integrations.prompts.github.issue_commenter_prompt import ISSUE_COMMENTER_PROMPT, ISSUE_COMMENTER_SCHEMA
+from haystack_integrations.tools.github.utils import deserialize_handlers, serialize_handlers
 
 
 class GitHubIssueCommenterTool(ComponentTool):
     """
     A tool for commenting on GitHub issues.
+
+    :param name: Optional name for the tool.
+    :param description: Optional description.
```
sjrl, line 19: I do realize this is a bit confusing in our tools, but it seems that if we define a `__init__` then these docstrings are put under the `__init__` def. If there is no `__init__` defined like in `Tool` then we put it in the class description.

julian-risch: Ah, of course! Do you think we should a usage example here then (in addition to moving the param docstrings to the init)? I realized that's missing too.

.github/workflows/github.yml
```diff
@@ -0,0 +1,75 @@
...
+      - name: Lint
+        if: matrix.python-version == '3.9' && runner.os == 'Linux'
+        run: hatch run lint:all
+
+      - name: Generate docs
+        if: matrix.python-version == '3.9' && runner.os == 'Linux'
+        run: hatch run docs
+
+      - name: Run tests
+        run: hatch run cov-retry
```
sjrl, line 61: Oh it could also be good to set up the new testing Stefano and David have been adding to the other integrations. Here is an example https://github.com/deepset-ai/haystack-core-integrations/pull/1828

Issue #1654, opened 2025-04-16
In LangfuseConnector expose more parameters to the Langfuse client (e.g. host)


sjrl:
It would be nice if we could expose more parameters to the underlying Langfuse client like `host`, `threads`, etc. so we can more easily customize the Langfuse client. 

I think it would make sense to make some dedicated init variables like `host`, but for the remainder we can add a new parameter called `langfuse_client_kwargs` so users can pass additional options.

PR #1657, opened 2025-04-17
feat: Add more options to LangfuseConnector

sjrl:
### Related Issues

- fixes https://github.com/deepset-ai/haystack-core-integrations/issues/1654

### Proposed Changes:

 
 

- added a `host` variable to the init method. Defaults to `None` which means it will use the default value from Langfuse.
- added a generic `langfuse_client_kwargs` that can be used to pass any additional kwargs to the underlying Langfuse client

### How did you test it?

- Updated tests

### Checklist


vblagoje:
@sjrl when exactly should we introduce this change? Should it be part of a major release when we accumulate more changes? This change will cause backward compatibility issues when components serialized with a newer library version are deserialized by an older version and I'm reluctant to go there just for fun and not really bad need. WDYT?

sjrl:
> This change will cause backward compatibility issues when components serialized with a newer library version are deserialized by an older version and I'm reluctant to go there just for fun and not really bad need. WDYT?

Hmm I don't think we need to worry about that. What you're describing isn’t a backward compatibility issue which usually means newer versions can read data from older versions. What you're describing is forward incompatibility — older versions failing to read newer data — which is generally expected and not usually a blocker.

Or am I missing something?

vblagoje:
> > This change will cause backward compatibility issues when components serialized with a newer library version are deserialized by an older version and I'm reluctant to go there just for fun and not really bad need. WDYT?
> 
> Hmm I don't think we need to worry about that. What you're describing isn’t a backward compatibility issue which usually means newer versions can read data from older versions. What you're describing is forward incompatibility — older versions failing to read newer data — which is generally expected and not usually a blocker.
> 
> Or am I missing something?

Yes, apologies for the confusion.

vblagoje (approved):
Would you please also update readme while we are at it. It's a bit outdated!

Issue #1653, opened 2025-04-16
Langfuse: Remove warning "Creating a new trace without a parent span is not recommended ..."


sjrl:
The warning message isn't helpful since the user can't take any action to "fix" anything. 

Also often times the message isn't valid, there isn't a requirement that a parent span has to be started from any particular operation. In this case it popped up again when running the Agent in isolation with the warning message `"Creating a new trace without a parent span is not recommended for operation 'haystack.agent.run'."` which actually is fine.

PR #1661, opened 2025-04-21
fix: Langfuse - remove warning "Creating a new trace without a parent span is not recommended ..."

vblagoje:
We observed "Creating a new trace without a parent span is not recommended for operation `haystack.agent.run`" when running agent only examples - no pipeline. This warning is confusing to users. This fix removes it by adding it to allowed span roots.

### Related Issues

- fixes https://github.com/deepset-ai/haystack-core-integrations/issues/1653

### Proposed Changes:

Allow `haystack.agent.run` to be span root 

### How did you test it?

Manual testing with agent only examples

### DIFF
--- a/integrations/langfuse/src/haystack_integrations/tracing/langfuse/tracer.py
+++ b/integrations/langfuse/src/haystack_integrations/tracing/langfuse/tracer.py
@@ -52,7 +52,6 @@ _PIPELINE_INPUT_KEY = "haystack.pipeline.input_data"
 _PIPELINE_OUTPUT_KEY = "haystack.pipeline.output_data"
 _ASYNC_PIPELINE_RUN_KEY = "haystack.async_pipeline.run"
 _PIPELINE_RUN_KEY = "haystack.pipeline.run"
-_AGENT_RUN_KEY = "haystack.agent.run"
 _COMPONENT_NAME_KEY = "haystack.component.name"
 _COMPONENT_TYPE_KEY = "haystack.component.type"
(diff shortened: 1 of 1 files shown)


vblagoje:
@julian-risch @sjrl has all the context - we talked about this internally

integrations/langfuse/src/haystack_integrations/tracing/langfuse/tracer.py
```diff
@@ -262,7 +263,8 @@ def create_span(self, context: SpanContext) -> LangfuseSpan:
 
         tracing_ctx = tracing_context_var.get({})
         if not context.parent_span:
-            if context.operation_name not in [_PIPELINE_RUN_KEY, _ASYNC_PIPELINE_RUN_KEY]:
+            allowed_span_roots = [_PIPELINE_RUN_KEY, _ASYNC_PIPELINE_RUN_KEY, _AGENT_RUN_KEY]
+            if context.operation_name not in allowed_span_roots:
```
sjrl, line 267: hey @vblagoje I think we agreed to just remove this check since it's not that useful and is often wrong. So think I we should just remove the check instead of adding Agent. This way when we add any new spans in the future this won't trigger.

vblagoje: Ok deal, I'll make an update to remove it 👍

sjrl (approved):
Looks good!

vblagoje:
> Looks good!

Yes, and I tested it on several examples including agent for any discernable difference - none found!

PR #1662, opened 2025-04-21
feat!: Deprecate SSEServerInfo base_url in favour of url

vblagoje:
### Why

This PR deprecates the `base_url` parameter in favor of a single `url` argument for configuring the SSE client. It adds deprecation warnings and updates relevant classes to reflect this change. 

This change aligns with common usage patterns—users are typically more familiar with providing the **full URL including the `/sse` endpoint**, and omitting it (via `base_url`) can be unintuitive or error-prone. Simplifying the interface improves clarity and consistency.

### Details

- Adds a deprecation warning if `base_url` is used.
- Ensures that only one of `url` or `base_url` can be passed.
- Updates both `SSEClient` and `SSEServerInfo` to enforce and reflect this change.

### Deprecation Notice

`base_url` is now **deprecated** and will be removed in a future release. Please switch to using the full `url` (including the `/sse` path) when initializing the client.

### Checklist

- Backward compatibility maintained (with warning)
- Clear error handling for conflicting or missing parameters
- Docstrings and inline comments updated


vblagoje:
@anakin87 @mpangrazzi is already familiar with this discussion - it is a small change that will improve devX as users are mostly used to providing url with /sse at the end rather than just `base_url` and intuitively seem more future proof.

vblagoje:
cc @bilgeyucel this is the devX upgrade we talked about

vblagoje:
@anakin87 @mpangrazzi can you please take another look now? There are some minor internal simplifications that are not public and make everything much simpler

anakin87:
I am ignoring this PR because I thought that Michele had more context. In case it is urgent, please let me know.

vblagoje:
@anakin87 np, @mpangrazzi has all the context and @oryx1729 can also chime in with his UI/UX experience around this matter

integrations/mcp/src/haystack_integrations/tools/mcp/mcp_tool.py
```diff
@@ -432,22 +434,51 @@ class SSEServerInfo(MCPServerInfo):
...
+        elif not is_valid_http_url(self.url):
+            message = f"Invalid url: {self.url}"
+            raise ValueError(message)
+
     def create_client(self) -> MCPClient:
         """
         Create an SSE MCP client.
 
-        :returns: Configured HttpMCPClient instance
+        :returns: Configured SSEClient instance
```
mpangrazzi, line 478: shouldn't be:

```
:returns: Configured MCPClient instance
```

?

mpangrazzi (approved):
LGTM! - I've left only a minor comment

PR #1691, opened 2025-05-02
chore: Upgrade default model to command-r-plus

vblagoje:
### Why:

This PR upgrades the Cohere integration to use the `command-r-plus` chat model, Cohere’s most advanced LLM as of August 2024. Compared to `command-r`, the `command-r-plus` model delivers significantly better performance for complex tasks such as multi-step agent workflows and RAG-heavy pipelines. It is optimized for long-context conversational use cases and is intended for production-grade applications, making it a better fit for most Haystack usage scenarios. 

The previous `command-r` model remains suitable for lightweight tasks and latency-sensitive applications, but `command-r-plus` is now the recommended default for more capable, reliable outputs in realistic production settings.

Cohere has even newer model [Command A](https://docs.cohere.com/docs/command-a) the performant model to date, excelling at real world enterprise tasks including tool use, retrieval augmented generation (RAG), agents, and multilingual use cases. At 111B parameters, Command A has a context length of 256K. But we can update to this one once it stabilizes as the current version seems more beta - i.e. it doesn't have full name as command-r-plus but rather command-a-03-2025

### What:

* Updated the Cohere integration to reference `command-r-plus` instead of `command-r`.
* Adjusted associated test cases to validate against the updated model.

### How can it be used:

No changes needed from the user—new pipelines using Cohere will default to the upgraded model automatically.

### How did you test it:

* Ran unit tests to validate output structure and model invocation.
* Confirmed compatibility with existing tool use cases and RAG patterns.

### Notes for the reviewer:

* Focus on verifying model name substitution and test coverage.

### DIFF
--- a/integrations/cohere/src/haystack_integrations/components/generators/cohere/chat/chat_generator.py
+++ b/integrations/cohere/src/haystack_integrations/components/generators/cohere/chat/chat_generator.py
@@ -236,7 +236,7 @@ class CohereChatGenerator:
     from haystack.utils import Secret
     from haystack_integrations.components.generators.cohere import CohereChatGenerator
 
-    client = CohereChatGenerator(model="command-r", api_key=Secret.from_env_var("COHERE_API_KEY"))
+    client = CohereChatGenerator(model="command-r-plus", api_key=Secret.from_env_var("COHERE_API_KEY"))
     messages = [ChatMessage.from_user("What's Natural Language Processing?")]
     client.run(messages)
 
@@ -296,7 +296,7 @@ class CohereChatGenerator:
     def __init__(
         self,
         api_key: Secret = Secret.from_env_var(["COHERE_API_KEY", "CO_API_KEY"]),
-        model: str = "command-r",
+        model: str = "command-r-plus",
         streaming_callback: Optional[Callable[[StreamingChunk], None]] = None,
         api_base_url: Optional[str] = None,
         generation_kwargs: Optional[Dict[str, Any]] = None,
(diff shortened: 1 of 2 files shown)


vblagoje:
@julian-risch in addition manually verified to work nicely with agents.

julian-risch:
Updating only the usage example and the https://haystack.deepset.ai/integrations/cohere integration page seems good to me though.

vblagoje:
Ok, I don't mind,  but command-r is two generations behind and the new model cost is inline with top of the line models from various [providers](https://livechatai.com/anthropic-claude-3-sonnet-pricing-calculator)

julian-risch:
Let's get an opinion from another team member. We could release a new major version cohere-v5.0.0 with that change.

davidsbatista:
What about a warning, just something in the stdout warning the user that this current release changed the default model, and the price is 10x more? We would then remove it in the next release.

vblagoje:
> What about a warning, just something in the stdout warning the user that this current release changed the default model, and the price is 10x more? We would then remove it in the next release.

It is very common that such old models are cheap and new ones are more expensive. See the price chat above - nothing out of the ordinary here requiring warnings IMHO

davidsbatista:
Even if all other models have the same base price as the new default, we should make clear to the user that this integration has changed, and the default model cost is now 10x the previous default.

julian-risch (approved):
Looks good to me! 👍 
command-r is an alias for command-r-03-2024 ([Cohere docs](https://docs.cohere.com/v2/docs/command-r))
and while command-r-03-2024 is not deprecated yet, support for models fine-tuned with Command-R-03-2024 ended on March 08, 2025. ([Cohere docs](https://docs.cohere.com/docs/deprecations#deprecation-history))
For that reason, I believe it's a good idea to update our default from command-r-03-2024 to command-r-08-2024. And my understanding is that the price doesn't change at all.

vblagoje:
Included in https://pypi.org/project/cohere-haystack/4.1.0/ release

PR #1701, opened 2025-05-05
feat: adding an `HybridRetriever` as a `Supercomponent` having `OpenSearch` as the document store

davidsbatista:
### Related Issues

- fixes #issue-number

### Checklist


integrations/opensearch/src/haystack_integrations/components/retrievers/open_search_hybrid_retriever.py
```diff
@@ -0,0 +1,444 @@
...
+        and pass the rest as kwargs. This is to keep the constructor clean and easy to read.
+
+        If you need to pass extra parameters to the components, you can do so by passing them as kwargs. It expects
+        a dictionary with the component name as the key and the parameters as the value.
+
+        text_embedder -> SentenceTransformersTextEmbedder
+        bm25_retriever -> OpenSearchBM25Retriever
+        embedding_retriever -> OpenSearchEmbeddingRetriever
+        document_joiner -> DocumentJoiner
+        chat_prompt_builder -> ChatPromptBuilder
```
julian-risch, line 104: Let's not include a ChatPromptBuilder, OpenAIChatGenerator and AnswerBuilder in the HybridRetriever. The HybridRetriever should only retrieve documents from OpenSearch using both embedding-based and keyword-based retrieval methods. That way it can be re-used more flexibly.

julian-risch (changes requested):
The LazyImport part looks good to me. However, we should remove those components from the OpenSearchHybridRetriever that do more than the retrieval. The last component of the OpenSearchHybridRetriever should be the DocumentJoiner. ChatPromptBuilder, ChatGenerator and AnswerBuilder should be left out.

integrations/opensearch/src/haystack_integrations/components/retrievers/open_search_hybrid_retriever.py
```diff
@@ -0,0 +1,322 @@
...
+    A hybrid retriever that combines embedding-based and keyword-based retrieval from OpenSearch.
+
+    This component requires haystack-ai>=2.13.0 to work properly.
+    """
+
+    def __init__(
+        self,
+        document_store: OpenSearchDocumentStore,
+        *,
+        # SentenceTransformer
```
julian-risch, line 42: Let's use the full name
```suggestion
        # SentenceTransformersTextEmbedder
```

integrations/opensearch/src/haystack_integrations/components/retrievers/open_search_hybrid_retriever.py
```diff
@@ -0,0 +1,322 @@
...
+from haystack.utils import ComponentDevice
+
+from haystack_integrations.components.retrievers.opensearch import OpenSearchBM25Retriever, OpenSearchEmbeddingRetriever
+from haystack_integrations.document_stores.opensearch import OpenSearchDocumentStore
+
+logger = logging.getLogger(__name__)
+
+# Use LazyImport to conditionally import haystack super_component, which is only available in newer versions
+with LazyImport(
+    "To use the OpenSearchHybridRetriever you need more recent version of haystack. Run 'pip install haystack-ai>=2.13.0'"  # noqa: E501
```
julian-risch, line 22:
```suggestion
    "To use the OpenSearchHybridRetriever you need a more recent version of haystack. Run 'pip install haystack-ai>=2.13.0'"  # noqa: E501
```

integrations/opensearch/src/haystack_integrations/components/retrievers/open_search_hybrid_retriever.py
```diff
@@ -0,0 +1,322 @@
...
+        all_terms_must_match: bool = False,
+        filter_policy_bm25: Union[str, FilterPolicy] = FilterPolicy.REPLACE,
+        custom_query_bm25: Optional[Dict[str, Any]] = None,
+        # OpenSearchEmbeddingRetriever
+        filters_embedding: Optional[Dict[str, Any]] = None,
+        top_k_embedding: int = 10,
+        filter_policy_embedding: Union[str, FilterPolicy] = FilterPolicy.REPLACE,
+        custom_query_embedding: Optional[Dict[str, Any]] = None,
+        # DocumentJoiner
+        join_mode: Union[str, JoinMode] = JoinMode.CONCATENATE,
```
julian-risch, line 65: I'd say `RECIPROCAL_RANK_FUSION` is a better default for a hybrid retriever because scores of the OpenSearchBM25Retriever are not scaled to [0,1] by default and thus are not comparable to the scores of the OpenSearchEmbeddingRetriever.

integrations/opensearch/src/haystack_integrations/components/retrievers/open_search_hybrid_retriever.py
```diff
@@ -0,0 +1,322 @@
...
+        hybrid_retrieval.connect("text_embedder", "embedding_retriever")
+        hybrid_retrieval.connect("bm25_retriever", "document_joiner")
+        hybrid_retrieval.connect("embedding_retriever", "document_joiner")
+
+        # Define how pipeline inputs/outputs map to subcomponent inputs/outputs
+        self.input_mapping = {
+            # The pipeline input "query" feeds into each of the retrievers
+            "query": ["text_embedder.text", "bm25_retriever.query"],
+        }
+        # The pipeline output "answers" comes from "answer_builder.answers"
```
julian-risch, line 272: This comment refers to the answer_builder that was removed.
```suggestion
```

integrations/opensearch/src/haystack_integrations/components/retrievers/open_search_hybrid_retriever.py
```diff
@@ -0,0 +1,322 @@
...
+            init_args["embedding_retriever"].update(kwargs["embedding_retriever"])
+            init_args["embedding_retriever"]["document_store"] = self.document_store
+
+        self.pipeline = self._create_pipeline(init_args)
+
+    def _create_pipeline(self, data: dict[str, Any]) -> Pipeline:
+        """
+        Create the pipeline for the OpenSearchHybridRetriever.
+
+        If the template for the ChatPromptBuilder is not provided, a default template is used.
```
julian-risch, line 250: This comment refers to the removed ChatPromptBuilder.
```suggestion
```

integrations/opensearch/src/haystack_integrations/components/retrievers/open_search_hybrid_retriever.py
```diff
@@ -0,0 +1,322 @@
...
+    "To use the OpenSearchHybridRetriever you need more recent version of haystack. Run 'pip install haystack-ai>=2.13.0'"  # noqa: E501
+) as haystack_imports:
+    from haystack import super_component
+
+# Trigger an error message when the OpenSearchHybridRetriever is imported without haystack-ai>=2.13.0
+haystack_imports.check()
+
+
+@super_component
+class OpenSearchHybridRetriever:
```
julian-risch, line 31: I assume we can't add OpenSearchHybridRetriever to the init file where the other two retrievers are also listed because then users who import OpenSearchBM25Retriever also are required to have haystack>=2.13 installed? https://github.com/deepset-ai/haystack-core-integrations/blob/main/integrations/opensearch/src/haystack_integrations/components/retrievers/opensearch/__init__.py

integrations/opensearch/src/haystack_integrations/components/retrievers/open_search_hybrid_retriever.py
```diff
@@ -0,0 +1,322 @@
...
+
+    This component requires haystack-ai>=2.13.0 to work properly.
+    """
+
+    def __init__(
+        self,
+        document_store: OpenSearchDocumentStore,
+        *,
+        # SentenceTransformer
+        text_embedder_model: str = "sentence-transformers/all-mpnet-base-v2",
```
julian-risch, line 43: as there is only one model, I suggest we call this just `model` instead of `text_embedder_model`

integrations/opensearch/src/haystack_integrations/components/retrievers/open_search_hybrid_retriever.py
```diff
@@ -0,0 +1,322 @@
...
+        return hybrid_retrieval
+
+    def to_dict(self):
+        """
+        Serialize OpenSearchHybridRetriever to a dictionary.
+
+        :returns:
+            Dictionary with serialized data.
+        """
+        return default_to_dict(
```
julian-risch, line 284: We're not serializing the extra kwargs right now. I wonder if we could do that. For example by listing all parameters explicitly with the default that is used in the respective definitions of the components.
The `token` parameter of the `SentenceTransformersTextEmbedder` is a good example where serialization is not trivial but currently requires `token=self.token.to_dict() if self.token else None`

julian-risch (changes requested):
Thanks for removing the ChatPromptBuilder, AnswerBuilder and ChatGenerator! I have a few more, smaller change requests. Besides that, I suggest we briefly discuss with the team if we want to expose all parameters or not and if the current way of exposing extra kwargs is good or not.

integrations/opensearch/src/haystack_integrations/components/retrievers/open_search_hybrid_retriever.py
davidsbatista: > For example by listing all parameters explicitly with the default that is used in the respective definitions of the components

I'm a bit confused by this - the idea of the extra kwargs is to allow the user to customise other parameters that the super component doesn't expose, or missed what you were trying to suggest.

I have a solution to serialise the extra kwargs and then use them in the initialisation. I'm writing some tests now to make sure nothing breaks.

I think the `token` is a good example of a parameter that should be exposed, due to not being a regular parameter in serialisation/deserialisation, not having in the kwargs and as an explicit parameter might make things easier.

integrations/opensearch/src/haystack_integrations/components/retrievers/opensearch/open_search_hybrid_retriever.py
```diff
@@ -0,0 +1,347 @@
...
+
+    This component requires haystack-ai>=2.13.0 to work properly.
+    """
+
+    def __init__(
+        self,
+        document_store: OpenSearchDocumentStore,
+        *,
+        # SentenceTransformersTextEmbedder
+        model: str = "sentence-transformers/all-mpnet-base-v2",
```
sjrl, line 43: Sorry to derail this a bit, but I'm unsure if we should hardcode the TextEmbedder users have to use in this SupreComponent. For example, it would make it unusable for the majority of deepset Platforms use cases since we use a custom text embedder in all of our templates. 

I'd prefer if we allow a user to pass in a fully initialized Text Embedder as init param so we don't restrict users into having to use SentenceTransformers.

sjrl: This should also hopefully reduce the init signature and simplify the sede by not needing to support all the params of the TextEmbedder

davidsbatista: I understand that makes sense - I briefly discussed it with Julian, and we agreed to use the SentenceTransformer as it was the most popular one. 

I opened a ticket here: https://github.com/deepset-ai/haystack/issues/9351

julian-risch: Thanks for creating the issue! I added one bullet point to the issue. I admit that I didn't take into account that multiple TextEmbedders could be supported easily with a TextEmbedder protocol. As it simplifies the HybridRetriever and will allow more users to benefit from it, I believe we should add the protocol before continuing with the HybridRetriever. @davidsbatista We can talk about later today.

integrations/opensearch/src/haystack_integrations/components/retrievers/opensearch/open_search_hybrid_retriever.py
```diff
@@ -0,0 +1,288 @@
...
+from haystack_integrations.components.retrievers.opensearch import OpenSearchBM25Retriever, OpenSearchEmbeddingRetriever
+from haystack_integrations.document_stores.opensearch import OpenSearchDocumentStore
+
+logger = logging.getLogger(__name__)
+
+# Use LazyImport to conditionally import haystack super_component, which is only available in newer versions
+with LazyImport(
+    "To use the OpenSearchHybridRetriever you need a more recent version of haystack. Run 'pip install haystack-ai>=2.13.0'"  # noqa: E501
+) as haystack_imports:
+    from haystack import super_component
```
sjrl, line 24: Do we need this anymore if we are going to bump the required version of this package to haystack>=2.14.0?

integrations/opensearch/src/haystack_integrations/components/retrievers/opensearch/open_search_hybrid_retriever.py
```diff
@@ -0,0 +1,288 @@
...
+        top_k: Optional[int] = None,
+        sort_by_score: bool = True,
+        # extra kwargs
+        **kwargs,
+    ):
+        """
+        Initialize the OpenSearchHybridRetriever, a super component to retrieve documents from OpenSearch using
+        both embedding-based and keyword-based retrieval methods.
+
+        This super component is tied to a SentenceTransformersTextEmbedder.
```
sjrl, line 68: Remove docstring since no longer true

integrations/opensearch/src/haystack_integrations/components/retrievers/opensearch/open_search_hybrid_retriever.py
```diff
@@ -0,0 +1,288 @@
...
+
+        We don't explicitly define all the init parameters of the components in the constructor, for each
+        of the components, since that would be around 20+ parameters. Instead, we define the most important ones
+        and pass the rest as kwargs. This is to keep the constructor clean and easy to read.
```
sjrl, line 72: Is this still true?

integrations/opensearch/src/haystack_integrations/components/retrievers/opensearch/open_search_hybrid_retriever.py
```diff
@@ -0,0 +1,288 @@
...
+            The weights for the joiner.
+
+        :param top_k:
+            The number of results to return from the joiner.
+
+        :param sort_by_score:
+            Whether to sort the results by score.
+
+        :param **kwargs:
+            Additional keyword arguments.
```
sjrl, line 133: Could we tell the user what structure we expect for kwargs. It looks like we expect top-level keys to be `bm25_retriever` and `embedding_retriever`

integrations/opensearch/src/haystack_integrations/components/retrievers/opensearch/open_search_hybrid_retriever.py
```diff
@@ -0,0 +1,288 @@
...
+    A hybrid retriever that combines embedding-based and keyword-based retrieval from OpenSearch.
+
+    This component requires haystack-ai>=2.13.0 to work properly.
+    """
+
+    def __init__(
+        self,
+        document_store: OpenSearchDocumentStore,
+        *,
+        embedder: TextEmbedder = None,
```
sjrl, line 42: It seems like `embedder` is actually required 
```suggestion
        self,
        document_store: OpenSearchDocumentStore,
        embedder: TextEmbedder,
        *,
```
perhaps we should update to something like this?

integrations/opensearch/src/haystack_integrations/components/retrievers/opensearch/open_search_hybrid_retriever.py
```diff
@@ -0,0 +1,288 @@
...
+
+        self.extra_args = kwargs
+
+        # handle extra kwargs for the bm25 and embedding retrievers and the doc store as init param
+        if "bm25_retriever" in kwargs:
+            init_args["bm25_retriever"].update(kwargs["bm25_retriever"])
+            init_args["bm25_retriever"]["document_store"] = self.document_store
+        if "embedding_retriever" in kwargs:
+            init_args["embedding_retriever"].update(kwargs["embedding_retriever"])
+            init_args["embedding_retriever"]["document_store"] = self.document_store
```
sjrl, line 193: It might be nice to run a check on kwargs as in if we notice the user passed something like 
```python
kwargs = {"fuzzines": "AUTO"}
```
we could warn them saying `kwargs must contain 'bm25_retriever' and 'embedding_retriever' as top-level keys ...`

integrations/opensearch/src/haystack_integrations/components/retrievers/opensearch/open_search_hybrid_retriever.py
sjrl: Ahh never mind I missed that you provide this explanation up in the description of the docstring. Maybe better to move it here? But up to you.

sjrl:
@davidsbatista a few remaining comments, otherwise looks good!

integrations/opensearch/src/haystack_integrations/components/retrievers/opensearch/open_search_hybrid_retriever.py
davidsbatista: we can remove it yes, since now we have a dependency on the `TextEmbedder` Protocol which will be in `haystack>=2.14.0` we can remove it.

integrations/opensearch/src/haystack_integrations/components/retrievers/opensearch/open_search_hybrid_retriever.py
davidsbatista: yes, we are only exposing a few parameters for the retrieveres

integrations/opensearch/src/haystack_integrations/components/retrievers/opensearch/open_search_hybrid_retriever.py
davidsbatista: > We could branch dependencies depending on which component you want to import (note: just an idea, would have to explore how to implement that)

Applying the lazyimport only when we try to import (OpenSearch) HybridRetriever could work.

But, unless there's an issue, I still would vote on making this dependent on the upcoming haystack release.

julian-risch: In my opinion, the reason for the LazyImport and not requiring >=2.14.0 is still the same as when we discussed LazyImport for the OpenSearchHybridRetriever in the beginning: If we make any bug fixes or other changes in future to OpenSearchDocumentStore or one of the other retrievers, all users can benefit from those changes. Even if they can't/don't want to upgrade their Haystack version to >=2.14.0.

`haystack_imports.check()` is still in the implementation by the way [here](https://github.com/deepset-ai/haystack-core-integrations/pull/1701/files#diff-095ae23eb2de32c9f28370202fa6a9c147e6b73d5b1cfd6519995bf77ca3db38R20)

davidsbatista: reverted to 2.11 and only HybridRetriever for 2.14

julian-risch (approved):
LGTM! 👍 We can merge and release a new version of the integration as soon as Haystack 2.14.0 has been released. In the mean time, we can prepare a documentation page for the new component and draft an extension of the integration page. We can also add the new retriever to https://docs.haystack.deepset.ai/docs/retrievers 
Could you please sync with @dfokina on these next steps?

integrations/opensearch/src/haystack_integrations/components/retrievers/opensearch/open_search_hybrid_retriever.py
```diff
@@ -0,0 +1,286 @@
...
+
+logger = logging.getLogger(__name__)
+
+
+@super_component
+class OpenSearchHybridRetriever:
+    """
+    A hybrid retriever that combines embedding-based and keyword-based retrieval from OpenSearch.
+
+    This component requires haystack-ai>=2.13.0 to work properly.
```
julian-risch, line 24:
```suggestion
    This component requires haystack-ai>=2.14.0 to work properly.
```

integrations/opensearch/src/haystack_integrations/components/retrievers/opensearch/open_search_hybrid_retriever.py
```diff
@@ -0,0 +1,286 @@
...
+    """
```
julian-risch, line 25: Let's add a usage example here. Can be the same as in the integrations page, once we have it.

julian-risch (changes requested):
Found two more things that we should improve.

integrations/opensearch/src/haystack_integrations/components/retrievers/opensearch/open_search_hybrid_retriever.py
```diff
@@ -0,0 +1,353 @@
...
+
+logger = logging.getLogger(__name__)
+
+
+@super_component
+class OpenSearchHybridRetriever:
+    """
+    A hybrid retriever that combines embedding-based and keyword-based retrieval from OpenSearch.
+
+    This component requires haystack-ai>=2.14.0 to work properly.
```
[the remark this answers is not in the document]
davidsbatista: I don't think it hurts either, but I will remove it.

integrations/opensearch/src/haystack_integrations/components/retrievers/opensearch/open_search_hybrid_retriever.py
```diff
@@ -0,0 +1,353 @@
...
+        # deserialize the embedder
+        try:
+            text_embedder_class = import_class_by_name(data["init_parameters"]["embedder"]["type"])
+        except ImportError as e:
+            msg = f"Class '{data['init_parameters']['embedder']['type']}' not correctly imported"
+            raise DeserializationError(msg) from e
+
+        data["init_parameters"]["embedder"] = component_from_dict(
+            cls=text_embedder_class, data=data["init_parameters"]["embedder"], name="embedder"
+        )
```
sjrl, line 318: It would be nice if we could add a `deserialize_text_embedder_inplace` like we have for `deserialize_chatgenerator_inplace` in haystack main. 

So probably not to further block this, let's open an issue to create `deserialize_text_embedder_inplace` in main and use that here in a later version.

integrations/opensearch/src/haystack_integrations/components/retrievers/opensearch/open_search_hybrid_retriever.py
```diff
@@ -0,0 +1,353 @@
...
+        Serialize OpenSearchHybridRetriever to a dictionary.
+
+        :returns:
+            Dictionary with serialized data.
+        """
+        return default_to_dict(
+            self,
+            # DocumentStore
+            document_store=self.document_store.to_dict(),
+            embedder=self.embedder.to_dict(),
```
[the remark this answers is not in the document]
sjrl: Noted! Let's tackle that in the resolution for the other issue, but for now let's use component_to_dict :)

PR #1708, opened 2025-05-06
fix: hard coded embedding field for mongodb

ArzelaAscoIi:
### Related Issues

- make embedding filed configurable 

We should make this field configurable since we can't assume a data format of our customers. 

Even the sample dataset from mongodb you can use in atlas uses a different key to store embeddings: https://huggingface.co/datasets/MongoDB/embedded_movies

### Checklist


ArzelaAscoIi:
@sjrl We don't have a way to integration test the integration against atlas, right ?

sjrl (approved):
Looks good!

Issue #1681, opened 2025-04-30
fix: Inconsistent behavior between streaming and non-streaming in AmazonBedrockChatGenerator


sjrl:
This was found while working on this PR https://github.com/deepset-ai/haystack-core-integrations/pull/1671

Basically the problem is that streaming a response with mixed content such as text and a tool call results in two ChatMessages when using streaming and when not using streaming you only get one ChatMessage. 

The two functions in question are:
- `_parse_streaming_response` and `_parse_completion_response`

Here is an example with `_parse_streaming_response`
```python
from haystack_integrations.components.generators.amazon_bedrock.chat.utils import _parse_streaming_response

events = [
    {"contentBlockStart": {"start": {"text": ""}}},
    {"contentBlockDelta": {"delta": {"text": "Let me "}}},
    {"contentBlockDelta": {"delta": {"text": "help you."}}},
    {"contentBlockStop": {}},
    {"contentBlockStart": {"start": {"toolUse": {"toolUseId": "123", "name": "search_tool"}}}},
    {"contentBlockDelta": {"delta": {"toolUse": {"input": '{"query":'}}}},
    {"contentBlockDelta": {"delta": {"toolUse": {"input": '"test"}'}}}},
    {"contentBlockStop": {}},
    {"messageStop": {"stopReason": "complete"}},
    {"metadata": {"usage": {"inputTokens": 10, "outputTokens": 20, "totalTokens": 30}}},
]

replies = _parse_streaming_response(events, None, "model-name")
assert len(replies) == 2  # is True
```

Here is using non-streaming so with `_parse_completion_response`
```python
from haystack_integrations.components.generators.amazon_bedrock.chat.utils import _parse_completion_response

mixed_response = {
    "output": {
        "message": {
            "role": "assistant",
            "content": [
                {"text": "Let me help you."},
                {"toolUse": {"toolUseId": "123", "name": "search_tool", "input": {"query": "test"}}},
            ],
        }
    },
    "stopReason": "complete",
    "usage": {"inputTokens": 25, "outputTokens": 35, "totalTokens": 60},
}

replies = _parse_completion_response(mixed_response, "model-name")
assert len(replies) == 1 # is True
```

These two functions should return consistent output. I'm not sure which one is the one we should go with.

sjrl:
cc @tstadel

tstadel:
Interesting. I guess this is due to how streaming API works and the tool. Is there a practical reason for them having to be consistent?

[...]

Issue #1680, opened 2025-04-30
fix: AmazonBedrockChatGenerator does not pass Tool Call streaming chunk to streaming_callback


sjrl:
The ``

For example, in the OpenAIChatGenerator we handle the stream response in this [function](https://github.com/deepset-ai/haystack/blob/201becd40040392646245df1f26515ecb50ba4fc/haystack/components/generators/chat/openai.py#L417) which is shown below:
```python
    def _handle_stream_response(self, chat_completion: Stream, callback: SyncStreamingCallbackT) -> List[ChatMessage]:
        chunks: List[StreamingChunk] = []
        chunk = None
        chunk_delta: StreamingChunk

        for chunk in chat_completion:  # pylint: disable=not-an-iterable
            assert len(chunk.choices) <= 1, "Streaming responses should have at most one choice."
            # NOTE: We convert both Tool Call and Text chunks into a chunk_delta here
            chunk_delta = self._convert_chat_completion_chunk_to_streaming_chunk(chunk)
            chunks.append(chunk_delta)
            # NOTE: Here we always pass that chunk_delta to the callback
            callback(chunk_delta)
        return [self._convert_streaming_chunks_to_chat_message(chunk, chunks)]
```
as you can see we pass all chunk deltas whether they are text or tool calls to the callback allowing streaming of the tool call. 

We should update the `AmazonBedrockChatGenerator` to allow for streaming of tool calls as well by passing the tool call chunks to the streaming callback.

sjrl:
cc @tstadel

Issue #1677, opened 2025-04-29
AmazonBedrockChatGenerator + ToolInvoker don't handle multiple tool calls at once properly


vblagoje:
**Description:**

When using `AmazonBedrockChatGenerator` with a model capable of requesting multiple tool calls in a single turn (like Anthropic Claude 3.5 Sonnet), the component fails when attempting to send the results of these multiple tool executions back to the Bedrock API.

**Steps Observed:**

1.  The Haystack Agent calls `AmazonBedrockChatGenerator`.
2.  The Bedrock model (`anthropic.claude-3-5-sonnet-20240620-v1:0`) responds with a message containing *multiple* distinct `tool_calls` requests, each with a unique ID. For example:
    *   Call 1: `{ "id": "tooluse_TIB5o2PCQHW_fpWTOwvROA", "function": { "name": "get_weather_forecast", ... } }`
    *   Call 2: `{ "id": "tooluse_45K4_G22SEaolJj1FAD14w", "function": { "name": "qdrant-find", ... } }`
3.  The `ToolInvoker` (within the Agent) successfully receives these requests and executes the corresponding tools (`get_weather_forecast` and `qdrant-find`).
4.  `AmazonBedrockChatGenerator` attempts to continue the conversation by sending the results of *both* tool executions back to the Bedrock `ConverseStream` API.
5.  The call to `client.converse_stream` fails with a `botocore.errorfactory.ValidationException`.

**Error Message:**

```
botocore.errorfactory.ValidationException: An error occurred (ValidationException) when calling the ConverseStream operation: Expected toolResult blocks at messages.4.content for the following Ids: tooluse_45K4_G22SEaolJj1FAD14w, but found: tooluse_TIB5o2PCQHW_fpWTOwvROA
```

**Analysis:**

The error message clearly indicates that the Bedrock API was expecting the result associated with the `qdrant-find` tool call (`tooluse_45K4_G22SEaolJj1FAD14w`) but instead received the result associated with the `get_weather_forecast` tool call (`tooluse_TIB5o2PCQHW_fpWTOwvROA`) at that position in the message list (`messages.4.content`).

It seems that when `AmazonBedrockChatGenerator` prepares the message containing the tool results to send back to Bedrock, it is **incorrectly ordering or formatting the `toolResult` blocks** when multiple results need to be sent simultaneously. The Bedrock API requires the results to be provided and correctly matched to their corresponding `toolUseId` from the previous turn.

[...]

PR #1711, opened 2025-05-07
fix: Fix BedrockChatGenerator to be able to handle multiple tool calls from one response

sjrl:
### Related Issues

- fixes https://github.com/deepset-ai/haystack-core-integrations/issues/1677
- fixes https://github.com/deepset-ai/haystack-core-integrations/issues/1680
- fixes https://github.com/deepset-ai/haystack-core-integrations/issues/1681

### Proposed Changes:

 
 

**Fixing Multi Tool Calling**
Repairs the expected formatting of the bedrock messages.

Basically if a ChatMessage has multiple Tool Calls in it then the corresponding Tool Result ChatMessage must also contain both tool results. This is currently not possible because the ToolInvoker in Haystack returns a separate ChatMessage per Tool Call no matter if the original Tool Calls came one from one or multiple ChatMessages.

As a workaround, we retroactively repair the bedrock formatted messages to regroup the Tool Results based on how the Tool Calls are grouped. 

This could also be "fixed" in the ToolInvoker, but I'm not sure if that would have adverse effects on our other integrations. **Update:** Nothing to fix in the ToolInvoker, most providers expect separate messages per tool result. So it's better to modify the behavior only in this integration.

**Fixing Tool Call Streaming**
Changed the internal logic of the integration to take Events from Bedrock, convert them into Haystack `StreamingChunk`s that follow the same format as `OpenAIChatGenerator` and then pass the streaming chunk to the streaming callback. Now we will be able to reuse the `print_streaming_chunk` in Haystack main after a slight change is made. Here is an example of the output with the updated  `print_streaming_chunk` coming in a separate PR.
Screenshot 2025-05-09 at 09 54 38

**Fixing Consistency Between Streaming and Non-Streaming**
I updated the streaming processing to produce one ChatMessage like the non-streaming processing. This makes the most sense since this singular ChatMessage is converted into a single Bedrock formatted message. 

### How did you test it?

- Added unit tests
- Updated integration tests

### Checklist

### DIFF
--- a/integrations/amazon_bedrock/src/haystack_integrations/components/generators/amazon_bedrock/chat/utils.py
+++ b/integrations/amazon_bedrock/src/haystack_integrations/components/generators/amazon_bedrock/chat/utils.py
@@ -23,7 +23,8 @@ def _format_tools(tools: Optional[List[Tool]] = None) -> Optional[Dict[str, Any]
     Format Haystack Tool(s) to Amazon Bedrock toolConfig format.
 
     :param tools: List of Tool objects to format
-    :return: Dictionary in Bedrock toolConfig format or None if no tools
+    :returns:
+        Dictionary in Bedrock toolConfig format or None if no tools are provided
     """
     if not tools:
         return None
@@ -38,6 +39,13 @@ def _format_tools(tools: Optional[List[Tool]] = None) -> Optional[Dict[str, Any]
 
 
 def _format_tool_call_message(tool_call_message: ChatMessage) -> Dict[str, Any]:
+    """
+    Format a Haystack ChatMessage containing tool calls into Bedrock format.
+
+    :param tool_call_message: ChatMessage object containing tool calls to be formatted.
+    :returns:
+        Dictionary representing the tool call message in Bedrock's expected format
+    """
     content = []
     # Tool call message can contain text
     if tool_call_message.text:
@@ -51,6 +59,12 @@ def _format_tool_call_message(tool_call_message: ChatMessage) -> Dict[str, Any]:
 
 
 def _format_tool_result_message(tool_call_result_message: ChatMessage) -> Dict[str, Any]:
+    """
+    Format a Haystack ChatMessage containing tool call results into Bedrock format.
+
+    :param tool_call_result_message: ChatMessage object containing tool call results to be formatted.
+    :returns: Dictionary representing the tool result message in Bedrock's expected format
+    """
     # Assuming tool call result messages will only contain tool results
     tool_results = []
     for result in tool_call_result_message.tool_call_results:
@@ -74,10 +88,19 @@ def _format_tool_result_message(tool_call_result_message: ChatMessage) -> Dict[s
 
 
 def _repair_tool_result_messages(bedrock_formatted_messages: List[Dict[str, Any]]) -> List[Dict[str, Any]]:
+    """
+    Repair and reorganize tool result messages to maintain proper ordering and grouping.
+
+    Ensures tool result messages are properly grouped in the same way as their corresponding tool call messages
+    and maintains the original message ordering.
+
+    :param bedrock_formatted_messages: List of Bedrock-formatted messages that may need repair.
+    :returns: List of properly organized Bedrock-formatted messages with correctly grouped tool results.
+    """
     tool_call_messages = []
     tool_result_messages = []
     for idx, msg in enumerate(bedrock_formatted_messages):
-        content = msg.get("content", None)
+        content = msg.get("content", [])
         if content:
             if any("toolUse" in c for c in content):
                 tool_call_messages.append((idx, msg))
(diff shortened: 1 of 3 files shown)


sjrl:
@anakin87 I've assigned @vblagoje as the reviewer since he has more context. 

However, @anakin87 I think you'd find the root of the problem to be potentially interesting which is described in the PR description. Basically this could have been also overcome by updating how the ToolInvoker works and by adding support for multiple tool results being stored in a ChatMessage dataclass.

anakin87:
I'm not sure if I understand this correctly.

I think the choices we made in the past stem from the fact that most model providers (OpenAI, Google, Ollama) only accept messages from `tool` containing a single tool call result.
In this perspective, I think that handling this aspect specifically in the affected integrations is a viable option.

Please add more details if you think I have missed the point.

sjrl:
> I'm not sure if I understand this correctly.
> 
> I think the choices we made in the past stem from the fact that most model providers (OpenAI, Google, Ollama) only accept messages from `tool` containing a single tool call result. In this perspective, I think that handling this aspect specifically in the affected integrations is a viable option.
> 
> Please add more details if you think I have missed the point.

No I don't think you have. I ultimately agree that fixing it here makes the most sense. I just wasn't sure about this point

> I think the choices we made in the past stem from the fact that most model providers (OpenAI, Google, Ollama) only accept messages from `tool` containing a single tool call result.

so thanks for the clarification! 

In that case I agree leaving ToolInvoker as is makes the most sense.

Amnah199:
@sjrl As discussed in the [other PR,](https://github.com/deepset-ai/haystack-core-integrations/pull/1716) if this also addresses #1680 and #1681, please link them in the PR description for clarity.

integrations/amazon_bedrock/src/haystack_integrations/components/generators/amazon_bedrock/chat/utils.py
```diff
@@ -25,6 +37,95 @@ def _format_tools(tools: Optional[List[Tool]] = None) -> Optional[Dict[str, Any]
...
+                    "content": content,
+                    **({"status": "error"} if result.error else {}),
+                }
+            }
+        )
+    # role must be user
+    return {"role": "user", "content": tool_results}
+
+
+def _repair_tool_result_messages(bedrock_formatted_messages: List[Dict[str, Any]]) -> List[Dict[str, Any]]:
```
vblagoje, line 76: The only improvement I could find here is to introduce two helper methods:

```
def _find_content_blocks_by_key(message: Dict[str, Any], key_name: str) -> List[Dict[str, Any]]:
    """Returns a list of content blocks from the message that contain the given key_name."""
    content = message.get("content")
    found_blocks = []
    if content and isinstance(content, list):
        for c_block in content:
            if isinstance(c_block, dict) and key_name in c_block:
                found_blocks.append(c_block)
    return found_blocks


def _tool_use_content_blocks(message: Dict[str, Any]) -> List[Dict[str, Any]]:
    """Returns a list of 'toolUse' content blocks from the given Bedrock formatted message."""
    return _find_content_blocks_by_key(message, "toolUse")


def _tool_result_content_blocks(message: Dict[str, Any]) -> List[Dict[str, Any]]:
    """Returns a list of 'toolResult' content blocks from the given Bedrock formatted message."""
    return _find_content_blocks_by_key(message, "toolResult")
    ```
    
   And then reuse it throughout  _repair_tool_result_messages

**Review by vblagoje — COMMENTED** [review]
I've checked with agent and it works in both streaming and non-streaming method. ✅
The repair method is crucial and it deserves a bit more pydoc.
Suggested a small improvement for repair and other helper methods
I'll still take a look into streaming in detail, and verify tests.

File: integrations/amazon_bedrock/src/haystack_integrations/components/generators/amazon_bedrock/chat/utils.py [hunk 2]
```diff
@@ -143,78 +403,18 @@ def _parse_streaming_response(
...
-            for reply in replies:
-                if "usage" in metadata:
-                    usage = metadata["usage"]
-                    reply.meta["usage"] = {
-                        "prompt_tokens": usage.get("inputTokens", 0),
-                        "completion_tokens": usage.get("outputTokens", 0),
-                        "total_tokens": usage.get("totalTokens", 0),
-                    }
-
+        streaming_chunk = _convert_event_to_streaming_chunk(event=event, model=model)
```
vblagoje, line 408: I'd put somewhere here a short pydoc comment or a note to ourselves here I'd put that we receive these chunks from Bedrock and sometimes they are guaranteed to have atomic units with specific fields toolUseId, name etc (that's why we are guaranteed to be able to access them from event dict)  yet some blocks will contain only arguments for a function that we need to accumulate.

integrations/amazon_bedrock/tests/test_chat_generator.py
```diff
@@ -325,90 +325,94 @@ def streaming_callback(chunk: StreamingChunk):
...
+        assert tool_call_message is not None, "No message with tool call found"
+        assert isinstance(tool_call_message, ChatMessage), "Tool message is not a ChatMessage instance"
+        assert ChatMessage.is_from(tool_call_message, ChatRole.ASSISTANT), "Tool message is not from the assistant"
 
-        tool_call = tool_message.tool_call
-        assert tool_call.id, "Tool call does not contain value for 'id' key"
-        assert tool_call.tool_name == "weather"
-        assert tool_call.arguments == {"city": "Paris"}
-        assert tool_message.meta["finish_reason"] == "tool_use"
+        tool_calls = tool_call_message.tool_calls
```
vblagoje, line 350: Is len(tool_calls) == 2 here?

vblagoje (approved):
I've completed the review. Finally we get this one 🔥 
I'd use this opportunity to add some comments throughout the code so we can maintain it easier in the future. Great work 👍

integrations/amazon_bedrock/src/haystack_integrations/components/generators/amazon_bedrock/chat/utils.py
sjrl: That' doesn't work for me since I need to keep track of the original message idx not just accumulate the content blocks

integrations/amazon_bedrock/tests/test_chat_generator.py
sjrl: yup!

PR #1716, opened 2025-05-07
fix: ensure consistent behaviour for non/streaming in `AmazonBedrockChatGenerator`

### LINKED ISSUE #1680: fix: AmazonBedrockChatGenerator does not pass Tool Call streaming chunk to streaming_callback

### LINKED ISSUE #1681: fix: Inconsistent behavior between streaming and non-streaming in AmazonBedrockChatGenerator

Amnah199:
### Related Issues

- fixes #1681 
- fixes #1680 

### Proposed Changes:

- Updated `_parse_completion_response` to return text and `ToolCall` in separate `ChatMessage` objects, aligning the behavior with streaming responses.
- Pass the `tool_call` as streaming chunks to callback function.
- Add a `print_streaming_chunk` method similar to one we have for generators in haystack. That function cannot be used with bedrock (and possibly other gens) due to varied structuring in chunks.

### How did you test it?

Updated some existing tests
### Notes for the reviewer

NA
### Checklist


integrations/amazon_bedrock/src/haystack_integrations/components/generators/amazon_bedrock/utils.py
```diff
@@ -0,0 +1,40 @@
+from haystack.dataclasses import StreamingChunk
+
+
+def print_streaming_chunk(chunk: StreamingChunk):
```
Amnah199, line 4: @sjrl While updating the print_streaming_chunk function in the Haystack repo, we started thinking about how to standardize tool_call streaming across generators, since different APIs return data in different formats.

One possible solution could be to introduce a dedicated utility function for handling streaming in each generator. We already have a function for OpenAI in the core repo. For generators in `core-integrations`, we could fadd a separate streaming utility function within each integration.

sjrl:
While working on https://github.com/deepset-ai/haystack-core-integrations/pull/1711 I realized to get multi tool calling to work I also had to refactor the streaming parsing to actually match how parse completion response only produces one chat message. 

@Amnah199 sorry I didn’t say this earlier but I think it’d be best to close this PR and we’ll resolve this issue in my branch.

integrations/amazon_bedrock/src/haystack_integrations/components/generators/amazon_bedrock/utils.py
sjrl: I see what you’re saying but I’d rather create one util function that can handle all of our integrations. 

This way users can easily switch generators without also needing to switch streaming callbacks. It’s also something that @tstadel recently opened an issue about in the main repo.

sjrl: The issue he opened is here https://github.com/deepset-ai/haystack/issues/9347

Amnah199:
Closing this as its being handled here https://github.com/deepset-ai/haystack-core-integrations/pull/1711

PR #1723, opened 2025-05-11
feat: Add OpenRouter integration

### LINKED ISSUE #1158: Support for OpenRouter
[OpenRouter ](https://openrouter.ai/docs) is a great API alternative to the original OpenAI API as you already have access to all the models, including gpt-4-32k for example. 

I was able to make it work locally:

```
prompt_node = PromptNode(
    "openai/gpt-4-32k",
    api_key=openai_api_key,
    default_prompt_template=prompt_template,
    model_kwargs={"api_base": "https://openrouter.ai/api/v1"},
)
```

[https://github.com/deepset-ai/haystack/blob/main/haystack/nodes/prompt/invocation_layer/open_ai.py](https://github.com/deepset-ai/haystack/blob/main/haystack/nodes/prompt/invocation_layer/open_ai.py)
```
    @property
    def headers(self) -> Dict[str, str]:
        headers = {"Authorization": f"Bearer {self.api_key}", "Content-Type": "application/json"}
        if self.openai_organization:
            headers["OpenAI-Organization"] = self.openai_organization

        if "openrouter.ai" in self.api_base:
            headers["HTTP-Referer"] = "http://test.test"
            headers["X-Title"] = "Test"

        return headers
```

See their [Docs ](https://openrouter.ai/docs#api-keys) for explanation of the difference to the official API.

Amnah199:
### Related Issues

- fixes https://github.com/deepset-ai/haystack/issues/9274

### Proposed Changes:

- Add a `OpenRouterChatGenerator` that inherits from `OpenAIChatGenerator`

### How did you test it?

To be added

### Checklist


integrations/openrouter/src/haystack_integrations/components/generators/openrouter/chat/chat_generator.py
```diff
@@ -0,0 +1,193 @@
...
+    def __init__(
+        self,
+        api_key: Secret = Secret.from_env_var("OPENROUTER_API_KEY"),
+        model: str = "openai/gpt-4o-mini",
+        streaming_callback: Optional[Callable[[StreamingChunk], None]] = None,
+        api_base_url: Optional[str] = "https://openrouter.ai/api/v1",
+        generation_kwargs: Optional[Dict[str, Any]] = None,
+        tools: Optional[List[Tool]] = None,
+        timeout: Optional[int] = None,
+        extra_headers: Optional[Dict[str, Any]] = None,
```
Amnah199, line 70: I’ve added an `extra_headers` parameter so you can pass custom HTTP headers to the `OpenRouter` API. These headers are OpenRouter-specific and will be ignored by the official `OpenAI` endpoint. If we also want to expose `extra_headers` on the `run` (and `run_async`) methods, I can extend the implementation to override those calls accordingly.

integrations/openrouter/src/haystack_integrations/components/generators/openrouter/chat/chat_generator.py
```diff
@@ -0,0 +1,193 @@
...
+from haystack import component, default_to_dict, logging
+from haystack.components.generators.chat import OpenAIChatGenerator
+from haystack.dataclasses import ChatMessage, StreamingChunk
+from haystack.tools.tool import Tool, _check_duplicate_tool_names
+from haystack.utils import serialize_callable
+from haystack.utils.auth import Secret
+
+logger = logging.getLogger(__name__)
+
+StreamingCallbackT = Callable[[StreamingChunk], None]
```
sjrl, line 16: Hey @Amnah199 could we use the type defined from 
```python
from haystack.dataclasses.streaming_chunk import StreamingCallbackT
```
since the type is actually defined as 
```python
SyncStreamingCallbackT = Callable[[StreamingChunk], None]
AsyncStreamingCallbackT = Callable[[StreamingChunk], Awaitable[None]]

StreamingCallbackT = Union[SyncStreamingCallbackT, AsyncStreamingCallbackT]
```
since we allow for a Sync or Async version of the callback

integrations/openrouter/src/haystack_integrations/components/generators/openrouter/chat/chat_generator.py
```diff
@@ -0,0 +1,193 @@
...
+
+    client = OpenRouterChatGenerator()
+    response = client.run(messages)
+    print(response)
+
+    >>{'replies': [ChatMessage(content='Natural Language Processing (NLP) is a branch of artificial intelligence
+    >>that focuses on enabling computers to understand, interpret, and generate human language in a way that is
+    >>meaningful and useful.', role=<ChatRole.ASSISTANT: 'assistant'>, name=None,
+    >>meta={'model': 'openai/gpt-4o-mini', 'index': 0, 'finish_reason': 'stop',
+    >>'usage': {'prompt_tokens': 15, 'completion_tokens': 36, 'total_tokens': 51}})]}
```
sjrl, line 57: Could we update this example based on the feedback from this issue https://github.com/deepset-ai/haystack/issues/9366 namely that when printing a ChatMessage now the fields have underscores in them.

integrations/openrouter/src/haystack_integrations/components/generators/openrouter/chat/chat_generator.py
```diff
@@ -0,0 +1,193 @@
...
+    ```
+    """
+
+    def __init__(
+        self,
+        api_key: Secret = Secret.from_env_var("OPENROUTER_API_KEY"),
```
sjrl, line 63: I think we've started to prefer keyword only arguments, so if @julian-risch agrees I think changing this to 
```suggestion
        self,
        *,
        api_key: Secret = Secret.from_env_var("OPENROUTER_API_KEY"),
```
would be a good idea

integrations/openrouter/src/haystack_integrations/components/generators/openrouter/chat/chat_generator.py
```diff
@@ -0,0 +1,193 @@
...
+        model: str = "openai/gpt-4o-mini",
+        streaming_callback: Optional[Callable[[StreamingChunk], None]] = None,
```
sjrl, line 65: Let's use the type imported above here for the streaming callback so `StreamingCallbackT`

integrations/openrouter/src/haystack_integrations/components/generators/openrouter/chat/chat_generator.py
```diff
@@ -0,0 +1,193 @@
...
+        api_base_url: Optional[str] = "https://openrouter.ai/api/v1",
+        generation_kwargs: Optional[Dict[str, Any]] = None,
+        tools: Optional[List[Tool]] = None,
+        timeout: Optional[int] = None,
```
sjrl, line 69: The timeout in `OpenAIChatGenerator` is defined as `Optional[float]`
```suggestion
        timeout: Optional[float] = None,
```

integrations/openrouter/src/haystack_integrations/components/generators/openrouter/chat/chat_generator.py
```diff
@@ -0,0 +1,193 @@
...
+    """
+
+    def __init__(
+        self,
+        api_key: Secret = Secret.from_env_var("OPENROUTER_API_KEY"),
+        model: str = "openai/gpt-4o-mini",
+        streaming_callback: Optional[Callable[[StreamingChunk], None]] = None,
+        api_base_url: Optional[str] = "https://openrouter.ai/api/v1",
+        generation_kwargs: Optional[Dict[str, Any]] = None,
+        tools: Optional[List[Tool]] = None,
```
sjrl, line 68: We should probably update this to 
```suggestion
        tools: Optional[Union[List[Tool], Toolset]] = None,
```
The OpenAIChatgenerator already accepts `Toolset` and @vblagoje has been working on adding support for `Toolset` in our other integrations.

integrations/openrouter/src/haystack_integrations/components/generators/openrouter/chat/chat_generator.py
```diff
@@ -0,0 +1,193 @@
...
+        timeout: Optional[int] = None,
+        extra_headers: Optional[Dict[str, Any]] = None,
+        max_retries: Optional[int] = None,
+    ):
```
sjrl, line 72: Would it also make sense to allow for the `http_client_kwargs` that the `OpenAIChatGenerator` also supports?

integrations/openrouter/src/haystack_integrations/components/generators/openrouter/chat/chat_generator.py
```diff
@@ -0,0 +1,193 @@
...
+            max_retries=self.max_retries,
+        )
+
+    def _prepare_api_call(
+        self,
+        *,
+        messages: List[ChatMessage],
+        streaming_callback: Optional[StreamingCallbackT] = None,
+        generation_kwargs: Optional[Dict[str, Any]] = None,
+        tools: Optional[List[Tool]] = None,
```
sjrl, line 157: Let's add support for `Toolset` here as well
```suggestion
        tools: Optional[Union[List[Tool], Toolset]] = None,
```

integrations/openrouter/src/haystack_integrations/components/generators/openrouter/chat/chat_generator.py
Amnah199: Good point!

sjrl:
> @sjrl Thanks for the helpful review! I initially used `MistralChatGenerator` as a starting point for this integration — should we also update it with your suggested changes? I’ll create a small issue for it.

Yes that would be great!

integrations/openrouter/pyproject.toml
```diff
@@ -0,0 +1,173 @@
...
+  "Programming Language :: Python",
+  "Programming Language :: Python :: 3.9",
+  "Programming Language :: Python :: 3.10",
+  "Programming Language :: Python :: 3.11",
+  "Programming Language :: Python :: 3.12",
+  "Programming Language :: Python :: 3.13",
+  "Programming Language :: Python :: Implementation :: CPython",
+  "Programming Language :: Python :: Implementation :: PyPy",
+]
+dependencies = ["haystack-ai>=2.12.0"]
```
sjrl, line 26:
```suggestion
dependencies = ["haystack-ai>=2.13.1"]
```
We should update this to `2.13.1` since I believe that's when `Toolset` was added to Haystack

integrations/openrouter/src/haystack_integrations/components/generators/openrouter/__init__.py
```diff
@@ -0,0 +1,3 @@
+from .chat.chat_generator import OpenRouterChatGenerator
```
sjrl, line 1: I think we also should include the license header in this file too right?

sjrl:
@Amnah199 the rest looks good! I'd appreciate if @julian-risch could give it a look over in terms of reviewing what a new integration needs @dfokina for the docstrings on the new component :)

integrations/openrouter/src/haystack_integrations/components/generators/openrouter/__init__.py
Amnah199: These are demo files so it shouldn't matter but I'll add the header for consistency and we can keep it a standard for future examples.

julian-risch (changes requested):
Looks very good to me already! Just two things to be done before we can merge this PR. We need to add the new integration to
- https://github.com/deepset-ai/haystack-core-integrations/blob/main/.github/labeler.yml
- https://github.com/deepset-ai/haystack-core-integrations/blob/main/README.md

I already created the new label.
By the way, I used the checklist for new integrations that we have here: https://github.com/deepset-ai/haystack-core-integrations/blob/main/.github/ISSUE_TEMPLATE/new-integration-proposal.md#tasks

integrations/openrouter/src/haystack_integrations/components/generators/openrouter/chat/chat_generator.py
Amnah199: This isn't a critical parameter—it simply allows passing custom HTTP headers to `OpenRouter`, such as tags like X`-Title `or `X-Org-ID`. It might be useful in cases where the chat generator serves multiple front-ends and the user wants to assign a different title or identifier per request.

Since it's not an immediate need, we can leave it as is for now. If users later request the ability to set `extra_headers` at runtime, we can extend the run method to support it.

sjrl: sounds good!

integrations/openrouter/src/haystack_integrations/components/generators/openrouter/chat/chat_generator.py
```diff
@@ -0,0 +1,200 @@
...
+
+
+@component
+class OpenRouterChatGenerator(OpenAIChatGenerator):
+    """
+    Enables text generation using OpenRouter generative models.
+    For supported models, see [OpenRouter docs](https://openrouter.ai/models).
+
+    Users can pass any text generation parameters valid for the OpenRouter chat completion API
+    directly to this component via the `generation_kwargs` parameter in `__init__` or the `generation_kwargs`
```
dfokina, line 24:
```suggestion
    directly to this component using the `generation_kwargs` parameter in `__init__` or the `generation_kwargs`
```

integrations/openrouter/src/haystack_integrations/components/generators/openrouter/chat/chat_generator.py
```diff
@@ -0,0 +1,200 @@
...
+
+    Key Features and Compatibility:
+    - **Primary Compatibility**: Designed to work seamlessly with the OpenRouter chat completion endpoint.
+    - **Streaming Support**: Supports streaming responses from the OpenRouter chat completion endpoint.
+    - **Customizability**: Supports all parameters supported by the OpenRouter chat completion endpoint.
+
+    This component uses the ChatMessage format for structuring both input and output,
+    ensuring coherent and contextually relevant responses in chat-based text generation scenarios.
+    Details on the ChatMessage format can be found in the
+    [Haystack docs](https://docs.haystack.deepset.ai/v2.0/docs/data-classes#chatmessage)
```
dfokina, line 35:
```suggestion
    [Haystack docs](https://docs.haystack.deepset.ai/docs/chatmessage)
```

integrations/openrouter/src/haystack_integrations/components/generators/openrouter/chat/chat_generator.py
```diff
@@ -0,0 +1,200 @@
...
+        api_base_url: Optional[str] = "https://openrouter.ai/api/v1",
+        generation_kwargs: Optional[Dict[str, Any]] = None,
+        tools: Optional[Union[List[Tool], Toolset]] = None,
+        timeout: Optional[float] = None,
+        extra_headers: Optional[Dict[str, Any]] = None,
+        max_retries: Optional[int] = None,
+        http_client_kwargs: Optional[Dict[str, Any]] = None,
+    ):
+        """
+        Creates an instance of OpenRouterChatGenerator. Unless specified otherwise
```
dfokina, line 74:
```suggestion
        Creates an instance of OpenRouterChatGenerator. Unless specified otherwise,
```

dfokina (commented):
Adding some small fixes for the docstrings. Also, is the `run` method supposed to be somewhere in the `chat_generator.py`?

dfokina:
Also, after this is merged, could you please create a documentation issue 🙏

Amnah199:
@dfokina you are right. Because of inheritance, the component uses the `run` method from `OpenAIChatGenerator`. Plus updated the docstrings 👍 .

julian-risch (approved):
LGTM! And a GitHub secret OPENROUTER_API_KEY is set now. ✔️

PR #1727, opened 2025-05-12
fix: fix error in Tasks checklist

Amnah199:
### Proposed Changes:

The Tasks checklist in `new integration `issue template is currently displayed as a code block. Updated it to render as an interactive checklist instead.


anakin87 (approved):
Thanks!

PR #1729, opened 2025-05-13
fix: correct Azure AI search _in delimiter

sggerard:
https://learn.microsoft.com/en-us/azure/search/search-query-odata-search-in-function

The azure search.in() filter accepts a list of strings to match against. However the way that the function to normalize the haystack 'in' operator to the azure 'in' operator appears to be incorrect. Despite some of the documentation Microsoft has provided it appears we do need to provide the delimiter to use (In this case ',') as the third parameter.

image

### Proposed Changes:

Specify the delimiter to use when building the search.in string.
Remove the extra space used in the join function that would prevent matching.

### How did you test it?

I identified this bug when some of my project filters were failing and I couldn't see any reason as to why. I identified this as the issue and after some local debugging and hot fixing the library, and have concluded that at least for my project/ use case this is the appropriate fix. 

### Notes for the reviewer

This has only been tested against my haystack project and Azure AI search and is no means exhaustive.


Amnah199 (approved):
LG! @sggerard Thanks for the fix!

PR #1743, opened 2025-05-14
feat: Add user-agent to `AzureAISearchDocumentStore`

Amnah199:
### Proposed Changes:

As requested by the Azure AI team, I’ve added a user agent policy to the client requests so they can track how our integration is being used.


mpangrazzi (approved):
LGTM! I was also wondering if there are other HTTPS calls we are doing without an User Agent specified and if we have a common one (not anything urgent, just asking)

Amnah199:
I don’t think we consistently pass a user agent across all integrations. I know we include one in the Pinecone integration [here](https://github.com/deepset-ai/haystack-core-integrations/blob/ba0a310269262517bce2b702e4f5bd679b391863/integrations/pinecone/src/haystack_integrations/document_stores/pinecone/document_store.py#L81), but we don’t have a standard. Good point though, maybe we can have a standard for this.

PR #1758, opened 2025-05-16
feat: MCPTool and MCPToolset async resource management improvements

vblagoje:
## Why
This PR addresses a critical issue with AnyIO/asyncio cancel scopes in MCP Tools. The core problem was that connection opening and closing occurred in different tasks, leading to the dreaded `Attempted to exit cancel scope in a different task than it was entered in` error. This issue affects proper resource cleanup and connection management when MCP tools are used in synchronous contexts, causing intermittent ☝️ failures and potential resource leaks.

## What
- **In `mcp_tool.py`**:
  - Implemented `_MCPClientSessionManager` to ensure open/close operations happen in the same async task
  - Added `AsyncExecutor.run_background()` method for proper cross-thread coroutine management
  - Created a lifecycle that connects, gets tools, and waits on an event before closing
  - Improved error handling, particularly for connection errors

- **In `mcp_toolset.py`**:
  - Applied similar connection management improvements

- **In test files**:
  - Added in-memory MCP client implementation for better testing (as mcp pyhon sdk does)
  - Created a central `mcp_tool_cleanup` fixture for consistent tool resource management
  - Enhanced test cleanup logic to maintain proper shutdown sequence

## How can it be used
These changes are transparent to users, with the API remaining unchanged. However:
- Explicit calls to `sync_close()` on tools and toolsets are recommended, if possible, rather than relying on garbage collection

## How did you test it
- Introduced an in-memory MCP client implementation for faster and more isolated unit testing
- Integration tests use real HTTP/SSE servers to verify actual connectivity
- Extensively manually tested with the itinerary agent in real-world usage scenarios

## Notes for the reviewer
- The core solution is in `_MCPClientSessionManager`, which ensures connect/close operations occur in the same async task
- Review how the class uses a long-running coroutine with an event-based shutdown mechanism to ensure connect and close run in the same task
- Note how we use `Future` and `asyncio.Event` to coordinate between threads
- The comment in `_MCPClientSessionManager` explaining the life-cycle is key to understanding the solution


mpangrazzi (commented):
I've left two comments. 

I think this approach definitely makes this first version of MCPTool more solid!

However, it's still a bit complex to me (i.e. you have to carefully manage a daemon thread and an event loop). I think it would worth to review it in the future, since running `MCPTool` from a sync context shouldn't be the preferred way.

One minor note: consider add `AsyncExecutor.__init__` return type to `None`, so you'll avoid linter warnings (`Mypy(annotation-unchecked)`).

vblagoje:
> I've left two comments.
> 
> I think this approach definitely makes this first version of MCPTool more solid!
> 
> However, it's still a bit complex to me (i.e. you have to carefully manage a daemon thread and an event loop). I think it would worth to review it in the future, since running `MCPTool` from a sync context shouldn't be the preferred way.
> 
> One minor note: consider add `AsyncExecutor.__init__` return type to `None`, so you'll avoid linter warnings (`Mypy(annotation-unchecked)`).

I agree - this is a temp bridge for sync-to-async, not a preferred, but we are forced to be in this box solution until full transition to async world.

mpangrazzi:
@vblagoje I agree as well. To me it looks good, don't know if @sjrl wants to take a look too. If not I would approve!

sjrl:
> @vblagoje I agree as well. To me it looks good, don't know if @sjrl wants to take a look too. If not I would approve!

@mpangrazzi I trust you on this one!

vblagoje:
@mpangrazzi I made small changes for consistency and added more comments to clarify ideas 🙏

PR #1765, opened 2025-05-19
feat: OllamaChatGenerator - add Toolset support

vblagoje:
WIP - dry run on CI

### DIFF
--- a/integrations/ollama/src/haystack_integrations/components/generators/ollama/chat/chat_generator.py
+++ b/integrations/ollama/src/haystack_integrations/components/generators/ollama/chat/chat_generator.py
@@ -2,7 +2,8 @@ from typing import Any, Callable, Dict, List, Literal, Optional, Union
 
 from haystack import component, default_from_dict, default_to_dict
 from haystack.dataclasses import ChatMessage, StreamingChunk, ToolCall
-from haystack.tools import Tool, _check_duplicate_tool_names, deserialize_tools_or_toolset_inplace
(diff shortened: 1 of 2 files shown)


vblagoje:
@mpangrazzi let's have @anakin87 look at this changes - he's familiar with the code and the context of https://github.com/deepset-ai/haystack/issues/9418

sjrl:
@vblagoje lets make sure to pin Haystack in the deps to when we added Toolset

anakin87:
Looks good in general.

I would only update this aspect in the PR description
> Automatically generates UUIDs for ToolCall.id if missing (e.g., in Ollama tool responses)

I understand that this is something you had to change to make Langfuse work, but in the current version of this PR we are not introducing this change.

vblagoje:
> Looks good in general.
> 
> I would only update this aspect in the PR description
> 
> > Automatically generates UUIDs for ToolCall.id if missing (e.g., in Ollama tool responses)
> 
> I understand that this is something you had to introduce to make Langfuse work, but in the current version of this PR we are not introducing this change.

Right, spot on! I had to remove it pending a proper resolution of https://github.com/deepset-ai/haystack/issues/9418

anakin87 (changes requested):
Apart from pinning Haystack (see https://github.com/deepset-ai/haystack-core-integrations/pull/1765#issuecomment-2897030739), looks good

anakin87 (approved):
Ah, sorry. Looks good!

Issue #1760, opened 2025-05-12
Add `streamable_http` transport support to `MCPTool` and `MCPToolset`


vblagoje:
**Is your feature request related to a problem? Please describe.**
Currently, our `MCPTool` and `MCPToolset` support SSE transport. With the deprecation notice for SSE and the introduction of `streamable_http` in v1.8.0 of the Python SDK, we need to upgrade our mcp integration as well.

**Describe the solution you'd like**
Add native support for `streamable_http` in `MCPTool` and `MCPToolset` , allowing connection to server tools supporting new `streamable_http`

**Describe alternatives you've considered**
None

**Additional context**
See MCP Python SDK v1.8.0 release notes: [https://github.com/modelcontextprotocol/python-sdk/releases/tag/v1.8.0](https://github.com/modelcontextprotocol/python-sdk/releases/tag/v1.8.0) — it introduces `streamable_http` as a first-class streaming transport alongside SSE.

PR #1777, opened 2025-05-22
feat: Add streamable-http transport MCP support

vblagoje:
### Why:
Add support for streamable-http transport to MCP tools and implement proper resource management

- fixes https://github.com/deepset-ai/haystack-core-integrations/issues/1760

### What:
- Added `StreamableHttpClient` and `StreamableHttpServerInfo` classes
- Added support for streamable-http transport in addition to existing SSE transport
- Updated error messages with more detailed information including arguments that caused failures
- Standardized client and server examples to use consistent command-line arguments

### How can it be used:
The MCP tools and examples now support both SSE and streamable-http transports:

```python
# SSE transport
server_info = SSEServerInfo(url="http://localhost:8000/sse")

# Streamable HTTP transport
server_info = StreamableHttpServerInfo(url="http://localhost:8000/mcp")

# Create tool with selected transport
tool = MCPTool(name="tool_name", server_info=server_info)
```

### How did you test it:
Updated all examples to demonstrate both transport types and proper resource management.  Tested all examples. 

Unit tests include:
- Server and client connections using both transport methods
- Toolset initialization with different transport configurations
- Error handling and resource cleanup in all examples

Manual tests include:
- Itinerary agent with a mix of streamable-http, sse and stdio servers

### Notes for the reviewer:
- See if there are any patterns that were broken, or any potential oddities compared to other two transports


vblagoje:
@Amnah199 please have a look as well, @mpangrazzi is familiar with the codebase and he'l review but it would be nice to have another set of eyes go through it 🙏

Amnah199:
@vlad I have a question about a code section not part of this PR.
Shouldn't we be closing the event_loop in `AsyncExecutor.shutdown()` after stopping it.

```
def shutdown(self, timeout: float = 2):
    """
    Shut down the background event loop and thread.

    :param timeout: Timeout in seconds for shutting down the event loop
    """
    if self._loop.is_closed():
        return

    def _stop_loop():
        self._loop.stop()

    try:
        asyncio.run_coroutine_threadsafe(asyncio.sleep(0), self._loop).result(timeout=timeout)
        self._loop.call_soon_threadsafe(_stop_loop)
        self._thread.join(timeout=timeout)
    finally:
        # Always close the loop to prevent resource leaks
        if not self._loop.is_closed():
            self._loop.close()
```

vblagoje:
> @vlad I have a question about a code section not part of this PR. Shouldn't we be closing the event_loop in `AsyncExecutor.shutdown()` after stopping it.
> 
> Updated
> 
> ```
> def shutdown(self, timeout: float = 2):
>     """
>     Shut down the background event loop and thread.
> 
>     :param timeout: Timeout in seconds for shutting down the event loop
>     """
>     if self._loop.is_closed():  ## this can be removed if we want to raise a RunTimeError
>         return
> 
>     def _stop_loop():
>         self._loop.stop()
> 
>     try:
>         asyncio.run_coroutine_threadsafe(asyncio.sleep(0), self._loop).result(timeout=timeout)
>         self._loop.call_soon_threadsafe(_stop_loop)
>         self._thread.join(timeout=timeout)
>     finally:
>         # Always close the loop to prevent resource leaks
>         if not self._loop.is_closed():
>             self._loop.close()
> ```

Hey @Amnah199 I'm @vblagoje 😄 

Yes good point - but this event loop is intended to run for the entire lifetime of the interpreter where synchronous versions of MCPTool and toolset are used - at least for now. The shutdown here just in case for potential future use cases where that might not be true.

integrations/mcp/examples/mcp_client.py
```diff
@@ -0,0 +1,68 @@
...
+
+    # Construct the appropriate URL based on transport type
+    if args.transport == "sse":
+        server_info = SSEServerInfo(url="http://localhost:8000/sse")
+    else:  # streamable-http
+        server_info = StreamableHttpServerInfo(url="http://localhost:8000/mcp")
+
+    try:
+        tool = MCPTool(name="add", server_info=server_info)
+        tool_subtract = MCPTool(name="subtract", server_info=server_info)
```
Amnah199, line 47: If an exception is raised, the `tool` and `tool_subtract` will not be initialized to be used in `finally` block, and will instead throw an error.
```suggestion
    tool = None
    tool_subtract = None
    try:
        tool = MCPTool(name="add", server_info=server_info)
        tool_subtract = MCPTool(name="subtract", server_info=server_info)
```

integrations/mcp/examples/mcp_filtered_tools.py
```diff
@@ -17,7 +17,7 @@ def main():
         print("Creating toolset with all available tools:")
         # Create a toolset with all available tools
         full_toolset = MCPToolset(
-            server_info=SSEServerInfo(base_url="http://localhost:8000"),
+            server_info=SSEServerInfo(url="http://localhost:8000/sse"),
```
Amnah199, line 20: I would also recommend to initialize `full_toolset=None` and `filtered_toolset=None` outside of `try` block.

Issue #1786, opened 2025-05-26
QDrantDocumentStore serializes plaintext API key when to_dict is called


mathislucka:
**Describe the bug**
I noticed this when calling pipeline.dumps but it should be the case when simply calling `to_dict` on the QDrantDocumentStore as well. The YAML config for the document store will contain a `metadata` key which holds:
- `User-Agent`
- `api-key`

The `api-key` attribute has the api key (loaded via Secret.from_env_var) in plaintext.

Expected behavior would be that the api key does not get serialized in plain text. 

**To Reproduce**
Create an instance of QDrantDocumentStore and call to_dict

**Describe your environment (please complete the following information):**
 - Haystack version:
 - Integration version:

sjrl:
Hey @mathislucka could you provide some example code that causes this issue. I'm checking locally and I'm unable to reproduce currently.

mathislucka:
Updated the example accordingly

sjrl:
@mathislucka thanks for the additional context! I was able to reproduce when running a local version of QDrant using docker. If using the in memory version this issue doesn't occur.

sjrl:
@mathislucka new version with the fix has been released https://pypi.org/project/qdrant-haystack/9.1.2/

PR #1813, opened 2025-05-27
fix: Fix exposing Qdrant api-key in `metadata` field when running `to_dict`

sjrl:
### Related Issues

- fixes https://github.com/deepset-ai/haystack-core-integrations/issues/1786#issuecomment-2909698240

### Proposed Changes:

 
 
Deepcopy the returned client params in `_prepare_client_params` to prevent changing the original class attributes.

This specifically only occurs when you run a hosted version of QDrant (but not when running the in memory version), initialize the client and then run `to_dict`.

### How did you test it?

Added a unit test

### Checklist

### DIFF
--- a/integrations/qdrant/src/haystack_integrations/document_stores/qdrant/document_store.py
+++ b/integrations/qdrant/src/haystack_integrations/document_stores/qdrant/document_store.py
@@ -1,6 +1,6 @@
 import inspect
 from itertools import islice
-from typing import Any, AsyncGenerator, ClassVar, Dict, Generator, List, Optional, Set, Union
+from typing import Any, AsyncGenerator, ClassVar, Dict, Generator, List, Optional, Set, Tuple, Union
 
 import numpy as np
 import qdrant_client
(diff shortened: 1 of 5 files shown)


integrations/qdrant/src/haystack_integrations/document_stores/qdrant/document_store.py
```diff
@@ -1537,26 +1535,31 @@ def _prepare_collection_params(self):
...
+                "port": self.port,
+                "grpc_port": self.grpc_port,
+                "prefer_grpc": self.prefer_grpc,
+                "https": self.https,
+                "api_key": self.api_key.resolve_value() if self.api_key else None,
+                "prefix": self.prefix,
+                "timeout": self.timeout,
+                "host": self.host,
+                "path": self.path,
+                "metadata": self.metadata,
```
mathislucka, line 1559: Wouldn't it be enough to only deepcopy metadata? Or maybe something like: `"metadata": {**self.metadata}`?

All other params are primitives and they should not be affected when the QDrantClient changes them. Or am I missing something?

sjrl: yeah that's fair we could just target the metadata field in this case

PR #1841, opened 2025-05-27
build: pinning `unstructured` to lowest working versions

### LINKED ISSUE #1809: Set lower-bound pins for Unstructured
More details in the parent issue.

For each integration with errors, we should follow these steps:
- set correct lower-bound pins for direct dependencies in pyproject.toml (this should ensure that the script runs without errors)
- add a CI job in the relevant workflow to check this aspect (example in #1788)
- when the PR is merged, release a new version of the integration
  - if we are sure to have found the minimum lower bound, a new bugfix version can be released (X.Y.Z+1)
  - if for some reason we were not able to find the minimum lower bound, this is technically a breaking change, so we should release a new major version (X+1.Y.Z)

davidsbatista:
### Related Issues

- fixes #1809 

### Proposed Changes:

- "haystack-ai>=2.0.0",
- "unstructured>=0.15.0",
- "unstructured-client>=0.21.0"

### Checklist

### DIFF
--- a/integrations/unstructured/pyproject.toml
+++ b/integrations/unstructured/pyproject.toml
@@ -24,9 +24,11 @@ classifiers = [
   "Programming Language :: Python :: Implementation :: PyPy",
 ]
 dependencies = [
-  "haystack-ai",
+  "haystack-ai>=2.0.0",
   "unstructured",
-  "unstructured-client<0.30.0"]
+  "unstructured-client"
+]
+
 
 [project.urls]
 Documentation = "https://github.com/deepset-ai/haystack-core-integrations/tree/main/integrations/unstructured#readme"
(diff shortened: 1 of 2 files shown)


integrations/unstructured/pyproject.toml
```diff
@@ -23,7 +23,12 @@ classifiers = [
   "Programming Language :: Python :: Implementation :: CPython",
   "Programming Language :: Python :: Implementation :: PyPy",
 ]
-dependencies = ["haystack-ai", "unstructured", "unstructured-client<0.30.0"]
+dependencies = [
+  "haystack-ai>=2.0.0",
+  "unstructured>=0.15.0",
+  "unstructured-client>=0.21.0"
```
anakin87, line 29:
```suggestion
  "unstructured-client>=0.21.0, <0.30.0"
```

If I'm not wrong, we are still incompatible with newer versions of the client and we want to prevent users from installing them

anakin87: no I see that the tests pass, so for some reason, it seems that it works

SORRY

anakin87 (approved):

Issue #1780, opened 2025-05-21
`ToolCall` ID requirement in `to_openai_dict_format` causes failure with non-OpenAI models and Langfuse


vblagoje:
**Describe the bug**
Using Langfuse tracing with chat models that don't require `ToolCall` IDs (e.g., Google Gemini, Ollama) leads to a `ValueError` when serializing chat messages using `to_openai_dict_format()` from `ChatMessage`.

Currently, `chat_message.py` enforces `ToolCall.id` presence when calling `to_openai_dict_format`, with this hardcoded assumption:

```python
raise ValueError("`ToolCall` must have a non-null `id` attribute to be used with OpenAI.")
```

However, Langfuse when serializing ChatMessage instances invokes  to_openai_dict_format regardless of model provider, which breaks serialization for valid messages from other providers.

---

**Error message**

```
ValueError: `ToolCall` must have a non-null `id` attribute to be used with OpenAI.
```

---

**Expected behavior**
Serialization should not fail when using non-OpenAI models that don't require `ToolCall.id`. The function should either:

* Not raise by default
* Or Langfuse should be updated to avoid calling this method when not using OpenAI

---

**Additional context**
This was encountered when running an itinerary agent using Google Gemini + Langfuse tracing. Confirmed on Ollama as well. 
---

**To Reproduce**

1. Enable Langfuse tracing
2. Use any non-OpenAI chat model that doesn't provide `ToolCall.id`
3. Run agent
4. Observe the serialization error from Langfuse integration

---

**FAQ Check**

* [x] Have you had a look at [[our new FAQ page](https://docs.haystack.deepset.ai/docs/faq)](https://docs.haystack.deepset.ai/docs/faq)?

PR #1885, opened 2025-06-02
feat: Use Langfuse local to_openai_dict_format function to serialize messages

vblagoje:
## Why:
The Langfuse integration was using `ChatMessage.to_openai_dict_format()` from Haystack core, which has strict validation requirements (e.g., requiring ToolCall IDs). This led to several ChatGenerators that don't support tool call IDs failing during Langfuse tracing. To maintain stability and avoid potential breaking changes from upstream modifications to the core method, we need local control over ChatMessage serialization for Langfuse tracing.

- fixes https://github.com/deepset-ai/haystack-core-integrations/issues/1780

## What:
- **Core component**: Local `to_openai_dict_format` function in Langfuse tracer
- **Key features**: 
  - Less strict validation than core Haystack method (doesn't require ToolCall ID presence)
  - Maintains OpenAI API format compatibility for Langfuse ingestion
- **Integration**: Replaces calls to `ChatMessage.to_openai_dict_format()` with local implementation in span content tagging

## How can it be used:
- The change is internal to langfuse tracing - no user-facing API changes

## How did you test it:
- **Manual testing**: Verified with multiple chat generators to ensure proper serialization
- **Compatibility**: Confirmed OpenAI format compatibility for Langfuse ingestion
- **Edge cases**: Tested with tool calls, tool call results, and various message types using itinerary agent

## Notes for the reviewer:
- **Dependencies**: No new packages added - uses existing `json` module
- **Compatibility**: Maintains backward compatibility while providing more flexible serialization
- **Limitations**: Local implementation needs to be maintained separately from core Haystack changes, we now have duplication of code, albeit minimal
- **Risk mitigation**: Isolates Langfuse integration from potential breaking changes in core serialization logic

### DIFF
--- a/integrations/langfuse/src/haystack_integrations/tracing/langfuse/tracer.py
+++ b/integrations/langfuse/src/haystack_integrations/tracing/langfuse/tracer.py
@@ -65,9 +65,14 @@ _COMPONENT_INPUT_KEY = "haystack.component.input"
 tracing_context_var: ContextVar[Dict[Any, Any]] = ContextVar("tracing_context")
 
 
-def to_openai_dict_format(chat_message: ChatMessage) -> Dict[str, Any]:
+def _to_openai_dict_format(chat_message: ChatMessage) -> Dict[str, Any]:
     """
     Convert a ChatMessage to the dictionary format expected by OpenAI's chat completion API.
+
+    Note: We already have such a method in Haystack's ChatMessage class.
+    However, the original method doesn't tolerate None values for ids of ToolCall and ToolCallResult.
+    Some generators, like GoogleGenAIChatGenerator, return None values for ids of ToolCall and ToolCallResult.
+    To seamlessly support these generators, we use this, Langfuse local, version of the method.
 
     :param chat_message: The ChatMessage instance to convert.
     :return: Dictionary in OpenAI Chat API format.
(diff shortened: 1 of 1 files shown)


vblagoje:
@mpangrazzi Seb and I talked about this one already internally. He has all the context already so I asked him to review

integrations/langfuse/src/haystack_integrations/tracing/langfuse/tracer.py
```diff
@@ -64,6 +65,56 @@
 tracing_context_var: ContextVar[Dict[Any, Any]] = ContextVar("tracing_context")
 
 
+def to_openai_dict_format(chat_message: ChatMessage) -> Dict[str, Any]:
```
sjrl: @vblagoje what do you think of making this private for now. Since it's only internal to Langfuse and I think we'd prefer users to still use the `ChatMessage.to_openai_dict_format` in general.

integrations/langfuse/src/haystack_integrations/tracing/langfuse/tracer.py
```diff
@@ -64,6 +65,56 @@
...
+    """
+    Convert a ChatMessage to the dictionary format expected by OpenAI's chat completion API.
```
sjrl, line 70: I'd update this docstring to mention the difference between this method and the `ChatMessage` util one.

sjrl (approved):
Looks good!

PR #1906, opened 2025-06-05
fix: Ollama tools with streaming

Sheripov:
## Proposed Changes
- Bug fix – streaming + tools
- Re-implemented _handle_streaming_response to accumulate tool-call argument fragments that arrive as either strings or dicts, preventing the TypeError: can only concatenate str (not "dict") to str.
- Deduplicates tool-calls by ID and rebuilds ToolCall objects with complete arguments before passing them to Haystack.
- API & docs polish
- Added full :param / :returns docstrings for every public method and for the class constructor (plain-text style, no Markdown).
- Renamed constructor parameter url → api_base_url to match Haystack conventions; kept previous default.
- Clarified example in class docstring; shows attaching a demo tool and streaming output.
- Resilience & minor refactor
- Unified metadata translation via _convert_ollama_meta_to_openai_format.
- Ensured serialization handles callbacks and tools correctly.
- Simplified chunk-building; preserved role and tool-call meta for debugging.

## How did you test it?
- Verified that run(..., stream=False) returns a ChatMessage with correct tool_calls.
- Verified that run(..., streaming_callback=…) streams tokens, merges them, and returns identical final message.
- Manual verification:
  1. Started local Ollama with mistral-small3.1:24b and a dummy function-calling model.
  2. Invoked component with a tool that echoes queries; confirmed callback prints tokens live and PR returns final reply ## with populated tool_calls.
  3. Re-checked behaviour with multiple tools and with no tools specified.
  
Code for test:
```python
from haystack.components.generators.utils import print_streaming_chunk
from haystack.components.agents import Agent
from components.OlamaChatGenerator import OllamaChatGenerator
def echo(query: str) -> str:
    print("Tool executed with QUERY: {🔍 {query}}")
    return f"🔍 {query}"
echo_tool = Tool(
    name="echo_tool",
    description="Echoes the query (demo tool).",
    function=echo,
    parameters={"query": {"type": "string", "description": "Search query"}},
)
agent = Agent(
    chat_generator=OllamaChatGenerator(model="mistral-small3.1:24b"),
    tools=[echo_tool],
    system_prompt="Use tool to print the query to test tools. Do not answer the question, just send the query to the tool",
    max_agent_steps=5,
    raise_on_tool_invocation_failure=True,
    streaming_callback=lambda ch: print(ch.content, end="", flush=True),  # comment out for test without stream
)
messages = [ChatMessage.from_user("This is stream test of tool usage")]
result = agent.run(messages=messages)
for message in result["messages"]:
    print("\n======")
    if message.role == "system":
        continue
    elif message.role == "tool":
        print(f"{message.role}:")
        print(f"Tool Results: {[tool.result for tool in message.tool_call_results]}")
        print(f"Used Tools: {[tool.origin.tool_name for tool in message.tool_call_results]}\n")
    else:
        print(f"{message.role}: {message.text}")
        print(f"Used Tools: {[tool.tool_name for tool in message.tool_calls]}\n")
```

## Notes for the reviewer
- Focus on _handle_streaming_response; that’s where most of the logic changed.
- Docstring updates are extensive but mechanical—no behavioural impact.
- If you have an Ollama instance handy, run the example in the class docstring to sanity-check streaming.

### DIFF
--- a/integrations/ollama/src/haystack_integrations/components/generators/ollama/chat/chat_generator.py
+++ b/integrations/ollama/src/haystack_integrations/components/generators/ollama/chat/chat_generator.py
@@ -107,6 +107,31 @@ class OllamaChatGenerator:
 
     * Fully supports streaming.
     * Correctly passes tool-calls to Haystack when `stream=True`.
+
+    Self-test / demo:
+    ```python
+    from haystack.dataclasses import ChatMessage
+    from haystack.tools import Tool
+
+    def echo_tool(query: str) -> str:
+        return f"🔍 {query}"
+
+    rag_search = Tool(
+        name="rag_search",
+        description="Echoes the query (demo tool).",
+        func=echo_tool,
+        parameters={"query": {"type": "string", "description": "Search query"}},
+    )
+
+    gen = OllamaChatGenerator(model="phi3", tools=[rag_search])
+
+    msgs = [ChatMessage.from_user("Explain linear interpolation (G1)")]
+    result = gen.run(
+        messages=msgs,
+        streaming_callback=lambda ch: print(ch.content, end="", flush=True),
+    )
+    print("\n\nAssistant:", result["replies"][0].texts[0])
+    ```
     """
 
     # ------------------------------------------------------------------
@@ -115,7 +140,7 @@ class OllamaChatGenerator:
     def __init__(
             self,
             model: str = "orca-mini",
-            url: str = "http://localhost:11434",
+            api_base_url: str = "http://localhost:11434",
             generation_kwargs: Optional[Dict[str, Any]] = None,
             timeout: int = 120,
             keep_alive: Optional[Union[float, str]] = None,
@@ -124,19 +149,35 @@ class OllamaChatGenerator:
             response_format: Optional[Union[None, Literal["json"], JsonSchemaValue]] = None,
     ):
         """
-        :param model: Name of the model inside the running Ollama instance.
-        :param url:   Base URL of that instance.
-        :param generation_kwargs: Default inference-options (temperature, top_p …).
-        :param timeout:  API timeout in seconds.
-        :param keep_alive: Ollama “keep-alive” setting (seconds, duration string, -1 or '0').
-        :param streaming_callback: Optional token callback when `stream=True`.
-        :param tools: List[Tool] **or** Toolset; duplicate names are rejected.
-        :param response_format: None / "json" / JSON Schema dict (Ollama ≥ 0.1.34).
+        :param model:
+            The name of the model to use. The model must already be present (pulled) in the running Ollama instance.
+        :param api_base_url:
+            The base URL of the Ollama server (default ``"http://localhost:11434"``).
+        :param generation_kwargs:
+            Default inference options—e.g. ``temperature``, ``top_p``,
+            ``num_predict``—merged with any ``generation_kwargs`` passed to
+            :meth:`run`.
+        :param timeout:
+            Socket timeout *in seconds* for HTTP calls to Ollama.
+        :param keep_alive:
+            How long Ollama should keep the model in memory after the request.
+            Accepts seconds (int/float), a duration string (``"10m"``, ``"24h"``),
+            ``-1`` (never unload), or ``'0'`` (unload immediately).
+        :param streaming_callback:
+            Callable invoked **once per :class:`StreamingChunk`** when streaming is
+            active.  Supply either here *or* in :meth:`run` to enable streaming.
+        :param tools:
+            A list of :class:`haystack.tools.Tool` **or** a
+            :class:`haystack.tools.Toolset`. Duplicate tool names raise a
+            ``ValueError``.
+        :param response_format:
+            ``None`` for free-form text, ``"json"`` to require valid JSON, or a
+            JSON-Schema dict (needs Ollama ≥ 0.1.34).
         """
         _check_duplicate_tool_names(tools)
 
         self.model = model
-        self.url = url
+        self.url = api_base_url
         self.generation_kwargs = generation_kwargs or {}
         self.timeout = timeout
         self.keep_alive = keep_alive
@@ -273,11 +314,21 @@ class OllamaChatGenerator:
             streaming_callback: Optional[Callable[[StreamingChunk], None]] = None,
     ):
         """
-        Generate a reply given the full `messages` history.
+        Generate a reply from the model.
 
-        * If `streaming_callback` **or** the instance was created with one,
-          streaming is enabled.
-        * `tools` overrides the instance’s default list for this call only.
+        :param messages:
+            Complete chat history, including the user’s **latest** message.
+        :param generation_kwargs:
+            Per-call overrides for Ollama inference options.  These are merged
+            on top of the instance-level ``generation_kwargs``.
+        :param tools:
+            Temporarily override the component’s tool list for *this* call.
+        :param streaming_callback:
+            A callable to receive :class:`StreamingChunk` objects as they
+            arrive.  Supplying a callback (here or in the constructor) switches
+            the component into streaming mode.
+        :returns:
+            ``{"replies": List[ChatMessage]}`` – ready for Haystack pipelines.
         """
         generation_kwargs = {**self.generation_kwargs, **(generation_kwargs or {})}
         tools = tools or self.tools
--- a/integrations/ollama/src/haystack_integrations/components/generators/ollama/chat/chat_generator.py
+++ b/integrations/ollama/src/haystack_integrations/components/generators/ollama/chat/chat_generator.py
@@ -28,24 +28,25 @@ def _convert_chatmessage_to_ollama_format(message: ChatMessage) -> Dict[str, Any
     tool_call_results = message.tool_call_results
 
     if not text_contents and not tool_calls and not tool_call_results:
-        raise ValueError(
-            "A `ChatMessage` must contain at least one `TextContent`, `ToolCall`, or `ToolCallResult`."
-        )
+        msg = "A `ChatMessage` must contain at least one `TextContent`, `ToolCall`, or `ToolCallResult`."
+        raise ValueError(msg)
     elif len(text_contents) + len(tool_call_results) > 1:
-        raise ValueError("A `ChatMessage` can only contain one `TextContent` or one `ToolCallResult`.")
+        msg = "A `ChatMessage` can only contain one `TextContent` or one `ToolCallResult`."
+        raise ValueError(msg)
 
     ollama_msg: Dict[str, Any] = {"role": message._role.value}
 
     if tool_call_results:
-        # Ollama does not expose an error-field for failed tool invocations.
+        # Ollama does not provide a way to communicate errors in tool invocations, so we ignore the error field
         ollama_msg["content"] = tool_call_results[0].result
         return ollama_msg
 
     if text_contents:
         ollama_msg["content"] = text_contents[0]
     if tool_calls:
+        # Ollama does not support tool call id, so we ignore it
         ollama_msg["tool_calls"] = [
-            {"type": "function", "function": {"name": tc.tool_name, "arguments": tc.arguments}}
+            {"type": "function", "function": {"name": tc.tool_name, "arguments": tc.arguments}} 
             for tc in tool_calls
         ]
     return ollama_msg
@@ -55,7 +56,7 @@ def _convert_ollama_meta_to_openai_format(input_response_dict: Dict) -> Dict:
     """
     Map Ollama metadata keys onto the OpenAI-compatible names Haystack expects.
     All fields that are not part of the OpenAI metadata are left unchanged in the returned dict.
-    
+
     Example Ollama metadata:
     {
         'model': 'phi4:14b-q4_K_M',
@@ -105,7 +106,7 @@ def _convert_ollama_meta_to_openai_format(input_response_dict: Dict) -> Dict:
 
 def _convert_ollama_response_to_chatmessage(ollama_response: "ChatResponse") -> ChatMessage:
     """
-    Convert **non-streaming** Ollama Chat API response to Haystack ChatMessage.
+    Convert non-streaming Ollama Chat API response to Haystack ChatMessage with assistant role..
     """
     response_dict = ollama_response.model_dump()
     ollama_message = response_dict["message"]
@@ -133,12 +134,12 @@ def _convert_ollama_response_to_chatmessage(ollama_response: "ChatResponse") -> 
 @component
 class OllamaChatGenerator:
     """
-    Haystack generator for models served by **Ollama** (https://ollama.ai).
+    Haystack generator for models served by Ollama (https://ollama.ai).
 
     * Fully supports streaming.
     * Correctly passes tool-calls to Haystack when `stream=True`.
 
-    Self-test / demo:
+    Usage example:
     ```python
     from haystack.dataclasses import ChatMessage
     from haystack.tools import Tool
@@ -182,27 +183,33 @@ class OllamaChatGenerator:
         :param model:
             The name of the model to use. The model must already be present (pulled) in the running Ollama instance.
         :param api_base_url:
-            The base URL of the Ollama server (default ``"http://localhost:11434"``).
+            The base URL of the Ollama server (default "http://localhost:11434").
         :param generation_kwargs:
-            Default inference options—e.g. ``temperature``, ``top_p``,
-            ``num_predict``—merged with any ``generation_kwargs`` passed to
-            :meth:`run`.
+            Optional arguments to pass to the Ollama generation endpoint, such as temperature,
+            top_p, and others. See the available arguments in
+            [Ollama docs](https://github.com/jmorganca/ollama/blob/main/docs/modelfile.md#valid-parameters-and-values).
         :param timeout:
             Socket timeout *in seconds* for HTTP calls to Ollama.
         :param keep_alive:
-            How long Ollama should keep the model in memory after the request.
-            Accepts seconds (int/float), a duration string (``"10m"``, ``"24h"``),
-            ``-1`` (never unload), or ``'0'`` (unload immediately).
+            The option that controls how long the model will stay loaded into memory following the request.
+            If not set, it will use the default value from the Ollama (5 minutes).
+            The value can be set to:
+            - a duration string (such as "10m" or "24h")
+            - a number in seconds (such as 3600)
+            - any negative number which will keep the model loaded in memory (e.g. -1 or "-1m")
+            - '0' which will unload the model immediately after generating a response.
         :param streaming_callback:
-            Callable invoked **once per :class:`StreamingChunk`** when streaming is
-            active.  Supply either here *or* in :meth:`run` to enable streaming.
+            A callback function that is called when a new token is received from the stream.
+            The callback function accepts StreamingChunk as an argument.
         :param tools:
-            A list of :class:`haystack.tools.Tool` **or** a
-            :class:`haystack.tools.Toolset`. Duplicate tool names raise a
-            ``ValueError``.
+            A list of `haystack.tools.Tool` or a `haystack.tools.Toolset`. Duplicate tool names raise a `ValueError`.
+            Not all models support tools. For a list of models compatible with tools, see the
+            [models page](https://ollama.com/search?c=tools).
         :param response_format:
-            ``None`` for free-form text, ``"json"`` to require valid JSON, or a
-            JSON-Schema dict (needs Ollama ≥ 0.1.34).
+            The format for structured model outputs. The value can be:
+            - None: No specific structure or format is applied to the response. The response is returned as-is.
+            - "json": The response is formatted as a JSON object.
+            - JSON Schema: The response is formatted as a JSON object that adheres to the specified JSON Schema. (needs Ollama ≥ 0.1.34)
         """
         _check_duplicate_tool_names(tools)
 
@@ -219,6 +226,14 @@ class OllamaChatGenerator:
 
     # Serialisation helpers
     def to_dict(self) -> Dict[str, Any]:
+        """
+        Serializes the component to a dictionary.
+
+        :returns:
+              Dictionary with serialized data.
+        """
+        callback_name = serialize_callable(self.streaming_callback) if self.streaming_callback else None
+
         return default_to_dict(
             self,
             model=self.model,
@@ -226,15 +241,22 @@ class OllamaChatGenerator:
             generation_kwargs=self.generation_kwargs,
             timeout=self.timeout,
             keep_alive=self.keep_alive,
-            streaming_callback=serialize_callable(self.streaming_callback)
-            if self.streaming_callback
-            else None,
+            streaming_callback=callback_name,
             tools=serialize_tools_or_toolset(self.tools),
             response_format=self.response_format,
         )
 
     @classmethod
     def from_dict(cls, data: Dict[str, Any]) -> "OllamaChatGenerator":
+        """
+        Deserializes the component from a dictionary.
+
+        :param data:
+            Dictionary to deserialize from.
+
+        :returns:
+            Deserialized component.
+        """
         deserialize_tools_or_toolset_inplace(data["init_parameters"], key="tools")
         if callback_ser := data["init_parameters"].get("streaming_callback"):
             data["init_parameters"]["streaming_callback"] = deserialize_callable(callback_ser)
@@ -263,7 +285,9 @@ class OllamaChatGenerator:
         response_iter: Any,
         streaming_callback: Optional[Callable[[StreamingChunk], None]],
     ) -> Dict[str, List[ChatMessage]]:
-
+        """
+        Handles streaming response and converts it to Haystack format
+        """
         chunks: List[StreamingChunk] = []
 
         # For every tool-call id store EITHER
@@ -326,6 +350,7 @@ class OllamaChatGenerator:
             tool_calls=completed_tool_calls or None,
         )
 
+        # Convert the last chunk's metadata to OpenAI format and attach it to the ChatMessage
         if chunks:
             reply._meta = _convert_ollama_meta_to_openai_format(chunks[-1].meta)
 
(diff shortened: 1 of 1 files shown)


integrations/ollama/src/haystack_integrations/components/generators/ollama/chat/chat_generator.py
```diff
@@ -103,74 +106,107 @@ def _convert_ollama_meta_to_openai_format(intput_response_dict: Dict) -> Dict:
...
-    message._meta = _convert_ollama_meta_to_openai_format(response_dict)
-    return message
+    chat_msg = ChatMessage.from_assistant(text=text, tool_calls=tool_calls or None)
+    chat_msg._meta = _convert_ollama_meta_to_openai_format(response_dict)
+    return chat_msg
 
 
+# --------------------------------------------------------------------------------------
+# Main component
+# --------------------------------------------------------------------------------------
```
sjrl: Could we remove these style comment blocks? For consistency in our code we prefer not to have decorative parts like the `# --------------------------------------------------------------------------------------`

integrations/ollama/src/haystack_integrations/components/generators/ollama/chat/chat_generator.py
```diff
@@ -103,74 +106,107 @@ def _convert_ollama_meta_to_openai_format(intput_response_dict: Dict) -> Dict:
...
-                                "num_predict": 100,
-                                "temperature": 0.9,
-                                })
-
-    messages = [ChatMessage.from_system("\nYou are a helpful, respectful and honest assistant"),
-    ChatMessage.from_user("What's Natural Language Processing?")]
-
-    print(generator.run(messages=messages))
+    from haystack.tools import Tool
+    def echo(query: str) -> str:
```
sjrl:
```suggestion
    from haystack.tools import Tool

    def echo(query: str) -> str:
```

integrations/ollama/src/haystack_integrations/components/generators/ollama/chat/chat_generator.py
Sheripov: Sure

integrations/ollama/src/haystack_integrations/components/generators/ollama/chat/chat_generator.py
```diff
@@ -103,74 +106,107 @@ def _convert_ollama_meta_to_openai_format(intput_response_dict: Dict) -> Dict:
...
+        print("Tool executed with QUERY: {🔍 {query}}")
+        return f"🔍 {query}"
+    echo_tool = Tool(
```
sjrl:
```suggestion
        return f"🔍 {query}"

    echo_tool = Tool(
```

integrations/ollama/src/haystack_integrations/components/generators/ollama/chat/chat_generator.py
```diff
@@ -103,74 +106,107 @@ def _convert_ollama_meta_to_openai_format(intput_response_dict: Dict) -> Dict:
...
+        function=echo,
+        parameters={"query": {"type": "string", "description": "Search query"}},
+    )
+    agent = Agent(
+        chat_generator=OllamaChatGenerator(model="mistral-small3.1:24b"),
+        tools=[echo_tool],
+        system_prompt="Use tool to print the query to test tools. Do not answer the question, just send the query to the tool",
+        max_agent_steps=5,
+        raise_on_tool_invocation_failure=True,
+        streaming_callback=lambda ch: print(ch.content, end="", flush=True),
```
sjrl: Could we just use `print_streaming_chunk` here?

integrations/ollama/src/haystack_integrations/components/generators/ollama/chat/chat_generator.py
```diff
@@ -103,74 +106,107 @@ def _convert_ollama_meta_to_openai_format(intput_response_dict: Dict) -> Dict:
...
-        self,
-        model: str = "orca-mini",
-        url: str = "http://localhost:11434",
-        generation_kwargs: Optional[Dict[str, Any]] = None,
-        timeout: int = 120,
-        keep_alive: Optional[Union[float, str]] = None,
-        streaming_callback: Optional[Callable[[StreamingChunk], None]] = None,
-        tools: Optional[Union[List[Tool], Toolset]] = None,
-        response_format: Optional[Union[None, Literal["json"], JsonSchemaValue]] = None,
+            self,
```
sjrl: Let's fix the tab formatting here. It should just be one tab. For an easy formatting fix you should be able to `hatch run lint:fmt`

integrations/ollama/src/haystack_integrations/components/generators/ollama/chat/chat_generator.py
```diff
@@ -103,74 +106,107 @@ def _convert_ollama_meta_to_openai_format(intput_response_dict: Dict) -> Dict:
...
+            model: str = "orca-mini",
+            url: str = "http://localhost:11434",
+            generation_kwargs: Optional[Dict[str, Any]] = None,
+            timeout: int = 120,
+            keep_alive: Optional[Union[float, str]] = None,
+            streaming_callback: Optional[Callable[[StreamingChunk], None]] = None,
+            tools: Optional[Union[List[Tool], Toolset]] = None,
+            response_format: Optional[Union[None, Literal["json"], JsonSchemaValue]] = None,
+            api_base_url: str | None = None,
```
sjrl: Thanks for looking to make things more consistent, but in favor of not making future breaking changes I'd say let's leave `url` and remove `api_base_url`

integrations/ollama/src/haystack_integrations/components/generators/ollama/chat/chat_generator.py
```diff
@@ -183,29 +219,29 @@ def __init__(
...
+        self.generation_kwargs = generation_kwargs or {}
+        self.timeout = timeout
         self.keep_alive = keep_alive
         self.streaming_callback = streaming_callback
         self.tools = tools
         self.response_format = response_format
+
         self._client = Client(host=self.url, timeout=self.timeout)
 
+    # Serialisation helpers
```
sjrl: I don't think this comment is necessary, could we remove it?

integrations/ollama/src/haystack_integrations/components/generators/ollama/chat/chat_generator.py
```diff
@@ -234,118 +270,168 @@ def from_dict(cls, data: Dict[str, Any]) -> "OllamaChatGenerator":
...
+    # Internal helpers
+    # ------------------------------------------------------------------
+    @staticmethod
+    def _build_chunk(chunk_response: Any) -> StreamingChunk:
         """
-        Converts the response from the Ollama API to a StreamingChunk.
+        Convert one Ollama stream-chunk to Haystack StreamingChunk.
         """
-        chunk_response_dict = chunk_response.model_dump()
+        d = chunk_response.model_dump()
```
sjrl: Following pep8 we don't want to use one letter variable names in this case. Could we go back to the previous name?

integrations/ollama/src/haystack_integrations/components/generators/ollama/chat/chat_generator.py
```diff
@@ -234,118 +270,168 @@ def from_dict(cls, data: Dict[str, Any]) -> "OllamaChatGenerator":
...
 
-        content = chunk_response_dict["message"]["content"]
-        meta = {key: value for key, value in chunk_response_dict.items() if key != "message"}
-        meta["role"] = chunk_response_dict["message"]["role"]
+        content = d["message"]["content"]
+        meta = {k: v for k, v in d.items() if k != "message"}
+        meta["role"] = d["message"]["role"]  # keep for debugging
```
sjrl:
```suggestion
        meta["role"] = d["message"]["role"]
```

integrations/ollama/src/haystack_integrations/components/generators/ollama/chat/chat_generator.py
```diff
@@ -234,118 +270,168 @@ def from_dict(cls, data: Dict[str, Any]) -> "OllamaChatGenerator":
...
-        generation_kwargs: Optional[Dict[str, Any]] = None,
-        tools: Optional[Union[List[Tool], Toolset]] = None,
-        *,
-        streaming_callback: Optional[Callable[[StreamingChunk], None]] = None,
+            self,
+            messages: List[ChatMessage],
+            generation_kwargs: Optional[Dict[str, Any]] = None,
+            tools: Optional[Union[List[Tool], Toolset]] = None,
+            *,
+            streaming_callback: Optional[Callable[[StreamingChunk], None]] = None,
```
sjrl: Fix formatting
```suggestion
        self,
        messages: List[ChatMessage],
        generation_kwargs: Optional[Dict[str, Any]] = None,
        tools: Optional[Union[List[Tool], Toolset]] = None,
        *,
        streaming_callback: Optional[Callable[[StreamingChunk], None]] = None,
```

integrations/ollama/src/haystack_integrations/components/generators/ollama/chat/chat_generator.py
```diff
@@ -234,118 +270,168 @@ def from_dict(cls, data: Dict[str, Any]) -> "OllamaChatGenerator":
...
     ):
         """
         Runs an Ollama Model on a given chat history.
 
         :param messages:
-            A list of ChatMessage instances representing the input messages.
+            Complete chat history, including the user’s latest message.
```
sjrl: I think the previous docstring makes more sense. At this point in time we don't know what messages are being given to the run method. And there is no gurantee the last message in the list is a user message.

integrations/ollama/src/haystack_integrations/components/generators/ollama/chat/chat_generator.py
```diff
@@ -234,118 +270,168 @@ def from_dict(cls, data: Dict[str, Any]) -> "OllamaChatGenerator":
...
         :param generation_kwargs:
-            Optional arguments to pass to the Ollama generation endpoint, such as temperature,
-            top_p, etc. See the
-            [Ollama docs](https://github.com/jmorganca/ollama/blob/main/docs/modelfile.md#valid-parameters-and-values).
+            Per-call overrides for Ollama inference options.  These are merged
+            on top of the instance-level `generation_kwargs`.
```
sjrl: I see the additional detail you'd like to add, but let's also keep the previous docstring that includes a link to the Ollama docs. So please combine the two together.

integrations/ollama/src/haystack_integrations/components/generators/ollama/chat/chat_generator.py
```diff
@@ -234,118 +270,168 @@ def from_dict(cls, data: Dict[str, Any]) -> "OllamaChatGenerator":
...
             during component initialization.
         :param streaming_callback:
-            A callback function that is called when a new token is received from the stream.
-        :returns: A dictionary with the following keys:
-            - `replies`: The responses from the model
+            A callable to receive `StreamingChunk` objects as they
+            arrive.  Supplying a callback (here or in the constructor) switches
+            the component into streaming mode.
+        :returns:
+            {"replies": List[ChatMessage]} – ready for Haystack pipelines.
```
sjrl:
```suggestion
        :returns: A dictionary with the following keys:
            - `replies`: A list of ChatMessages containing the model's response
```

integrations/ollama/src/haystack_integrations/components/generators/ollama/chat/chat_generator.py
```diff
@@ -234,118 +270,168 @@ def from_dict(cls, data: Dict[str, Any]) -> "OllamaChatGenerator":
...
+                if arg_delta is None:
+                    continue
+
+                prev = tool_call_args_by_id.get(tc_id)
+
+                if isinstance(arg_delta, str):
+                    if isinstance(prev, str):
+                        tool_call_args_by_id[tc_id] = prev + arg_delta
+                    else:  # prev is dict or None → start fresh with str
+                        tool_call_args_by_id[tc_id] = arg_delta
```
sjrl: Could you explain how this would happen? If previous is a dict how would we still get a new arg_delta for the same tool?

integrations/ollama/src/haystack_integrations/components/generators/ollama/chat/chat_generator.py
```diff
@@ -234,118 +270,168 @@ def from_dict(cls, data: Dict[str, Any]) -> "OllamaChatGenerator":
...
+
+        # ── merge text ────────────────────────────────────────────────
+        combined_text = "".join(c.content for c in chunks)
+
+        # ── rebuild completed ToolCall objects ────────────────────────
+        completed_tool_calls: List[ToolCall] = []
+        if last_seen_calls:
+            seen_ids = set()
+            # walk backwards so earlier duplicates are skipped
+            for tc in reversed(last_seen_calls):
```
sjrl: Also a bit confused here. How is it possible to get duplicates for tool calls? It should be possible to get one tool to be requested twice with different args. E.g. if you had a weather tool and the user asked "What is the weather in Paris and Berlin?" then LLM's like OpenAI would return two tool calls with the same tool name but different args.

integrations/ollama/src/haystack_integrations/components/generators/ollama/chat/chat_generator.py
```diff
@@ -234,118 +270,168 @@ def from_dict(cls, data: Dict[str, Any]) -> "OllamaChatGenerator":
...
+        for raw_chunk in response_iter:
+            chunk = self._build_chunk(raw_chunk)
+            chunks.append(chunk)
+
+            if streaming_callback is not None:
+                streaming_callback(chunk)
+
+            # ── collect tool-call deltas ──────────────────────────────
+            for tc in chunk.meta.get("tool_calls", []):
+                tc_id = tc.get("id") or tc["function"]["name"]
```
sjrl: Out of curiosity is not guaranteed that Ollama provides an `id` for a tool?

integrations/ollama/src/haystack_integrations/components/generators/ollama/chat/chat_generator.py
```diff
@@ -234,118 +270,168 @@ def from_dict(cls, data: Dict[str, Any]) -> "OllamaChatGenerator":
...
+        reply = ChatMessage.from_assistant(
+            text=combined_text,
+            tool_calls=completed_tool_calls or None,
+        )
 
         # Convert the last chunk's metadata to OpenAI format and attach it to the ChatMessage
         if chunks:
-            meta = _convert_ollama_meta_to_openai_format(chunks[-1].meta)
-            reply._meta = meta
+            reply._meta = _convert_ollama_meta_to_openai_format(chunks[-1].meta)
```
sjrl: I realized it was this way before but could we add this directly to the creation of `ChatMessage` so 
```python
        reply = ChatMessage.from_assistant(
            text=combined_text,
            tool_calls=completed_tool_calls or None,
            meta=_convert_ollama_meta_to_openai_format(chunks[-1].meta) if chunks else None
        )
```

sjrl:
@Sheripov thanks for you work on this! 

A few more comments on the review:
* It would be great if you could add tests for this. Tests in the format like the ones below from the Bedrock integration would be extremely helpful to see how the ollama streaming chunks look and how we convert them. For this integration testing the function `_handle_streaming_response` to see how ollama chunks are converted into a ChatMessage would be good to have.
  * https://github.com/deepset-ai/haystack-core-integrations/blob/main/integrations/amazon_bedrock/tests/test_chat_generator_utils.py#L337 
  * https://github.com/deepset-ai/haystack-core-integrations/blob/main/integrations/amazon_bedrock/tests/test_chat_generator_utils.py#L421
* Also adding an integration test to test live streaming with tool calls would be great! So like this [one](https://github.com/deepset-ai/haystack-core-integrations/blob/cfd52540ac277284eb7b3d3e735fe56eda3c33da/integrations/ollama/tests/test_chat_generator.py#L522) but enable streaming and tool calling

integrations/ollama/src/haystack_integrations/components/generators/ollama/chat/chat_generator.py
Sheripov: No — the Ollama server doesn’t guarantee an id field in every tool_calls entry.
- OpenAI-compatible endpoint (/v1/chat/completions) – recent releases do add an auto-generated id when the model produces multiple tool calls, so that clients can map results back.
- Native Ollama endpoint (/api/chat) and older builds – the JSON often contains only function.name + arguments; many users have reported that id is missing even with several calls, making client-side resolution harder.  ￼ ￼[issue-1](https://github.com/ollama/ollama/issues/6708?utm_source=chatgpt.com), [issue-2](https://github.com/ollama/ollama/issues/7510?utm_source=chatgpt.com)

integrations/ollama/src/haystack_integrations/components/generators/ollama/chat/chat_generator.py
Sheripov: I have added as a comment in the code:
```python
# Update the argument accumulator for this tool_call_id.

# • Ollama may stream the same `arguments` field in *two* different forms:
#   1) as one or more **str fragments**  -- characters of a JSON string, delivered chunk-by-chunk;
#   2) as a complete **dict**            -- fully-parsed JSON in a single chunk.

# • A dict always represents the *final* state (it is already parsed JSON),
#   so it should **overwrite** anything collected before.

# • If we are still receiving str fragments *and* we have not yet seen a
#   dict, we concatenate them in arrival order.

# • If a dict has already been stored (`prev` is dict) and another string
#   fragment arrives, we ignore it by skipping the concat - this prevents
#   `TypeError: can only concatenate str (not "dict") to str` and keeps the
#   fully-formed JSON intact.
```

integrations/ollama/src/haystack_integrations/components/generators/ollama/chat/chat_generator.py
Sheripov: Yes, you’re absolutely right — it is entirely valid for an LLM like OpenAI (and Ollama, if the model supports it) to return multiple tool calls to the same tool, e.g.:

`What is the weather in Paris and Berlin?`

…might result in two tool calls like:

```json
[
  {"id": "abc", "function": {"name": "get_weather", "arguments": "{\"city\": \"Paris\"}"}},
  {"id": "def", "function": {"name": "get_weather", "arguments": "{\"city\": \"Berlin\"}"}}
]
```
So why do we guard against “duplicates”?

Because Ollama doesn’t always provide an id, especially in native streaming mode. If it omits the id field:

```json
{"function": {"name": "get_weather", "arguments": "{\"city\": \"Paris\"}"}}
{"function": {"name": "get_weather", "arguments": "{\"city\": \"Berlin\"}"}}
```

…you’d have no way to tell if these are:
- two tool calls to the same tool (intended to be separate), or
- multiple chunks of the same tool call (streaming JSON fragment-by-fragment).

So, for robustness, we fall back to using function.name as a pseudo-ID, which can cause accidental overwriting if there are multiple intended tool calls to the same tool name without distinct IDs — that’s a limitation of the fallback.

Summary
- If id is provided → you can distinguish multiple calls to the same tool.
- If id is missing → fallback to function.name works only when there’s one call.
- That’s why the deduplication logic is cautious and assumes one logical call per name when id is absent.

integrations/ollama/src/haystack_integrations/components/generators/ollama/chat/chat_generator.py
Sheripov: done

integrations/ollama/src/haystack_integrations/components/generators/ollama/chat/chat_generator.py
Sheripov: done

integrations/ollama/src/haystack_integrations/components/generators/ollama/chat/chat_generator.py
Sheripov: done

integrations/ollama/src/haystack_integrations/components/generators/ollama/chat/chat_generator.py
Sheripov: done

integrations/ollama/src/haystack_integrations/components/generators/ollama/chat/chat_generator.py
Sheripov: done

integrations/ollama/src/haystack_integrations/components/generators/ollama/chat/chat_generator.py
Sheripov: done

integrations/ollama/src/haystack_integrations/components/generators/ollama/chat/chat_generator.py
Sheripov: done

integrations/ollama/src/haystack_integrations/components/generators/ollama/chat/chat_generator.py
Sheripov: done

integrations/ollama/src/haystack_integrations/components/generators/ollama/chat/chat_generator.py
Sheripov: done

integrations/ollama/src/haystack_integrations/components/generators/ollama/chat/chat_generator.py
Sheripov: done

integrations/ollama/src/haystack_integrations/components/generators/ollama/chat/chat_generator.py
Sheripov: done

integrations/ollama/src/haystack_integrations/components/generators/ollama/chat/chat_generator.py
Sheripov: done

integrations/ollama/src/haystack_integrations/components/generators/ollama/chat/chat_generator.py
Sheripov: done

integrations/ollama/src/haystack_integrations/components/generators/ollama/chat/chat_generator.py
Sheripov: done

integrations/ollama/src/haystack_integrations/components/generators/ollama/chat/chat_generator.py
sjrl: Thanks for the explanation! 

Could we add a dev comment here that briefly explains the above and perhaps links to the issues you mention [here](https://github.com/deepset-ai/haystack-core-integrations/pull/1906#discussion_r2131969979). This way it could be easier for a future dev to change/improve this if ollama resolves these issues.

integrations/ollama/src/haystack_integrations/components/generators/ollama/chat/chat_generator.py
sjrl: Great! Thanks for the detailed comment. Sounds like a lot of edge cases to work around!

integrations/ollama/src/haystack_integrations/components/generators/ollama/chat/chat_generator.py
```diff
@@ -143,12 +136,14 @@ class OllamaChatGenerator:
     ```python
     from haystack.components.generators.utils import print_streaming_chunk
     from haystack.components.agents import Agent
-    from components.OlamaChatGenerator import OllamaChatGenerator
+    from components.OllamaChatGenerator import OllamaChatGenerator
```
sjrl: I think this should be 
```suggestion
    from haystack_integrations.components.generators.ollama.chat import OllamaChatGenerator
```

Sheripov: done

Sheripov:
> Hey @Sheripov the failing test (link [here](https://github.com/deepset-ai/haystack-core-integrations/actions/runs/15489719324/job/43612224571?pr=1906#step:9:91)) will need to be removed since it specifically tested for a ValueError being raised when using tools + streaming.

I have changed those tests so, now they are marked as an integration tests.

Sheripov:
> @Sheripov thanks for you work on this!
> 
> A few more comments on the review:
> 
> * It would be great if you could add tests for this. Tests in the format like the ones below from the Bedrock integration would be extremely helpful to see how the ollama streaming chunks look and how we convert them. For this integration testing the function `_handle_streaming_response` to see how ollama chunks are converted into a ChatMessage would be good to have.
>   
>   * https://github.com/deepset-ai/haystack-core-integrations/blob/main/integrations/amazon_bedrock/tests/test_chat_generator_utils.py#L337
>   * https://github.com/deepset-ai/haystack-core-integrations/blob/main/integrations/amazon_bedrock/tests/test_chat_generator_utils.py#L421
> * Also adding an integration test to test live streaming with tool calls would be great! So like this [one](https://github.com/deepset-ai/haystack-core-integrations/blob/cfd52540ac277284eb7b3d3e735fe56eda3c33da/integrations/ollama/tests/test_chat_generator.py#L522) but enable streaming and tool calling

Done

integrations/ollama/src/haystack_integrations/components/generators/ollama/chat/chat_generator.py
Sheripov: done

integrations/ollama/src/haystack_integrations/components/generators/ollama/chat/chat_generator.py
```diff
@@ -297,55 +388,47 @@ def run(
...
-            msg = "Ollama does not support tools and streaming at the same time. Please choose one."
-            raise ValueError(msg)
-
-        if self.response_format and tools:
-            msg = "Ollama does not support tools and response_format at the same time. Please choose one."
-            raise ValueError(msg)
-
-        if self.response_format and stream:
-            msg = "Ollama does not support streaming and response_format at the same time. Please choose one."
-            raise ValueError(msg)
```
sjrl, line 329: Could you explain this change briefly. Does this work now?

Sheripov: Yes, this also works well now, I have added integration tests for it.

sjrl (approved):
Thanks!

PR #1912, opened 2025-06-09
fix: fix Bedrock types + add py.typed

### LINKED ISSUE #1771: Investigate better typing configuration for core integrations
Most of our core integrations have a `[[tool.mypy.overrides]]` section in pyproject, containing several libraries.

This is usually done to avoid mypy errors like
`error: Cannot find implementation or library stub for module named "..."  [import-not-found]`.

The error above appears because we are running mypy with the `--explicit-package-bases` flag. This was likely done to support namespace packages. More info in mypy docs: https://mypy.readthedocs.io/en/stable/running_mypy.html#mapping-paths-to-modules.

My impression is that using these workarounds does not guarantee type safety: we may treat the types coming from haystack and other libraries as `Any`.
Now that we are adding explicit support for types (https://github.com/deepset-ai/haystack/pull/9329), we should evaluate how to better configure typing in core integrations.

anakin87:
### Related Issues

- part of #1771

### Proposed Changes:
- run `hatch run test:types ` and fix type errors
  (this means testing with all dependencies installed, differently from `hatch run lint:typing` that did not install dependencies)
- introduce stricter configuration in pyproject.toml (and fix small errors)
  - as shown in #1910, these files should be put in submodules of the namespace packages
  - the namespace package is `haystack_integrations`
  - in this case, I put py.typed in 
    - `integrations/amazon_bedrock/src/haystack_integrations/common/py.typed`
    - `integrations/amazon_bedrock/src/haystack_integrations/components/embedders/py.typed`
    - `integrations/amazon_bedrock/src/haystack_integrations/components/generators/py.typed`
    - `integrations/amazon_bedrock/src/haystack_integrations/components/rankers/py.typed`
    - as a rule of thumb, py.typed files should be placed at the same level as the inner folder that corresponds to the integration name.

### How did you test it?
CI

### Notes for the reviewer
This PR can be seen as a complete example of the work we need to do for each integration to fix #1771.

### Checklist

### DIFF
--- a/integrations/amazon_bedrock/src/haystack_integrations/common/amazon_bedrock/utils.py
+++ b/integrations/amazon_bedrock/src/haystack_integrations/common/amazon_bedrock/utils.py
@@ -1,4 +1,4 @@
-from typing import Literal, Optional, Union, overload
+from typing import Any, Optional, Union
 
 import aioboto3
 import boto3
@@ -15,24 +15,6 @@ AWS_CONFIGURATION_KEYS = [
 ]
 
 
-@overload
-def get_aws_session(
-    aws_access_key_id: Optional[str] = None,
-    aws_secret_access_key: Optional[str] = None,
-    aws_session_token: Optional[str] = None,
-    aws_region_name: Optional[str] = None,
-    aws_profile_name: Optional[str] = None,
-    async_mode: Literal[True] = True,
-) -> aioboto3.Session: ...
-@overload
-def get_aws_session(
-    aws_access_key_id: Optional[str] = None,
-    aws_secret_access_key: Optional[str] = None,
-    aws_session_token: Optional[str] = None,
-    aws_region_name: Optional[str] = None,
-    aws_profile_name: Optional[str] = None,
-    async_mode: Literal[False] = False,
-) -> boto3.Session: ...
 def get_aws_session(
     aws_access_key_id: Optional[str] = None,
     aws_secret_access_key: Optional[str] = None,
@@ -40,7 +22,7 @@ def get_aws_session(
     aws_region_name: Optional[str] = None,
     aws_profile_name: Optional[str] = None,
     async_mode: bool = False,
-    **kwargs,
+    **kwargs: Any,
 ) -> Union[boto3.Session, aioboto3.Session]:
     """
     Creates an AWS Session with the given parameters.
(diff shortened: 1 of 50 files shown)


integrations/amazon_bedrock/src/haystack_integrations/common/amazon_bedrock/utils.py
```diff
@@ -22,7 +22,7 @@ def get_aws_session(
     aws_region_name: Optional[str] = None,
     aws_profile_name: Optional[str] = None,
     async_mode: bool = False,
-    **kwargs,
+    **kwargs: Any,
```
sjrl, line 25: I think we can make this a bit more strict with `Dict[str, Any]`. Or did you run into problems with that?

integrations/amazon_bedrock/src/haystack_integrations/components/generators/amazon_bedrock/adapters.py
```diff
@@ -44,7 +45,9 @@ def get_responses(self, response_body: Dict[str, Any]) -> List[str]:
         responses = [completion.lstrip() for completion in completions]
         return responses
 
-    def get_stream_responses(self, stream, streaming_callback: Callable[[StreamingChunk], None]) -> List[str]:
+    def get_stream_responses(
+        self, stream: EventStream, streaming_callback: Callable[[StreamingChunk], None]
```
sjrl: Should we use the `StreamingCallbackT` here?

sjrl (approved):
Two small comments otherwise looks good!

integrations/amazon_bedrock/src/haystack_integrations/common/amazon_bedrock/utils.py
anakin87: I have the impression that the two formulations are equivalent
https://peps.python.org/pep-0484/#arbitrary-argument-lists-and-default-argument-values
https://mypy.readthedocs.io/en/stable/cheat_sheet_py3.html#functions

integrations/amazon_bedrock/src/haystack_integrations/components/generators/amazon_bedrock/adapters.py
anakin87: I'm now using `SyncStreamingCallbackT`

integrations/amazon_bedrock/src/haystack_integrations/common/amazon_bedrock/utils.py
anakin87: Yes, I tried to do `reveal_type(kwargs)` and I get
`src/haystack_integrations/common/amazon_bedrock/utils.py:42: note: Revealed type is "builtins.dict[builtins.str, Any]"`
(as expected)

PR #1917, opened 2025-06-10
fix!: Use model_dump_json for tool results instead of returning mcp-sdk types.CallToolResult

vblagoje:
### Why:
Fixed incorrect serialization of `MCPTool` responses. Previously, when tools returned TextContent or ImageContent objects, they were being converted to strings using `str()` which produced raw Python object representations instead of proper JSON. This made the responses difficult to parse and use, especially for LLMs and users expecting structured JSON data.

*Note*: this is a breaking change as we are switching from raw mcp-sdk CallToolResult types being returned from tool calls to JSON strings

### What:
- Proper JSON serialization of TextContent and ImageContent using `model_dump_json()`
- Consistent JSON string responses for all tool calls
- Fixed string representation of error messages

### How can it be used:

No changes in usage patterns:

```python
from haystack.tools import MCPTool, SSEServerInfo
import json

tool = MCPTool(
    name="image_tool",
    server_info=SSEServerInfo(url="http://localhost:8000/sse")
)

# Now returns proper JSON string instead of raw string representation
result_json = tool.invoke(param="value")
result = json.loads(result_json)
```

### How did you test it:
- **Unit tests**: Verified proper JSON serialization in tool response tests
- **Integration tests**: Confirmed correct handling of TextContent serialization
- **Feature tests**: Tested error cases ensure proper JSON formatting

### Notes for the reviewer:
- **Compatibility**: breaking change as now we return json object string representation instead of mcp-sdk objects

### DIFF
--- a/integrations/mcp/src/haystack_integrations/tools/mcp/mcp_tool.py
+++ b/integrations/mcp/src/haystack_integrations/tools/mcp/mcp_tool.py
@@ -4,6 +4,7 @@
 
 import asyncio
 import concurrent.futures
+import json
 import threading
 import warnings
 from abc import ABC, abstractmethod
@@ -233,16 +234,16 @@ class MCPClient(ABC):
         """
         pass
 
-    async def call_tool(self, tool_name: str, tool_args: dict[str, Any]) -> Any:
+    async def call_tool(self, tool_name: str, tool_args: dict[str, Any]) -> str:
         """
         Call a tool on the connected MCP server.
 
         :param tool_name: Name of the tool to call
         :param tool_args: Arguments to pass to the tool
-        :returns: Result of the tool invocation
+        :returns: JSON string representation of the tool invocation result
         :raises MCPConnectionError: If not connected to an MCP server
         :raises MCPInvocationError: If the tool invocation fails
-        :raises MCPResponseTypeError: If response type is not TextContent
+        :raises MCPResponseTypeError: If response type is not TextContent or ImageContent
         """
         if not self.session:
             message = "Not connected to an MCP server"
@@ -250,8 +251,10 @@ class MCPClient(ABC):
 
         try:
             result = await self.session.call_tool(tool_name, tool_args)
-            validated_result = self._validate_response(tool_name, result)
-            return validated_result
+            if result.isError:
+                message = f"Tool '{tool_name}' returned an error: {result.content!s}"
+                raise MCPInvocationError(message=message, tool_name=tool_name)
+            return result.model_dump_json()
         except MCPError:
             # Re-raise specific MCP errors directly
             raise
@@ -259,44 +262,6 @@ class MCPClient(ABC):
             # Wrap other exceptions with context about which tool failed
             message = f"Failed to invoke tool '{tool_name}' with args: {tool_args} , got error: {e!s}"
             raise MCPInvocationError(message, tool_name, tool_args) from e
-
-    def _validate_response(self, tool_name: str, result: types.CallToolResult) -> types.CallToolResult:
-        """
-        Validate response from an MCP tool call, accepting only TextContent.
-
-        :param tool_name: Name of the called tool (for error messages)
-        :param result: CallToolResult from MCP tool call
-        :returns: The original CallToolResult object
-        :raises MCPResponseTypeError: If content type is not TextContent
-        :raises MCPInvocationError: If the tool call resulted in an error
-        """
-
-        # Check for error response
-        if result.isError:
-            if len(result.content) > 0 and isinstance(result.content[0], types.TextContent):
-                # Get the error message from the first item
-                first_item = result.content[0]
-                message = f"Tool '{tool_name}' returned an error: {first_item.text}"
-            else:
-                message = f"Tool '{tool_name}' returned an error: {result.content!s}"
-            raise MCPInvocationError(
-                message=message,
-                tool_name=tool_name,
-            )
-
-        # Validate content types - only allow TextContent for now
-        if result.content:
-            for item in result.content:
-                if not isinstance(item, types.TextContent):
-                    # Reject any non-TextContent
-                    message = (
-                        f"Unsupported content type in response from tool '{tool_name}'. "
-                        f"Only TextContent is currently supported."
-                    )
-                    raise MCPResponseTypeError(message, result, tool_name)
-
-        # Return the original result object
-        return result
 
     async def aclose(self) -> None:
         """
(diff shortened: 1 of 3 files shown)


integrations/mcp/src/haystack_integrations/tools/mcp/mcp_tool.py
```diff
@@ -235,25 +235,27 @@ async def connect(self) -> list[Tool]:
...
         Call a tool on the connected MCP server.
 
         :param tool_name: Name of the tool to call
         :param tool_args: Arguments to pass to the tool
-        :returns: Result of the tool invocation
+        :returns: JSON string representation of the tool invocation result
         :raises MCPConnectionError: If not connected to an MCP server
         :raises MCPInvocationError: If the tool invocation fails
-        :raises MCPResponseTypeError: If response type is not TextContent
+        :raises MCPResponseTypeError: If response type is not TextContent or ImageContent
```
Amnah199: I couldn't find a check to raise this error if the response type is a different type. Maybe we should add a check.

integrations/mcp/tests/test_mcp_integration.py
```diff
@@ -86,21 +86,29 @@ def subtract(a: int, b: int) -> int:
             tool = MCPTool(name="add", server_info=server_info)
 
             # Invoke the tool
-            result = tool.invoke(a=5, b=3)
+            result_json = tool.invoke(a=5, b=3)
+
+            # Parse the JSON result
+            import json
```
Amnah199: I would move this with other imports.

integrations/mcp/tests/test_mcp_tool.py
```diff
@@ -66,15 +66,15 @@ def test_mcp_tool_invoke(self, mcp_add_tool, mcp_echo_tool):
         """Test invoking MCPTools connected to in-memory servers."""
         # Test add tool invocation
         add_result = mcp_add_tool.invoke(a=25, b=17)
-        assert not add_result.isError
-        assert len(add_result.content) == 1
-        assert add_result.content[0].text == "42"  # 25 + 17
+        import json
```
Amnah199: same for this, lets move it to the top

integrations/mcp/src/haystack_integrations/tools/mcp/mcp_tool.py
```diff
@@ -235,25 +235,27 @@ async def connect(self) -> list[Tool]:
...
             raise MCPConnectionError(message=message, operation="call_tool")
 
         try:
             result = await self.session.call_tool(tool_name, tool_args)
-            validated_result = self._validate_response(tool_name, result)
-            return validated_result
+            if result.isError:
+                message = f"Tool '{tool_name}' returned an error: {result.content!s}"
+                raise MCPInvocationError(message=message, tool_name=tool_name)
+            return result.model_dump_json()
```
Amnah199, line 241: Any serialization issues here will also be raised as `MCPInvocationError`. If its not ideal, maybe we can handle this separately.

integrations/mcp/tests/test_mcp_toolset.py
```diff
@@ -308,6 +315,7 @@ def subtract(a: int, b: int) -> int:
     @pytest.mark.skipif(sys.platform == "win32", reason="Windows fails for some reason")
     def test_toolset_with_streamable_http_connection(self):
         """Test MCPToolset with a streamable-http connection to a simple server."""
+        import json
         import socket
         import subprocess
         import tempfile
```
Amnah199: these can be removed

integrations/mcp/tests/test_mcp_toolset.py
```diff
@@ -210,6 +210,7 @@ class TestMCPToolsetIntegration:
     @pytest.mark.skipif(sys.platform == "win32", reason="Windows fails for some reason")
     def test_toolset_with_sse_connection(self):
         """Test MCPToolset with an SSE connection to a simple server."""
+        import json
         import socket
         import subprocess
         import tempfile
```
Amnah199: Lets move these to the top.

integrations/mcp/src/haystack_integrations/tools/mcp/mcp_tool.py
vblagoje: Yes, good catch 🙏 
Removed this pydoc

integrations/mcp/src/haystack_integrations/tools/mcp/mcp_tool.py
vblagoje: I think it is ok for now, we can improve later if really needed

integrations/mcp/src/haystack_integrations/tools/mcp/mcp_tool.py
```diff
@@ -848,12 +817,12 @@ async def invoke():
...
-    async def ainvoke(self, **kwargs: Any) -> Any:
+    async def ainvoke(self, **kwargs: Any) -> str:
         """
         Asynchronous tool invocation.
 
         :param kwargs: Arguments to pass to the tool
-        :returns: Result of the tool invocation, processed based on content type
+        :returns: JSON string representation of the tool invocation result
         :raises MCPInvocationError: If the tool invocation fails
         :raises MCPResponseTypeError: If response type is not supported
```
Amnah199: @vblagoje Do we not need to check anymore if the response type is not supported? In that case, we can remove `MCPResponseTypeError` because I dont see where it will be raised.

vblagoje: Right, but I removed it. Is that the latest version? One sec

Amnah199: I guess we can the remove it completely from the file.

vblagoje: Ok, forgot about that one. Fixed now 🙏

integrations/mcp/src/haystack_integrations/tools/mcp/mcp_tool.py
```diff
@@ -692,11 +655,14 @@ class MCPTool(Tool):
     compatibility with the Haystack tool ecosystem.
 
     Response handling:
-    - Text content is supported and returned as strings
-    - Unsupported content types (like binary/images) will raise MCPResponseTypeError
+    - Text and image content are supported and returned as JSON strings
+    - The JSON contains the structured response from the MCP server
+    - Use json.loads() to parse the response into a dictionary
+    - Unsupported content types will raise MCPResponseTypeError
```
Amnah199: we got one here as well.

Amnah199 (approved):
LG! Perhaps we also need to update the documentation about the JSON outputs.
https://docs.haystack.deepset.ai/docs/mcptool

vblagoje:
Additions sent to @dfokina for review. 🙏 all.

PR #1948, opened 2025-06-14
feat: Add "think" parameter for Ollama

### LINKED ISSUE #1881: Add Ollamas Thinking capabilities
When using Models with Thinking capabilities e.g. Deepseek Haystack currently does Not Support Ollamas latest Thinking capabilities (e.g. disabling Thinking or Splitting the Output of Thinking) 

Additional Tag when submitting a prompt to Ollama to disable/enable Thinking and extaction of the thought process after the query ist completed.

**Additional context**
https://ollama.com/blog/thinking

Ryzhtus:
### Related Issues

- fixes #1881 

### Proposed Changes:

Added an additional `think` parameter in accordance with Ollama’s current interface, and included a test to cover this feature.

### How did you test it?

I added unit tests

### Checklist


Ryzhtus:
@anakin87 Hi! What's your opinion about storing `thinking` field in the `_meta` attribute of `ChatMessage`. In `OllamaChatGenerator` the response is converted to a `ChatMessage` object, where its `_meta` value is formatted to be compatible with OpenAI API. The thing is OpenAI's ChatCompletion API doesn't support storing reasoning messages, so there's no appropriate field to store it and so this breaks compatibility in some way. But having it probably will be useful for users who want to track reasoning process of their LLMs

integrations/ollama/src/haystack_integrations/components/generators/ollama/chat/chat_generator.py
```diff
@@ -156,6 +161,7 @@ def __init__(
         url: str = "http://localhost:11434",
         generation_kwargs: Optional[Dict[str, Any]] = None,
         timeout: int = 120,
+        think=False,
```
anakin87: I would put this new parameter at the end, to make this change non-breaking.

integrations/ollama/src/haystack_integrations/components/generators/ollama/chat/chat_generator.py
```diff
@@ -172,6 +178,8 @@ def __init__(
             [Ollama docs](https://github.com/jmorganca/ollama/blob/main/docs/modelfile.md#valid-parameters-and-values).
         :param timeout:
             The number of seconds before throwing a timeout error from the Ollama API.
+        :param think
+            Enables the model's "thinking" process.
```
anakin87: I would expand this explanation to something like
```suggestion
            If True, the modell will "think" before producing a response. 
            Only [thinking models](https://ollama.com/search?c=thinking) support this feature.
            The intermediate "thinking" output can be found in the `meta` property of the returned `ChatMessage`.
```

integrations/ollama/src/haystack_integrations/components/generators/ollama/generator.py
```diff
@@ -36,6 +36,7 @@ def __init__(
         template: Optional[str] = None,
         raw: bool = False,
         timeout: int = 120,
+        think: bool = False,
```
anakin87: We are trying to introduce new features in the Chat Generators only. In the long run, we may deprecate Generators and keep only Chat Generators.

For this reason, I won't introduce support for thinking in Generators.

integrations/ollama/tests/test_chat_generator.py
```diff
@@ -508,6 +508,17 @@ def test_run_with_chat_history(self):
             city.lower() in response["replies"][-1].text.lower() for city in ["Manchester", "Birmingham", "Glasgow"]
         )
 
+    @pytest.mark.integration
+    def test_live_run_with_thinking(self):
+        chat_generator = OllamaChatGenerator(model="qwen3:1.7b", think=True)
```
anakin87: to use this model in an integration test, you should also change the following line
https://github.com/deepset-ai/haystack-core-integrations/blob/18e19f58598299a52af3369e32ec4aa8b371e6ce/.github/workflows/ollama.yml#L25

However, I would recommend using `qwen3:0.6b` if possible: based on my experiment, this would work quite well with our tests and being very small, it can speed up download and inference times.

Ryzhtus: Ok! Thank you for telling me about this, let's use qwen3:0.6b as it performs well

Ryzhtus:
> Thanks for working on this...
> 
> * I agree with your approach. For the moment, putting the thinking output in `ChatMessage._meta.thinking` is reasonable.
> * Please rebase your branch, fix conflicts and run tests.
> * I left some other comments .

Hey @anakin87 I’ve made fixes and also run the tests. I have the following question: since you suggested modifying llama3.2:3b to qwen3:0.6b to make the integration test work, do we also need to modify other mentions of llama3.2:3b in the tests? Do we have to open a separate PR for it?

integrations/ollama/src/haystack_integrations/components/generators/ollama/chat/chat_generator.py
```diff
@@ -186,6 +192,7 @@ def __init__(
         streaming_callback: Optional[Callable[[StreamingChunk], None]] = None,
         tools: Optional[Union[List[Tool], Toolset]] = None,
         response_format: Optional[Union[None, Literal["json"], JsonSchemaValue]] = None,
+        think=False,
```
anakin87:
```suggestion
        think: bool=False,
```

integrations/ollama/tests/test_chat_generator.py
```diff
@@ -565,6 +565,17 @@ def test_run_with_chat_history(self):
...
+    def test_live_run_with_thinking(self):
+        chat_generator = OllamaChatGenerator(model="qwen3:0.6b", think=True)
+
+        message = ChatMessage.from_user("How many times does the letter 'r' appear in the word 'strawberry'?")
+        response = chat_generator.run([message])
+
+        assert isinstance(response, dict)
+        assert isinstance(response["replies"], list)
+        assert "thinking" in response["replies"][0]._meta
+
```
anakin87, line 654: let's also add a simple unit test (`test_run_with_thinking`), similar to `test_run`

integrations/ollama/src/haystack_integrations/components/generators/ollama/chat/chat_generator.py
```diff
@@ -197,7 +204,11 @@ def __init__(
             top_p, and others. See the available arguments in
             [Ollama docs](https://github.com/jmorganca/ollama/blob/main/docs/modelfile.md#valid-parameters-and-values).
         :param timeout:
-            Socket timeout *in seconds* for HTTP calls to Ollama.
+            The number of seconds before throwing a timeout error from the Ollama API.
+        :param think
+            If True, the modell will "think" before producing a response.
+            Only [thinking models](https://ollama.com/search?c=thinking) support this feature.
+            The intermediate "thinking" output can be found in the `meta` property of the returned `ChatMessage`.
```
anakin87, line 211: currently, linting is failing because this parameter is unused.

I would set `self.think` in `__init__` and then use its value in `run`

anakin87:
> ... I have the following question: since you suggested modifying llama3.2:3b to qwen3:0.6b to make the integration test work, do we also need to modify other mentions of llama3.2:3b in the tests? Do we have to open a separate PR for it?

Yes, let's always use qwen3:0.6b.
Doing that in the current PR is fine.

anakin87 (approved):
Looks good!

Since the model we use is very small, I relaxed the constraint for some tests that sometimes failed in my local machine.

PR #1968, opened 2025-06-17
feat: add support for GitHubRepoForkerTool

### LINKED ISSUE #1869: Add GitHubRepoForkerTool to GitHub integration
In the newly added GitHub integration we added the following components

- GitHubIssueViewer
- GitHubIssueCommenter
- GitHubRepoViewer
- GitHubRepoForker
- GitHubFileEditor
- GitHubPRCreator

We have a Tool for each of them except for GitHubRepoForker. We could add a GitHubRepoForkerTool based on the GitHubRepoForker component.

srini047:
### Related Issues

- fixes #1869 

### Proposed Changes:

 
 
Added support for `GitHubRepoForkerTool` to the GitHub intergrations.

### How did you test it?

Added a test file to test the tool.

### Checklist


integrations/github/src/haystack_integrations/components/connectors/github/repo_forker.py
```diff
@@ -61,7 +61,7 @@ def __init__(
         :param create_branch: If True, creates a fix branch based on the issue number
         """
         error_message = "github_token must be a Secret"
-        if not isinstance(github_token, Secret):
+        if github_token is not None and not isinstance(github_token, Secret):
```
mpangrazzi: I would revert this change: this logic allows `None` value for `github_token` to pass validation (skips this condition basically). And this would be problematic later, since all API calls requires authentication.

So I would revert to `if not isinstance(github_token, Secret):`. WDYT?

integrations/github/src/haystack_integrations/tools/github/repo_forker_tool.py
```diff
@@ -0,0 +1,127 @@
...
+
+    def __init__(
+        self,
+        *,
+        name: Optional[str] = "repo_forker",
+        description: Optional[str] = REPO_FORKER_PROMPT,
+        parameters: Optional[Dict[str, Any]] = REPO_FORKER_SCHEMA,
+        github_token: Optional[Secret] = None,
+        repo: Optional[str] = None,
+        branch: str = "main",
```
mpangrazzi: Why are you passing `repo` and `branch` here? They doesn't seem to be used (nor required in `GitHubRepoForker`).

mpangrazzi (commented):
Looks good, I've left some comments!

integrations/github/src/haystack_integrations/components/connectors/github/repo_forker.py
srini047: @mpangrazzi I'm using `and` hence both the sides must be evaluated to `true` for entering the statement. One side I'm checking if `is not None` hence we won't end up in that condition.

integrations/github/src/haystack_integrations/tools/github/repo_forker_tool.py
srini047: @mpangrazzi Yup will remove it.

integrations/github/src/haystack_integrations/components/connectors/github/repo_forker.py
mpangrazzi: @srini047 If `github_token` is `None`, we should expect the method to return a `TypeError` early.

Your condition is: `if github_token is not None and not isinstance(github_token, Secret):`

If `github_token` is `None`, the first part (`github_token is not None`) evaluates to `False`. Due to short-circuit evaluation in the `and` operator, the entire condition becomes `False`. Therefore, the `TypeError` is not raised (you can easily replicate this with a test) but since any Github operation requires authentication, calling the `forker.run` would always return a 401 and a wrong answer from the method (since it expects a valid value for `github_token`).

Instead, if the condition is `if not isinstance(github_token, Secret):` and `github_token` is `None`, the method immediately returns a `TypeError: github_token must be a Secret` (which is the desired behaviour).

So I recommend reverting to `if not isinstance(github_token, Secret):`. WDYT?

srini047: @mpangrazzi Makes sense now. I didn't think this forward. Let me do the change.

srini047:
@mpangrazzi Can you please review it.

integrations/github/src/haystack_integrations/components/connectors/github/repo_forker.py
mpangrazzi: It seems you still have to revert this change

mpangrazzi (commented):
@srini047 there's a change still need to be reverted. Apart from this, everything else looks ok to me.

srini047:
@mpangrazzi Can you help me unblock in this PR. Many cases are failing for GithubRepoForker. I wasn't to debug the issue.

integrations/github/src/haystack_integrations/components/connectors/github/repo_forker.py
```diff
@@ -41,7 +41,7 @@ class GitHubRepoForker:
     def __init__(
         self,
         *,
-        github_token: Secret = Secret.from_env_var("GITHUB_TOKEN"),
+        github_token: Optional[Secret] = None,
```
sjrl: I believe the change here needs to be reverted to fix the failing tests
```suggestion
        github_token: Secret = Secret.from_env_var("GITHUB_TOKEN"),
```

integrations/github/src/haystack_integrations/tools/github/repo_forker_tool.py
```diff
@@ -0,0 +1,116 @@
...
+    A tool for forking Github repository.
+    """
+
+    def __init__(
+        self,
+        *,
+        name: Optional[str] = "repo_forker",
+        description: Optional[str] = REPO_FORKER_PROMPT,
+        parameters: Optional[Dict[str, Any]] = REPO_FORKER_SCHEMA,
+        github_token: Optional[Secret] = None,
```
sjrl: Typically to fork a repo you need a GH account so it makes sense to require authentication
```suggestion
        github_token: Secret = Secret.from_env_var("GITHUB_TOKEN"),
```

PR #1970, opened 2025-06-18
fix Nvidia types

anakin87:
### Related Issues

- fixes #issue-number

### Checklist

### DIFF
--- a/integrations/nvidia/src/haystack_integrations/components/embedders/nvidia/document_embedder.py
+++ b/integrations/nvidia/src/haystack_integrations/components/embedders/nvidia/document_embedder.py
@@ -11,7 +11,7 @@ from haystack.utils import Secret, deserialize_secrets_inplace
 from tqdm import tqdm
 
 from haystack_integrations.components.embedders.nvidia.truncate import EmbeddingTruncateMode
-from haystack_integrations.utils.nvidia import DEFAULT_API_URL, Model, NimBackend, url_validation
+from haystack_integrations.utils.nvidia import DEFAULT_API_URL, Client, Model, NimBackend, url_validation
 
 logger = logging.getLogger(__name__)
 
@@ -145,7 +145,7 @@ class NvidiaDocumentEmbedder:
             api_url=self.api_url,
             api_key=self.api_key,
             model_kwargs=model_kwargs,
-            client=self.__class__.__name__,
+            client=Client.NVIDIA_DOCUMENT_EMBEDDER,
             timeout=self.timeout,
         )
         if not self.model and self.backend.model:
--- a/integrations/nvidia/src/haystack_integrations/components/embedders/nvidia/document_embedder.py
+++ b/integrations/nvidia/src/haystack_integrations/components/embedders/nvidia/document_embedder.py
@@ -122,7 +122,9 @@ class NvidiaDocumentEmbedder:
                 UserWarning,
                 stacklevel=2,
             )
-            self.model = self.backend.model = name
+            self.model = name
+            if self.backend:
+                self.backend.model = name
         else:
             error_message = "No locally hosted model was found."
             raise ValueError(error_message)
@@ -232,7 +234,7 @@ class NvidiaDocumentEmbedder:
         return all_embeddings, {"usage": {"prompt_tokens": usage_prompt_tokens, "total_tokens": usage_total_tokens}}
 
     @component.output_types(documents=List[Document], meta=Dict[str, Any])
-    def run(self, documents: List[Document]):
+    def run(self, documents: List[Document]) -> Dict[str, Union[List[Document], Dict[str, Any]]]:
         """
         Embed a list of Documents.
 
--- a/integrations/nvidia/src/haystack_integrations/components/rankers/nvidia/ranker.py
+++ b/integrations/nvidia/src/haystack_integrations/components/rankers/nvidia/ranker.py
@@ -10,7 +10,7 @@ from haystack import Document, component, default_from_dict, default_to_dict, lo
 from haystack.utils import Secret, deserialize_secrets_inplace
 
 from haystack_integrations.components.rankers.nvidia.truncate import RankerTruncateMode
-from haystack_integrations.utils.nvidia import DEFAULT_API_URL, NimBackend, is_hosted, url_validation
+from haystack_integrations.utils.nvidia import DEFAULT_API_URL, Client, NimBackend, is_hosted, url_validation
 
 logger = logging.getLogger(__name__)
 
@@ -172,7 +172,7 @@ class NvidiaRanker:
                 api_key=self.api_key,
                 model_kwargs=model_kwargs,
                 timeout=self.timeout,
-                client=self.__class__.__name__,
+                client=Client.NVIDIA_RANKER,
             )
             if not self.is_hosted and not self.model:
                 if self.backend.model:
--- a/integrations/nvidia/src/haystack_integrations/components/rankers/nvidia/ranker.py
+++ b/integrations/nvidia/src/haystack_integrations/components/rankers/nvidia/ranker.py
@@ -162,7 +162,7 @@ class NvidiaRanker:
         :raises ValueError: If the API key is required for hosted NVIDIA NIMs.
         """
         if not self._initialized:
-            model_kwargs = {}
+            model_kwargs: Dict[str, Any] = {}
             if self.truncate is not None:
                 model_kwargs.update(truncate=str(self.truncate))
             self.backend = NimBackend(
@@ -174,7 +174,7 @@ class NvidiaRanker:
                 timeout=self.timeout,
                 client=self.__class__.__name__,
             )
-            if not self.is_hosted and not self._model:
+            if not self.is_hosted and not self.model:
                 if self.backend.model:
                     self.model = self.backend.model
             self._initialized = True
--- a/integrations/nvidia/src/haystack_integrations/utils/nvidia/client.py
+++ b/integrations/nvidia/src/haystack_integrations/utils/nvidia/client.py
@@ -1,0 +1,15 @@
+from enum import Enum
+
+
+class Client(Enum):
+    """
+    Client to use for NVIDIA NIMs.
+    """
+
+    NVIDIA_GENERATOR = "nvidia_generator"
+    NVIDIA_TEXT_EMBEDDER = "nvidia_text_embedder"
+    NVIDIA_DOCUMENT_EMBEDDER = "nvidia_document_embedder"
+    NVIDIA_RANKER = "nvidia_ranker"
+
+    def __str__(self):
+        return self.value
--- a/integrations/nvidia/src/haystack_integrations/utils/nvidia/models.py
+++ b/integrations/nvidia/src/haystack_integrations/utils/nvidia/models.py
@@ -1,6 +1,8 @@
 from dataclasses import dataclass
 from typing import Literal, Optional
 
+from .client import Client
+
 
 @dataclass
 class Model:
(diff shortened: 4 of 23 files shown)


integrations/nvidia/src/haystack_integrations/components/embedders/nvidia/document_embedder.py
```diff
@@ -143,7 +145,7 @@ def warm_up(self):
             api_url=self.api_url,
             api_key=self.api_key,
             model_kwargs=model_kwargs,
-            client=self.__class__.__name__,
+            client=Client.NVIDIA_DOCUMENT_EMBEDDER,
```
julian-risch, line 148: This change looks okay to me. Let's just not forget that this could result in a change for users who created a component that inherits from NvidiaDocumentEmbedder. Now, `client` will be `Client.NVIDIA_DOCUMENT_EMBEDDER` for them too. Previously it was the name of their custom component.

integrations/nvidia/src/haystack_integrations/components/rankers/nvidia/ranker.py
```diff
@@ -172,9 +172,9 @@ def warm_up(self):
                 api_key=self.api_key,
                 model_kwargs=model_kwargs,
                 timeout=self.timeout,
-                client=self.__class__.__name__,
+                client=Client.NVIDIA_RANKER,
             )
-            if not self.is_hosted and not self._model:
+            if not self.is_hosted and not self.model:
```
julian-risch, line 177: Good bug fix. 👍 I wonder why it didn't cause any trouble so far.

integrations/nvidia/src/haystack_integrations/utils/nvidia/nim_backend.py
```diff
@@ -21,14 +22,12 @@
 class NimBackend:
     def __init__(
         self,
-        model: str,
         api_url: str,
-        model_type: Optional[Literal["chat", "embedding", "ranking"]] = None,
+        model_type: Literal["chat", "embedding", "ranking"],
```
julian-risch: `model_type` was optional before. Do we really need to make it mandatory now?

julian-risch: My understanding is that previously, `validate_hosted_model` tried to infer `model_type` automatically if the user didn't specify it

integrations/nvidia/tests/conftest.py
```diff
@@ -12,9 +12,17 @@
 
 
 class MockBackend(NimBackend):
-    def __init__(self, model: str, api_key: Optional[Secret] = None, model_kwargs: Optional[Dict[str, Any]] = None):
+    def __init__(
+        self,
+        model: str,
+        model_type: str,
```
julian-risch: I'm in favor of keeping `model_type` optional and not passing it as a parameter value in the tests. That way, the document embedder and text embedder tests would remain completely unchanged for example.

integrations/nvidia/src/haystack_integrations/utils/nvidia/client.py
```diff
@@ -0,0 +1,15 @@
...
+    """
+    Client to use for NVIDIA NIMs.
+    """
+
+    NVIDIA_GENERATOR = "nvidia_generator"
+    NVIDIA_TEXT_EMBEDDER = "nvidia_text_embedder"
+    NVIDIA_DOCUMENT_EMBEDDER = "nvidia_document_embedder"
+    NVIDIA_RANKER = "nvidia_ranker"
+
+    def __str__(self):
```
julian-risch:
```suggestion
    def __str__(self) -> str:
    """
    Convert a Client enum to a string.
    """
```

integrations/nvidia/src/haystack_integrations/utils/nvidia/client.py
```diff
@@ -0,0 +1,15 @@
...
+        return self.value
```
julian-risch, line 16: We could add 
```suggestion
        return self.value
        
    @staticmethod
    def from_str(string: str) -> "Client":
        """
        Convert a string to a Client enum.
        """
        enum_map = {e.value: e for e in Client}
        mode = enum_map.get(string)
        if mode is None:
            msg = f"Unknown client '{string}' to use for NVIDIA NIMs. Supported modes are: {list(enum_map.keys())}"
            raise ValueError(msg)
        return mode
```

julian-risch: It's similar to what we do with PyPDFExtractionMode in Haystack.

integrations/nvidia/src/haystack_integrations/utils/nvidia/nim_backend.py
```diff
@@ -21,14 +22,12 @@
...
+        model: Optional[str] = None,
         api_key: Optional[Secret] = Secret.from_env_var("NVIDIA_API_KEY"),
         model_kwargs: Optional[Dict[str, Any]] = None,
-        client: Optional[
-            Literal["NvidiaGenerator", "NvidiaTextEmbedder", "NvidiaDocumentEmbedder", "NvidiaRanker"]
-        ] = None,
+        client: Optional[Client] = None,
```
julian-risch: For backwards compatibility, we could do the following and then check the type in the init. Again, I took inspiration from PyPDFExtractionMode in Haystack.
```suggestion
        client: Optional[Union[str, Client]] = None,
```

integrations/nvidia/src/haystack_integrations/utils/nvidia/nim_backend.py
```diff
@@ -43,23 +42,24 @@ def __init__(
         self.session.headers.update(headers)
 
         self.api_url = api_url
+        validated_model: Optional[Model] = None
```
julian-risch, line 47:
```suggestion
        if isinstance(client, Client):
            client = Client.from_str(client)
        validated_model: Optional[Model] = None
```

integrations/nvidia/src/haystack_integrations/utils/nvidia/utils.py
```diff
@@ -84,25 +85,24 @@ def determine_model(name: str) -> Optional[Model]:
 
 def validate_hosted_model(
     model_name: str,
-    client: Optional[Literal["NvidiaGenerator", "NvidiaTextEmbedder", "NvidiaDocumentEmbedder", "NvidiaRanker"]] = None,
-) -> Any:
+    client: Optional[Client] = None,
```
julian-risch, line 88: I suggest we use `Optional[Union[str, Client]]` here too and then in the method check `isinstance` and use `Client.from_str`

integrations/nvidia/tests/test_nim_backend.py
```diff
@@ -74,12 +74,17 @@ def mock_rank_post_response(*args, **kwargs):  # noqa: ARG001
 class TestNimBackend:
     def test_init_default(self, monkeypatch):
         monkeypatch.setenv("NVIDIA_API_KEY", "fake-api-key")
-        backend = NimBackend(model="nvidia/nv-embedqa-e5-v5", api_url=DEFAULT_API_URL, client="NvidiaTextEmbedder")
+        backend = NimBackend(
+            model="nvidia/nv-embedqa-e5-v5",
+            api_url=DEFAULT_API_URL,
+            client=Client.NVIDIA_TEXT_EMBEDDER,
```
julian-risch: If we ensure backwards compatibility as I suggested, we won't need to change existing tests. We could add additional tests.

integrations/nvidia/src/haystack_integrations/utils/nvidia/models.py
```diff
@@ -19,7 +21,7 @@ class Model:
 
     id: str
     model_type: Optional[Literal["chat", "embedding", "ranking"]] = None
-    client: Optional[Literal["NvidiaGenerator", "NvidiaTextEmbedder", "NvidiaDocumentEmbedder", "NvidiaRanker"]] = None
+    client: Optional[Client] = None
```
julian-risch: I suggest we keep this backwards compatible.

integrations/nvidia/src/haystack_integrations/utils/nvidia/nim_backend.py
anakin87: I initially discarded this idea because I think that `NimBackend` is used only internally, so I did not care.

But I can do this.

integrations/nvidia/src/haystack_integrations/utils/nvidia/nim_backend.py
anakin87: I would say that `model_type` should be mandatory.

In the codebase, `NimBackend` is always initialized passing this argument.

Let's try to analyze the existing implementation.
The following lines are triggered only if `model_type` is not null:
https://github.com/deepset-ai/haystack-core-integrations/blob/6fe3cf738584e186848c5cafca47f01a2e976ccf/integrations/nvidia/src/haystack_integrations/utils/nvidia/nim_backend.py#L53-L54

If it's null, it's also impossible to validate the model: `validate_hosted_model` expects a string
https://github.com/deepset-ai/haystack-core-integrations/blob/6fe3cf738584e186848c5cafca47f01a2e976ccf/integrations/nvidia/src/haystack_integrations/utils/nvidia/nim_backend.py#L56

With the existing implementation, I get the following mypy errors
```
src/haystack_integrations/utils/nvidia/nim_backend.py:58: error: Incompatible types in assignment (expression has type "Model", variable has type "str | None")  [assignment]
src/haystack_integrations/utils/nvidia/nim_backend.py:58: error: Argument 1 to "validate_hosted_model" has incompatible type "str | None"; expected "str"  [arg-type]
src/haystack_integrations/utils/nvidia/nim_backend.py:67: error: Cannot determine type of "model_type"  [has-type]
src/haystack_integrations/utils/nvidia/nim_backend.py:164: error: Cannot determine type of "model_type"  [has-type]
```

For these reasons, I am inclined to think that the existing implementation can be improved.

integrations/nvidia/src/haystack_integrations/utils/nvidia/utils.py
anakin87: Since this method is only called in `NimBackend.__init__`, where we already perform the conversion, I would say that it's not needed to use Union.

integrations/nvidia/src/haystack_integrations/utils/nvidia/models.py
anakin87: done

integrations/nvidia/src/haystack_integrations/utils/nvidia/nim_backend.py
julian-risch: I completely agree with you that this part of the code is quite chaotic and the implementation can for sure be improved. Here is how I'd summarize the old behavior.
If `is_hosted` is False then `model_type` is mandatory.
If `is_hosted` is True, meaning an `api_url` was specified by the user, there are two cases to distinguish:
* If `model` is not provided then `model_type` is mandatory. It is used to infer a default model of that type.
* If `model` is provided then `model_type` is **optional** and any value provided by the user will not be used because we lookup a `Model` with its pre-defined `model_type` via `validate_hosted_model`, `determine_model`, `lookup_model`. If the lookup fails, `validate_hosted_model` will raise an error.

julian-risch: In short, if a user specifies `model` and `api_url` to use a hosted model then `model_type` does not need to be specified and any user-defined value will be ignored. If it helps, we can also talk about that part of the code in a call. Everything else in the PR looks good to me.

anakin87: Ok I reverted some of my changes and now `model_type` is optional.
The internal implementation of `__init__` is changed a bit, to make mypy happy and the code (hopefully) better.

integrations/nvidia/tests/conftest.py
anakin87: done

integrations/nvidia/tests/test_nim_backend.py
anakin87: I added several additional tests for `__init__`. Some of them are AI-generated but manually curated 😄

julian-risch (approved):
LGTM! 👍

PR #1980, opened 2025-06-20
fix: rermove None key

Ra0R:
### Related Issues

- fixes #issue-number

### Proposed Changes:

Access in "Document.to_dict" expects "meta" to not be present or a dict (not None).

### How did you test it?

Manual test.


Ra0R:
I am using PgvectorKeywordRetriever in this setup:
(Some settings don't make much sense at the moment, since I am not using the embedding retriever.)
`
document_store = PgvectorDocumentStore(
    table_name=JOB_DOCUMENT_TABLE,
    embedding_dimension=768,
    vector_function="cosine_similarity",
    recreate_table=False,
    search_strategy="hnsw",
    connection_string=Secret.from_token(settings.PG_CONN_STR),
    hnsw_recreate_index_if_exists=True
)
retriever = PgvectorKeywordRetriever(document_store=document_store, top_k=10)
job_documents = retriever.run(query="Sicherheit", top_k=10)  # Adjust top_k as needed
print(f"Retrieved {job_documents} job documents.")
`


`Document.from_dict` then fails `return cls(**data, meta={**meta, **flatten_meta})` with:  'NoneType' object is not a mapping".

Meta is "NULL" in PgSQL.

Issue #1976, opened 2025-06-20
Tracing Context Corruption When (Nested Pipeline) Component Fails During Span Cleanup


lumi-xhs:
[Note: this whole thing is generated by cursor. I think it is accurate, but I apologize if there are some minor issues somewhere]

[Edit: This issue involves nested pipeline component failures]

## Summary

When a Haystack component contains its own internal `AsyncPipeline` (nested pipeline) and that internal pipeline fails during execution, the tracing context can become corrupted. This specifically occurs when a component like `IntentClassifier` creates and runs its own sub-pipeline within the context of a larger parent pipeline. If the nested pipeline fails and exceptions occur during span cleanup, Langfuse tracing context becomes permanently corrupted.

## Environment

- **haystack-integrations[langfuse]**: 1.1.2, but later versions may be also affected
- **haystack**: 2.13.x
- **langfuse**: 2.x
- **Python**: 3.10

## Root Cause Analysis

The issue occurs with this specific architecture:

```
Main AsyncPipeline (Parent Trace Context)
└── IntentClassifier Component
    └── Internal Pipeline (Nested Trace Context)
        ├── ChatPromptBuilder
        ├── ChatGenerator  
        └── IntentParser ❌ (Fails here)
```

**Failure Sequence:**
1. Main pipeline starts execution with parent trace context
2. `IntentClassifier` component is invoked as part of main pipeline
3. `IntentClassifier.run()` creates and executes its own internal `Pipeline`
4. Internal pipeline's `IntentParser` fails (e.g., JSON parsing error from LLM response)
5. Exception propagates to the nested pipeline's `LangfuseTracer.trace()` context manager
6. During span cleanup, if `span_handler.handle()` or `raw_span.end()` fail with the exception data
7. `self._context.pop()` is never executed in the tracer
8. The failed span remains stuck in the tracer's context stack
9. **All subsequent runs of the main pipeline use the stuck nested span as parent**

<img width="370" alt="Image" src="https://github.com/user-attachments/assets/99a2a269-96c0-4d66-9512-3ba6a0b1dc7a" />

The key insight is that this isn't just a component failure - it's specifically about **nested pipeline execution** corrupting the Langfuse tracing context, especially the parent pipeline trace.

## Reproduction Steps

Here's a minimal reproduction that demonstrates the nested pipeline issue:

```python
from haystack import component, Pipeline, AsyncPipeline
from haystack_integrations.tracing.langfuse import LangfuseConnector
[...]

PR #1999, opened 2025-06-26
fix: Properly cleanup Langfuse tracing context after pipeline run failures

vblagoje:
## Why:
Fixes a critical context corruption bug where failing nested operations would leave stale spans in the tracing context, causing subsequent operations to have incorrect parent-child relationships and corrupted trace hierarchies.

- fixes https://github.com/deepset-ai/haystack-core-integrations/issues/1976

## What:
- Modified `LangfuseTracer.trace()` method with nested try/finally blocks to guarantee context cleanup
- Added context corruption detection with error logging when stack inconsistencies occur
- Implemented graceful error handling for span processing failures without propagating exceptions
- Created comprehensive test `test_context_cleanup_after_nested_failures` using real Haystack components

## How can it be used:
These are internal improvements to the tracing infrastructure. No client-facing API changes.

## How did you test it:
- Created test that reproduces the exact issue user provided
- Validated context length equals 0 after both failed and successful operations
- Tested with real LangfuseConnector and Pipeline.run() to match production conditions

## Notes for the reviewer:
Please pay attention to the nested try/finally structure in `LangfuseTracer.trace()` - the outer `finally` guarantees context cleanup while the inner try/except handles span processing failures gracefully. The context corruption detection logic is critical for debugging stack inconsistencies.

### DIFF
--- a/integrations/langfuse/tests/test_tracer.py
+++ b/integrations/langfuse/tests/test_tracer.py
@@ -13,8 +13,7 @@ import pytest
 from haystack import Pipeline, component
 from haystack.dataclasses import ChatMessage, ToolCall
 
-from haystack_integrations.components.connectors.langfuse import \
-LangfuseConnector
+from haystack_integrations.components.connectors.langfuse import LangfuseConnector
 from haystack_integrations.tracing.langfuse.tracer import (
     _COMPONENT_OUTPUT_KEY, DefaultSpanHandler, LangfuseSpan, LangfuseTracer,
     SpanContext)


vblagoje:
@mpangrazzi I know you are not as familiar with this code as @sjrl - so I asked him to have a look as well.

integrations/langfuse/src/haystack_integrations/tracing/langfuse/tracer.py
```diff
@@ -449,22 +449,31 @@ def trace(
...
+            except Exception as cleanup_error:
+                # Log cleanup errors but don't let them corrupt context
+                logger.warning(f"Error during span cleanup for {operation_name}: {cleanup_error}")
+            finally:
+                # CRITICAL: Always pop context to prevent corruption
+                # This is especially important for nested pipeline scenarios
+                if self._context and self._context[-1] == span:
+                    self._context.pop()
+                else:
+                    logger.error(f"Context corruption detected: expected {span} at top of stack")
```
sjrl: Only question I have is if these if checks are needed/warranted. Do you have a test where we can see how this logger.error could be raised?

integrations/langfuse/tests/test_tracer.py
```diff
@@ -397,3 +404,61 @@ def test_init_with_tracing_disabled(self, monkeypatch, caplog):
...
+            pass  # Expected to fail
+
+        # Critical assertion: context should be empty after failed operation
+        assert len(tracer.tracer._context) == 0
+
+        # Test 2: Second run should work normally with clean context
+        try:
+            result = main_pipeline.run({"nested_component": {"input_data": '{"key": "valid"}'}})
+        except Exception:
+            assert False, "Second run should not fail"
```
sjrl: If this is expected to run we can get rid of the try except block. If the run fails the test will fail with an exception

integrations/langfuse/src/haystack_integrations/tracing/langfuse/tracer.py
```diff
@@ -449,22 +449,31 @@ def trace(
...
+                # Process span data (may fail with nested pipeline exceptions)
+                self._span_handler.handle(span, component_type)
+
+                # End span (may fail if span data is corrupted)
+                raw_span = span.raw_span()
+                if isinstance(raw_span, (StatefulSpanClient, StatefulGenerationClient)):
+                    raw_span.end()
+            except Exception as cleanup_error:
+                # Log cleanup errors but don't let them corrupt context
+                logger.warning(f"Error during span cleanup for {operation_name}: {cleanup_error}")
```
sjrl: Please use the correct logger formatting for Haystack
```suggestion
                logger.warning("Error during span cleanup for {operation_name}: {cleanup_error}", operation_name=operation_name, cleanup_error=cleanup_error)
```

sjrl: Also could we add a test checking for this warning message? It would be helpful to understand how an exception could occur at this point.

integrations/langfuse/src/haystack_integrations/tracing/langfuse/tracer.py
vblagoje: I removed the error

sjrl (approved):
Looks good to me!

integrations/langfuse/tests/test_tracer.py
```diff
@@ -367,7 +373,8 @@ def test_update_span_flush_disable(self, monkeypatch):
         monkeypatch.setenv("HAYSTACK_LANGFUSE_ENFORCE_FLUSH", "false")
         tracer_mock = Mock()
 
-        from haystack_integrations.tracing.langfuse.tracer import LangfuseTracer
+        from haystack_integrations.tracing.langfuse.tracer import \
+            LangfuseTracer
```
sjrl, line 376: Small nit, why add the line break?

vblagoje: Hm, must have been black or ruff, let me see

Issue #1992, opened 2025-06-25
AnthropicChatGenerator does not get correct prompt_tokens when using streaming.


mathislucka:
**Describe the bug**
Anthropic returns streaming events where the first chunk has the `input_tokens` and the last chunk has the `completion_tokens`.

However, in this line we try to get usage only from the last chunk: https://github.com/deepset-ai/haystack-core-integrations/blob/d804dd78a314917c2bd9b960fc46ace9e36d292b/integrations/anthropic/src/haystack_integrations/components/generators/anthropic/chat/chat_generator.py#L416

We need to change the conversion so that we get input tokens from the first chunk and output tokens from the last chunk.

**To Reproduce**
Use the AnthropicChatGenerator with streaming and look at `ChatMessage.meta["usage"]`.

**Describe your environment (please complete the following information):**
 - Haystack version:
 - Integration version:

PR #2001, opened 2025-06-26
fix: Anthropic reports input tokens in first message delta

mathislucka:
### Related Issues

- fixes [#issue-number
](https://github.com/deepset-ai/haystack-core-integrations/issues/1992)
### Proposed Changes:

 Get the input tokens from the first message delta.

### Checklist

### DIFF
--- a/integrations/anthropic/src/haystack_integrations/components/generators/anthropic/chat/chat_generator.py
+++ b/integrations/anthropic/src/haystack_integrations/components/generators/anthropic/chat/chat_generator.py
@@ -414,7 +414,29 @@ class AnthropicChatGenerator:
 
         # Update meta information
         last_chunk_meta = chunks[-1].meta
-        usage = self._get_openai_compatible_usage(last_chunk_meta)
+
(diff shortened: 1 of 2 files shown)


integrations/anthropic/src/haystack_integrations/components/generators/anthropic/chat/chat_generator.py
```diff
@@ -414,7 +414,29 @@ def _convert_streaming_chunks_to_chat_message(
 
         # Update meta information
         last_chunk_meta = chunks[-1].meta
-        usage = self._get_openai_compatible_usage(last_chunk_meta)
+
```
vblagoje, line 417: Looks good but let's just create first_chunk_meta = chunks[0].meta so we can use it multiple times below. While we guarantee that chunks is not None perhaps we should add a check if chunks conditional expression here.

vblagoje (commented):
Looks good, small optional comment to lower cognitive load even further

vblagoje (approved):
Looks good!

PR #2011, opened 2025-06-27
fix: Astra - fix types + add py.typed

anakin87:
### Related Issues

- part of #1771

### Proposed Changes:
- run `hatch run test:types`
  (this means testing with all dependencies installed, differently from `hatch run lint:typing` that did not install dependencies) and fix type errors

### How did you test it?
CI

### Checklist

### DIFF
--- a/integrations/astra/pyproject.toml
+++ b/integrations/astra/pyproject.toml
@@ -64,17 +64,14 @@ unit = 'pytest -m "not integration" {args:tests}'
 integration = 'pytest -m "integration" {args:tests}'
 all = 'pytest {args:tests}'
 cov-retry = 'all --cov=haystack_integrations --reruns 3 --reruns-delay 30 -x'
-types = "mypy --install-types --non-interactive --explicit-package-bases {args:src/ tests}"
+types = """mypy -p haystack_integrations.document_stores.astra \
+-p haystack_integrations.components.retrievers.astra {args}"""
 
-# TODO: remove lint environment once this integration is properly typed
-# test environment should be used instead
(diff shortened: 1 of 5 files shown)


integrations/astra/pyproject.toml
```diff
@@ -156,16 +153,4 @@ exclude_lines = ["no cov", "if __name__ == .__main__.:", "if TYPE_CHECKING:"]
...
-[[tool.mypy.overrides]]
-module = [
-  "astra_client.*",
-  "astrapy.*",
-  "pydantic.*",
-  "haystack.*",
-  "haystack_integrations.*",
-  "pytest.*",
-  "openpyxl.*",
-]
```
Amnah199, line 170: For my own knowledge, can you explain why we are removing these?

anakin87: Yes, this is discussed in #1771 in a long form.

In short, this was a wrong workaround to avoid mypy errors.

Errors stemmed from the fact that the old `lint` environment was set as detached: it did not install the project dependencies.
This led to right mypy `[import-not-found]` errors, that we wrongly silenced with `ignore_missing_imports`: the effect is treating types from these libraries as `Any`, which is practically equivalent to not doing type checking.

So now, I am using an environment with these dependencies (`test`), finding a lot of mypy issues and trying to fix them.

(`ignore_missing_imports` should only be used for libraries that do not provide type hints.)

Amnah199: Ahh amazing!! Thanks

Amnah199 (approved):
Thanks for the fixes!

PR #2018, opened 2025-06-30
fix: fix Optimum types + add py.typed

anakin87:
### Related Issues

- part of #1771

### Proposed Changes:
- run `hatch run test:types`
  (this means testing with all dependencies installed, differently from `hatch run lint:typing` that did not install dependencies) and fix type errors

### How did you test it?
CI

### Checklist

### DIFF
--- a/integrations/optimum/src/haystack_integrations/components/embedders/optimum/_backend.py
+++ b/integrations/optimum/src/haystack_integrations/components/embedders/optimum/_backend.py
@@ -2,7 +2,7 @@ import copy
 import json
 from dataclasses import dataclass
 from pathlib import Path
-from typing import Any, Dict, List, Optional, Tuple, Union
+from typing import Any, Dict, List, Optional, Tuple, Union, overload
 
 import numpy as np
 import torch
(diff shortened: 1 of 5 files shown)


integrations/optimum/src/haystack_integrations/components/embedders/optimum/_backend.py
```diff
@@ -188,6 +188,12 @@ def pool_embeddings(self, model_output: torch.Tensor, attention_mask: torch.Tens
         pooled_outputs = self.pooling_layer.forward(features)
         return pooled_outputs["sentence_embedding"]
 
+    @overload
```
davidsbatista, line 191: interesting, did not know about this `@overload` decorator.

I notice that we are using the decorator here, but below we are still keeping the typing with the Union - do need to keep both, or we can remove the one below?

davidsbatista (approved):
LGTM - just a small doubt comment about duplicate typing information

integrations/optimum/src/haystack_integrations/components/embedders/optimum/_backend.py
anakin87: > In regular modules, a series of @overload-decorated definitions must be followed by exactly one non-@overload-decorated definition (for the same function/method). The @overload-decorated definitions are for the benefit of the type checker only, since they will be overwritten by the non-@overload-decorated definition, while the latter is used at runtime but should be ignored by a type checker.

from https://typing.python.org/en/latest/spec/overload.html

davidsbatista: ok

PR #2038, opened 2025-06-30
fix!: Weave - move Connector, fix types, add py.typed

anakin87:
### Related Issues
- part of  #1771

Unfortunately, I found some additional issues in this integration, that I am trying to fix in some way.

### Proposed Changes:
- for consistency with other integrations and for namespace packages to work, I moved
src/haystack_integrations/components/connectors/weave_connector.py -> src/haystack_integrations/components/connectors/weave/weave_connector.py
*After releasing this change (which is breaking), I will also update docs and integration page*

- I had to ignore a violation of the Liskov Substitution Principle. The alternative would be rewriting the whole Tracer. (Langfuse might have similar problems, but I am not sure).

- Usual type fixes:
  - run `hatch run test:types` and fix type errors

### How did you test it?
CI

### Notes for the reviewer
The only positive aspect is that currently this integration has low usage, so breaking compatibility would not affect many users.

### Checklist

### DIFF
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -174,10 +174,10 @@ integration:weaviate:
       - any-glob-to-any-file: "integrations/weaviate/**/*"
       - any-glob-to-any-file: ".github/workflows/weaviate.yml"
 
-integration:weight_bias:
+integration:weights_and_biases_weave:
   - changed-files:
-      - any-glob-to-any-file: "integrations/weight_bias/**/*"
-      - any-glob-to-any-file: ".github/workflows/weight_bias.yml"
+      - any-glob-to-any-file: "integrations/weights_and_biases_weave/**/*"
+      - any-glob-to-any-file: ".github/workflows/weights_and_biases_weave.yml"
 
 # Topics
 topic:CI:


integrations/weights_and_biases_weave/src/haystack_integrations/tracing/weave/tracer.py
```diff
@@ -151,7 +151,12 @@ def _finish_regular_call(self, span: WeaveSpan) -> None:
 
     @contextlib.contextmanager
     def trace(
-        self, operation_name: str, tags: Optional[dict[str, Any]] = None, parent_span: Optional[WeaveSpan] = None
+        self,
+        operation_name: str,
+        tags: Optional[dict[str, Any]] = None,
+        # the current implementation violates the Liskov Substitution Principle by using WeaveSpan instead of Span
+        # unfortunately, it seems hard to fix without rewriting the Tracer
```
Amnah199, line 158: Is there a future action item for this. If this is meant for tracking then lets create an issue and remove this from here?

anakin87: I prefer to leave the comment here: it explains why we are doing `type:ignore[override]`.
In any case, I can also create an issue, if you prefer.

Amnah199: Right! Sure create an issue, if there is an intention to fix it. Otherwise we can just leave it.

anakin87: https://github.com/deepset-ai/haystack-core-integrations/issues/2045

PR #2051, opened 2025-07-02
Fix/prompt caching support

### LINKED ISSUE #1942: Update needed on Anthropic prompt caching support
While working on #1940, I found that the support for [prompt caching in Anthropic](https://docs.anthropic.com/en/docs/build-with-claude/prompt-caching) is outdated and does not work as expected.

- When introduced, this feature was in beta and our implementation is based on this aspect (e.g. if specific beta headers are not found, cache support is disabled)
- The [prompt caching example](https://github.com/deepset-ai/haystack-core-integrations/blob/main/integrations/anthropic/example/prompt_caching.py) does not work.
- Our implementation seems to apply to system messages only, and our successful tests focus on this aspect.

The best solution would be to update our Chat Generator; this would require some effort. Alternatively, rather than keeping a broken or misleading feature, it would be better to (temporarily) remove it altogether.

gulbaki:
### Related Issues

- fixes #1942 

### Proposed Changes:

Anthropic beta cache usage is no longer behind a beta flag and can now be used without specifying a beta header.
Additionally, users can now include multiple beta features using a comma-separated format, e.g., `beta1,beta2,` as described in [[Anthropic’s documentation](https://docs.anthropic.com/en/api/beta-headers#multiple-beta-features)](https://docs.anthropic.com/en/api/beta-headers#multiple-beta-features).

However, if this approach is not suitable for us, we can revert to the previous behavior where only a single beta feature was supported.

Here’s how I approached it: with each newly introduced beta feature, we can add a new conditional block to make it easier for users to adopt them.
But as mentioned, if this is not desirable, I can also simplify the logic to enforce single-beta usage again.

Additionally, caching is now set to a default of 5 minutes, but it has been updated to allow usage of a 1-hour TTL as a beta feature.
### How did you test it?

I added tests that cover the work I’ve done, but some of the existing test cases have either become invalid or are now breaking the updated behavior. That’s why, instead of directly removing or editing them, I wanted to consult with you first and proceed accordingly.

* `test_prompt_caching_enabled`: The content of this test can remain unchanged, but its name could potentially be updated to something more global like `test_beta_enabled`.
* `test_prompt_caching_cache_control_without_extra_headers`: I had to make changes here because caching can now function even without the beta header. As a result, the original purpose of the test has been compromised.
* `test_run_with_prompt_caching`: The function name could also be revised. The content remains valid since we are essentially testing whether the beta feature is applied or not.
* `test_to_dict_with_prompt_caching`: Similarly, the name could be updated.
* `test_from_dict_with_prompt_caching`: Same here — renaming might make it more accurate.

### Notes for the reviewer
I made sure not to delete or modify any existing work without consulting you first. That’s why I’m open to updating the test code according to your preferences.

Additionally, regarding the beta headers accepting multiple values like beta,beta2, I can revert and adjust this behavior as well based on your feedback.

I haven't made any changes to the example usages for now; I will update them based on the decision we make.

### Checklist

x I have read the [contributors guidelines](https://github.com/deepset-ai/haystack-core-integrations/blob/main/CONTRIBUTING.md) and the [code of conduct](https://github.com/deepset-ai/haystack-core-integrations/blob/main/CODE_OF_CONDUCT.md)
x I have updated the related issue with new insights and changes

### DIFF
--- a/.github/workflows/CI_license_compliance.yml
+++ b/.github/workflows/CI_license_compliance.yml
@@ -13,7 +13,7 @@ on:
 env:
   CORE_DATADOG_API_KEY: ${{ secrets.CORE_DATADOG_API_KEY }}
   PYTHON_VERSION: "3.10"
-  EXCLUDE_PACKAGES: "(?i)^(deepeval|fastembed|ollama|ragas|tqdm|psycopg|typing_extensions).*"
+  EXCLUDE_PACKAGES: "(?i)^(deepeval|fastembed|ollama|ragas|tqdm|psycopg|typing_extensions|adbc_driver_snowflake).*"
 
   # Exclusions must be explicitly motivated
   #
@@ -23,6 +23,7 @@ env:
   # - ragas is Apache 2.0 but the license is not available on PyPI
   # - typing_extensions>=4.13.0 has a Python Software Foundation License 2.0 but pip-license-checker does not recognize it
   #   (https://github.com/pilosus/pip-license-checker/issues/143)
+  # - adbc_driver_snowflake is Apache 2.0 but the license is not correctly specified on PyPI
 
   # - tqdm is MLP but there are no better alternatives
   # - psycopg is LGPL-3.0 but FOSSA is fine with it
--- a/integrations/deepeval/CHANGELOG.md
+++ b/integrations/deepeval/CHANGELOG.md
@@ -1,5 +1,12 @@
 # Changelog
 
+## [integrations/deepeval-v0.3.0] - 2025-07-08
+
+### 🚀 Features
+
+- Update deepeval integration to be unpinned (#2068)
+
+
 ## [integrations/deepeval-v0.2.0] - 2025-06-30
 
 ### 🐛 Bug Fixes
--- a/integrations/deepeval/pyproject.toml
+++ b/integrations/deepeval/pyproject.toml
@@ -23,7 +23,7 @@ classifiers = [
   "Programming Language :: Python :: Implementation :: CPython",
   "Programming Language :: Python :: Implementation :: PyPy",
 ]
-dependencies = ["haystack-ai", "deepeval==0.20.57"]
+dependencies = ["haystack-ai", "deepeval"]
 
 [project.urls]
 Source = "https://github.com/deepset-ai/haystack-core-integrations/tree/main/integrations/deepeval"
--- a/integrations/deepeval/src/haystack_integrations/components/evaluators/deepeval/evaluator.py
+++ b/integrations/deepeval/src/haystack_integrations/components/evaluators/deepeval/evaluator.py
@@ -3,7 +3,8 @@ from typing import Any, Callable, Dict, List, Optional, Union
 
 from haystack import DeserializationError, component, default_from_dict, default_to_dict
 
-from deepeval.evaluate import TestResult, evaluate
+from deepeval.evaluate import evaluate
+from deepeval.evaluate.types import EvaluationResult
 from deepeval.metrics import BaseMetric
 from deepeval.test_case import LLMTestCase
 
@@ -46,7 +47,7 @@ class DeepEvalEvaluator:
 
     _backend_metric: BaseMetric
     # Wrapped for easy mocking.
-    _backend_callable: Callable[[List[LLMTestCase], BaseMetric], List[TestResult]]
+    _backend_callable: Callable[[List[LLMTestCase], BaseMetric], EvaluationResult]
 
     def __init__(
         self,
@@ -93,7 +94,9 @@ class DeepEvalEvaluator:
         converted_inputs: List[LLMTestCase] = list(self.descriptor.input_converter(**inputs))  # type: ignore
 
         results = self._backend_callable(converted_inputs, self._backend_metric)
-        converted_results = [[result.to_dict() for result in self.descriptor.output_converter(x)] for x in results]
+        converted_results = [
+            [result.to_dict() for result in self.descriptor.output_converter(x)] for x in results.test_results
+        ]
 
         return {"results": converted_results}
 
@@ -137,8 +140,8 @@ class DeepEvalEvaluator:
         return default_from_dict(cls, data)
 
     @staticmethod
-    def _invoke_deepeval(test_cases: List[LLMTestCase], metric: BaseMetric) -> List[TestResult]:
-        return evaluate(test_cases, [metric])
+    def _invoke_deepeval(test_cases: List[LLMTestCase], metric: BaseMetric) -> EvaluationResult:
+        return evaluate(test_cases=test_cases, metrics=[metric])
 
     def _init_backend(self):
         """
--- a/integrations/deepeval/src/haystack_integrations/components/evaluators/deepeval/metrics.py
+++ b/integrations/deepeval/src/haystack_integrations/components/evaluators/deepeval/metrics.py
@@ -5,14 +5,14 @@ from enum import Enum
 from functools import partial
 from typing import Any, Callable, Dict, Iterable, List, Mapping, Optional, Type
 
-from deepeval.evaluate import TestResult
-from deepeval.metrics import (  # type: ignore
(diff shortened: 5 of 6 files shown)


anakin87:
@gulbaki thanks for the PR.
Please fix type errors. You can reproduce them locally by running `hatch run test:types`.

anakin87:
I spoke with @vblagoje, who will take care of reviewing this PR.

vblagoje:
Thanks for the updates @gulbaki - I'll try your branch today and verify that it all works as intended on real examples. We also have examples dir in this integration with [prompt_caching.py](https://github.com/deepset-ai/haystack-core-integrations/blob/main/integrations/anthropic/example/prompt_caching.py) that we can us to verify prompt caching still works as expected. Let me update that file while I experiment with your new prompt caching implementation.

gulbaki:
@vblagoje
I hadn’t realized that we’re skipping the real API key tests for Anthropic. Sorry about that.
I’ll probably get the same output as well.
I haven't tried it myself because I don’t have an API key.
Do you know if Anthropic offers any kind of free access or test credits?
If so, I’d be happy to try it out and investigate further.

vblagoje:
Ah, I was using updated https://github.com/deepset-ai/haystack-core-integrations/blob/main/integrations/anthropic/example/prompt_caching.py example and the `token_stats` never shows cache usage or cache hit. Please have a look @gulbaki

gulbaki:
@vblagoje 
https://docs.anthropic.com/en/docs/build-with-claude/prompt-caching#cache-limitations
`The minimum cacheable prompt length is:
1024 tokens for Claude Opus 4, Claude Sonnet 4, Claude Sonnet 3.7, Claude Sonnet 3.5 and Claude Opus 3`
`2048 tokens for Claude Haiku 3.5 and Claude Haiku 3`

If the input doesn't exceed this value(1024), caching won't be triggered, there's a minimum token threshold that must be met.
Since our messages didn't reach that limit, that's why we're seeing this issue.

I’ve added a small example below, and The [prompt_caching.py](https://github.com/deepset-ai/haystack-core-integrations/blob/main/integrations/anthropic/example/prompt_caching.py)  example is working against this rule — I can fix it if you’d like.

```
system_message = ChatMessage.from_system("Ab" * 4000)  # ~1000+ token
system_message._meta["cache_control"] = {"type": "ephemeral"}

messages = [
    system_message,
    ChatMessage.from_user("What is this?")
]


result1 = claude_llm.run(messages)
result2 = claude_llm.run(messages)

print(result1["replies"][0].meta["usage"])
print(result2["replies"][0].meta["usage"])
```
RESULT:
```

{'cache_creation_input_tokens': 4001, 'cache_read_input_tokens': 0, 'server_tool_use': None, 'service_tier': 'standard', 'prompt_tokens': 11, 'completion_tokens': 158}
{'cache_creation_input_tokens': 0, 'cache_read_input_tokens': 4001, 'server_tool_use': None, 'service_tier': 'standard', 'prompt_tokens': 11, 'completion_tokens': 149}
```

mathislucka:
I've got a question regarding this PR, maybe @gulbaki or @vblagoje can help:

The tests look like you need to set the cache control meta on the ChatMessages manually. The messages will be cached if `meta["cache_control"]` is set. Is this how it works?

The most important use case for prompt caching is agents. How would this work with an `Agent` where we can't set the cache control meta on the messages.

I think best would be to have a way to enable caching for all messages when we initialize the `AnthropicChatGenerator`. This way, we don't need to introduce Anthropic specific params in the Agent component.

Or am I missing something?

vblagoje:
> I've got a question regarding this PR, maybe @gulbaki or @vblagoje can help:
> 
> The tests look like you need to set the cache control meta on the ChatMessages manually. The messages will be cached if `meta["cache_control"]` is set. Is this how it works?
> 
> The most important use case for prompt caching is agents. How would this work with an `Agent` where we can't set the cache control meta on the messages.
> 
> I think best would be to have a way to enable caching for all messages when we initialize the `AnthropicChatGenerator`. This way, we don't need to introduce Anthropic specific params in the Agent component.
> 
> Or am I missing something?

Yes yes @mathislucka one needs to set cache_control on messages. Let's talk about it on Monday in the office - higher fidelity than Github. @gulbaki please continue with this task as we previously agreed and then after @mathislucka and I talk we'll loop you in more detail about our reasoning for eventual modifications.

gulbaki:
> Aha ok, ok. previously it worked and I checked that the message is "pretty long" but I never expected it to be this short

@vblagoje 
I made a mistake before — it was pretty long, sorry 😄
After checking the issue more carefully, I saw that the problem happens when streaming is turned on. In that case, caching doesn’t work. But if you turn off streaming, it works fine.
If you try it with streaming disabled, it should work.
In my  example I used, I did not experience any problems because the stream was closed, so I was wrong.

vblagoje:
@gulbaki thanks for your patience while I reviewed the details of why streaming didn't work. In fact it does but we didn't record it in usage. They key to prompt caching with streaming on is at https://docs.anthropic.com/en/docs/build-with-claude/prompt-caching#tracking-cache-performance That is in our code we should replace https://github.com/deepset-ai/haystack-core-integrations/blob/main/integrations/anthropic/src/haystack_integrations/components/generators/anthropic/chat/chat_generator.py#L429 with capturing all the key/value pairs, including prompt caching stats! What worked for me in quick experiments is
```
combined_usage = first_chunk_usage
```
Then these stats will show up in https://github.com/deepset-ai/haystack-core-integrations/blob/main/integrations/anthropic/example/prompt_caching.py 

Here are my suggestions what to do next:
1. Surgically update chat_generator and prompt_caching.py example and verify it works in streaming
2. Experiment with cache message being not the system prompt but the regular message, let's see the effect
3. Summarize our learnings and adjust the PR/tests (if needed) so we can move this PR forward to integration stage

gulbaki:
@vblagoje
Thanks for the detailed explanations. I'm working on my updates based on what you mentioned.

gulbaki:
> explanations. I'm working on my updates based on what you mentioned. By the way, my API credit has decreased a lot with the tests. We talked about this. Can you send me an API key for testing?
> Sure email me at dovlex at gmail.

I sent mail. my mail address bakigul at outlook

integrations/anthropic/example/prompt_caching.py
```diff
@@ -92,9 +92,12 @@ def stream_callback(chunk: StreamingChunk) -> None:
...
     # tokens read from the cache
     token_stats = result["llm"]["replies"][0].meta.get("usage")
-    if token_stats.get("cache_creation_input_tokens", 0) > 0:
+    cache_created = token_stats.get("cache_creation_input_tokens")
+    cache_read = token_stats.get("cache_read_input_tokens")
+
+    if isinstance(cache_created, int) and cache_created > 0:
         print("Cache created! ", end="")
-    elif token_stats.get("cache_read_input_tokens", 0) > 0:
+    elif isinstance(cache_read, int) and cache_read > 0:
```
vblagoje, line 100: since we have elif here we always see
Cache created! Cache usage details: {'cache_creation_input_tokens': 21, 'cache_read_input_tokens': 20791, 'server_tool_use': None, 'service_tier': 'standard', 'prompt_tokens': 3, 'completion_tokens': 594}
even though cache_read_input_tokens is 20791

vblagoje:
Interesting that sonnet creates cache for 21 tokens of user message, why? Do you understand?

gulbaki:
> Interesting that sonnet creates cache for 21 tokens of user message, why? Do you understand?

Very interesting, could there be a situation like this; If cache_ is added to the system message or user, it may accept the system prompt and the user prompt together if their total exceeds 1024. In other words, if cache control is added to any of them and the total prompts are greater than 1024, it seems to be running the cache wdyt ?

vblagoje (approved):
Looks good, thanks @gulbaki

integrations/anthropic/example/prompt_caching.py
vblagoje: Ah @gulbaki is this correct or we don't have to change anything here?

PR #2064, opened 2025-07-04
feat: Add llama-stack integration

Amnah199:
### Proposed Changes:

- Add a new LlamaStackChatGenerator
- Fixes https://github.com/deepset-ai/haystack-private/issues/153

### Checklist


ConnorHack:
I was able to successfully test this Llama Stack integration using vLLM for inference, albeit with the following changes:
- Updating the model name to use the format: `meta-llama/Llama-3.2-3B-Instruct`
- Adding the `api_key` field to the `LlamaStackChatGenerator` constructor to add a placeholder API for OpenAI API compatibility

Amnah199:
@ConnorHack Thanks for the feedback! I’m glad to hear it’s working well for you. I’ll add support for an optional `api_key`. The `model` param will need to be adjusted based on the inference provider, as the default currently assumes Ollama.

integrations/llama_stack/src/haystack_integrations/components/generators/llama_stack/chat/chat_generator.py
```diff
@@ -0,0 +1,126 @@
...
+        http_client_kwargs: Optional[Dict[str, Any]] = None,
+    ):
+        """
+        Creates an instance of LlamaStackChatGenerator. You need to setup Llama Stack Server and have a model available.
+        By default, the model is set to `llama3.2:3b`, that is available
+        on Ollama as inference provider. For a complete list of models,
+        see [LlamaStack docs](https://llama-stack.readthedocs.io/en/latest/models/index.html).
+
+        :param model:
+            The name of the LlamaStack chat completion model to use.
```
sjrl: Mention that this will change depending on the provider

integrations/llama_stack/src/haystack_integrations/components/generators/llama_stack/chat/chat_generator.py
```diff
@@ -0,0 +1,126 @@
...
+    >>that focuses on enabling computers to understand, interpret, and generate human language in a way that is
+    >>meaningful and useful.')], _role=<ChatRole.ASSISTANT: 'assistant'>, _name=None,
+    >>_meta={'model': 'llama3.2:3b', 'index': 0, 'finish_reason': 'stop',
+    >>'usage': {'prompt_tokens': 15, 'completion_tokens': 36, 'total_tokens': 51}})]}
+    """
+
+    def __init__(
+        self,
+        *,
+        model: str = "llama3.2:3b",
```
sjrl: perhaps set no default value

ConnorHack:
> @ConnorHack for this integration can we set a placeholder api key inside the init method to prevent the OpenAI client from raising an error? Or does the api key need to be exposed in the init method so users can pass one in?
> 
> From what we can see in the docs it looks like secret management is meant to be handled through configuration files and not necessarily through this component. If that's the case we'd be fine with setting a placeholder key to prevent the OpenAI client from throwing an error.

@Amnah199 This is a fair question to raise, and a good point to mention how secret management is supposed to be handled. I'm okay with setting a placeholder `api_key` inside the init method for this integration.

integrations/llama_stack/src/haystack_integrations/components/generators/llama_stack/chat/chat_generator.py
```diff
@@ -132,3 +124,41 @@ def __init__(
...
+            self,
+            model=self.model,
+            streaming_callback=callback_name,
+            api_base_url=self.api_base_url,
+            organization=self.organization,
+            generation_kwargs=self.generation_kwargs,
+            timeout=self.timeout,
+            max_retries=self.max_retries,
+            tools=serialize_tools_or_toolset(self.tools),
+            tools_strict=self.tools_strict,
```
julian-risch, line 150: We should add `tools_strict` to the init too

julian-risch (approved):
LGTM! 👍

PR #2078, opened 2025-07-08
Google GenAI - refactor streaming

anakin87:
### Related Issues

- fixes #issue-number

### Checklist

### DIFF
--- a/integrations/google_genai/src/haystack_integrations/components/generators/google_genai/chat/chat_generator.py
+++ b/integrations/google_genai/src/haystack_integrations/components/generators/google_genai/chat/chat_generator.py
@@ -2,6 +2,7 @@
 #
 # SPDX-License-Identifier: Apache-2.0
 
+import json
 from datetime import datetime, timezone
 from typing import Any, AsyncIterator, Dict, Iterator, List, Optional, Union
 
@@ -12,9 +13,11 @@ from haystack.core.component import component
 from haystack.core.serialization import default_from_dict, default_to_dict
 from haystack.dataclasses import (
     AsyncStreamingCallbackT,
+    FinishReason,
     StreamingCallbackT,
     StreamingChunk,
     ToolCall,
+    ToolCallDelta,
     select_streaming_callback,
 )
 from haystack.dataclasses.chat_message import ChatMessage, ChatRole
@@ -29,6 +32,14 @@ from haystack.utils import Secret, deserialize_callable, deserialize_secrets_inp
 from jsonref import replace_refs
 
 from haystack_integrations.components.generators.google_genai.chat.utils import remove_key_from_schema
+
+# Mapping from Google GenAI finish reasons to Haystack FinishReason values
+FINISH_REASON_MAPPING = {
+    "STOP": "stop",
+    "MAX_TOKENS": "length",
+    "SAFETY": "content_filter",
+    "FUNCTION_CALL": "tool_calls",
+}
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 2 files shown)


integrations/google_genai/src/haystack_integrations/components/generators/google_genai/chat/chat_generator.py
```diff
@@ -359,37 +374,65 @@ def from_dict(cls, data: Dict[str, Any]) -> "GoogleGenAIChatGenerator":
...
+                                id=tool_call.id,
+                                tool_name=tool_call.tool_name,
+                                arguments=json.dumps(tool_call.arguments) if tool_call.arguments else None,
+                            )
+                        )
 
-        return chunk_text
+        # Prioritize tool calls over content when both are present
+        return StreamingChunk(
+            content="" if tool_calls else content,
```
anakin87: I see that having both `content` and `tool_calls` in the same `StreamingChunk` is not supported by the design of the dataclass and this won't probably happen. Is my approach OK?

sjrl: Yeah the original idea was not to support this under the assumption that this wouldn't happen from providers. However, happy to reconsider/update StreamingChunk if this does indeed happen.

sjrl:
This is looking good! One note is that in my opinion ideally we'd try and always use the util method [_convert_streaming_chunks_to_chat_message](https://github.com/deepset-ai/haystack/blob/4b515433d9ba18de36f4d0649b3d7dfbfcd2b0da/haystack/components/generators/utils.py#L72-L131) to construct the final chat message. 

This would hopefully help reduce code duplication across our integrations and make us confident that our StreamingChunks do indeed have all the info needed to construct a final chat message. What do you think?

integrations/google_genai/src/haystack_integrations/components/generators/google_genai/chat/chat_generator.py
```diff
@@ -359,37 +374,65 @@ def from_dict(cls, data: Dict[str, Any]) -> "GoogleGenAIChatGenerator":
...
                             id=part.function_call.id,
                         )
                         all_tool_calls.append(tool_call)
+                        tool_calls.append(
+                            ToolCallDelta(
+                                # Google GenAI does not provide index, but it is required for tool calls
+                                index=tc_index,
+                                id=tool_call.id,
+                                tool_name=tool_call.tool_name,
+                                arguments=json.dumps(tool_call.arguments) if tool_call.arguments else None,
```
sjrl: okay interesting you needed to use `json.dumps` here because Google returns a dict already for arguments?

integrations/google_genai/tests/test_chat_generator.py
```diff
@@ -170,6 +171,141 @@ def test_convert_message_to_google_genai_format_complex(self):
         assert google_content.parts[0].function_response.name == "math"
         assert google_content.parts[0].function_response.response == {"result": "4"}
 
+    def test_process_streaming_chunk_text_only(self, monkeypatch):
```
sjrl: In addition to these tests I think it might be useful to also have one like https://github.com/deepset-ai/haystack/blob/87c08bbab742be41d3e8b4eba6622a4b782bb890/test/components/generators/chat/test_openai.py#L1173-L1180 

where we take an actual list of chunks from Google GenAI and see how they all convert to StreamingChunks. Usually I've been trying to get the most complicated stream as possible which for OpenAI is getting it to return two tool calls in one response. For Claude it'd be to get a though process and then two tool calls.

integrations/google_genai/src/haystack_integrations/components/generators/google_genai/chat/chat_generator.py
anakin87: yes

```python
class FunctionCall(_common.BaseModel):
  """A function call."""
...
  args: Optional[dict[str, Any]] = Field(
      default=None,
      description="""Optional. The function parameters and values in JSON object format. See [FunctionDeclaration.parameters] for parameter details.""",
  )
```

integrations/google_genai/tests/test_chat_generator.py
anakin87: Added a real-world example in https://github.com/deepset-ai/haystack-core-integrations/pull/2078/commits/fe0c34b98aaf99fca9cc2b6e0b4564f63e36f9ba

anakin87:
I'll also add `component_info` in this PR.

integrations/google_genai/src/haystack_integrations/components/generators/google_genai/chat/chat_generator.py
```diff
@@ -397,82 +397,51 @@ def _process_streaming_chunk(
...
+                    )
+
+        # start is only used by print_streaming_chunk. We try to make a reasonable assumption here but it should not be
+        # a problem if we change it in the future.
+        start = index == 0 or len(tool_calls) > 0
 
-        # Prioritize tool calls over content when both are present
         return StreamingChunk(
-            content="" if tool_calls else content,
+            content="" if tool_calls else content,  # prioritize tool calls over content when both are present
```
sjrl, line 431: Is it possible for both to be present at the same time?

anakin87: generally, this does not happen for Google

sjrl (approved):
Looks good!

PR #2088, opened 2025-07-11
fix: Fix the error in versioning of llama-stack integration

Amnah199:
### Proposed Changes:
 The pypi release is failing because of a typo in release tags. This PR fixes the error.


integrations/llama_stack/pyproject.toml
```diff
@@ -34,7 +34,7 @@ packages = ["src/haystack_integrations"]
 
 [tool.hatch.version]
 source = "vcs"
-tag-pattern = 'integrations\/llama-stack-v(?P<version>.*)'
+tag-pattern = 'integrations\/llama_stack-v(?P<version>.*)'
```
julian-risch, line 37: We could also push a tag with `llama-stack` instead of `llama_stack` but the latter is more consistent with most other integrations.

integrations/llama_stack/pyproject.toml
```diff
@@ -22,7 +22,7 @@ classifiers = [
   "Programming Language :: Python :: Implementation :: CPython",
   "Programming Language :: Python :: Implementation :: PyPy",
 ]
-dependencies = ["haystack-ai>=2.13.1", "llama-stack"]
+dependencies = ["haystack-ai>=2.14", "llama-stack"]
```
julian-risch, line 25: Just curious, any specific reason?

julian-risch (approved):
LGTM! 👍

PR #2096, opened 2025-07-18
feat: adopt new `StreamingChunk` in Anthropic

### LINKED ISSUE #2073: feat: Update `AnthropicChatGenerator` to use the `StreamingChunk` fields

Amnah199:
### Related Issues

- fixes #2073

### Proposed Changes:

adopt the new `StreamingChunk` and stream also tool calls
pass component info to `StreamingChunk`

### How did you test it?

Updated the tests 
Added new tests for streaming chunks
### Checklist

### DIFF
--- a/integrations/anthropic/tests/test_chat_generator.py
+++ b/integrations/anthropic/tests/test_chat_generator.py
@@ -14,10 +14,8 @@ from anthropic.types import (
     MessageDeltaUsage,
     RawContentBlockDeltaEvent,
     RawContentBlockStartEvent,
-    RawContentBlockStopEvent,
     RawMessageDeltaEvent,
     RawMessageStartEvent,
-    RawMessageStopEvent,
     TextBlock,
     TextBlockParam,
     TextDelta,
@@ -86,6 +84,7 @@ def mock_anthropic_completion():
         )
         mock_anthropic.return_value = completion
         yield mock_anthropic
+
 
 class TestAnthropicChatGenerator:
     def test_init_default(self, monkeypatch):


integrations/anthropic/src/haystack_integrations/components/generators/anthropic/chat/chat_generator.py
```diff
@@ -356,110 +372,50 @@ def _convert_chat_completion_to_chat_message(
...
+        chunk: RawMessageStreamEvent, index: int, component_info: ComponentInfo
     ) -> StreamingChunk:
         """
         Converts an Anthropic StreamEvent to a StreamingChunk.
-        """
-        content = ""
-        if chunk.type == "content_block_delta" and chunk.delta.type == "text_delta":
-            content = chunk.delta.text
 
-        return StreamingChunk(content=content, meta=chunk.model_dump(), component_info=component_info)
```
Amnah199, line 368: @sjrl I had a question about meta. In the new approach we create our own meta, and no longer do chunk.model_dump right?

integrations/anthropic/src/haystack_integrations/components/generators/anthropic/chat/chat_generator.py
```diff
@@ -356,110 +372,50 @@ def _convert_chat_completion_to_chat_message(
...
+                    index=chunk.index,
+                    id=chunk.content_block.id,
+                    tool_name=chunk.content_block.name,
+                ))
+        elif chunk.type == "content_block_delta":
+            if chunk.delta.type == "text_delta":
+                content = chunk.delta.text
+            elif chunk.delta.type == "input_json_delta":
+                tool_calls.append(ToolCallDelta(
+                    index=chunk.index,
```
Amnah199: @sjrl The tool_call chunks in anthropic have an `index` but its always `index=1`. If the purpose of our `ToolCallDelta.index` is to keep a counter, then this can be confusing.

integrations/anthropic/src/haystack_integrations/components/generators/anthropic/chat/chat_generator.py
sjrl: So the way I've been approaching meta currently is that I more or less left it alone for our ChatGenerators in Haystack main. I do realize this led to duplicate information in different formats, but I thought it would be the least breaking way to make this change. So I'd say up to you if you keep it as is or change it.

integrations/anthropic/src/haystack_integrations/components/generators/anthropic/chat/chat_generator.py
sjrl: To help answer your question it would be helpful if you could create a test like I describe here https://github.com/deepset-ai/haystack-core-integrations/pull/2078#discussion_r2194627929

sjrl: Here you can also find a similar one for Bedrock https://github.com/deepset-ai/haystack-core-integrations/blob/8d593f433b8fcff9ec8856b5bd8855744ae4476e/integrations/amazon_bedrock/tests/test_chat_generator_utils.py#L337 with an Anthropic model. 

There I can see in the raw_chunk from Bedrock they do indeed have something called `contentBlockIndex` which keeps track of the index we are interested in.

integrations/anthropic/src/haystack_integrations/components/generators/anthropic/chat/chat_generator.py
```diff
@@ -356,110 +372,49 @@ def _convert_chat_completion_to_chat_message(
...
-                    full_content += delta.get("text", "")
-                elif delta.get("type") == "input_json_delta" and current_tool_call:
-                    current_tool_call["arguments"] += delta.get("partial_json", "")
-            elif chunk_type == "message_delta":
-                if chunk.meta.get("delta", {}).get("stop_reason") == "tool_use" and current_tool_call:
-                    try:
-                        # When calling a tool with no arguments, the `arguments` field is an empty string.
-                        # We handle this by checking if `arguments` is empty and setting it to an empty dict.
-                        arguments = (
-                            json.loads(current_tool_call.get("arguments", "{}"))
```
Amnah199, line 406: @sjrl 
Context: I replaced this function to use your `_convert_streaming_chunks_to_chat_message` utility function. Mostly it works great except some subtle differences.

I saw[ here](https://github.com/deepset-ai/haystack/blob/e5810005ed7c833247763a37f890019857750fc4/haystack/components/generators/utils.py#L107), you do not pass an empty {} if arguments are missing and you skip adding the toolcall to `ChatMessage`. 
I am wondering if this same behaviour should be reflected in Anthropic.

sjrl: Ahh that is a good catch and the behavior here is probably what we should adopt generally. I think this came up when Mathis extensively used this component in his Agent testing. The solution to the problem of calling a tool with no args was either to instruct the LLM to output `{}` or for us to do that automatically like we do here. 

I think it's worth adding this change to `_convert_streaming_chunks_to_chat_message` in Haystack main.

Amnah199: The old function was still appending the ToolCall with empty arguments to `ChatMessage`. I believe its our choice, but I tried to read some Anthropic docs and it allows to create tools that dont require arguments 🤔 .

Sorry for the confusion

integrations/anthropic/tests/test_chat_generator.py
```diff
@@ -79,6 +96,90 @@ def mock_anthropic_completion():
...
+                input={},
+                name="calculator",
+                type="tool_use"
+            ),
+            index=1,
+            type="content_block_start"
+        ),
+        RawContentBlockDeltaEvent(
+            delta=InputJSONDelta(partial_json="", type="input_json_delta"),
+            index=1,
```
Amnah199: @sjrl these are the raw chunks from anthropic and adding to my older comment, the index in tool calls is always `index=1`.

integrations/anthropic/pyproject.toml
```diff
@@ -23,7 +23,7 @@ classifiers = [
   "Programming Language :: Python :: Implementation :: CPython",
   "Programming Language :: Python :: Implementation :: PyPy",
 ]
-dependencies = ["haystack-ai>=2.15.1", "anthropic>=0.47.0"]
+dependencies = ["haystack-ai @ git+https://github.com/deepset-ai/haystack.git@main", "anthropic>=0.47.0"]
```
Amnah199: I added this for now, because we need the latest `_convert_streaming_chunks_to_chat_message` function from main.

integrations/anthropic/tests/test_chat_generator.py
sjrl: Hmm I see that there is only one tool call in this stream. Is it possible to get Anthropic to send two tool calls in one streaming response? When that happens I'd imagine then the index will increment to 2 for the second tool call.

integrations/anthropic/pyproject.toml
sjrl: Sounds good, let's wait for the release then and update this dep when that happens

integrations/anthropic/src/haystack_integrations/components/generators/anthropic/chat/chat_generator.py
```diff
@@ -527,19 +487,25 @@ def _process_response(
...
             model: Optional[str] = None
+            finish_reason = None
             component_info = ComponentInfo.from_component(self)
             for chunk in response:
                 if chunk.type in ["message_start", "content_block_start", "content_block_delta", "message_delta"]:
                     # Extract model from message_start chunks
                     if chunk.type == "message_start":
                         model = chunk.message.model
+                    elif chunk.type == "message_delta":
+                        finish_reason = getattr(chunk.delta, "stop_reason", None)
```
sjrl: Just to confirm, we've now agreed this isn't needed anymore?

integrations/anthropic/src/haystack_integrations/components/generators/anthropic/chat/chat_generator.py
```diff
@@ -567,21 +533,33 @@ async def _process_response_async(
...
             async for chunk in response:
                 if chunk.type == "message_start":
                     model = chunk.message.model
+                elif chunk.type == "message_delta":
+                    finish_reason = getattr(chunk.delta, "stop_reason", None)
                 elif chunk.type in [
                     "content_block_start",
                     "content_block_delta",
                     "message_delta",
                 ]:
```
sjrl: Could we align the for loops between the sync and async versions of this function? I noticed that they have a slightly different implementation.

integrations/anthropic/tests/test_chat_generator.py
```diff
@@ -422,12 +517,22 @@ def test_convert_streaming_chunks_to_chat_message(self):
                     },
                 },
                 component_info=ComponentInfo.from_component(self),
+                index=0,
```
sjrl: So this chunk shouldn't have index set, since we should only set it if this chunk belongs to a content block. Since this is a message_start type it's not part of a content block so index should be left as None.

integrations/anthropic/tests/test_chat_generator.py
```diff
@@ -399,6 +485,15 @@ def test_convert_anthropic_chunk_to_streaming_chunk(self):
...
         assert streaming_chunk.component_info.type.endswith("chat_generator.AnthropicChatGenerator")
 
+    def test_convert_anthropic_completion_chunks_to_streaming_chunks(self, mock_anthropic_completion_chunks):
+        component = AnthropicChatGenerator(api_key=Secret.from_token("test-api-key"))
+        component_info = ComponentInfo.from_component(component)
+        for chunk in mock_anthropic_completion_chunks:
+            streaming_chunk = component._convert_anthropic_chunk_to_streaming_chunk(
+                chunk, component_info=component_info
+            )
+            assert streaming_chunk.meta["type"] == chunk.type
```
sjrl: We should expand this test so we compare each converted anthropic completion chunk into the expected streaming chunk. This will help us determine if the conversion is working as expected across an entire response.

integrations/anthropic/tests/test_chat_generator.py
```diff
@@ -422,12 +517,22 @@ def test_convert_streaming_chunks_to_chat_message(self):
...
+                start=True,
+                finish_reason=None,
             ),
             # Initial text content
             StreamingChunk(
                 content="",
                 meta={"type": "content_block_start", "index": 0, "content_block": {"type": "text", "text": ""}},
                 component_info=ComponentInfo.from_component(self),
+                index=0,
+                tool_calls=[],
```
sjrl: `tool_calls` should be `None` if there are no tool calls.

sjrl: Also just as a future note, if the values of the fields are just the default you could leave them out so you don't have to worry about filling them in every chunk in the tests.

integrations/anthropic/tests/test_chat_generator.py
```diff
@@ -481,15 +613,19 @@ def test_convert_streaming_chunks_to_chat_message(self):
                     "delta": {"stop_reason": "tool_use", "stop_sequence": None},
                     "usage": {"completion_tokens": 40},
                 },
+                component_info=ComponentInfo.from_component(self),
+                index=1,
```
sjrl: So I'd say this streaming chunk is also not part of the content block since it's of type `message_delta` instead of `content_block_delta`. So I'd also assign the index to None here.

integrations/anthropic/tests/test_chat_generator.py
```diff
@@ -498,209 +634,141 @@ def test_convert_streaming_chunks_to_chat_message(self):
...
                         "content": [],
                         "model": "claude-sonnet-4-20250514",
                         "stop_reason": None,
                         "stop_sequence": None,
-                        "usage": {"input_tokens": 50, "output_tokens": 0},
+                        "usage": {"input_tokens": 25, "output_tokens": 0},
                     },
                 },
+                component_info=ComponentInfo.from_component(self),
+                index=0,
```
sjrl, line 634: Same comment [here](https://github.com/deepset-ai/haystack-core-integrations/pull/2096/files#r2250991154)

integrations/anthropic/tests/test_chat_generator.py
```diff
@@ -498,209 +634,141 @@ def test_convert_streaming_chunks_to_chat_message(self):
...
             StreamingChunk(
                 content="",
-                meta={
-                    "content_block": {"citations": None, "text": "", "type": "text"},
-                    "index": 0,
-                    "type": "content_block_start",
-                },
+                meta={"type": "content_block_start", "index": 0, "content_block": {"type": "text", "text": ""}},
+                component_info=ComponentInfo.from_component(self),
+                index=1,
```
sjrl, line 642: I believe this should be index=0 since this is the first content block in the stream from what I can tell.

integrations/anthropic/tests/test_chat_generator.py
```diff
@@ -498,209 +634,141 @@ def test_convert_streaming_chunks_to_chat_message(self):
...
-                        "name": "hello_world",
-                        "type": "tool_use",
-                    },
-                    "index": 1,
                     "type": "content_block_start",
+                    "index": 1,
+                    "content_block": {"type": "tool_use", "id": "toolu_123", "name": "weather", "input": {}},
                 },
+                component_info=ComponentInfo.from_component(self),
+                index=1,
```
sjrl, line 674: And this looks to be the correct start of `index=1`

integrations/anthropic/tests/test_chat_generator.py
```diff
@@ -455,6 +570,12 @@ def test_convert_streaming_chunks_to_chat_message(self):
                     "index": 1,
                     "content_block": {"type": "tool_use", "id": "toolu_123", "name": "weather", "input": {}},
                 },
+                component_info=ComponentInfo.from_component(self),
+                index=1,
+                tool_calls=[ToolCallDelta(index=1, id="toolu_123", tool_name="weather", arguments=None)],
```
sjrl: I think we talked about this, but this should be `index=0` since it's the first and only tool call in this chunk. 
```suggestion
                tool_calls=[ToolCallDelta(index=0, id="toolu_123", tool_name="weather", arguments=None)],
```

sjrl: Same goes for the rest of the chunks in this test.

integrations/anthropic/tests/test_chat_generator.py
```diff
@@ -498,209 +634,141 @@ def test_convert_streaming_chunks_to_chat_message(self):
...
-                        "type": "tool_use",
-                    },
-                    "index": 1,
                     "type": "content_block_start",
+                    "index": 1,
+                    "content_block": {"type": "tool_use", "id": "toolu_123", "name": "weather", "input": {}},
                 },
+                component_info=ComponentInfo.from_component(self),
+                index=1,
+                tool_calls=[ToolCallDelta(index=1, id="toolu_123", tool_name="weather", arguments=None)],
```
sjrl: Same comment [here](https://github.com/deepset-ai/haystack-core-integrations/pull/2096/files#r2251013946)

integrations/anthropic/tests/test_chat_generator.py
```diff
@@ -498,209 +634,141 @@ def test_convert_streaming_chunks_to_chat_message(self):
...
-                        "cache_read_input_tokens": None,
-                        "input_tokens": None,
-                        "output_tokens": 69,
-                        "server_tool_use": None,
-                    },
+                    "delta": {"stop_reason": "tool_use", "stop_sequence": None},
+                    "usage": {"completion_tokens": 40},
                 },
+                component_info=ComponentInfo.from_component(self),
+                index=1,
```
sjrl, line 698: Same comment [here](https://github.com/deepset-ai/haystack-core-integrations/pull/2096/files#r2251003392)

integrations/anthropic/tests/test_chat_generator.py
Amnah199: Sorry if its already discussed.
Anthropic response chunks iteratively increase the `index` for each content block regardless of text and tool calls. So here` index=1` is because its the second content block in stream that has tool call.
[See the raw response here](https://github.com/deepset-ai/haystack-core-integrations/blob/668dd9c5cf69aa39fbf384d89ba48ba2ff03e78e/integrations/anthropic/tests/test_chat_generator.py#L130)
 To achieve what you suggest we will have to handle this on our end by separately keeping track of tool calls indexes.
 WDYT?

sjrl: > So here index=1 is because its the second content block in stream that has tool call.
[Reference](https://github.com/deepset-ai/haystack-core-integrations/blob/668dd9c5cf69aa39fbf384d89ba48ba2ff03e78e/integrations/anthropic/tests/test_chat_generator.py#L130)

Right! So the overall index of the StreamingChunk should be 1 which you have already. So `StreamingChunk.index == 1`, but you're right we should then separately keep track of the tool call indexes. Remember though the tool call index is only relevant for the number of ToolCallDeltas present in a single StreamingChunk. So if there is only ever one ToolCallDelta in a StreamingChunk then the tool call index will always be 0.

Amnah199: Right, so a single `StreamingChunk` always has a single `ToolCallDelta` as each `InputJSONDelta` is passed as a separate `RawContentBlockDeltaEvent`.
So maybe we can always assign tool call index = 0.

sjrl: Sounds good!

Amnah199: I couldn't trigger multiple ToolCalls, as anthropic api kept producing malformed JSON args, like you mentioned.

integrations/anthropic/tests/test_chat_generator.py
```diff
@@ -317,87 +395,172 @@ def test_run_with_params(self, chat_messages, mock_anthropic_completion):
...
-        assert response["replies"][0].meta["finish_reason"] == "end_turn"
+        assert response["replies"][0].meta["finish_reason"] == "stop"
 
     def test_check_duplicate_tool_names(self, tools):
         """Test that the AnthropicChatGenerator component fails to initialize with duplicate tool names."""
         with pytest.raises(ValueError):
             AnthropicChatGenerator(tools=tools + tools)
 
-    def test_convert_anthropic_chunk_to_streaming_chunk(self):
+    def test_convert_anthropic_completion_chunks_with_tools_to_streaming_chunks(self):
```
Amnah199, line 333: @sjrl I updated this test to be more through in checking the conversion of each chunk to streaming chunk. Additionally I removed the old test as it should be covered here.

The `mock_anthropic_completion_chunks` are also not needed anymore, so I'll remove them. But they were helpful in understanding the flow of response chunks from anthropic.

PR #2108, opened 2025-07-25
feat: `LlamaCppChatGenerator` streaming support

### LINKED ISSUE #730: Add streaming support in LlamaCPPGenerator AND LlamaCPPChatGenerator
Llama CPP generators don't support streaming.

Implement support for streaming, as done for other generators (see `streaming_callback`).

**Additional context**
Drafted in https://github.com/deepset-ai/haystack-core-integrations/pull/723#issuecomment-2099831919
Related: #678

### LINKED ISSUE #1825: feat: pass `component_info` to `StreamingChunk` in `LlamaCppChatGenerator`
In this PR, https://github.com/deepset-ai/haystack/pull/9426
we added support for streaming a component's name and type by passing them as `ComponentInfo` to the generated `StreamingChunk` in chat generators in main repo. This support should also be extended to `LlamaCppChatGenerator`

### LINKED ISSUE #2075: feat: Update `LlamaCppChatGenerator` to use the `StreamingChunk` fields

anakin87:
### Related Issues

- fixes  #730: streaming support only added for ChatGenerator. I don't think we should add new features to Generator.
- fixes #1825
- fixes #2075

### Proposed Changes:
- introduce streaming support for `LlamaCppChatGenerator`, using the new `StreamingChunk` and passing `ComponentInfo`

### How did you test it?
- CI, several new tests
- locally tested with `print_streaming_chunk`: both standard completions and tool calls are correctly printed

### Notes for the reviewer
Don't get scared by the size of this PR. Most of the code consist of llama.cpp chunk objects for tests.

### Checklist

### DIFF
--- a/integrations/llama_cpp/src/haystack_integrations/components/generators/llama_cpp/chat/chat_generator.py
+++ b/integrations/llama_cpp/src/haystack_integrations/components/generators/llama_cpp/chat/chat_generator.py
@@ -335,7 +335,7 @@ class LlamaCppChatGenerator:
             # They are often spread across multiple chunks.
             new_tool_call_ids = set()
 
-            if chunk.get("choices"):
+            if chunk.get("choices") and len(chunk["choices"]) > 0:
                 choice = chunk["choices"][0]
                 delta = choice.get("delta", {})
 
@@ -388,7 +388,11 @@ class LlamaCppChatGenerator:
             streaming_chunks.append(streaming_chunk)
 
             # Stream the chunk
-            streaming_callback(streaming_chunk)
+            try:
+                streaming_callback(streaming_chunk)
+            except Exception as e:
+                logger.error(f"Error in streaming callback invocation: {e}")
+                continue
 
         message = _convert_streaming_chunks_to_chat_message(streaming_chunks)
         return {"replies": [message]}
--- a/integrations/llama_cpp/src/haystack_integrations/components/generators/llama_cpp/chat/chat_generator.py
+++ b/integrations/llama_cpp/src/haystack_integrations/components/generators/llama_cpp/chat/chat_generator.py
@@ -290,7 +290,6 @@ class LlamaCppChatGenerator:
                 response_stream=response_stream,  # type: ignore[arg-type]
                 streaming_callback=streaming_callback,
                 component_info=ComponentInfo.from_component(self),
-                model_path=self.model_path,
             )  # we know that response_stream is Iterator[CreateChatCompletionStreamResponse]
             # because create_chat_completion was called with stream=True, but mypy doesn't know that
 
--- a/integrations/llama_cpp/src/haystack_integrations/components/generators/llama_cpp/chat/chat_generator.py
+++ b/integrations/llama_cpp/src/haystack_integrations/components/generators/llama_cpp/chat/chat_generator.py
@@ -1,8 +1,18 @@
 import json
-from typing import Any, Dict, List, Optional, Union
+from datetime import datetime, timezone
+from typing import Any, Dict, Iterator, List, Optional, Union
 
 from haystack import component, default_from_dict, default_to_dict, logging
-from haystack.dataclasses import ChatMessage, ToolCall
+from haystack.components.generators.utils import _convert_streaming_chunks_to_chat_message
+from haystack.dataclasses import (
+    ChatMessage,
+    ComponentInfo,
+    StreamingCallbackT,
+    ToolCall,
+    ToolCallDelta,
+    select_streaming_callback,
+)
+from haystack.dataclasses.streaming_chunk import FinishReason, StreamingChunk, SyncStreamingCallbackT
 from haystack.tools import (
     Tool,
     Toolset,
@@ -10,6 +20,7 @@ from haystack.tools import (
     deserialize_tools_or_toolset_inplace,
     serialize_tools_or_toolset,
 )
+from haystack.utils import deserialize_callable, serialize_callable
 from llama_cpp import (
     ChatCompletionMessageToolCall,
     ChatCompletionRequestAssistantMessage,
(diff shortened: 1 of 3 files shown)


integrations/llama_cpp/tests/test_chat_generator.py
```diff
@@ -123,6 +124,487 @@ def test_convert_message_to_llamacpp_invalid():
...
+        0
+    ]
+    assert message.text == " France is located in Western Europe, bordered by Luxembourg, Germany, Switzerland"
+    assert message.tool_calls == []
+    assert message.meta["finish_reason"] == "stop"
+    assert message.meta["model"] == "tests/models/openchat-3.5-1210.Q3_K_S.gguf"
+    assert "completion_start_time" in message.meta
+
+
+def test_handle_streaming_response_tool_calls():
```
anakin87, line 270: Tools support in llama.cpp python bindings (not llama.cpp) is quite outdated and I was not able to force the compatible model (`functionary`) to produce more than one call. So the following is the most complex example I could come up with.

integrations/llama_cpp/src/haystack_integrations/components/generators/llama_cpp/chat/chat_generator.py
```diff
@@ -254,9 +304,95 @@ def run(
...
+                meta={
+                    "model": chunk["model"],
+                    "received_at": datetime.fromtimestamp(chunk["created"], tz=timezone.utc).isoformat(),
+                },  # llama.cpp does not provide usage metadata during streaming
+            )
+
+            streaming_chunks.append(streaming_chunk)
+
+            # Stream the chunk
+            streaming_callback(streaming_chunk)
```
mpangrazzi: One thing I've always wondered (in other Generators too): if `streaming_callback(streaming_chunk)` fail with an exception, it will probably break the streaming I guess. Has this ever happened? WDYT about wrapping this in try/except?

integrations/llama_cpp/src/haystack_integrations/components/generators/llama_cpp/chat/chat_generator.py
```diff
@@ -254,9 +304,95 @@ def run(
...
+            tool_calls = []
+            mapped_finish_reason = None
+
+            # Track new tool call IDs in this chunk.
+            # Considering tool call ID is the only reliable way to recognize tool calls in llama.cpp streaming.
+            # They are often spread across multiple chunks.
+            new_tool_call_ids = set()
+
+            if chunk.get("choices"):
+                choice = chunk["choices"][0]
```
mpangrazzi, line 339: Can it happen to have 0-length `choices`?

mpangrazzi (commented):
Looks already good! Left two minor comments.

integrations/llama_cpp/src/haystack_integrations/components/generators/llama_cpp/chat/chat_generator.py
anakin87: Yes. It was never reported but can happen.
I added a try/except block.

integrations/llama_cpp/src/haystack_integrations/components/generators/llama_cpp/chat/chat_generator.py
anakin87: Should not happen but I now adopted a more defensive approach.

sjrl: @anakin87 should we open an issue to update this in our other chat generators?

anakin87: Are you referring to `choices` length? I think it depends on how this part is implemented in different Chat Generators...

sjrl: ahh sorry wrong thread I meant for the try/except block [here](https://github.com/deepset-ai/haystack-core-integrations/pull/2108#discussion_r2236202351)

integrations/llama_cpp/src/haystack_integrations/components/generators/llama_cpp/chat/chat_generator.py
```diff
@@ -1,28 +1,47 @@
...
 from llama_cpp.llama_tokenizer import LlamaHFTokenizer
 
 logger = logging.getLogger(__name__)
 
+FINISH_REASON_MAPPING: Dict[str, FinishReason] = {
+    "stop": "stop",
+    "length": "length",
+    "tool_calls": "tool_calls",
+    "function_call": "tool_calls",
+}
```
sjrl, line 43: @Amnah199 brought up a good point here. 

If we convert the finish reason when streaming is active then the final ChatMessage ends up with the converted finish reason and not the original. However, if in non-streaming mode we end up with the original finish reason which is inconsistent. 

Perhaps we should also convert the finish reason in non-streaming mode as well? WDYT?

anakin87: Ok I can do that

sjrl: Only unsure if that should constitute a breaking change, what do you think?

anakin87: For comprehensiveness, I've added the mapping `function_call`>`tool_calls` (because this is a possible value) but `function_call` is deprecated in llama.cpp python bindings and in my experience I've never seen it.
So I think we can avoid releasing a major version in this case. Agree?

integrations/llama_cpp/src/haystack_integrations/components/generators/llama_cpp/chat/chat_generator.py
anakin87: This seems like a good idea and it's also a "good first issue". I'll create an issue.

integrations/llama_cpp/src/haystack_integrations/components/generators/llama_cpp/chat/chat_generator.py
sjrl: Yup, sounds good! No need for a major release.

Amnah199: Just an update for the `finish_reason` issue, I created a parent issue to track this in future.
https://github.com/deepset-ai/haystack-core-integrations/issues/2111

I will refine this further after looking into the behaviour in various integrations.

PR #2109, opened 2025-07-25
feat: adopt new `StreamingChunk` in Ollama

### LINKED ISSUE #2072: feat: Update `OllamaChatGenerator` to use the `StreamingChunk` fields

Amnah199:
### Related Issues

- fixes #2072 

### Proposed Changes:

adopt the new `StreamingChunk` and stream also tool calls

### How did you test it?

Updated the tests
### Checklist

### DIFF
--- a/integrations/ollama/src/haystack_integrations/components/generators/ollama/chat/chat_generator.py
+++ b/integrations/ollama/src/haystack_integrations/components/generators/ollama/chat/chat_generator.py
@@ -311,7 +311,9 @@ class OllamaChatGenerator:
                     ToolCallDelta(
                         index=tool_call_index,
                         tool_name=tool_call["function"]["name"],
-                        arguments=json.dumps(tool_call["function"]["arguments"]) if tool_call["function"]["arguments"] else "",
+                        arguments=json.dumps(tool_call["function"]["arguments"])
+                        if tool_call["function"]["arguments"]
+                        else "",
                     )
                 )
 


integrations/ollama/src/haystack_integrations/components/generators/ollama/chat/chat_generator.py
```diff
@@ -312,81 +337,18 @@ def _handle_streaming_response(
...
-        # Accumulators
-        arg_by_id: Dict[str, Union[str, dict]] = {}
-        name_by_id: Dict[str, str] = {}
-        id_order: List[str] = []
-
         # Stream
-        for raw in response_iter:
-            chunk = self._build_chunk(chunk_response=raw, component_info=component_info)
+        for index, raw in enumerate(response_iter):
+            chunk = self._build_chunk(chunk_response=raw, component_info=component_info, index=index)
```
Amnah199: I am creating index because response chunks from ollama dont have index.

integrations/ollama/src/haystack_integrations/components/generators/ollama/chat/chat_generator.py
```diff
@@ -312,81 +337,18 @@ def _handle_streaming_response(
...
-                # the Ollama server doesn't guarantee an id field in every tool_calls entry.
-                # OpenAI-compatible endpoint (/v1/chat/completions) - recent releases do add an auto-generated id
-                # when the model produces multiple tool calls, so that clients can map results back.
-                # Native Ollama endpoint (/api/chat) and older builds
-                # - the JSON often contains only function.name + arguments;
-                # many users have reported that id is missing even with several calls,
-                # making client-side resolution harder:
-                # https://github.com/ollama/ollama/issues/6708
-                # https://github.com/ollama/ollama/issues/7510
-                # - If id is provided → we can distinguish multiple calls to the same tool.
```
Amnah199, line 338: @anakin87 lets discuss if this is still relevant?

anakin87: This was introduced in https://github.com/deepset-ai/haystack-core-integrations/pull/1906. It's worth taking a look at that PR because the contributor seems to have experimented with several different models that we can't all test in the CI. (@sjrl was the reviewer)
My impression is that it is relevant since the linked Ollama issues are still open.

integrations/ollama/pyproject.toml
```diff
@@ -27,7 +27,7 @@ classifiers = [
     "Programming Language :: Python :: Implementation :: CPython",
     "Programming Language :: Python :: Implementation :: PyPy",
 ]
-dependencies = ["haystack-ai>=2.15.1", "ollama>=0.4.0", "pydantic"]
+dependencies = ["haystack-ai>=2.16.0", "ollama>=0.4.0", "pydantic"]
```
anakin87: nit: let's use 2.16.1 (since 2.16.0 was buggy)

integrations/ollama/src/haystack_integrations/components/generators/ollama/chat/chat_generator.py
```diff
@@ -283,19 +285,42 @@ def from_dict(cls, data: Dict[str, Any]) -> "OllamaChatGenerator":
...
+
         meta = {key: value for key, value in chunk_response_dict.items() if key != "message"}
         meta["role"] = chunk_response_dict["message"]["role"]
         if tool_calls := chunk_response_dict["message"].get("tool_calls"):
-            meta["tool_calls"] = tool_calls
+            for tool_index, tool_call in enumerate(tool_calls):
+                arg = tool_call["function"]["arguments"]
+                if isinstance(arg, dict):
+                    arg = str(arg)
+                arg = arg.replace("'", '"')
```
anakin87: I would suggest using `json.dumps` if possible

integrations/ollama/src/haystack_integrations/components/generators/ollama/chat/chat_generator.py
```diff
@@ -283,19 +285,42 @@ def from_dict(cls, data: Dict[str, Any]) -> "OllamaChatGenerator":
...
+                    )
+                )
 
-        return StreamingChunk(content=content, meta=meta, component_info=component_info)
+        return StreamingChunk(
+            content=content,
+            meta=meta,
+            index=index,
+            # we ignore the type error because ollama returns openai-compatible finish_reason
+            finish_reason=finish_reason,  # type: ignore[arg-type]
```
anakin87: I see that in specific cases other `done_reason`s are used, so I would use a dictionary mapping
https://github.com/ollama/ollama/blob/main/docs/api.md#response-20

anakin87 (commented):
I left some comments.

I would also suggest to:
- locally test some use cases (chat, tool calls) with `print_streaming_chunk` and confirm that the output is good
- write unit tests with real Ollama chunks
  - chat - similar to [this](https://github.com/deepset-ai/haystack-core-integrations/blob/4734dc28dfb4f4b7360cb6aff640769d5e2a8786/integrations/llama_cpp/tests/test_chat_generator.py#L127)
  - tool calls - similar to [this](https://github.com/deepset-ai/haystack-core-integrations/blob/4734dc28dfb4f4b7360cb6aff640769d5e2a8786/integrations/llama_cpp/tests/test_chat_generator.py#L270) (but with multiple tool calls if possible; in llama.cpp it was not possible)

integrations/ollama/src/haystack_integrations/components/generators/ollama/chat/chat_generator.py
Amnah199: I didnt test the openai endpoint of Ollama but this implementation made sense given the issues you mentioned.

integrations/ollama/src/haystack_integrations/components/generators/ollama/chat/chat_generator.py
```diff
@@ -309,69 +338,74 @@ def _handle_streaming_response(
...
 
         # Stream
-        for raw in response_iter:
-            chunk = self._build_chunk(chunk_response=raw, component_info=component_info)
+        for index, raw in enumerate(response_iter):
+            if raw.message.tool_calls:
+                tool_call_index += 1
+            chunk = self._build_chunk(
+                chunk_response=raw, component_info=component_info, index=index, tool_call_index=tool_call_index
+            )
```
Amnah199, line 355: Ollama doesnt provide a ToolCall index, so one way is to track it ourselves iteratively.

integrations/ollama/tests/test_chat_generator.py
```diff
@@ -587,6 +587,131 @@ def streaming_callback(_: StreamingChunk) -> None:
...
+            ),
+        ]
+
+        generator = OllamaChatGenerator()
+        result = generator._handle_streaming_response(ollama_chunks, print_streaming_chunk)
+        assert result["replies"][0].text == ""
+        assert result["replies"][0].tool_calls[0].tool_name == "calculator"
+        assert result["replies"][0].tool_calls[0].arguments == {"expression": "7 * (4 + 2)"}
+        assert result["replies"][0].tool_calls[1].tool_name == "factorial"
+        assert result["replies"][0].tool_calls[1].arguments == {"n": 5}
```
Amnah199, line 711: Note that both ToolCalls are part of the same ChatMessage.

integrations/ollama/src/haystack_integrations/components/generators/ollama/chat/chat_generator.py
```diff
@@ -14,6 +15,13 @@
 
 from ollama import ChatResponse, Client
 
+FINISH_REASON_MAPPING: Dict[str, FinishReason] = {
+    "stop": "stop",
+    "tool_calls": "tool_calls",
+    "load": "stop",
+    "unload": "stop",
```
anakin87: I would not explicitly map `load` and `unload`.
If we use `FINISH_REASON_MAPPING.get(finish_reason)`, this would return `None` which is an acceptable value
https://github.com/deepset-ai/haystack/blob/f2012a4521f4fad35ea8e0dd10530c9688e6eb12/haystack/dataclasses/streaming_chunk.py#L126

integrations/ollama/src/haystack_integrations/components/generators/ollama/chat/chat_generator.py
```diff
@@ -283,19 +291,40 @@ def from_dict(cls, data: Dict[str, Any]) -> "OllamaChatGenerator":
...
         if tool_calls := chunk_response_dict["message"].get("tool_calls"):
-            meta["tool_calls"] = tool_calls
+            for tool_call in tool_calls:
+                arg = tool_call["function"]["arguments"]
+
+                tool_calls_list.append(
+                    ToolCallDelta(
+                        index=tool_call_index,
+                        tool_name=tool_call["function"]["name"],
+                        arguments=arg,
```
anakin87: it seems strange that mypy does not complain
In Ollama python, `arguments: Mapping[str, Any]` (a dict)
https://github.com/ollama/ollama-python/blob/fe91357d4b9c86d79efe4fabbdfabf9a1e68b07f/ollama/_types.py#L303

while in our `ToolCallDelta`, `arguments: Optional[str]`
https://github.com/deepset-ai/haystack/blob/f2012a4521f4fad35ea8e0dd10530c9688e6eb12/haystack/dataclasses/streaming_chunk.py#L30

Do you think that a conversion is needed? Could you verify?

---

UPDATE

Reading the original code, I see that the contributor assumed that Ollama can return `str` `arguments` (incomplete) or `dict` `arguments`

integrations/ollama/src/haystack_integrations/components/generators/ollama/chat/chat_generator.py
```diff
@@ -283,19 +291,40 @@ def from_dict(cls, data: Dict[str, Any]) -> "OllamaChatGenerator":
...
+                    )
+                )
 
-        return StreamingChunk(content=content, meta=meta, component_info=component_info)
+        return StreamingChunk(
+            content=content,
+            meta=meta,
+            index=index,
+            finish_reason=FINISH_REASON_MAPPING.get(finish_reason, "stop") if finish_reason else "stop",
```
anakin87:
```suggestion
            FINISH_REASON_MAPPING.get(finish_reason or "")
```

If we do something like this, we get either a correct `FinishReason` or `None` (which is acceptable)

integrations/ollama/src/haystack_integrations/components/generators/ollama/chat/chat_generator.py
```diff
@@ -309,69 +338,74 @@ def _handle_streaming_response(
...
+        for index, raw in enumerate(response_iter):
+            if raw.message.tool_calls:
+                tool_call_index += 1
+            chunk = self._build_chunk(
+                chunk_response=raw, component_info=component_info, index=index, tool_call_index=tool_call_index
+            )
             chunks.append(chunk)
 
+            if chunk.tool_calls:
+                chunk.start = True
```
anakin87, line 359: I would agree with the `start` choice if each chunk contains a complete Tool Call.

But this conflicts with the assumption below
> • Ollama may stream the same `arguments` field in *two* different forms:
> 1) as one or more **str fragments**  -- characters of a JSON string, delivered chunk-by-chunk;
> 2) as a complete **dict**            -- fully-parsed JSON in a single chunk.
> ...

In case 1, we would probably end up with a duplicate streamed tool call.

integrations/ollama/src/haystack_integrations/components/generators/ollama/chat/chat_generator.py
```diff
@@ -384,8 +418,10 @@ def _handle_streaming_response(
         reply = ChatMessage.from_assistant(
             text=text,
             tool_calls=tool_calls or None,
-            meta=_convert_ollama_meta_to_openai_format(chunks[-1].meta) if chunks else None,
+            meta=_convert_ollama_meta_to_openai_format(chunks[-1].meta) if chunks else {},
         )
+        reply._meta["finish_reason"] = reply._meta.get("done_reason", "stop")
```
anakin87: this is related to https://github.com/deepset-ai/haystack-core-integrations/issues/2111
I would prefer to use the mapped `finish_reason`, but let's decide in the related issue

integrations/ollama/src/haystack_integrations/components/generators/ollama/chat/chat_generator.py
```diff
@@ -309,69 +338,74 @@ def _handle_streaming_response(
...
-                if isinstance(args, dict):
-                    # Dict beats anything seen so far (final, authoritative version).
-                    arg_by_id[tool_call_id] = args
-                elif isinstance(args, str):
-                    # Append only when we are still in "string mode".
-                    if not isinstance(arg_by_id.get(tool_call_id), dict):
-                        prev = arg_by_id.get(tool_call_id, "")
-                        arg_by_id[tool_call_id] = f"{prev}{args}"
-
         # Compose final reply
```
anakin87, line 387: if possible, we would like to use `_convert_streaming_chunks_to_chat_message` everywhere to perform this conversion

https://github.com/deepset-ai/haystack/blob/f2012a4521f4fad35ea8e0dd10530c9688e6eb12/haystack/components/generators/utils.py#L72

anakin87 (commented):
I spent significant time trying to understand how Ollama behaves when it comes to tool calls + streaming, since it seems a crucial point for this PR.

In all my experiments, I've found that single tool calls are included in a single chunk, but I am not sure that this always holds true. I've tried mistral-small3.1:24b, llama3.2:3b, llama3.1:8b, qwen3:0.6b, and qwen3:1.7b.

Also looking for online resources, I cannot find a clear source of truth. At this point, I would open an issue on ollama-python and ask.

---

Most of the comments I left are related, some are not.

integrations/ollama/src/haystack_integrations/components/generators/ollama/chat/chat_generator.py
Amnah199: Good point. This was added to allow `print_streaming_chunk` to correctly stream the ToolCall. Without chunk.start thats not possible.
Then at the moment, we dont have a way to check when a new tool call is starting.

integrations/ollama/src/haystack_integrations/components/generators/ollama/chat/chat_generator.py
Amnah199: Hmm thats difficult given we are recreating tool_calls by mapping their args using ids. The `_convert_streaming_chunks_to_chat_message` as is, cannot handle that.

integrations/ollama/src/haystack_integrations/components/generators/ollama/chat/chat_generator.py
Amnah199: Resolved because we assume complete `ToolCalls` in each chunk.

integrations/ollama/src/haystack_integrations/components/generators/ollama/chat/chat_generator.py
anakin87: outdated - moving this comment where relevant

integrations/ollama/src/haystack_integrations/components/generators/ollama/chat/chat_generator.py
```diff
@@ -283,19 +290,38 @@ def from_dict(cls, data: Dict[str, Any]) -> "OllamaChatGenerator":
...
         meta = {key: value for key, value in chunk_response_dict.items() if key != "message"}
         meta["role"] = chunk_response_dict["message"]["role"]
         if tool_calls := chunk_response_dict["message"].get("tool_calls"):
-            meta["tool_calls"] = tool_calls
+            for tool_call in tool_calls:
+                tool_calls_list.append(
+                    ToolCallDelta(
+                        index=tool_call_index,
+                        tool_name=tool_call["function"]["name"],
+                        arguments=tool_call["function"]["arguments"],
```
anakin87: it seems strange that mypy does not complain
In Ollama python, `arguments: Mapping[str, Any]` (a dict)
https://github.com/ollama/ollama-python/blob/fe91357d4b9c86d79efe4fabbdfabf9a1e68b07f/ollama/_types.py#L303

while in our `ToolCallDelta`, `arguments: Optional[str]`
https://github.com/deepset-ai/haystack/blob/f2012a4521f4fad35ea8e0dd10530c9688e6eb12/haystack/dataclasses/streaming_chunk.py#L30

Do you think that a conversion is needed (`json.dumps`)? Could you verify?

integrations/ollama/src/haystack_integrations/components/generators/ollama/chat/chat_generator.py
anakin87: Ok. Maybe we might find a way to use the existing function but would require a complex refactoring.

Let's add a comment in the code explaining why we don't use `_convert_streaming_chunks_to_chat_message`, differently from other Chat Generators.

integrations/ollama/src/haystack_integrations/components/generators/ollama/chat/chat_generator.py
Amnah199: Added a fix. I guess this was missed by mypy because earlier we do `chunk_response_dict = chunk_response.model_dump()` which returns generic `Dict[str, Any]`.

anakin87 (approved):
Once the format error is fixed, feel free to merge.

I am not completely sure that everything is perfect, but we'll figure out.

anakin87:
From https://github.com/ollama/ollama/issues/11633#issuecomment-3152357213

> Tool calls should be coming back fully parsed

Issue #2054, opened 2025-07-03
Bug Report: ClientSession expires after idle periods causing ClosedResourceError with empty error message


LastRemote:
[Note: I am too lazy, so this issue is also generated by cursor. I did take a look, but I would apologize in advance if there are some minor confusions somewhere]

### **Environment**
- **mcp-haystack version**: 0.4.0
- **mcp version**: 1.10.1  
- **Transport**: SSE
- **Environment**: Client: same issue whether it's inside a docker or local. Server: from url.

### **Description**
When using `MCPTool` with SSE transport, the underlying `ClientSession` expires after approximately 5~10 minutes of idle time, causing tool invocations to fail with `anyio.ClosedResourceError`. The error is then wrapped in `MCPInvocationError` but the original error message is lost, resulting in an empty error message.

### **Root Cause**
The issue occurs in the `call_tool` method at line 237:
```python
result = await self.session.call_tool(tool_name, tool_args)
```

The `ClientSession` from the MCP SDK becomes closed after idle periods, but mcp-haystack doesn't detect this condition or attempt to reconnect. When `call_tool` is invoked on a closed session, it raises `anyio.ClosedResourceError()`, which gets caught by the generic exception handler at line 248 and wrapped in `MCPInvocationError` with an empty error message.

### **Steps to Reproduce**
1. Initialize an `MCPTool` with SSE transport:
```python
from haystack.tools import MCPTool, SSEServerInfo

tool = MCPTool(
    name="some_tool",
    server_info=SSEServerInfo(url="http://localhost:8000/sse")
)
```

2. Use the tool successfully initially
3. Wait approximately 10 minutes without invoking the tool
4. Attempt to invoke the tool again

### **Expected Behavior**
- The tool should either automatically reconnect when the session expires, or
- Provide a clear error message indicating the session has expired and needs to be reconnected

### **Actual Behavior**
Tool invocation fails with:
```
MCPInvocationError("Failed to invoke tool 'tool_name' with args: {...} , got error: ")
```

Note the empty error message after "got error: " - this is because `anyio.ClosedResourceError()` converts to an empty string.

### **Technical Details**
The current architecture uses `_MCPClientSessionManager` to maintain a long-running connection in the background, but there's no mechanism to:
1. Detect when the session becomes closed/expired
2. Automatically reconnect when this occurs  
3. Provide meaningful error messages when connection issues occur

### **Proposed Solutions**
[...]

PR #2115, opened 2025-07-29
fix: ClientSession expires after idle periods causing ClosedResourceError with empty error message

vblagoje:
## Why:
Fixes critical issue where MCP SSE connections timeout after 5-10 minutes of idle time, causing tool invocations to fail with `anyio.ClosedResourceError` wrapped in unhelpful empty `MCPInvocationError` messages, requiring manual MCPTool recreation.

- fixes https://github.com/deepset-ai/haystack-core-integrations/issues/2054

## What:
- Added automatic reconnection logic with exponential backoff to `MCPClient.call_tool()` for SSE/HTTP transports
- Enhanced error handling to provide meaningful messages even when underlying exceptions have empty string representations  
- Created integration test that simulates real SSE timeout scenarios by killing/restarting server processes

## How can it be used:
```python
# Before: Manual recreation required after timeout
try:
    result = tool.invoke(data="some_data")
except MCPInvocationError as e:
    print(e)  # Empty error message: "Failed to invoke tool 'name' with args: {...}, got error: "
    # User had to recreate MCPTool instance manually

# After: Automatic reconnection with clear errors
try:
    result = tool.invoke(data="some_data") 
    # Works transparently even after 10+ minute idle periods
except MCPInvocationError as e:
    print(e)  # Clear message: "Tool 'name' failed and all 3 reconnection attempts failed. ClosedResourceError: Connection closed unexpectedly"
```

## How did you test it:
- Added integration test `test_real_sse_reconnection_after_server_restart` that uses real SSE server processes
- Tested manually in idle itinerary agent that restarts after 10 min
- Tested manually in isolated script with MCPToolset only
- Confirmed existing functionality remains unchanged

### DIFF
--- a/integrations/mcp/src/haystack_integrations/tools/mcp/mcp_tool.py
+++ b/integrations/mcp/src/haystack_integrations/tools/mcp/mcp_tool.py
@@ -14,7 +14,7 @@ from dataclasses import dataclass, fields
 from datetime import timedelta
 from typing import Any, cast
 
-import anyio  # Added for ClosedResourceError detection
+import anyio
 from anyio.streams.memory import MemoryObjectReceiveStream, MemoryObjectSendStream
 from exceptiongroup import ExceptionGroup
 from haystack import logging
@@ -264,7 +264,7 @@ class MCPClient(ABC):
 
                 # Only attempt reconnection for SSE/HTTP transports
                 if isinstance(self, SSEClient| StreamableHttpClient):
-                    logger.warning(f"Connection lost during tool call '{tool_name}' (attempt {attempt + 1}/{max_retries + 1}): {error_type}: {error_msg}")
+                    logger.warning(f"Connection lost during tool call '{tool_name}': {error_type}: {error_msg}")
 
                     try:
                         # Exponential backoff before reconnection attempt
--- a/integrations/mcp/src/haystack_integrations/tools/mcp/mcp_tool.py
+++ b/integrations/mcp/src/haystack_integrations/tools/mcp/mcp_tool.py
@@ -14,6 +14,7 @@ from dataclasses import dataclass, fields
 from datetime import timedelta
 from typing import Any, cast
 
+import anyio  # Added for ClosedResourceError detection
 from anyio.streams.memory import MemoryObjectReceiveStream, MemoryObjectSendStream
 from exceptiongroup import ExceptionGroup
 from haystack import logging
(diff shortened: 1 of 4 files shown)


integrations/mcp/src/haystack_integrations/tools/mcp/mcp_tool.py
```diff
@@ -234,19 +235,75 @@ async def call_tool(self, tool_name: str, tool_args: dict[str, Any]) -> str:
...
-        except MCPError:
-            # Re-raise specific MCP errors directly
-            raise
-        except Exception as e:
-            # Wrap other exceptions with context about which tool failed
-            message = f"Failed to invoke tool '{tool_name}' with args: {tool_args} , got error: {e!s}"
-            raise MCPInvocationError(message, tool_name, tool_args) from e
+        # Implement retry logic with exponential backoff for connection-related errors
+        max_retries = 3
+        base_delay = 1  # Start with 1 second
```
mpangrazzi: WDYT about adding those values in init and make them configurable with defaults? Something like:

```python
def __init__(self, ..., max_retries: int = 3, base_delay: float = 1.0):
    self.max_retries = max_retries
    self.base_delay = base_delay
```

vblagoje:
Thanks @mpangrazzi will do follow your advice 🙏

vblagoje:
There seems to be something wrong with BRAVE mcp either API keys or the docker image we use because the main fails as well all of a sudden https://github.com/deepset-ai/haystack-core-integrations/pull/2124 so the error experienced here is unrelated. 

I'll continue polishing this PR and disable the BRAVE stdio test, then I'll proceed to investigate why that test failed and reenable it later again.

integrations/mcp/src/haystack_integrations/tools/mcp/mcp_tool.py
```diff
@@ -234,19 +237,74 @@ async def call_tool(self, tool_name: str, tool_args: dict[str, Any]) -> str:
...
+                # Only attempt reconnection for SSE/HTTP transports (if available)
+                if isinstance(self, SSEClient | StreamableHttpClient) and (
+                    sse_client is not None or streamablehttp_client is not None
+                ):
+                    logger.warning(f"Connection lost during tool call '{tool_name}': {error_type}: {error_msg}")
+
+                    try:
+                        # Exponential backoff before reconnection attempt
+                        if attempt > 0:
+                            delay = min(self.base_delay * (2 ** (attempt - 1)), 30)  # Cap at 30 seconds
```
mpangrazzi: Would it make sense to make the 30 seconds hardcoded cap here as e.g. `max_delay` (between retries) in `__init__` (as we did for other vars)?

PR #2142, opened 2025-08-04
fix!: `GoogleGenAIChatGenerator` - align `finish_reason` between streaming and non-streaming

### LINKED ISSUE #2138: Update finish_reason in `GoogleGenAIChatGenerator`
In `GoogleGenAIChatGenerator`, there is inconsistency in value of `finish_reason` in final ChatMessage based on whether streaming is enabled or not.

Streaming Mode:
The `finish_reason` is mapped to a `FinishReason` enum. This used in streaming chunks and stored in the ChatMessage.
[Reference](https://github.com/deepset-ai/haystack-core-integrations/blob/d83b10515df1501dbac46dadf3262dfc0c226e43/integrations/google_genai/src/haystack_integrations/components/generators/google_genai/chat/chat_generator.py#L436)

Non-Streaming Mode:
The model's original `finish_reason` string is passed directly into the final `ChatMessage` without mapping.

To ensure consistency, we will map all `finish_reason` values—both in streaming and non-streaming ChatMessages and `meta` fields—to the `FinishReason` enum defined in `StreamingChunk`. This change will be included in a major version release.

anakin87:
### Related Issues

- fixes #2138

### Proposed Changes:
- use the same mapped `finish_reason` in streaming and non-streaming cases

### How did you test it?
CI, new assertions

### Notes for the reviewer
This is a breaking change and I will release a new major version

### Checklist

### DIFF
--- a/integrations/google_genai/src/haystack_integrations/components/generators/google_genai/chat/chat_generator.py
+++ b/integrations/google_genai/src/haystack_integrations/components/generators/google_genai/chat/chat_generator.py
@@ -44,7 +44,6 @@ FINISH_REASON_MAPPING: Dict[str, FinishReason] = {
     "PROHIBITED_CONTENT": "content_filter",
     "SPII": "content_filter",
     "IMAGE_SAFETY": "content_filter",
-    "FUNCTION_CALL": "tool_calls",
 }
 
 logger = logging.getLogger(__name__)
@@ -199,7 +198,7 @@ def _convert_google_genai_response_to_chatmessage(response: types.GenerateConten
         tool_calls=tool_calls,
         meta={
             "model": model,
-            "finish_reason": str(finish_reason) if finish_reason else None,
+            "finish_reason": FINISH_REASON_MAPPING.get(finish_reason or ""),
             "usage": {
                 "prompt_tokens": getattr(usage_metadata, "prompt_token_count", 0),
                 "completion_tokens": getattr(usage_metadata, "candidates_token_count", 0),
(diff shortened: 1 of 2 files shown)


integrations/google_genai/src/haystack_integrations/components/generators/google_genai/chat/chat_generator.py
```diff
@@ -44,7 +44,6 @@
     "PROHIBITED_CONTENT": "content_filter",
     "SPII": "content_filter",
     "IMAGE_SAFETY": "content_filter",
-    "FUNCTION_CALL": "tool_calls",
```
anakin87, line 47: I found that in #2078 I erroneously introduced this line, but Google GenAI does not have `genai.types.FinishReason.FUNCTION_CALL`. When the model produces a tool call, the finish reason is `genai.types.FinishReason.STOP`. https://googleapis.github.io/python-genai/genai.html#genai.types.FinishReason

I would prefer to stick to Google conventions and not implement logic to identify tool calls and remap the finish reason, but we can discuss this point.

sjrl: Huh that is a bit of a bummer about the finish reason being stop for tool calls. 

I think fine to leave as-is for now, but let's open an issue to discuss. I think ideally we would identify the finish reason as tool calls in that scenario but I could see that being some work.

anakin87: Ok. I will open an issue.

anakin87: https://github.com/deepset-ai/haystack-core-integrations/issues/2143

PR #2150, opened 2025-08-05
feat: Image support in GoogleGenAIChatGenerator

vblagoje:
## Why:
Enables multimodal capabilities for `GoogleGenAIChatGenerator` by adding ImageContent support through Google's Files API, allowing users to send images alongside text in conversations.

## What:
- Modified `_convert_message_to_google_genai_format` to handle ImageContent with Files API upload
- Added image format validation and base64 decoding with temporary file management

## How can it be used:
```python
from haystack.dataclasses import ChatMessage, ImageContent

image_content = ImageContent.from_file_path("image.jpg")
message = ChatMessage.from_user(content_parts=["What's in this image?", image_content])
generator = GoogleGenAIChatGenerator()
response = generator.run([message])
```

## How did you test it:
- Added integration test with real image file (`test_run_with_image_input`)
- Verified image format validation and error handling
- Added integration test mirroring Amazon Bedrock pattern


integrations/google_genai/src/haystack_integrations/components/generators/google_genai/chat/chat_generator.py
```diff
@@ -89,6 +106,54 @@ def _convert_message_to_google_genai_format(message: ChatMessage) -> types.Conte
                 )
             )
 
+    # Handle images (from user only)
+    if message.images:
+        if message.is_from(ChatRole.ASSISTANT):
+            msg = "Image content is not supported for assistant messages"
```
anakin87:
```suggestion
        if not message.is_from(ChatRole.USER):
            msg = "Image content is only supported for user messages"
```

Something like this should be more aligned with the comment above

integrations/google_genai/src/haystack_integrations/components/generators/google_genai/chat/chat_generator.py
```diff
@@ -89,6 +106,54 @@ def _convert_message_to_google_genai_format(message: ChatMessage) -> types.Conte
...
+            raise ValueError(msg)
+
+        for image_content in message.images:
```
anakin87: I would prefer to use a for-loop for all parts in `ChatMessage._content`.
This would preserve the order or textual and images part, which can be relevant.

I'm referring to something like
```
Here are some images. Image 1:
[IMAGE 1]
Image 2
[IMAGE 2]
Please describe them
```

(I'm not doing this in Ollama because it only allows passing a list of images instead of a list of multimodal parts)

integrations/google_genai/src/haystack_integrations/components/generators/google_genai/chat/chat_generator.py
```diff
@@ -89,6 +106,54 @@ def _convert_message_to_google_genai_format(message: ChatMessage) -> types.Conte
...
+                        file_data=types.FileData(file_uri=uploaded_file.uri, mime_type=image_content.mime_type)
+                    )
+                    parts.append(file_part)
+                finally:
+                    # Clean up temporary file
+                    os.unlink(temp_file_path)
+
+            except Exception as e:
+                msg = f"Failed to upload image via Google AI Files API: {e}"
+                raise RuntimeError(msg) from e
```
anakin87: For consistency with other implementations, I would prefer [passing inline image data](https://ai.google.dev/gemini-api/docs/image-understanding?lang=python#inline-image).

WDYT?

integrations/google_genai/tests/test_chat_generator.py
```diff
@@ -512,6 +514,27 @@ def test_live_run(self) -> None:
         assert "gemini-2.0-flash" in message.meta["model"]
         assert message.meta["finish_reason"] == "stop"
```
anakin87, line 594: I would add one/more unit tests for `_convert_message_to_google_genai_format` with image inputs

anakin87 (commented):
Luckily, the implementation does not seem complex

I left some initial comments

integrations/google_genai/src/haystack_integrations/components/generators/google_genai/chat/chat_generator.py
vblagoje: yeah, np, I decided for the File API approach as this inline approach has a limit of 20 MB, that sounded small so I opted for the other, wdyt @anakin87 ?

anakin87: I think that using inline data would simplify the logic.

About limits: Anthropic has 5 MB per image while OpenAI has 50 MB total, so they seem comparable.

(tagging also @sjrl for an opinion)

vblagoje: Ok, np @anakin87 - let's go with inline and later we can always switch if users ask for it.

sjrl: Yeah I'd also go with inline. Writing tempfiles like this can add a lot of latency to the pipeline.

integrations/google_genai/src/haystack_integrations/components/generators/google_genai/chat/chat_generator.py
```diff
@@ -580,12 +627,14 @@ def run(
             if streaming_callback:
                 # Use streaming
                 response_stream = self._client.models.generate_content_stream(
-                    model=self._model, contents=contents, config=config
+                    model=self._model, contents=cast(Any, contents), config=config
```
anakin87: It seems that `cast(Any)` is very similar to `type:ignore`. Why do we need this?

If possible, I would prefer to fix typing issues.
If not possible, we can use `type:ignore[specific-error-ignored]` with a comment explaining why we need this.

integrations/google_genai/tests/test_chat_generator.py
```diff
@@ -512,6 +592,73 @@ def test_live_run(self) -> None:
...
+
+        chat_message = ChatMessage.from_user(content_parts=["What's in the image? Max 5 words.", image_content])
+
+        response = client.run([chat_message])
+
+        first_reply = response["replies"][0]
+        assert isinstance(first_reply, ChatMessage)
+        assert ChatMessage.is_from(first_reply, ChatRole.ASSISTANT)
+        assert first_reply.text
+        assert "apple" in first_reply.text.lower()
```
anakin87: I would leave just one integration test to cut costs and times. Up to you which one to choose.

anakin87 (commented):
I like the general direction.
I left two minor comments.

integrations/google_genai/tests/test_chat_generator.py
vblagoje: The more complex one :-) Thanks @anakin87

Issue #2145, opened 2025-08-04
Add `redacted_thinking` support to Bedrock when using Claude


sjrl:
**Is your feature request related to a problem? Please describe.**
As a follow up to this issue https://github.com/deepset-ai/haystack-core-integrations/issues/2093 we should also add `redacted_thinking` support to Bedrock. 

More info here https://docs.aws.amazon.com/bedrock/latest/userguide/claude-messages-thinking-encryption.html

PR #2094, opened 2025-07-15
Adds support for thinking when using Claude with BedrockChatGenerator in multi-turn conversations

sjrl:
### Related Issues

- fixes https://github.com/deepset-ai/haystack-core-integrations/issues/2093

### Proposed Changes:

 
 
Adds an integration test (currently skipped) and a unit test for work on this issue https://github.com/deepset-ai/haystack-core-integrations/issues/2093

### Checklist

- I have read the [contributors guidelines](https://github.com/deepset-ai/haystack-core-integrations/blob/main/CONTRIBUTING.md) and the [code of conduct](https://github.com/deepset-ai/haystack-core-integrations/blob/main/CODE_OF_CONDUCT.md)
- I have updated the related issue with new insights and changes
- I added unit tests and updated the docstrings
- I've used one of the [conventional commit types](https://www.conventionalcommits.org/en/v1.0.0/) for my PR title: `fix:`, `feat:`, `build:`, `chore:`, `ci:`, `docs:`, `style:`, `refactor:`, `perf:`, `test:`.

PR #2153, opened 2025-08-05
feat: Add `reasoning_contents` to meta of BedrockChatGenerator to support normal thinking and redacted thinking

sjrl:
### Related Issues

- fixes https://github.com/deepset-ai/haystack-core-integrations/issues/2145

### Proposed Changes:

 
 
- Changes storage of reasoning content in ChatMessage.meta from `reasoning_content` to `reasoning_contents`. This isn't breaking b/c the previous PR that introduced `reasoning_content` has not been released. 
- Supports both `reasoning_text` and `redacted_thinking`. More info here https://docs.aws.amazon.com/bedrock/latest/userguide/claude-messages-thinking-encryption.html

### How did you test it?

- Updated unit tests
- Added integration tests

### Checklist

### DIFF
--- a/integrations/amazon_bedrock/tests/test_chat_generator.py
+++ b/integrations/amazon_bedrock/tests/test_chat_generator.py
@@ -33,7 +33,7 @@ MODELS_TO_TEST_WITH_IMAGE_INPUT = [
 ]
 
 MODELS_TO_TEST_WITH_THINKING = [
-    "arn:aws:bedrock:us-east-1::inference-profile/us.anthropic.claude-3-7-sonnet-20250219-v1:0",
+    "us.anthropic.claude-3-7-sonnet-20250219-v1:0",
     "us.anthropic.claude-sonnet-4-20250514-v1:0",
 ]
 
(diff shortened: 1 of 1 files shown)


sjrl:
@anakin87 I'd appreciate your review on this especially since you are looking into adding thinking content to ChatMessages

integrations/amazon_bedrock/tests/test_chat_generator.py
```diff
@@ -22,12 +22,21 @@
...
+    "us.anthropic.claude-sonnet-4-20250514-v1:0",
+    "cohere.command-r-plus-v1:0",
+]
 
 MODELS_TO_TEST_WITH_IMAGE_INPUT = [
     "us.anthropic.claude-sonnet-4-20250514-v1:0",
 ]
 
+MODELS_TO_TEST_WITH_THINKING = [
+    "arn:aws:bedrock:us-east-1::inference-profile/us.anthropic.claude-3-7-sonnet-20250219-v1:0",
```
anakin87: nit: not an AWS expert here, but can the `arn:aws:bedrock:us-east-1::inference-profile` part be avoided?

anakin87 (approved):
Looks good. I only left a minor comment.

I would also use this PR to reintroduce some comments that were deleted in #2094:
- "In the CI, those tests are skipped if AWS Authentication fails" - this was useful
- "This is always 0 b/c it represents the choice index" - IDK, I'll let you decide

integrations/amazon_bedrock/tests/test_chat_generator.py
sjrl: Not locally at least. I had a thread with David about this and some models were only available via this inference profile. I can try removing it and see if CI still passes since my local set up and the CI are different.

anakin87: OK to leave it as is. Better to have something that works both locally and in the CI.
Thanks for the explanation.

sjrl: I think it's fine to take out. It works and it's easy for me to add back locally when needed.

PR #2154, opened 2025-08-05
fix: error in `ToolCallDelta.index` for parallel tool calling

Amnah199:
### Related Issues

- fixes: The arguments for multiple tool calls are not being parsed correctly due to an incorrect indices in `ToolCallDelta`, which triggers a malformed JSON error [here](https://github.com/deepset-ai/haystack/blob/323274e17478394fdc73f9e47c4eeefded3fa657/haystack/components/generators/utils.py#L72)

### Proposed Changes:

[In this discussion](https://github.com/deepset-ai/haystack-core-integrations/pull/2096#discussion_r2251930603), we decided to keep `ToolCallDelta.index = 0`, which caused malformed JSON because all arguments were being assigned to the first `ToolCall` with (index 0).
When [converting streaming chunks](https://github.com/deepset-ai/haystack/blob/323274e17478394fdc73f9e47c4eeefded3fa657/haystack/components/generators/utils.py#L88), we rely on `ToolCallDelta.index` to map tool call arguments to the correct tool across chunks — even if a single chunk contains only one `ToolCallDelta`. 

### Checklist

### DIFF
--- a/integrations/anthropic/src/haystack_integrations/components/generators/anthropic/chat/chat_generator.py
+++ b/integrations/anthropic/src/haystack_integrations/components/generators/anthropic/chat/chat_generator.py
@@ -370,13 +370,14 @@ class AnthropicChatGenerator:
 
     @staticmethod
     def _convert_anthropic_chunk_to_streaming_chunk(
-        chunk: RawMessageStreamEvent, component_info: ComponentInfo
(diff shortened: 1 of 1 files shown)


mpangrazzi:
@Amnah199 Looks good to me, but I would wait for @sjrl review on this ;)

mpangrazzi (approved):

PR #2168, opened 2025-08-07
chore: Add framework name into UserAgent header for bedrock integration

0x-fang:
### Related Issues

This PR adds the framework name into User-Agent header of Bedrock API call

### Proposed Changes:

This change only append the framework name `haystack` into the User-Agent header field for Bedrock API call, which can be used to collect the usage. Similar change had been made to other framework such as [langchain](https://github.com/langchain-ai/langchain-aws/pull/558)

### How did you test it?

Run through all tests. 

### Checklist

### DIFF
--- a/integrations/amazon_bedrock/src/haystack_integrations/components/embedders/amazon_bedrock/document_embedder.py
+++ b/integrations/amazon_bedrock/src/haystack_integrations/components/embedders/amazon_bedrock/document_embedder.py
@@ -137,8 +137,7 @@ class AmazonBedrockDocumentEmbedder:
                 aws_profile_name=resolve_secret(aws_profile_name),
             )
             config = Config(
-                user_agent_extra="x-client-framework:haystack",
-                **(self.boto3_config if self.boto3_config else {})
+                user_agent_extra="x-client-framework:haystack", **(self.boto3_config if self.boto3_config else {})
             )
             self._client = session.client("bedrock-runtime", config=config)
         except Exception as exception:
--- a/integrations/amazon_bedrock/src/haystack_integrations/components/embedders/amazon_bedrock/document_embedder.py
+++ b/integrations/amazon_bedrock/src/haystack_integrations/components/embedders/amazon_bedrock/document_embedder.py
@@ -138,7 +138,9 @@ class AmazonBedrockDocumentEmbedder:
             )
             config: Optional[Config] = None
             if self.boto3_config:
-                config = Config(**self.boto3_config)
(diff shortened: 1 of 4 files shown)


integrations/amazon_bedrock/src/haystack_integrations/components/embedders/amazon_bedrock/document_embedder.py
```diff
@@ -138,7 +138,9 @@ def resolve_secret(secret: Optional[Secret]) -> Optional[str]:
             )
             config: Optional[Config] = None
             if self.boto3_config:
-                config = Config(**self.boto3_config)
+                config = Config(**self.boto3_config, user_agent_extra="x-client-framework:haystack")
+            else:
+                config = Config(user_agent_extra="x-client-framework:haystack")
```
anakin87: for simplicity, I would replace these lines with something like
```python
config_kwargs = {"user_agent_extra":"x-client-framework:haystack", **(self.boto3_config if self.boto3_config else {})}
```

I would do that everywhere in this PR

anakin87 (changes requested):
Thanks!

I left a comment

integrations/amazon_bedrock/src/haystack_integrations/components/embedders/amazon_bedrock/document_embedder.py
0x-fang: thanks, fixed in revision.

integrations/amazon_bedrock/src/haystack_integrations/components/generators/amazon_bedrock/chat/chat_generator.py
```diff
@@ -498,7 +499,12 @@ async def run_async(
             session = self._get_async_session()
             # Note: https://aioboto3.readthedocs.io/en/latest/usage.html
             # we need to create a new client for each request
-            async with session.client("bedrock-runtime", config=self.boto3_config) as async_client:
+            config = None
+            if self.boto3_config:
+                config = Config(**self.boto3_config, user_agent_extra="x-client-framework:haystack")
+            else:
+                config = Config(user_agent_extra="x-client-framework:haystack")
```
anakin87: please also adjust this, then it will be ready to merge

0x-fang: ah, thanks, fixed in the new revision.

anakin87 (approved):
Thanks! Merging now...

This change will be released as soon we have other features/bug fixes to release. Next week, I think...

PR #2169, opened 2025-08-08
fix: Pin openai to `<=1.99.2` avoid errors in Llama Stack integration

sjrl:
### Related Issues

- fixes failures due to openai's release of 1.99.3

### Proposed Changes:

 
 

We pin openai here since we are already pinning llama-stack so even when they fix the issue we won't be able to benefit from it unless we also unpin llama stack.

### How did you test it?

Existing tests

### Checklist


anakin87 (commented):
We should probably pin `<=1.99.1`

Not in this PR, but I would also unpin `llama-stack`.
I think that pinning a specific version is not a good practice, unless temporary.

PR #2173, opened 2025-08-08
fix: add `input_tokens` in usage of Anthropic messages

### LINKED ISSUE #2166: Usage for AnthropicChatGenerator in streaming mode is broken
**Describe the bug**
https://github.com/deepset-ai/haystack-core-integrations/pull/2096 breaks usage in AnthropicChatGenerator. When streaming, the first chunk has the usage for input tokens and the last one for output tokens. This was fixed here: https://github.com/deepset-ai/haystack-core-integrations/pull/2001 but the change was not taken into account for the refactoring.

**To Reproduce**
Steps to reproduce the behavior. Feel free to link a Colab we can run to investigate the issue.

 - Haystack version:
 - Integration version:

Amnah199:
### Related Issues

- fixes #2166 

### Proposed Changes:

Pass `input_tokens` from the first chunk.
### How did you test it?

Updated the tests to check for `input_tokens`
### Checklist

### DIFF
--- a/integrations/anthropic/tests/test_chat_generator.py
+++ b/integrations/anthropic/tests/test_chat_generator.py
@@ -994,7 +994,6 @@ class TestAnthropicChatGenerator:
 
         assert "claude-sonnet-4-20250514" in message.meta["model"]
         assert message.meta["finish_reason"] == "stop"
-        print(message.meta)
         assert callback.counter > 1
         assert "Paris" in callback.responses
         assert "input_tokens" in message.meta["usage"]


integrations/anthropic/tests/test_chat_generator.py
```diff
@@ -574,6 +587,16 @@ def test_convert_anthropic_completion_chunks_with_multiple_tool_calls_to_streami
         # message_stop_chunk = RawMessageStopEvent(type="message_stop")
         # but we don't stream it
 
+        generator = AnthropicChatGenerator()
```
sjrl: Since we've added this line we should make sure that this test has a pytest skip decorator if the API key is not provided. 

Or we should use monkeypatch to fake the API key or do this (example from above)
```python
        component = AnthropicChatGenerator(
            api_key=Secret.from_token("test-api-key")
        )
```
otherwise this test will fail locally

integrations/anthropic/tests/test_chat_generator.py
```diff
@@ -728,7 +751,7 @@ def test_convert_streaming_chunks_to_chat_message_with_multiple_tool_calls(self)
         # Verify meta information
         assert message._meta["index"] == 0
         assert message._meta["finish_reason"] == "tool_calls"
-        assert message._meta["usage"] == {"completion_tokens": 40}
+        assert message._meta["usage"] == {"output_tokens": 40, "completion_tokens": 40}
```
sjrl: Aren't we also missing `input_tokens` here?

integrations/anthropic/tests/test_chat_generator.py
```diff
@@ -703,7 +726,7 @@ def test_convert_streaming_chunks_to_chat_message_with_multiple_tool_calls(self)
                 meta={
                     "type": "message_delta",
                     "delta": {"stop_reason": "tool_calls", "stop_sequence": None},
-                    "usage": {"completion_tokens": 40},
+                    "usage": {"completion_tokens": 40, "output_tokens": 40},
```
sjrl: I'm not sure `output_tokens` should be added here. I believe Anthropics name for `output_tokens` is `completion_tokens`

sjrl:
@Amnah199 actually one more comment. I think we should update one of the integration tests to check for the presence of all expected keys in the usage metadata field.

integrations/anthropic/tests/test_chat_generator.py
Amnah199: Actually its just `output_tokens`, I'll remove `completion_tokens`.

integrations/anthropic/tests/test_chat_generator.py
Amnah199: We add the `input_tokens` after the conversion of streaming chunks [here](https://github.com/deepset-ai/haystack-core-integrations/blob/85df4220a777e43910b428697576a1690b1876b4/integrations/anthropic/src/haystack_integrations/components/generators/anthropic/chat/chat_generator.py#L521) so at this point its not in final message.

Issue #2140, opened 2025-08-04
Concurrency: `LangfuseTracer` is not thread-safe, causing mixed traces in async environments


LastRemote:
[P.S. This is auto-generated by Cursor but I think it is pretty well written]

### **Description**

There is a concurrency issue in the `haystack-integrations` implementation of `LangfuseTracer`. The tracer uses a shared instance-level list (`self._context`) to manage the span stack. In an asynchronous environment, such as a FastAPI application handling multiple requests concurrently, this shared state leads to a race condition.

As a result, spans from different, independent pipeline runs can be incorrectly nested or intertwined within the same trace in Langfuse. This corrupts the tracing data, making it unreliable for debugging and monitoring.

### **Affected Component**

`haystack-integrations.tracing.langfuse.tracer.LangfuseTracer`

### **Steps to Reproduce**

1.  Initialize a Haystack pipeline with `LangfuseTracer` enabled.
2.  Expose this pipeline through an async web framework like FastAPI.
3.  Send multiple requests to the pipeline endpoint concurrently.
4.  Observe the traces in the Langfuse UI. You will notice that spans from different requests are mixed together under a single trace instead of being isolated in their own respective traces.

### **Root Cause**

The `LangfuseTracer` instance maintains its span stack in an instance variable:

```python
class LangfuseTracer(Tracer):
    def __init__(...):
        # ...
        self._context: List[LangfuseSpan] = []
        # ...

    def trace(...):
        # ...
        self._context.append(span)
        # ...
        try:
            yield span
        finally:
            # ...
            if self._context and self._context[-1] == span:
                self._context.pop()
```

When a single `LangfuseTracer` instance is used for the entire application (which is standard practice), this `_context` list is shared across all concurrent requests, leading to the race condition.

### **Proposed Solution (Recommended)**

The standard Python solution for handling context-local state in asynchronous applications is to use `contextvars`. By replacing the instance-level list with a `ContextVar`, the span stack will be automatically isolated for each concurrent execution context (e.g., each incoming request).

The fix involves these changes:

1.  **Introduce a `ContextVar` for the span stack:**

    ```python
    # Add this at the module level
    span_context_var: ContextVar[List["LangfuseSpan"]] = ContextVar("span_context", default=[])
    ```

2.  **Modify `LangfuseTracer` to use the `ContextVar`:**

    ```python
    class LangfuseTracer(Tracer):
        def __init__(self, ...):
            # self._context is removed
            # ...

[...]

PR #2188, opened 2025-08-13
fix: LangfuseTracer is not thread-safe, causing mixed traces in async environments

vblagoje:
## Why:
**Purpose:** Fixes a race condition where spans from concurrent pipeline runs become intertwined, corrupting tracing data and making debugging unreliable in production web server environments.

- fixes https://github.com/deepset-ai/haystack-core-integrations/issues/2140

## What:
- Replaced instance-level `self._context` list with `ContextVar` for thread-safe span stack management
- Updated `LangfuseTracer.trace()` method to use context-local span storage
- Modified `current_span()` method to read from context-local state
- Added module-level `span_context_var` ContextVar for isolated span contexts per execution thread
- Fixed support for session IDs as well

## How can it be used:
Nothing changes in client code except that we don't corrupt traces in concurrent async pipelines. All existing usage patterns continue to work exactly as before, but now with proper thread safety.

## How did you test it:
- Added unit tests verifying context isolation between concurrent tracer instances
- Verified that spans are properly isolated per execution context
- Tested examples in examples directory and itinerary agent
- Ran existing test suite to ensure no regressions in single-threaded scenarios

## Notes for the reviewer:
@LastRemote would you please test this branch on your use case

### DIFF
--- a/integrations/langfuse/README.md
+++ b/integrations/langfuse/README.md
@@ -91,11 +91,43 @@ response = pipe.run(
 print(response["llm"]["replies"][0])
 print(response["tracer"]["trace_url"])
 print(response["tracer"]["trace_id"])
+
+# Using session_id to group related traces
+response = pipe.run(
+    data={"prompt_builder": {"template_variables": {"location": "Paris"}, "template": messages}},
+    tracer={"session_id": "user_123_conversation_456"}
+)
+print(response["llm"]["replies"][0])
+print(response["tracer"]["trace_url"])
+print(response["tracer"]["trace_id"])
 ```
 
 In this example, we add the `LangfuseConnector` to the pipeline with the name "tracer". 
 Each run of the pipeline produces one trace viewable on the Langfuse website with a specific URL. 
 The trace captures the entire execution context, including the prompts, completions, and metadata.
+
+### Using Session IDs
+
+You can group related traces together by passing a `session_id` when running the pipeline. This is useful for:
+- **Multi-turn conversations**: Group all messages in a chat session
+- **Workflow tracking**: Group all operations in a user workflow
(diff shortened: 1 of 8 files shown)


vblagoje:
> ```python
> # Using session_id to group related traces
> response = pipe.run(
>     data={"prompt_builder": {"template_variables": {"location": "Paris"}, "template": messages}},
>     tracer={"session_id": "user_123_conversation_456"}
> )
> ```
> 
> @vblagoje Is this supported in the current `main` branch of `haystack`? I found this quote in README but I don't see the `tracer` parameter being mentioned elsewhere.

Mistake, it should be tracer inside data! No not supported, I'm testing it now on this branch only

LastRemote:
Screenshot 2025-08-20 at 14 31 45
@vblagoje I did a real world scenario test and found unfortunate results. It looks like these three requests (the two async.run traces and the main trace) are still overlapping. I am using the test branch with ContextVar `span_stack_var`.

vblagoje:
> Looks good to me!
> 
> I have one question (not strictly related to this PR) about `_session_root_traces`: it seems that it will grow indefinitely without cleanup. Would you see this as a problem? I am thinking e.g. at long running applications.

Yes @mpangrazzi  spot on, there is another solution Seb and I are looking at now that would not require these _session_root_traces and would simplify things. Will update you soon.

vblagoje:
Dropping this PR in favour of https://github.com/deepset-ai/haystack-core-integrations/pull/2207
cc @LastRemote let's please test the other PR, it is much simpler change
cc @mpangrazzi let's wait for @LastRemote to confirm the fix in his particular scenario, he's the OP for the issue

PR #2190, opened 2025-08-14
Cohere Image Document Embedder - draft

anakin87:
### Related Issues

- fixes #issue-number

### Checklist


integrations/cohere/src/haystack_integrations/components/embedders/cohere/document_image_embedder.py
```diff
@@ -0,0 +1,341 @@
...
+            - `documents`: Documents with embeddings.
+        """
+
+        images_to_embed = self._extract_images_to_embed(documents)
+
+        embeddings = []
+
+        # The Cohere API only supports passing one image at a time
+        async for doc, image in tqdm_async(
+            zip(documents, images_to_embed), desc="Embedding images", disable=not self.progress_bar
```
Amnah199: Doesn't this raise a TypeError as zip is a sync iterator. Maybe we can use await inside the loop and remove async with for.

integrations/cohere/src/haystack_integrations/components/embedders/cohere/document_image_embedder.py
```diff
@@ -0,0 +1,341 @@
...
+            List of images to embed in the format expected by the Cohere API.
+
+        :raises TypeError:
+            If the input is not a list of `Documents`.
+        :raises ValueError:
+            If the input contains unsupported image MIME types.
+        :raises RuntimeError:
+            If the conversion of some documents fails.
+        """
+        if not isinstance(documents, list) or (documents and not isinstance(documents[0], Document)):
```
Amnah199: What if user passes string somewhere in the middle, this break in the mid of processing.
```suggestion
         if not isinstance(documents, list) or not all(isinstance(d, Document) for d in documents):
```

integrations/cohere/src/haystack_integrations/components/embedders/cohere/document_image_embedder.py
```diff
@@ -0,0 +1,341 @@
...
+                    model=self.model,
+                    images=[image],
+                    input_type="image",
+                    output_dimension=self.embedding_dimension,
+                    embedding_types=[self.embedding_type.value],
+                )
+
+                embedding = getattr(response.embeddings, self.embedding_type.value)[0]
+            except Exception as e:
+                msg = f"Error embedding Document {doc.id}. The Document will be skipped. \nException: {e}"
```
Amnah199: Maybe we can make it more clear that we are returning the document with embedding=None. Skipped is a little ambiguous. WDYT?

integrations/cohere/src/haystack_integrations/components/embedders/cohere/document_image_embedder.py
```diff
@@ -0,0 +1,341 @@
...
+from haystack.utils.auth import Secret, deserialize_secrets_inplace
+from tqdm import tqdm
+from tqdm.asyncio import tqdm as tqdm_async
+
+from cohere import AsyncClientV2, ClientV2
+
+from .embedding_types import EmbeddingTypes
+
+# PDF is not officially supported, but we convert PDFs to JPEG images
+SUPPORTED_IMAGE_MIME_TYPES = ["image/jpeg", "image/png", "application/pdf"]
```
Amnah199, line 24: Hmm what would happen to `image/jpg `? Is it not supported?

integrations/cohere/src/haystack_integrations/components/embedders/cohere/document_image_embedder.py
```diff
@@ -0,0 +1,341 @@
...
+                    model=self.model,
+                    images=[image],
+                    input_type="image",
+                    output_dimension=self.embedding_dimension,
+                    embedding_types=[self.embedding_type.value],
+                )
+
+                embedding = getattr(response.embeddings, self.embedding_type.value)[0]
+            except Exception as e:
+                msg = f"Error embedding Document {doc.id}. The Document will be skipped. \nException: {e}"
```
Amnah199: Same comment as above.

integrations/cohere/src/haystack_integrations/components/embedders/cohere/document_image_embedder.py
sjrl: Yes that's a good point. In other places we opted to only do a soft check here b/c users could theoretically pass a large amount of docs and doing a isinstance check on all of them could be uncessarily expensive (although tbh not sure how expensive it would be)

integrations/cohere/src/haystack_integrations/components/embedders/cohere/document_image_embedder.py
sjrl: @anakin87 I may advocate for just letting the exception raise here since that how most of our Document Embedders work (only exception is AzureDocumentEmbedder). If an embedding fails then the API error is raised so users can decide what to do.

integrations/cohere/src/haystack_integrations/components/embedders/cohere/document_image_embedder.py
sjrl: Yeah I agree, I think adding common variations which in this case is just `image/jpg` would be a good idea

integrations/cohere/src/haystack_integrations/components/embedders/cohere/document_image_embedder.py
```diff
@@ -0,0 +1,341 @@
...
+class CohereDocumentImageEmbedder:
+    """
+    A component for computing Document embeddings based on images using Cohere models.
+
+    The embedding of each Document is stored in the `embedding` field of the Document.
+
+    ### Usage example
+    ```python
+    from haystack import Document
+    from haystack_integrations.components.embedders.cohere import CohereDocumentImageEmbedder
```
sjrl, line 40: Should we follow the same structure as in Haystack and put these under a folder called `image`? Not a big deal just wondering if that's something we want to be consistent about. 

Or it could be in `components/embedders/image/cohere`.

integrations/cohere/src/haystack_integrations/components/embedders/cohere/document_image_embedder.py
```diff
@@ -0,0 +1,341 @@
...
+    def __init__(
+        self,
+        *,
+        file_path_meta_field: str = "file_path",
+        root_path: Optional[str] = None,
+        image_size: Optional[Tuple[int, int]] = None,
+        api_key: Secret = Secret.from_env_var(["COHERE_API_KEY", "CO_API_KEY"]),
+        model: str = "embed-v4.0",
+        api_base_url: str = "https://api.cohere.com",
+        timeout: int = 120,
```
sjrl: `timeout` is a float type in the `ClientV2` object.

integrations/cohere/src/haystack_integrations/components/embedders/cohere/document_image_embedder.py
anakin87: In general, I have followed what we do for Chat Generators

`from haystack.components.generators.chat import OpenAIChatGenerator`
vs
`from haystack_integrations.components.generators.cohere import CohereChatGenerator`

sjrl: Okay fair enough. Technically it's 
`from haystack_integrations.components.generators.cohere.chat import CohereChatGenerator` 
but we add an `__init__.py` to the `cohere/chat` folder so the import statement you provide works.

integrations/cohere/src/haystack_integrations/components/embedders/cohere/document_image_embedder.py
anakin87: I will add it to be on the safe side but `image/jpg` is not a valid MIME type: https://stackoverflow.com/questions/33692835/is-the-mime-type-image-jpg-the-same-as-image-jpeg.

integrations/cohere/src/haystack_integrations/components/embedders/cohere/document_image_embedder.py
anakin87: you are right. No need for `async for` here

integrations/cohere/src/haystack_integrations/components/embedders/cohere/document_image_embedder.py
anakin87: I think @sjrl is right about the motivation, but I extended the check

integrations/cohere/src/haystack_integrations/components/embedders/cohere/document_image_embedder.py
anakin87: I'm now letting the exception raise

integrations/cohere/src/haystack_integrations/components/embedders/cohere/document_image_embedder.py
anakin87: I'm now letting the exception raise

integrations/cohere/src/haystack_integrations/components/embedders/cohere/document_image_embedder.py
anakin87: fixed, also in other Cohere embedders. (I think that mypy was not complaining because `int` and `float` are compatible)

integrations/cohere/src/haystack_integrations/components/embedders/cohere/document_image_embedder.py
```diff
@@ -0,0 +1,334 @@
...
+
+from dataclasses import replace
+from typing import Any, Optional, Tuple
+
+from haystack import Document, component, default_from_dict, default_to_dict, logging
+from haystack.components.converters.image.image_utils import (
+    _batch_convert_pdf_pages_to_images,
+    _encode_image_to_base64,
+    _extract_image_sources_info,
+    _PDFPageInfo,
```
sjrl, line 13: Nothing to do now, but let's keep an eye on how often these are used in integrations. If we find that these are necessary for users to use to make custom image embedders we may want to consider making them public.

sjrl: Or create some sort of public interface that uses methods internally.

integrations/cohere/tests/test_document_image_embedder.py
```diff
@@ -0,0 +1,370 @@
...
+
+        with pytest.raises(RuntimeError, match="Error embedding Document"):
+            await embedder.run_async(documents=documents)
+
+    @pytest.mark.integration
+    @pytest.mark.skipif(
+        not os.environ.get("COHERE_API_KEY", None) and not os.environ.get("CO_API_KEY", None),
+        reason="Export an env var called COHERE_API_KEY/CO_API_KEY containing the Cohere API key to run this test.",
+    )
+    def test_live_run(self, test_files_path):
```
sjrl, line 347: Could we add an live `run_async` test as well?

anakin87: added

PR #2192, opened 2025-08-15
feat: add a new `S3Downloader` component

Amnah199:
### Related Issues

- fixes [#9264](https://github.com/deepset-ai/haystack/issues/9264)

### Proposed Changes:

- Add a new S3Downloader component in Bedrock integration
- Used the [`DeepsetDownloader`](https://github.com/deepset-ai/deepset-cloud-custom-nodes/blob/main/deepset_cloud_custom_nodes/augmenters/deepset_file_downloader.py) component as a guide 

### How did you test it?

Added a test file
### Checklist

### DIFF
--- a/integrations/amazon_bedrock/tests/test_s3_downloader.py
+++ b/integrations/amazon_bedrock/tests/test_s3_downloader.py
@@ -195,7 +195,6 @@ class TestS3Downloader:
         not os.environ.get("S3_DOWNLOADER_BUCKET", None),
         reason="Export an env var called `S3_DOWNLOADER_BUCKET` containing the S3 bucket to run this test.",
     )
-
     @pytest.mark.integration
     @pytest.mark.skipif(
         not os.environ.get("S3_DOWNLOADER_BUCKET", None),


integrations/amazon_bedrock/src/haystack_integrations/components/downloaders/s3/s3_downloader.py
```diff
@@ -0,0 +1,405 @@
...
+            ]
+
+    def _build_source_documents(self, sources: List[Union[ByteStream, UUID, str]]) -> List[Document]:
+        """
+        Builds the documents from the provided sources.
+
+        :param sources: Contains the information about the files to download. Can include:
+            - str: S3 URL in the form `s3://bucket/key`.
+            - ByteStream: `meta` must contain `s3_url` or (`s3_bucket` + `s3_key`).
+            - `UUID`/other types: ignored with a warning.
```
Amnah199: S3 buckets can also be passed `http` urls but its not supported in the current implementation. We can decide if its important to implement that right away.

integrations/amazon_bedrock/examples/s3_downloader.py
```diff
@@ -0,0 +1,50 @@
...
+    ),
+    # Using s3_url
+    Document(
+        content="",
+        meta={
+            "s3_url": "s3://bucket_name/key_name/file_name.json",
+            "file_name": "opensearch-website.json",
+        },
+    ),
+]
```
anakin87: My impression is that supporting all of these options (`sources` + `documents` with different combinations of fields) could make this component less clear for users and harder to maintain for us.

I would at least simplify this for `Documents`. E.g., we might only support `s3_url`.
WDYT?

integrations/amazon_bedrock/src/haystack_integrations/components/downloaders/s3/s3_downloader.py
```diff
@@ -0,0 +1,405 @@
...
+            "AWS_SECRET_ACCESS_KEY", strict=False
+        ),
+        aws_session_token: Optional[Secret] = Secret.from_env_var("AWS_SESSION_TOKEN", strict=False),  # noqa: B008
+        aws_region_name: Optional[Secret] = Secret.from_env_var("AWS_DEFAULT_REGION", strict=False),  # noqa: B008
+        aws_profile_name: Optional[Secret] = Secret.from_env_var("AWS_PROFILE", strict=False),  # noqa: B008
+        verify_ssl: bool = True,
+        boto3_config: Optional[Dict[str, Any]] = None,
+        download_dir: Optional[str] = None,
+        file_extensions: Optional[List[str]] = None,
+        sources_target_type: Literal["str", "pathlib.Path", "haystack.dataclasses.ByteStream"] = "str",
```
anakin87: Could you please why we want/need to support different output types?

integrations/amazon_bedrock/src/haystack_integrations/components/downloaders/s3/s3_downloader.py
```diff
@@ -0,0 +1,405 @@
...
+        self.max_workers = max_workers
+        self.max_cache_size = max_cache_size
+
+        # Set up download directory
+        if download_dir:
+            self.download_dir = Path(download_dir)
+        else:
+            self.download_dir = Path.cwd() / "s3_downloads"
+        self.download_dir.mkdir(parents=True, exist_ok=True)
+        self.file_root_path: Path = self.download_dir
```
anakin87: I would remove `self.file_root_path`: it just seems a copy of `self.download_dir`

integrations/amazon_bedrock/src/haystack_integrations/components/downloaders/s3/s3_downloader.py
```diff
@@ -0,0 +1,405 @@
...
+        except Exception as exception:
+            msg = (
+                "Could not connect to Amazon S3. Make sure the AWS environment is configured correctly. "
+                "See https://boto3.amazonaws.com/v1/documentation/api/latest/guide/quickstart.html#configuration"
+            )
+            raise AmazonS3ConfigurationError(msg) from exception
+
+    def warm_up(self) -> None:
+        """Ensure local cache directory exists."""
+        self.file_root_path.mkdir(parents=True, exist_ok=True)
```
anakin87:
```suggestion
```

It seems that we already do this in `__init__`

integrations/amazon_bedrock/src/haystack_integrations/components/downloaders/s3/s3_downloader.py
```diff
@@ -0,0 +1,405 @@
...
+
+    @component.output_types(
+        documents=List[Document],
+        sources=List[Union[str, Path, ByteStream]],
+    )
```
sjrl:
```suggestion
    @component.output_types(documents=List[Document], sources=List[Union[str, Path, ByteStream]])
```

integrations/amazon_bedrock/src/haystack_integrations/components/downloaders/s3/s3_downloader.py
```diff
@@ -0,0 +1,405 @@
...
+
+        You can pass existing `Document`s (with S3 coordinates in `meta`) and/or `sources`
+        (e.g., `s3://...` URLs or `ByteStream` with S3 metadata). Files are cached locally.
+
+        :param documents: Documents to download. Each must include one of:
+            - `meta['s3_bucket']` + `meta['s3_key']`, or
+            - `meta['s3_url']` (e.g., `s3://bucket/key`), or
+            - `meta['s3_key']` (with default bucket provided elsewhere).
+            Non-conforming documents are skipped.
+        :param sources: Optional heterogeneous inputs used to build `Document`s:
```
sjrl: We mention that `sources` are optional so we should probably mention that `documents` are also optional.

integrations/amazon_bedrock/src/haystack_integrations/components/downloaders/s3/s3_downloader.py
```diff
@@ -0,0 +1,405 @@
...
+            - `ByteStream` whose `meta` contains `s3_url` or (`s3_bucket` + `s3_key`),
+            - Other types are ignored with a warning.
+        :returns: A dictionary with:
+            - `documents`: The downloaded `Document`s; each has `meta['file_path']` and may include `mime_type`.
+            - `sources`: The sources in the requested target type (`str`, `Path`, or `ByteStream`).
+        :raises AmazonS3Error: If a download attempt fails.
+        :raises ValueError: If an S3 URL in inputs is invalid.
+        """
+        # Ensure cache dir
+        self.warm_up()
```
sjrl: Hmm, we never call `warm_up` in the run method. We should instead follow our normal pattern of throwing an error if we detect that the component has not warmed up.

anakin87 (commented):
Posting my first few comments (more will come)

integrations/amazon_bedrock/src/haystack_integrations/components/downloaders/s3/s3_downloader.py
```diff
@@ -0,0 +1,405 @@
...
+
+    @component.output_types(
+        documents=List[Document],
+        sources=List[Union[str, Path, ByteStream]],
+    )
+    def run(
+        self,
+        documents: Optional[List[Document]] = None,
+        sources: Optional[List[Union[ByteStream, UUID, str]]] = None,
+    ) -> Dict[str, Union[List[Document], List[Union[str, Path, ByteStream]]]]:
```
sjrl: I have a general question about the sources field. What use case do you have in mind for adding it?

- I’m not sure it makes sense to support `ByteStream` here. Since a `ByteStream` already contains the file’s content as a bytes object, there’s no need to download anything again.
- I don’t think we currently do anything with UUID. The docstring also only mentions support for `ByteStream` and str.
- For the str type, it looks like we’re doing two things: (1) downloading the S3 URL and (2) converting it into a Document. That seems like we’re mixing two separate concerns: a downloader and a converter. If the goal is to handle S3 URLs and turn them into Documents, I’d suggest making that a dedicated component, called `S3ToDocument`

What do you think?

integrations/amazon_bedrock/src/haystack_integrations/components/downloaders/s3/s3_downloader.py
```diff
@@ -0,0 +1,405 @@
...
+        :param max_cache_size: The maximum number of files to cache.
+        :raises AmazonS3ConfigurationError: If the AWS configuration is not correct.
+        """
+
+        self.aws_access_key_id = aws_access_key_id
+        self.aws_secret_access_key = aws_secret_access_key
+        self.aws_region_name = aws_region_name
+        self.aws_session_token = aws_session_token
+        self.aws_profile_name = aws_profile_name
+        self.verify_ssl = verify_ssl
```
sjrl: I wasn’t able to find where `verify_ssl` is being used in the code. Could you point me to where it’s referenced or explain how it’s intended to be used?

integrations/amazon_bedrock/src/haystack_integrations/components/downloaders/s3/s3_downloader.py
```diff
@@ -0,0 +1,405 @@
...
+        print(f"- {d.meta.get('file_name')} -> {d.meta.get('file_path')}")
+
+    ```
+    """
+
+    max_cache_size: int = 100
+
+    def __init__(
+        self,
+        aws_access_key_id: Optional[Secret] = Secret.from_env_var("AWS_ACCESS_KEY_ID", strict=False),  # noqa: B008
```
sjrl: Let's make this keyword-only
```suggestion
    def __init__(
        self,
        *,
        aws_access_key_id: Optional[Secret] = Secret.from_env_var("AWS_ACCESS_KEY_ID", strict=False),  # noqa: B008
```

integrations/amazon_bedrock/src/haystack_integrations/components/downloaders/s3/s3_downloader.py
```diff
@@ -0,0 +1,405 @@
...
+        self.file_extensions = [e.lower() for e in file_extensions] if file_extensions else None
+        self.sources_target_type = sources_target_type
+        self.max_workers = max_workers
+        self.max_cache_size = max_cache_size
+
+        # Set up download directory
+        if download_dir:
+            self.download_dir = Path(download_dir)
+        else:
+            self.download_dir = Path.cwd() / "s3_downloads"
```
sjrl: I'm not sure we should use the current working directory as the default file path. It's not guaranteed in a production setting that the current working directory can be written to so this could cause issues. 

Also it's important the the file root path is configurable from the environment like the original `DeepsetFileDownloader`. So I think I'd prefer removing `download_dir` as a param, replace it with `file_root_path` and then allow `file_root_path` to be loaded from an env variable if it's not provided as an init param.

integrations/amazon_bedrock/src/haystack_integrations/components/downloaders/s3/s3_downloader.py
sjrl: Ahh I see this came from the original `DeepsetFileDownloader`. I'd still prefer this capability to be split up rather than combining them together like this.

integrations/amazon_bedrock/examples/s3_downloader.py
```diff
@@ -0,0 +1,50 @@
+import os
```
sjrl: Thanks for adding an example! I would like to expand it though to show when you would use this component in a pipeline. The major use case we are trying to enable is a Visual RAG QA pipeline in a production environment. 

So it would be great if we could use this tutorial as an example https://haystack.deepset.ai/tutorials/46_multimodal_rag and expand the query pipeline under the header **Multimodal RAG on Image + Text** to use this Downloader component.  The S3Downloader should probably come between `doc_type_router` and `doc_to_image` to ensure that the image file is available for `doc_to_image` to load it.

integrations/amazon_bedrock/examples/s3_downloader.py
sjrl: This comment is related https://github.com/deepset-ai/haystack-core-integrations/pull/2192#discussion_r2300658197

integrations/amazon_bedrock/src/haystack_integrations/components/downloaders/s3/s3_downloader.py
sjrl: This comment is related https://github.com/deepset-ai/haystack-core-integrations/pull/2192#discussion_r2300681938

integrations/amazon_bedrock/src/haystack_integrations/components/downloaders/s3/s3_downloader.py
```diff
@@ -0,0 +1,405 @@
...
+        """Download S3-linked files and return enriched `Document`s and source paths.
+
+        You can pass existing `Document`s (with S3 coordinates in `meta`) and/or `sources`
+        (e.g., `s3://...` URLs or `ByteStream` with S3 metadata). Files are cached locally.
+
+        :param documents: Documents to download. Each must include one of:
+            - `meta['s3_bucket']` + `meta['s3_key']`, or
+            - `meta['s3_url']` (e.g., `s3://bucket/key`), or
+            - `meta['s3_key']` (with default bucket provided elsewhere).
+            Non-conforming documents are skipped.
```
sjrl: I don't think I like this idea of requiring S3 information to be stored in the meta of the documents. At least from a deepset perspective it was important that this s3 location information is not this exposed to the user. 

I'd prefer to make some of this information more hidden. I see that in our `DeepsetFileDownloader` implementation that there were helper classes like IndexableStorage and FileDownloaderSettings. Did we consider also using those in this integration as well?

integrations/amazon_bedrock/src/haystack_integrations/components/downloaders/s3/s3_downloader.py
anakin87: So you are proposing to only accept `documents`? This would be a good simplification.

sjrl: > So you are proposing to only accept `documents`? This would be a good simplification.

Yes that would be my suggestion.

integrations/amazon_bedrock/src/haystack_integrations/components/downloaders/s3/s3_downloader.py
anakin87: The outputs of this component seem a bit complex and have the same name as the inputs.

What about just returning the downloaded files as a list of strings?
(I might miss some details)

integrations/amazon_bedrock/src/haystack_integrations/components/downloaders/s3/s3_downloader.py
Amnah199: I understand and agree. The idea was to implement a similar functionality as `DeepsetFileDownloader` as a starting point and refine it from there. @vblagoje  you can also share the feedback you have here.

integrations/amazon_bedrock/src/haystack_integrations/components/downloaders/s3/s3_downloader.py
sjrl: I think we need to return the documents since in a query pipeline we need the output of this component to feed into the document to image component. At the very least to make sure the doc to image component doesn't trigger until after all the downloads are finished.

integrations/amazon_bedrock/src/haystack_integrations/components/downloaders/s3/s3_downloader.py
vblagoje: Yes thanks Amna and I had a private exchange about this issue. Personally, this API is somewhat confusing to me but I was reluctant to press back because it conflicted with the guide that existing deepset downloader component design should be followed. It seems like we need to sync on what the priority is here and then we'll easily agree on the API finesse. Looking forward to our sync.

integrations/amazon_bedrock/src/haystack_integrations/components/downloaders/s3/s3_downloader.py
anakin87: Ok, I see. What about returning `documents` and `downloaded_files` (list of strings). Would this be problematic?

sjrl: I think ideally we either update the document's `file_path` meta field (or add a new one) that contains the full download path of this file. This way our DocumentToImageContent converter can figure out how to load these files.

integrations/amazon_bedrock/src/haystack_integrations/components/downloaders/s3/s3_downloader.py
Amnah199: Agreed! Implemented it through `S3Storage` class instead.

integrations/amazon_bedrock/src/haystack_integrations/components/downloaders/s3/utils.py
```diff
@@ -0,0 +1,80 @@
...
+    """This class provides a storage class for downloading indexable files from S3."""
+
+    def __init__(
+        self,
+
+        bucket_name: str,
+        s3_folder: Optional[str] = None,
+        endpoint_url: Optional[str] = None,
+        session: Optional[Session] = None,
+    ) -> None:
```
Amnah199: @tstadel @abrahamy
I’d like your input before moving forward for the below questions:

The current code uses `S3Storage` instead of the `IndexableStorage` class. That raises the question:

1.  Do you actually need `org_id` or `workspace_id` in this component for your use cases?

2. I noticed that `IndexableStorage` constructs S3 object keys by combining `org_id`, `workspace_id`, and the given `file_id` (doesnt apply to general cases without naming conventions)

What I’ve done instead:

- The current implementation just expects a bucket name (and optionally a folder name) from environment variables.
- To download multiple files from this folder or bucket, you can pass their `file_name` in `Document.meta`. (which means file_names are exposed)

They’ll then be downloaded and saved locally under the given `Document.meta.file_path`.

Also tagging @sjrl for feedback.

integrations/amazon_bedrock/src/haystack_integrations/components/downloaders/s3/utils.py
```diff
@@ -0,0 +1,80 @@
...
+from botocore.exceptions import ClientError
+import os
+from dataclasses import dataclass
+
+
+class StorageError(RuntimeError):
+    """This exception is raised when an error occurs while interacting with a storage backend."""
+
+
+class S3Storage():
```
sjrl:
```suggestion
class S3Storage:
```

integrations/amazon_bedrock/src/haystack_integrations/components/downloaders/s3/utils.py
```diff
@@ -0,0 +1,80 @@
...
+            if error_code == HTTPStatus.NOT_FOUND:
+                raise StorageError(
+                    f"The object {s3_key!r} does not exist in the S3 bucket {self.bucket_name!r}."
+                ) from e
+
+            raise StorageError(f"Failed to download file {s3_key!r} from S3: {e}") from e
+
+@dataclass
+class S3DownloaderSettings:
+    """Settings for the S3Downloader component."""
```
sjrl: Still understanding how this dataclass is used, but since it's public let's put it under a folder called `dataclasses` at the same level as `components` to follow the folder structure of Haystack

integrations/amazon_bedrock/src/haystack_integrations/components/downloaders/s3/utils.py
```diff
@@ -0,0 +1,80 @@
...
+        :raises ValueError: Raised if the S3_BUCKET environment variable is not provided.
+        :return: An instance of S3DownloaderSettings
+        :rtype: S3DownloaderSettings
+        """
+
+        if not (s3_bucket := os.getenv("S3_BUCKET")):
+            raise ValueError("S3_BUCKET environment variable not provided")
+
+        if not (s3_folder := os.getenv("S3_FOLDER")):
+            raise ValueError("S3_FOLDER environment variable not provided")
```
sjrl: Why do we raise a ValueError if `S3_FOLDER` is missing but we have it as optional in it's definition, `s3_folder: Optional[str] = None`?

integrations/amazon_bedrock/src/haystack_integrations/components/downloaders/s3/utils.py
```diff
@@ -0,0 +1,80 @@
...
+
+    s3_bucket: str
+    s3_folder: Optional[str] = None
+    aws_endpoint_url: Optional[str] = None
+
+    @classmethod
+    def from_env(cls) -> "S3DownloaderSettings":
+        """Create a S3DownloaderSettings object from environment variables.
+
+        :raises ValueError: Raised if the S3_BUCKET environment variable is not provided.
```
sjrl:
```suggestion
        :raises ValueError: Raised if the S3_BUCKET or the S3_FOLDER environment variable is not provided.
```

integrations/amazon_bedrock/src/haystack_integrations/components/downloaders/s3/utils.py
```diff
@@ -0,0 +1,80 @@
...
+from pathlib import Path
+from typing import Optional
+
+from boto3.session import Session
+from botocore.exceptions import ClientError
+import os
+from dataclasses import dataclass
+
+
+class StorageError(RuntimeError):
```
sjrl: Maybe we should rename this to `S3StorageError`?

integrations/amazon_bedrock/src/haystack_integrations/common/amazon_bedrock/errors.py
```diff
@@ -18,3 +18,11 @@ class AmazonBedrockConfigurationError(AmazonBedrockError):
 
 class AmazonBedrockInferenceError(AmazonBedrockError):
     """Exception for issues that occur in the Bedrock inference node"""
+
+
+class AmazonS3Error(Exception):
+    """Exception for issues that occur in the S3 node"""
```
sjrl: Could we be specific about which S3 component this refers to?
```suggestion
    """Exception for issues that occur in the S3 component"""
```

integrations/amazon_bedrock/src/haystack_integrations/common/amazon_bedrock/errors.py
```diff
@@ -18,3 +18,11 @@ class AmazonBedrockConfigurationError(AmazonBedrockError):
...
+
+
+class AmazonS3ConfigurationError(AmazonS3Error):
+    """Exception raised when AmazonS3 node is not configured correctly"""
```
sjrl: These two error classes probably should live in a file specific to s3 and not amazon_bedrock right? 

We could put it into `common/s3`

integrations/amazon_bedrock/src/haystack_integrations/components/downloaders/s3/utils.py
sjrl: Could we also group this error with our other errors related to S3 issues?

integrations/amazon_bedrock/src/haystack_integrations/components/downloaders/s3/s3_downloader.py
```diff
@@ -0,0 +1,239 @@
...
+                settings.s3_bucket,
+                settings.s3_folder,
+                endpoint_url=settings.aws_endpoint_url,
+                session=self.session,
+            )
+
+    @component.output_types(documents=List[Document])
+    def run(
+        self,
+        documents: Optional[List[Document]] = None,
```
sjrl: Shouldn't documents be required?

integrations/amazon_bedrock/src/haystack_integrations/components/downloaders/s3/s3_downloader.py
```diff
@@ -0,0 +1,239 @@
...
+    ) -> Dict[str, List[Document]]:
+        """Download S3-linked files and return enriched `Document`s.
```
sjrl: Could we explain a bit more how the S3 download works? What meta information from the Document is used and how are the returned documents enriched?

integrations/amazon_bedrock/src/haystack_integrations/components/downloaders/s3/utils.py
sjrl: @Amnah199 just to clarify how are you constructing the key (or fetching the key) based on the passed in Documents?

Amnah199: @sjrl The S3 key is being built like this:

```python
# Example of S3 key:
# s3_key = myfolder/test_file.json

os.environ["S3_BUCKET"]   # The target S3 bucket 
os.environ["S3_FOLDER"]   # The folder name in the bucket,  which is optional

docs = [
    Document(
        content="", 
        meta={
            "file_id": "550e8400-e29b-41d4-a716-446655440000",  # Unique ID of the saved file
            "file_name": "test_file.json",  # Name of the file to download
        },
    )
]
```
I feel this can be done in a better way, so open to feedback.

abrahamy: @Amnah199 Thanks for your contribution. To answer your questions:

1. For a generic implementation, `org_id` and `workspace_id` are not relevant. These only make sense in the context of deepset Platform.
2. Since this is in Haystack core integrations, you're correct in using the more generic `S3Storage` class.

integrations/amazon_bedrock/src/haystack_integrations/components/downloaders/s3/utils.py
```diff
@@ -0,0 +1,78 @@
...
+
+            if error_code == HTTPStatus.NOT_FOUND:
+                raise StorageError(
+                    f"The object {s3_key!r} does not exist in the S3 bucket {self.bucket_name!r}."
+                ) from e
+
+            raise StorageError(f"Failed to download file {s3_key!r} from S3: {e}") from e
+
+@dataclass
+class S3DownloaderSettings:
```
abrahamy: To keep things simpler, I'd just include `s3_bucket_name` and `aws_endpoint_url` in the component's constructor and require filenames to be the full name (including folder name). So for a file `a.txt` in the folder `example`, I'll just use `example/a.txt` as the file name.

integrations/amazon_bedrock/src/haystack_integrations/components/downloaders/s3/utils.py
tstadel: I agree with @abrahamy. However one more thing: Can we make the name of the env variables configurable or at least a bit more specific (e.g. `FILE_S3_BUCKET`)? 
Note that on platform pipelines are accessing multiple buckets, not only the file buckets, so `S3_BUCKET` would be too generic. Same applies for `S3_FOLDER`.

integrations/amazon_bedrock/src/haystack_integrations/components/downloaders/s3/utils.py
tstadel: @abrahamy that would be a bummer for platform as the `file_name` is typically just the file_name without its path. So we'd need to manipulate that in order to make use of this component. Here's another suggestion:
Instead of having a `s3_folder` (which does not include a trailing `/`), how about calling it `s3_prefix` and expecting users to include the trailing `/`?
That way, both would work:
- full specified `file_name` meta
- platform's path-stripped file_names + env variable based prefix

integrations/amazon_bedrock/src/haystack_integrations/components/downloaders/s3/s3_downloader.py
```diff
@@ -0,0 +1,239 @@
...
+    """
+    A component for downloading files from S3 to local filesystem with caching and concurrent downloads.
+    Supports filtering by file extensions and returns documents with file paths.
+
+    """
+
+    def __init__(
+        self,
+        *,
+        file_extensions: List[str],
```
tstadel: From actual platform usage, I would allow for `None` values here and make this the default: don't filter for extensions in that case.

integrations/amazon_bedrock/src/haystack_integrations/components/downloaders/s3/s3_downloader.py
tstadel: Agree, on platform we made it optional just because we also accept bytestreams as well. But this is not the case here. Let's make it non-optional.

integrations/amazon_bedrock/src/haystack_integrations/components/downloaders/s3/s3_downloader.py
```diff
@@ -0,0 +1,239 @@
...
+                """Call "warm_up()" before calling "download_file()"."""
+            )
+
+        try:
+            file_id = UUID(document.meta["file_id"])
+        except KeyError:
+            logger.warning(f"Document with ID {document.id!r} does not have a file_id in the meta field", exc_info=True)
+            return None
+
+        file_name = Path(document.meta.get("file_name", ""))
```
tstadel: just a nit. But maybe worth the little tweak to make this more generic: how about making this meta key configurable?

tstadel (commented):
I left some smaller comments. Pls have a look. Ditching ORG_ID and WORKSPACE_ID is not a bummer.

integrations/amazon_bedrock/src/haystack_integrations/components/downloaders/s3/utils.py
Amnah199: Removed it as we dont actually need it. The `S3Storage` class is now configured through a class method instead.

integrations/amazon_bedrock/examples/s3_downloader_example.py
```diff
@@ -0,0 +1,82 @@
+# To run this example, you will need to
+# 1) set `AWS_ACCESS_KEY_ID`, `AWS_SECRET_ACCESS_KEY` and `AWS_DEFAULT_REGION` environment variables
+# 2) enabled access to the selected S3 bucket
+# 3) `S3_DOWNLOADER_BUCKET` environment variable should be set to the name of the S3 bucket.
+
+# The example shows how to use the S3Downloader component in a query pipeline to download files from an S3 bucket.
+# The files are then converted to images and used to answer a question.
```
sjrl: For this example to work do we also need the two files you mention on lines 23-24?

integrations/amazon_bedrock/src/haystack_integrations/components/downloaders/s3/s3_downloader.py
```diff
@@ -0,0 +1,235 @@
...
+        aws_secret_access_key: Optional[Secret] = Secret.from_env_var(  # noqa: B008
+            "AWS_SECRET_ACCESS_KEY", strict=False
+        ),
+        aws_session_token: Optional[Secret] = Secret.from_env_var("AWS_SESSION_TOKEN", strict=False),  # noqa: B008
+        aws_region_name: Optional[Secret] = Secret.from_env_var("AWS_DEFAULT_REGION", strict=False),  # noqa: B008
+        aws_profile_name: Optional[Secret] = Secret.from_env_var("AWS_PROFILE", strict=False),  # noqa: B008
+        boto3_config: Optional[Dict[str, Any]] = None,
+        file_root_path: Optional[str] = None,
+        file_extensions: Optional[List[str]] = None,
+        download_file_meta_key: str = "file_name",
```
sjrl: Perhaps we could call this `file_name_meta_key` instead? I think it better reflects your doc string explanation.

integrations/amazon_bedrock/src/haystack_integrations/components/downloaders/s3/s3_downloader.py
```diff
@@ -0,0 +1,235 @@
...
+        constructor. Three required parameters are `aws_access_key_id`, `aws_secret_access_key`,
+        and `aws_region_name`.
+
+        :param aws_access_key_id: AWS access key ID.
+        :param aws_secret_access_key: AWS secret access key.
+        :param aws_session_token: AWS session token.
+        :param aws_region_name: AWS region name.
+        :param aws_profile_name: AWS profile name.
+        :param boto3_config: The configuration for the boto3 client.
+        :param file_root_path: The path where the file will be downloaded.
```
sjrl, line 60: Let's mention in the docstring what happens if this is left as None and that it can be loaded from an env variable

integrations/amazon_bedrock/src/haystack_integrations/components/downloaders/s3/s3_downloader.py
```diff
@@ -0,0 +1,235 @@
...
+        :param max_cache_size: The maximum number of files to cache.
+        :param download_file_meta_key: The name of the meta key that contains the file name to download.
+            By default, the Document.meta["file_name"] is used. If you want to use a
+            different key in Document.meta, you can set it here.
+        :raises S3ConfigurationError: If the AWS environment is not configured correctly
+            to connect to the S3 bucket.
+        """
+
+        # Set up download directory
+        file_root_path = file_root_path or os.getenv("FILE_ROOT_PATH")
```
sjrl, line 76: As Thomas mentioned let's perhaps call this `S3_FILE_ROOT_PATH` otherwise `FILE_ROOT_PATH` is a little too generic

sjrl: Ahh sorry never mind just properly realized this is meant to be the directory of where to store the files once downloaded and nothing to do with S3

integrations/amazon_bedrock/src/haystack_integrations/components/downloaders/s3/s3_downloader.py
```diff
@@ -0,0 +1,235 @@
...
+
+        else:
+            self._storage.download(key=file_name, local_file_path=file_path)
+
+        document.meta["file_path"] = str(file_path)
+        return document
+
+    def cleanup_cache(self, documents: List[Document]) -> None:
+        """
+        Remove least-recently-accessed cache files when cache exceeds max_cache_size.
```
sjrl: let's add a docstring for `documents`

integrations/amazon_bedrock/src/haystack_integrations/components/downloaders/s3/s3_downloader.py
```diff
@@ -0,0 +1,235 @@
...
+
+        # cache budget = requested + misses; trim oldest misses if overflow
+        overflow = len(misses) + len(requested_ids) - self.max_cache_size
+        if overflow > 0:
+            misses.sort(key=lambda p: p.stat().st_atime)
+            for p in misses[:overflow]:
+                try:
+                    p.unlink()
+                except Exception:
+                    logger.warning(f"Failed to remove cache file {p}")
```
sjrl: Let's avoid f-strings in logging messages and do
```suggestion
                    logger.warning("Failed to remove cache file {path}", path=p)
```

integrations/amazon_bedrock/src/haystack_integrations/components/downloaders/s3/s3_downloader.py
```diff
@@ -0,0 +1,235 @@
...
+
+    def _filter_documents_by_extensions(self, documents: List[Document]) -> List[Document]:
+        """Filter documents by file extensions."""
+        return [
+            doc
+            for doc in documents
+            if Path(doc.meta.get(self.download_file_meta_key, "")).suffix.lower() in self.file_extensions
+        ]
+
+    def download_file(self, document: Document) -> Optional[Document]:
```
sjrl: I wonder whether we should make `download_file` and `cleanup_cache` private methods so we can be less worried about making breaking changes to them in the future. What do you think?

integrations/amazon_bedrock/src/haystack_integrations/components/downloaders/s3/s3_downloader.py
Amnah199: Sure, I wasnt able to find a good name for this parameter. I'll update.

integrations/amazon_bedrock/src/haystack_integrations/components/downloaders/s3/__init__.py
```diff
@@ -0,0 +1,7 @@
+# SPDX-FileCopyrightText: 2023-present deepset GmbH <info@deepset.ai>
```
sjrl, line 1: could we add an empty file called `py.typed` within the `downloaders` folder like we do for `embedders` and `generators`. This tells mypy and other type checkers that our files are typed

integrations/amazon_bedrock/src/haystack_integrations/components/downloaders/s3/s3_downloader.py
Amnah199: Agreed!

integrations/amazon_bedrock/src/haystack_integrations/components/downloaders/s3/s3_downloader.py
```diff
@@ -0,0 +1,235 @@
...
+    def run(
+        self,
+        documents: List[Document],
+    ) -> Dict[str, List[Document]]:
+        """Download files from AWS S3 Buckets to local filesystem.
+
+        Return enriched `Document`s with the path of the downloaded file.
+        :param documents: Document containing the name of the file to download in the meta field.
+        :returns: A dictionary with:
+            - `documents`: The downloaded `Document`s; each has `meta['file_path']` and may include `mime_type`.
```
sjrl: What do you mean here by may include `mime_type`?

integrations/amazon_bedrock/src/haystack_integrations/common/s3/errors.py
```diff
@@ -0,0 +1,10 @@
+class S3Error(Exception):
```
sjrl, line 6: I think we are missing the license header

integrations/amazon_bedrock/src/haystack_integrations/common/s3/utils.py
```diff
@@ -0,0 +1,99 @@
+import os
```
sjrl, line 5: Missing license header

integrations/amazon_bedrock/src/haystack_integrations/common/s3/errors.py
sjrl: Seems we are missing them in the `common/amazon_bedrock` files as well. Could you add them there too?

integrations/amazon_bedrock/src/haystack_integrations/common/s3/utils.py
```diff
@@ -0,0 +1,99 @@
...
+        self.file_prefix = file_prefix
+        self.endpoint_url = endpoint_url
+        self.session = session
+        self.config = config
+
+        try:
+            self._client = self.session.client("s3", endpoint_url=self.endpoint_url, config=self.config)
+        except Exception as e:
+            msg = f"Failed to create S3 session client: {e}"
+            raise S3ConfigurationError(msg) from e
```
Amnah199, line 54: @sjrl thanks for the review. There is one issue I cant understand. Even when I haven't set AWS keys, I dont get a `S3ConfigurationError` instead I get the error on line 78 which I find misleading.

`"Failed to download file {s3_key!r} from S3: {e}"`

I am wondering why `session.client`  isn't raising an exception.

integrations/amazon_bedrock/src/haystack_integrations/components/downloaders/s3/s3_downloader.py
Amnah199: Sorry thats a mistake.

integrations/amazon_bedrock/src/haystack_integrations/common/s3/utils.py
```diff
@@ -0,0 +1,103 @@
...
+
+@dataclass
+class S3Storage:
+    """This class provides a storage class for downloading files from an AWS S3 bucket."""
+
+    def __init__(
+        self,
+        s3_bucket: str,
+        session: Session,
+        file_prefix: Optional[str] = None,
```
sjrl: Could we call this `s3_prefix` instead like we call it in the `from_env` method?

integrations/amazon_bedrock/src/haystack_integrations/common/s3/utils.py
sjrl: Interesting, it seems that is the case. Looking at the source code of `sessiong.create_client` it seems like it should raise a `PartialCredentialsError` if some credentials are missing.

sjrl: Actually I think I might understand why. The init of this class requires a `Session` instance. It looks like creating a session instance also requires credentials and if they are missing it will throw an error. So my guess is if you have already successfully created a Session instance then the credential keys must already be available so perhaps the credential check doesn't work as expected at this step.

sjrl: Yeah confirmed by their docstrings 
>         :param aws_access_key_id: The access key to use when creating
>            the client.  This is entirely optional, and if not provided,
>            the credentials configured for the session will automatically
>            be used.  You only need to provide this argument if you want
>            to override the credentials used for this specific client.

integrations/amazon_bedrock/src/haystack_integrations/components/downloaders/s3/s3_downloader.py
```diff
@@ -0,0 +1,239 @@
...
+            The same `Document` with `meta` containing the `file_path` of the
+            downloaded file.
+        :raises S3Error: If the download or head request fails or the file does not exist in the S3 bucket.
+        """
+        if self._storage is None:
+            msg = (
+                f"The component {self.__class__.__name__} was not warmed up."
+                f" Call 'warm_up()' before calling 'download_file()'."
+            )
+            raise RuntimeError(msg)
```
sjrl: Is there a reason to put this in `_download_file` rather than in the `run` method? Seems unnecessary to perform this check every time `_download_file` is called

integrations/amazon_bedrock/src/haystack_integrations/components/downloaders/s3/s3_downloader.py
```diff
@@ -0,0 +1,239 @@
...
+        misses = [p for p in all_files if p.stem not in requested_ids]
+
+        overflow = len(misses) + len(requested_ids) - self.max_cache_size
+        if overflow > 0:
+            misses.sort(key=lambda p: p.stat().st_atime)
+            for p in misses[:overflow]:
+                try:
+                    p.unlink()
+                except Exception:
+                    logger.warning("Failed to remove cache file {path}", path=p)
```
sjrl: Could be good to also provide the error as part of the log. So something like 
```suggestion
                except Exception as error:
                    logger.warning("Failed to remove cache file at {path} with error: {e}", path=p, e=error)
```

sjrl (approved):
Just a few remaining comments, otherwise looks good!

integrations/amazon_bedrock/src/haystack_integrations/common/s3/utils.py
Amnah199: @sjrl thanks for this. After wasting a lot of time I realized I had some old credentials stored in `~/.aws/`. Because of this my local tests were not signaling for missing credentials.
Thanks for your help though! I added the check for credentials error specifically.

PR #2200, opened 2025-08-20
refactor!: `OllamaChatGenerator` - refine reasoning support + refactoring

### LINKED ISSUE #2175: Ollama - refine reasoning support

Reasoning already supported but we store it in `meta`.

anakin87:
### Related Issues

- fixes #2175 

### Proposed Changes:
- reasoning refinements
  - store reasoning in `ChatMessage` content instead of `meta`
  - changes to make reasoning be correctly returned when using streaming
  - support different thinking levels (for GPT-oss)

- refactoring: since we remove reasoning from `meta`, this is breaking change and we need to release a major version, so I took the opportunity to do some refactoring
  - change default model in Chat Generator: `orca-mini` is very outdated, while `qwen3:0.6b` is a small model, capable of tool calling and reasoning
  - refactor Chat Generator tests: better structure, parametrized several tests to reduce duplication
  - 5x smaller embedding model for tests: reduces download and inference times

### How did you test it?
CI

I have manually verified and written tests for:
- reasoning with multi-turn conversations
- reasoning with streaming
- reasoning with tool calls

### Checklist


integrations/ollama/src/haystack_integrations/components/generators/ollama/chat/chat_generator.py
```diff
@@ -131,81 +133,84 @@ def _convert_ollama_response_to_chatmessage(ollama_response: ChatResponse) -> Ch
...
+    finish_reason = FINISH_REASON_MAPPING.get(chunk_response.done_reason or "")
+    tool_calls_list = []
+
+    content = chunk_response_dict["message"]["content"]
+
+    meta = {key: value for key, value in chunk_response_dict.items() if key != "message"}
+    meta["role"] = chunk_response_dict["message"]["role"]
+
+    # until a specific field in StreamingChunk is available, we store the thinking in the meta
+    meta["reasoning"] = chunk_response_dict["message"].get("thinking", None)
```
anakin87, line 161: We temporarily store reasoning chunks in `StreamingChunk.meta`. When building the final `ChatMessage`, reasoning is removed from `meta`.

I might create an issue about a `reasoning` field in `StreamingChunk` and modifying `print_streaming_chunk` to print it.

sjrl: > I might create an issue about a reasoning field in StreamingChunk and modifying print_streaming_chunk to print it.

@anakin87 yes I agree that sounds like a good idea

anakin87: https://github.com/deepset-ai/haystack/issues/9728

integrations/ollama/src/haystack_integrations/components/generators/ollama/chat/chat_generator.py
```diff
@@ -131,81 +133,84 @@ def _convert_ollama_response_to_chatmessage(ollama_response: ChatResponse) -> Ch
...
+                )
+            )
+
+    return StreamingChunk(
+        content=content,
+        meta=meta,
+        index=index,
+        finish_reason=finish_reason,
+        component_info=component_info,
+        tool_calls=tool_calls_list,
```
sjrl, line 181: I realized we aren't utilizing the `start` in StreamingChunk. Any reason for that?

integrations/ollama/src/haystack_integrations/components/generators/ollama/chat/chat_generator.py
```diff
@@ -219,9 +224,11 @@ def __init__(
         :param timeout:
             The number of seconds before throwing a timeout error from the Ollama API.
         :param think
-            If True, the modell will "think" before producing a response.
+            If True, the model will "think" before producing a response.
             Only [thinking models](https://ollama.com/search?c=thinking) support this feature.
-            The intermediate "thinking" output can be found in the `meta` property of the returned `ChatMessage`.
+            Some models like gpt-oss support different levels of thinking: "low", "medium", "high".
+            The intermediate "thinking" output can be found by inspecting the `reasoning` property of the returned
+            `ChatMessage`.
```
sjrl, line 231: If a user specifieis gpt-oss and thinking=True what will happen?

integrations/ollama/src/haystack_integrations/components/generators/ollama/chat/chat_generator.py
sjrl: I see that we set `start` in `_handle_streaming_response` but only for `tool_calls`. It should also be set to True for the beginnings of normal text blocks and I can't immediately tell if that's happening or not.

integrations/ollama/tests/test_chat_generator.py
```diff
@@ -12,193 +14,210 @@
...
+        observed = _convert_ollama_response_to_chatmessage(ollama_response)
+
+        assert observed.role == "assistant"
+        assert observed.text is None
+        assert observed.tool_call == ToolCall(
+            tool_name="get_current_weather",
+            arguments={"format": "celsius", "location": "Paris, FR"},
+        )
+
+    def test_build_chunk(self):
```
sjrl, line 192: I realize this was missed in a previous PR but could we add a test like https://github.com/deepset-ai/haystack-core-integrations/blob/2fc25e2937f820e746b561590efd27d56d2b1497/integrations/amazon_bedrock/tests/test_chat_generator_utils.py#L611 that also includes reasoning content + a tool call. So similar to `test_handle_streaming_response_tool_calls` in this file. I'm realizing ideally we'd also test each streaming chunk individually and not just the final ChatMessage like I show in the linked tests.

integrations/ollama/src/haystack_integrations/components/generators/ollama/chat/chat_generator.py
anakin87: It works.

It's a bit hard to tell what happens.
Inspecting the [Ollama template](https://ollama.com/library/gpt-oss:20b/blobs/fa6710a93d78), it seems that the default level is medium.

When I set `think=true`, I get an answer which is very similar to `think=medium`, but not exactly the same.

sjrl: hmm okay, I guess fine to leave as-is then. If it was clearer marking a place in ollama docs what would happen would be nice, but if it works and it's not well documented I wouldn't worry about it.

integrations/ollama/src/haystack_integrations/components/generators/ollama/chat/chat_generator.py
anakin87: I now set `start`

integrations/ollama/tests/test_chat_generator.py
anakin87: added `test_handle_streaming_response_tool_calls_with_thinking` but also adjusted `test_handle_streaming_response` and `test_handle_streaming_response_tool_calls` to check the individual chunks

sjrl (approved):
Looks good!

PR #2207, opened 2025-08-21
fix: Avoid mixed Langfuse traces in async envs

vblagoje:
## Why:
**Purpose:** Fixes a race condition where spans from concurrent pipeline runs become intertwined, corrupting tracing data and making debugging unreliable in production web server environments.

- fixes https://github.com/deepset-ai/haystack-core-integrations/issues/2140

## What:
- Replaced instance-level `self._context` list with `ContextVar` for thread-safe span stack management
- Updated `LangfuseTracer.trace()` method to use context-local span storage
- Modified `current_span()` method to read from context-local state
- Added module-level `span_context_var` ContextVar for isolated span contexts per execution thread
- Fixed support for session IDs as well

## How can it be used:
Nothing changes in client code except that we don't corrupt traces in concurrent async pipelines. All existing usage patterns continue to work exactly as before, but now with proper thread safety.

## How did you test it:
- Added unit tests verifying context isolation between concurrent tracer instances
- Verified that spans are properly isolated per execution context
- Tested examples in examples directory and itinerary agent
- Ran existing test suite to ensure no regressions in single-threaded scenarios

### DIFF
--- a/integrations/langfuse/tests/test_tracer.py
+++ b/integrations/langfuse/tests/test_tracer.py
@@ -410,61 +410,7 @@ class TestLangfuseTracer:
 
             LangfuseTracer(tracer=MockTracer(), name="Haystack", public=False)
             assert "tracing is disabled" in caplog.text
-
-    def test_context_cleanup_after_nested_failures(self):
-        """
-        Test that tracer context is properly cleaned up even when nested operations fail.
-
-        This test addresses a critical bug where failing nested operations (like inner pipelines)
-        could corrupt the tracing context, leaving stale spans that affect subsequent operations.
-        The fix ensures proper cleanup through try/finally blocks.
-
-        Before the fix: context would retain spans after failures (length > 0)
-        After the fix: context is always cleaned up (length == 0)
-        """
-
-        @component
-        class FailingParser:
-            @component.output_types(result=str)
-            def run(self, data: str):
-                # This will fail with ValueError when data is not valid JSON
-                parsed = json.loads(data)
-                return {"result": parsed["key"]}
-
-        @component
-        class ComponentWithNestedPipeline:
-            def __init__(self):
-                # This simulates IntentClassifier's internal pipeline
-                self.internal_pipeline = Pipeline()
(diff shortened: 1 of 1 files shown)


vblagoje:
Replacing https://github.com/deepset-ai/haystack-core-integrations/pull/2188

integrations/langfuse/tests/test_tracer.py
```diff
@@ -400,3 +410,130 @@ def test_init_with_tracing_disabled(self, monkeypatch, caplog):
 
             LangfuseTracer(tracer=MockTracer(), name="Haystack", public=False)
             assert "tracing is disabled" in caplog.text
+
+    def test_context_cleanup_after_nested_failures(self):
```
anakin87: I recently moved this to test_tracing.py in #2203 and marked as an integration test (requiring API keys)
I'm also OK to moving back to this module and transforming it into a unit test. The important thing is not to have duplicates.

vblagoje: Aha, I can remove it here then, np!

integrations/langfuse/src/haystack_integrations/tracing/langfuse/tracer.py
```diff
@@ -414,10 +425,8 @@ def trace(
                     cleanup_error=cleanup_error,
                 )
             finally:
-                # CRITICAL: Always pop context to prevent corruption
-                # This is especially important for nested pipeline scenarios
-                if self._context and self._context[-1] == span:
-                    self._context.pop()
+                # Restore previous span stack using saved token - ensures proper cleanup
+                span_stack_var.reset(token)
```
mpangrazzi, line 429: Can we safely assume this line will never throw an error?

vblagoje: No mention of it in pydocs anywhere. Seems like it can't happen

vblagoje:
Fixed and released in https://pypi.org/project/langfuse-haystack/2.3.0/

PR #2220, opened 2025-08-25
feat: Add prompt routing to `AmazonBedrockChatGenerator`

### LINKED ISSUE #2167: Support AWS Bedrock Prompt Router Invocation in Generators
Yes. Currently, generators that interact with AWS Bedrock cannot invoke prompt router ARNs. Bedrock prompt routers provide a unified way to route prompts to different models (e.g., faster/cheaper vs. more powerful/expensive) using a single ARN. 

I’d like support for invoking AWS Bedrock prompt routers within generators. This would enable more flexible, centralized model routing logic without modifying the generator logic or hardcoding individual model ARNs. In a pipeline you'd just invoke a single ARN (the one for the router) and it would handle the decision under the hood and return the response.

**Additional context**
Bedrock prompt routers allow users to define custom routing logic (e.g., start with Claude for general Q&A, fall back to Llama for code-heavy prompts) under a single ARN. Supporting prompt routers would simplify pipeline configurations and allow users to experiment with dynamic model selection strategies. 

Prompt routing in Amazon Bedrock
: https://docs.aws.amazon.com/bedrock/latest/userguide/prompt-routing.html

Amnah199:
### Related Issues

- fixes #2167 

### Proposed Changes:

- Allow passing `prompt_router_arn` during initialization
- Make model param optional (shouldn't be a breaking change)
- Add an example

### How did you test it?

- Update unit tests
### Checklist


integrations/amazon_bedrock/src/haystack_integrations/components/generators/amazon_bedrock/chat/chat_generator.py
```diff
@@ -158,6 +161,7 @@ def __init__(
         streaming_callback: Optional[StreamingCallbackT] = None,
         boto3_config: Optional[Dict[str, Any]] = None,
         tools: Optional[Union[List[Tool], Toolset]] = None,
+        prompt_router_config: Optional[Dict[str, Any]] = None,
```
sjrl: Let's start enforcing kwargs
```suggestion
        tools: Optional[Union[List[Tool], Toolset]] = None,
        *,
        prompt_router_config: Optional[Dict[str, Any]] = None,
```

integrations/amazon_bedrock/src/haystack_integrations/components/generators/amazon_bedrock/chat/chat_generator.py
```diff
@@ -190,14 +195,22 @@ def __init__(
...
+        [bedrock documentation](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/bedrock/client/create_prompt_router.html).
+        Must be None if `model` is provided.
+        :raises ValueError: If the both `model` and `prompt_router_config` are provided or if both are None.
         :raises AmazonBedrockConfigurationError: If the AWS environment is not configured correctly or the model is
             not supported.
         """
-        if not model:
-            msg = "'model' cannot be None or empty string"
+        if not model and not prompt_router_config:
+            msg = "'model' can be None or empty string, only if 'prompt_router_config' is provided."
```
sjrl:
```suggestion
            msg = "'model' can be None or an empty string, only if 'prompt_router_config' is provided."
```

integrations/amazon_bedrock/src/haystack_integrations/components/generators/amazon_bedrock/chat/chat_generator.py
```diff
@@ -229,6 +243,24 @@ def resolve_secret(secret: Optional[Secret]) -> Optional[str]:
 
             self.client = session.client("bedrock-runtime", config=config)
 
+            if self.prompt_router_config:
+                bedrock_client = session.client("bedrock", config=config)
+
+                # check if prompt router already exists
+                router_name = self.prompt_router_config["promptRouterName"]
+                existing_routers = bedrock_client.list_prompt_routers(type="custom")["promptRouterSummaries"]
```
sjrl: Are there other types of prompt routers available than just `"custom"`?

integrations/amazon_bedrock/src/haystack_integrations/components/generators/amazon_bedrock/chat/chat_generator.py
```diff
@@ -229,6 +243,24 @@ def resolve_secret(secret: Optional[Secret]) -> Optional[str]:
...
+                for router in existing_routers:
+                    if router["promptRouterName"] == router_name:
+                        logger.warning(
+                            f'Prompt router with name "{router_name}" already exists. Fetching existing router.'
+                        )
+                        config_response = bedrock_client.get_prompt_router(promptRouterArn=router["promptRouterArn"])
+                        break
+                else:
+                    config_response = bedrock_client.create_prompt_router(**self.prompt_router_config)
```
sjrl: I think this else is wrongly indented
```suggestion
                    else:
                        config_response = bedrock_client.create_prompt_router(**self.prompt_router_config)
```

sjrl:
Hey @medsriha it would be great if you could check if this solution works for your use case!

integrations/amazon_bedrock/src/haystack_integrations/components/generators/amazon_bedrock/chat/chat_generator.py
Amnah199: Good point. There are also default prompt routers. I'll add them to the check.

sjrl:
@Amnah199 Taking a look at the code you added it seems like we are enabling two features here. 

1. Allow a user to create a new PromptRouter in Bedrock by passing a `prompt_router_config`
2. Allow a user to use an existing PromptRouter config in Bedrock by basically finding an already existing `promptRouterArn`. You do this currently by doing a lookup and comparing the name provided in `prompt_router_config`. 

Looking at the issue originally opened by @medsriha it seems like we really only need to support the second case currently. Basically allow a user to pass a `promptRouterArn` as input. I'm not sure if we need the fully prompt router creation to be added currently. WDYT?

integrations/amazon_bedrock/tests/test_chat_generator.py
```diff
@@ -137,6 +137,50 @@ def test_to_dict(self, mock_boto3_session, boto3_config):
...
+        }
+
+        assert generator.to_dict() == expected_dict
+
+    @pytest.mark.parametrize("boto3_config", [None, {"read_timeout": 1000}])
+    def test_to_dict_with_prompt_router_config(self, mock_boto3_session, boto3_config):
+        """
+        Test that the to_dict method returns the correct dictionary without aws credentials
+        """
+        generator = AmazonBedrockChatGenerator(
```
sjrl: In addition to this test it would be good to add an integration test where we actually run the generator with the prompt router and check that it returns the expected output.

integrations/amazon_bedrock/src/haystack_integrations/components/generators/amazon_bedrock/chat/chat_generator.py
```diff
@@ -228,6 +243,13 @@ def resolve_secret(secret: Optional[Secret]) -> Optional[str]:
             )
 
             self.client = session.client("bedrock-runtime", config=config)
+            if self.prompt_router_arn is not None:
+                resolved_router_arn = resolve_secret(self.prompt_router_arn)
+                bedrock_client = session.client("bedrock", config=config)
+                prompt_router = bedrock_client.get_prompt_router(
+                    promptRouterArn=resolved_router_arn,
+                )
```
Amnah199, line 251: @sjrl I believe we can show a specific error message in case of invalid ARN. If you agree I can raise an exception here with the message.

sjrl: Yeah that could be helpful. What did you have in mind?

integrations/amazon_bedrock/src/haystack_integrations/components/generators/amazon_bedrock/chat/chat_generator.py
```diff
@@ -228,6 +243,13 @@ def resolve_secret(secret: Optional[Secret]) -> Optional[str]:
...
+                self.model = prompt_router["promptRouterArn"]
```
sjrl, line 252: Just to clarify, is this value `prompt_router["promptRouterArn"]` the same as the one provided by the user so `prompt_router_arn` at init time?

integrations/amazon_bedrock/src/haystack_integrations/components/generators/amazon_bedrock/chat/chat_generator.py
```diff
@@ -158,6 +163,8 @@ def __init__(
         streaming_callback: Optional[StreamingCallbackT] = None,
         boto3_config: Optional[Dict[str, Any]] = None,
         tools: Optional[Union[List[Tool], Toolset]] = None,
+        *,
+        prompt_router_arn: Optional[Secret] = None,
```
sjrl, line 167: Just to double check, does `prompt_router_arn` need to be a Secret? What was the motivation to make it a secret?

integrations/amazon_bedrock/src/haystack_integrations/components/generators/amazon_bedrock/chat/chat_generator.py
Amnah199: Yes it would be the same. Here we verify if it actually is a valid ARN.

integrations/amazon_bedrock/src/haystack_integrations/components/generators/amazon_bedrock/chat/chat_generator.py
Amnah199: The custom router ARN contains your region and AWS account ID, which probably we dont want to expose.

integrations/amazon_bedrock/src/haystack_integrations/components/generators/amazon_bedrock/chat/chat_generator.py
sjrl: Okay I see. Perhaps I could propose an alternative approach. If `self.prompt_router_arn` works in the same way as we treat `self.model` do we really need to make a new variable?

Would specifying an `prompt_router_arn` in the `model` field at init time already work with the existing integration?

Amnah199:
Closing: The original issue is no longer valid.

PR #2224, opened 2025-08-28
Backsync new changes from main repo

nh-lh:
### Related Issues

- fixes #issue-number

### Checklist


nh-lh:
opened by mistake

integrations/elasticsearch/src/haystack_integrations/document_stores/elasticsearch/document_store.py
```diff
@@ -63,6 +64,12 @@ def __init__(
         hosts: Optional[Hosts] = None,
         custom_mapping: Optional[Dict[str, Any]] = None,
         index: str = "default",
+        api_key: Secret = Secret.from_env_var(
```
davidsbatista, line 67: this should `Optional` since it's not a mandatory  parameter.

integrations/elasticsearch/src/haystack_integrations/document_stores/elasticsearch/document_store.py
```diff
@@ -63,6 +64,12 @@ def __init__(
...
+            "DOCUMENT_STORE_PARAMS_API_KEY", strict=True
+        ),
+        api_key_id: Secret = Secret.from_env_var(
```
davidsbatista, line 70: the same thing here, it's not a mandatory parameter.

integrations/elasticsearch/src/haystack_integrations/document_stores/elasticsearch/document_store.py
```diff
@@ -114,11 +123,13 @@ def _ensure_initialized(self):
             # Initialize both sync and async clients
             self._client = Elasticsearch(
                 self._hosts,
+                api_key=(self.api_key_id.resolve_value(), self.api_key.resolve_value()),
```
davidsbatista, line 126: Since the `api_key` can accept either a single `str` or a `Tuple[str,str]` we need to make a check here to avoid having a tuple `(str, None)`. 

https://elasticsearch-py.readthedocs.io/en/v9.1.1/api/elasticsearch.html

As far as I understood a single string is when the user passes a base64-encoded string that encodes both id and secret (separated by “:”)

```python
api_key = None

# supply both the id of the key and the secret.
if self.api_key_id and self.api_key:
    api_key = (self.api_key_id.resolve_value(), self.api_key.resolve_value())
# a base64-encoded string that encodes both id and secret (separated by “:”)
if self.api_key:
    api_key = self.api_key.resolve_value()
# if you give only the id raise an error
if self.api_key_id and not self.api_key:
    raise ValueError("api_key_id is provided but api_key is missing.")
```

integrations/elasticsearch/src/haystack_integrations/document_stores/elasticsearch/document_store.py
```diff
@@ -114,11 +123,13 @@ def _ensure_initialized(self):
...
                 headers=headers,
                 **self._kwargs,
             )
             self._async_client = AsyncElasticsearch(
                 self._hosts,
+                api_key=(self.api_key_id.resolve_value(), self.api_key.resolve_value()),
```
davidsbatista, line 132: same thing as above

nh-lh:
Thank you @davidsbatista for taking over here, much appreciated

davidsbatista:
@nh-lh let me know if you run into any issues when using it.

PR #2228, opened 2025-08-28
feat: Adding optional keepalive parameter to Ollama embedders

### LINKED ISSUE #2225: Support `keep_alive` for OllamaTextEmbedder
When using OllamaTextEmbedder, there is no control over how long the model stays loaded.

When using OllamaTextEmbedder, I would like to specify how long the model stays loaded (just like I can with OllamaGenerator and OllamaChatGenerator).

The only alternative is to set the `OLLAMA_KEEP_ALIVE` environment variable, but it's not a reasonable solution to this problem since it sets the default for all models.

**Additional context**
I modified my local copy of [text_embedder.py](https://github.com/deepset-ai/haystack-core-integrations/blob/bc4a862506cb8dc18d9a7855a49330d996cb1b70/integrations/ollama/src/haystack_integrations/components/embedders/ollama/text_embedder.py) by adding in the `keep_alive` parameter (the underlying call to [Client.embeddings()](https://github.com/ollama/ollama-python/blob/da79e987f0ac0a4986bf396f043b36ef840370bc/ollama/_client.py#L385) already supports it).  After calling OllamaTextEmbedder with the new `keep_alive` parameter in my own code, I was able to confirm (executing `ollama ps` on the host) that the "UNTIL" column correctly showed the amount of time I specified.

I can submit a pull request if required, however, this is a very simple change to implement without it.

MaffooClock:
### Related Issues

- fixes #2225

### Proposed Changes:

 
 

This change enables `OllamaDocumentEmbedder` and `OllamaTextEmbedder` to accept an optional `keep_alive` parameter to allow for setting how long a model should remain loaded in the Ollama server (same as currently available in `OllamaGenerator` and `OllamaChatGenerator`).

### How did you test it?

1. Ran unit and integration tests: all passed
2. Manual verification in a project currently using Haystack: simply include the new `keep_alive` parameter when instantiating either `OllamaGenerator` or `OllamaChatGenerator`, then check the Ollama server with `ollama ps` to confirm that the value shown for "UNTIL" matches what was specified for `keep_alive`.

### Notes for the reviewer

The `.run()` method simply passes `keep_alive` directly to the underlying call to `_client.embeddings()`, so there are no changes to logic.

I only modified the existing unit tests, which confirm that the `keep_alive` attribute matched the value supplied at instantiation; they do not check the Ollama process to see if the keepalive was obeyed.  The integration tests seem to be concerned only with prompt and response, so I did not augment them.

### Checklist

- [x] I ~added~ augmented unit tests and updated the docstrings

### DIFF
--- a/integrations/ollama/src/haystack_integrations/components/embedders/ollama/document_embedder.py
+++ b/integrations/ollama/src/haystack_integrations/components/embedders/ollama/document_embedder.py
@@ -32,6 +32,7 @@ class OllamaDocumentEmbedder:
         url: str = "http://localhost:11434",
         generation_kwargs: Optional[Dict[str, Any]] = None,
         timeout: int = 120,
+        keep_alive: Optional[Union[float, str]] = None,
         prefix: str = "",
         suffix: str = "",
         progress_bar: bool = True,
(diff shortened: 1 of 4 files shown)


MaffooClock:
> Please fix the format errors

I already saw the failed lint check and pushed another commit.

anakin87 (approved):
Thanks!
I'll release a new version soon.

PR #2230, opened 2025-08-28
feat: `AnthropicChatGenerator`- refine reasoning support

### LINKED ISSUE #2178: Anthropic - refine reasoning support

Relevant docs: https://docs.anthropic.com/en/docs/build-with-claude/extended-thinking

Amnah199:
### Related Issues

- fixes #2178 

### Proposed Changes:

- Add thinking response to ChatMessage.reasoning

### Checklist

### DIFF
--- a/integrations/anthropic/src/haystack_integrations/components/generators/anthropic/chat/chat_generator.py
+++ b/integrations/anthropic/src/haystack_integrations/components/generators/anthropic/chat/chat_generator.py
@@ -1,8 +1,10 @@
 from datetime import datetime, timezone
 from typing import Any, ClassVar, Dict, List, Literal, Optional, Tuple, Union, cast, get_args
+import logging
+import json
 
 from haystack import component, default_from_dict, default_to_dict, logging
-from haystack.components.generators.utils import _convert_streaming_chunks_to_chat_message
+#from haystack.components.generators.utils import _convert_streaming_chunks_to_chat_message
 from haystack.dataclasses.chat_message import ChatMessage, ChatRole, TextContent, ToolCall, ToolCallResult
 from haystack.dataclasses.image_content import ImageContent
 from haystack.dataclasses.streaming_chunk import (
@@ -53,6 +55,70 @@ FINISH_REASON_MAPPING: Dict[str, FinishReason] = {
     "pause_turn": "stop",
     "tool_use": "tool_calls",
 }
+
+def _convert_streaming_chunks_to_chat_message(chunks: list[StreamingChunk]) -> ChatMessage:
+    """
+    Connects the streaming chunks into a single ChatMessage.
+
+    :param chunks: The list of all `StreamingChunk` objects.
+
+    :returns: The ChatMessage.
+    """
+    text = "".join([chunk.content for chunk in chunks])
+    tool_calls = []
+    reasoning = ""
+    # Process tool calls if present in any chunk
+    tool_call_data: dict[int, dict[str, str]] = {}  # Track tool calls by index
+    for chunk in chunks:
+        if chunk.tool_calls:
+            for tool_call in chunk.tool_calls:
+                # We use the index of the tool_call to track the tool call across chunks since the ID is not always
+                # provided
+                if tool_call.index not in tool_call_data:
+                    tool_call_data[tool_call.index] = {"id": "", "name": "", "arguments": ""}
+
+                # Save the ID if present
+                if tool_call.id is not None:
+                    tool_call_data[tool_call.index]["id"] = tool_call.id
+
+                if tool_call.tool_name is not None:
+                    tool_call_data[tool_call.index]["name"] += tool_call.tool_name
+                if tool_call.arguments is not None:
+                    tool_call_data[tool_call.index]["arguments"] += tool_call.arguments
+
+        if chunk.meta.get("reasoning"):
+            reasoning = reasoning + chunk.meta.get("reasoning")
+
+    # Convert accumulated tool call data into ToolCall objects
+    sorted_keys = sorted(tool_call_data.keys())
+    for key in sorted_keys:
+        tool_call_dict = tool_call_data[key]
+        try:
+            arguments = json.loads(tool_call_dict.get("arguments", "{}")) if tool_call_dict.get("arguments") else {}
+            tool_calls.append(ToolCall(id=tool_call_dict["id"], tool_name=tool_call_dict["name"], arguments=arguments))
+        except json.JSONDecodeError:
+            logger.warning(
+                "The LLM provider returned a malformed JSON string for tool call arguments. This tool call "
+                "will be skipped. To always generate a valid JSON, set `tools_strict` to `True`. "
+                "Tool call ID: {_id}, Tool name: {_name}, Arguments: {_arguments}",
+                _id=tool_call_dict["id"],
+                _name=tool_call_dict["name"],
+                _arguments=tool_call_dict["arguments"],
+            )
+
+    # finish_reason can appear in different places so we look for the last one
+    finish_reasons = [chunk.finish_reason for chunk in chunks if chunk.finish_reason]
+    finish_reason = finish_reasons[-1] if finish_reasons else None
+
+    meta = {
+        "model": chunks[-1].meta.get("model"),
+        "index": 0,
+        "finish_reason": finish_reason,
+        "completion_start_time": chunks[0].meta.get("received_at"),  # first chunk received
+        "usage": chunks[-1].meta.get("usage"),  # last chunk has the final usage data if available
+    }
+
+    return ChatMessage.from_assistant(text=text or None, tool_calls=tool_calls, meta=meta, reasoning=reasoning)
 
 
 def _update_anthropic_message_with_tool_call_results(
@@ -287,6 +353,8 @@ class AnthropicChatGenerator:
         *,
         timeout: Optional[float] = None,
         max_retries: Optional[int] = None,
+        think: Union[bool, Dict[str, Any]] = False,
+
     ):
         """
         Creates an instance of AnthropicChatGenerator.
@@ -319,6 +387,20 @@ class AnthropicChatGenerator:
         :param max_retries:
             Maximum number of retries to attempt for failed requests. If not set, it defaults to the default set by
             the Anthropic client.
+
+        :param think:
+            If True, the model will "think" before producing a response.
+            Only models with [extended thinking](https://docs.anthropic.com/en/docs/build-with-claude/extended-thinking) support this feature.
+            When enabled, anthropic requires you to pass `budget_tokens`.
+            Note: The budget_tokens must be greater than 10000 and less than max_tokens.
+            By default, the budget tokens is set to 10000 and max_tokens is set to 12000.
+
+            If you want to set your budget_tokens, you can pass a dictionary as:
+            think = {
+                "type": "enabled",
+                "budget_tokens": 17000,
+            }
+
         """
         _check_duplicate_tool_names(list(tools or []))  # handles Toolset as well
 
@@ -328,6 +410,18 @@ class AnthropicChatGenerator:
         self.streaming_callback = streaming_callback
         self.timeout = timeout
         self.max_retries = max_retries
+
+        if not think:
+            self.think = {"type": "disabled"}
+        elif think and isinstance(think, bool):
+            self.think = {"type": "enabled", "budget_tokens": 10000}
+            if "max_tokens" not in self.generation_kwargs:
+                self.generation_kwargs["max_tokens"] = 12000
+            elif self.generation_kwargs["max_tokens"] <= 10000:
+                raise ValueError("max_tokens must be greater than budget_tokens")
+        elif isinstance(think, dict):
+            self.think = think
+
 
         client_kwargs: Dict[str, Any] = {"api_key": api_key.resolve_value()}
         # We do this since timeout=None is not the same as not setting it in Anthropic
@@ -367,6 +461,7 @@ class AnthropicChatGenerator:
             tools=serialize_tools_or_toolset(self.tools),
             timeout=self.timeout,
             max_retries=self.max_retries,
+            think=self.think,
         )
 
     @classmethod
@@ -413,12 +508,14 @@ class AnthropicChatGenerator:
             if block.type == "tool_use"
         ]
 
+        thinking = " ".join(block.thinking for block in anthropic_response.content if block.type == "thinking")
+
         # Extract and join text blocks, respecting ignore_tools_thinking_messages
         text = ""
         if not (ignore_tools_thinking_messages and tool_calls):
             text = " ".join(block.text for block in anthropic_response.content if block.type == "text")
 
-        message = ChatMessage.from_assistant(text=text, tool_calls=tool_calls)
+        message = ChatMessage.from_assistant(text=text, tool_calls=tool_calls, reasoning=thinking)
 
         # Dump the chat completion to a dict
         response_dict = anthropic_response.model_dump()
@@ -449,7 +546,7 @@ class AnthropicChatGenerator:
         tool_calls = []
         start = False
         finish_reason = None
-
+        thinking = ""
         index = getattr(chunk, "index", None)
 
         # starting streaming message
@@ -467,15 +564,21 @@ class AnthropicChatGenerator:
                         tool_name=chunk.content_block.name,
                     )
                 )
+            
         # delta of a content block
         elif chunk.type == "content_block_delta":
             if chunk.delta.type == "text_delta":
                 content = chunk.delta.text
             elif chunk.delta.type == "input_json_delta":
                 tool_calls.append(ToolCallDelta(index=tool_call_index, arguments=chunk.delta.partial_json))
+            elif chunk.delta.type == "thinking_delta":
+                thinking = chunk.delta.thinking
         # end of streaming message
         elif chunk.type == "message_delta":
             finish_reason = FINISH_REASON_MAPPING.get(getattr(chunk.delta, "stop_reason" or ""))
+
+        meta = chunk.model_dump()
+        meta["reasoning"] = thinking
 
         return StreamingChunk(
             content=content,
@@ -484,7 +587,7 @@ class AnthropicChatGenerator:
             start=start,
             finish_reason=finish_reason,
             tool_calls=tool_calls if tool_calls else None,
-            meta=chunk.model_dump(),
+            meta=meta,
         )
 
     def _prepare_request_params(
@@ -664,7 +767,7 @@ class AnthropicChatGenerator:
         """
         system_messages, non_system_messages, generation_kwargs, anthropic_tools = self._prepare_request_params(
             messages, generation_kwargs, tools
-        )
+        )        
 
         streaming_callback = select_streaming_callback(
             init_callback=self.streaming_callback,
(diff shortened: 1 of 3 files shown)


integrations/anthropic/src/haystack_integrations/components/generators/anthropic/chat/chat_generator.py
```diff
@@ -287,6 +353,8 @@ def __init__(
         *,
         timeout: Optional[float] = None,
         max_retries: Optional[int] = None,
+        think: Union[bool, Dict[str, Any]] = False,
+
```
Amnah199: @anakin87 I’d like your input here. Are we aiming for a consistent `think` parameter in init for enabling thinking across all integrations?

In Ollama, you allow either a `bool` or a `Literal`, whereas [Anthropic expects](https://docs.anthropic.com/en/docs/build-with-claude/extended-thinking) a `dict` of the form:

`thinking = {"type": "enabled", "budget_tokens": 10000}
`

The `budget_tokens` key is required there. So the options I see are:

1) Also allow `think=True` and set a default `budget_tokens` internally, or
2) Only accept a `dict` and require the user to set `budget_tokens` explicitly.

The second option might differ from other integrations depending on API. WDYT?

anakin87: Since `thinking` seems a plain dictionary, I would simply let users inject it via `generation_kwargs`.
Something like
```python
AnthropicChatGenerator(
    generation_kwargs={"thinking": {
        "type": "enabled",
        "budget_tokens": 10000
    }}
)
```

(Unrelated: I also suggest taking a look at https://github.com/deepset-ai/haystack-core-integrations/pull/2226 for other aspects)

integrations/anthropic/src/haystack_integrations/components/generators/anthropic/chat/chat_generator.py
```diff
@@ -141,6 +151,16 @@ def _convert_messages_to_anthropic_format(
...
                     text_block["cache_control"] = cache_control
                 content.append(text_block)
+            elif isinstance(part, ReasoningContent):
+                signature_value = part.extra.get("signature") if part.extra else ""
+                reasoning_block = ThinkingBlockParam(
+                    type="thinking",
+                    thinking=part.reasoning_text,
+                )
+                if signature_value:
+                    reasoning_block["signature"] = signature_value
```
Amnah199: Added this check here otherwise mypy complains.

integrations/anthropic/src/haystack_integrations/components/generators/anthropic/chat/chat_generator.py
```diff
@@ -308,6 +328,10 @@ def __init__(
             - `top_p`: The top_p value to use for nucleus sampling.
             - `top_k`: The top_k value to use for top-k sampling.
             - `extra_headers`: A dictionary of extra headers to be passed to the model (i.e. for beta features).
+            - `thinking`: A dictionary of thinking parameters to be passed to the model.
+                The `budget_tokens` passed for thinking should be less than `max_tokens`.
+                For more details and supported models, see: [Anthropic Extended Thinking](https://docs.anthropic.com/en/docs/extended-results)
```
anakin87:
```suggestion
                For more details and supported models, see: [Anthropic Extended Thinking](https://docs.anthropic.com/en/docs/build-with-claude/extended-thinking)
```

integrations/anthropic/tests/test_chat_generator.py
```diff
@@ -743,6 +794,8 @@ def test_convert_streaming_chunks_to_chat_message_with_multiple_tool_calls(self)
 
         message = _convert_streaming_chunks_to_chat_message(chunks)
 
+        # Cannot Verify reasoning content
```
anakin87: Could you explain this comment?

integrations/anthropic/src/haystack_integrations/components/generators/anthropic/chat/chat_generator.py
```diff
@@ -184,7 +203,8 @@ def _convert_messages_to_anthropic_format(
             if cache_control:
                 for blk in content:
                     if blk.get("type") == "tool_result":
-                        blk["cache_control"] = cache_control
+                        # Only apply cache_control to block types that support it
+                        blk["cache_control"] = cache_control  # type: ignore [typeddict-unknown-key]
```
anakin87: why do we need the `type: ignore`?
(tip: If you insist with LLMs long enough, they are not bad at fixing type issues)

integrations/anthropic/tests/test_chat_generator.py
```diff
@@ -1363,6 +1418,55 @@ def test_live_run_with_tools_streaming(self, tools):
         assert len(final_message.text) > 0
         assert "paris" in final_message.text.lower()
 
+    @pytest.mark.skipif(
+        not os.environ.get("ANTHROPIC_API_KEY", None),
+        reason="Export an env var called ANTHROPIC_API_KEY containing the Anthropic API key to run this test.",
+    )
+    @pytest.mark.integration
+    def test_live_run_with_tools_streaming_and_reasoning(self, tools):
```
anakin87, line 1479: nit: I would prefer to be consistent in test names, so let's always use reasoning (or thinking)

integrations/anthropic/src/haystack_integrations/components/generators/anthropic/chat/chat_generator.py
```diff
@@ -413,12 +438,22 @@ def _convert_chat_completion_to_chat_message(
             if block.type == "tool_use"
         ]
```
anakin87: I see that we don't have unit tests for `_convert_chat_completion_to_chat_message`. 

It would be to add some simple unit tests, including one with thinking/reasoning.

integrations/anthropic/tests/test_chat_generator.py
Amnah199: We’re testing the` _convert_streaming_chunks_to_chat_message` function here, but the ReasoningContent is added outside of this function (at line [634](https://github.com/deepset-ai/haystack-core-integrations/blob/65d0ae10149b7bdfcfe5451bca89f4bc1bd2635a/integrations/anthropic/src/haystack_integrations/components/generators/anthropic/chat/chat_generator.py#L634)). Because of that, I wasn’t able to verify whether the chunks containing reasoning are being converted correctly.

anakin87: Ok, I now understand. I think we can remove this comment or explain better in the comment...

Amnah199: I’ll add a better one line explanation, so its not confusing.

integrations/anthropic/src/haystack_integrations/components/generators/anthropic/chat/chat_generator.py
Amnah199: mypy complains that `ThinkingBlockParam` doesn't have `cache_control`, although we have a check for the type. I tried to sort it with claude but it was spiraling and ended up telling me to add ignore 😄

integrations/anthropic/src/haystack_integrations/components/generators/anthropic/chat/chat_generator.py
Amnah199: Added some new tests.

anakin87: thanks!

integrations/anthropic/tests/test_chat_generator.py
```diff
@@ -743,6 +844,9 @@ def test_convert_streaming_chunks_to_chat_message_with_multiple_tool_calls(self)
 
         message = _convert_streaming_chunks_to_chat_message(chunks)
 
+        # Cannot conversion of reasoning content in streaming chunks
```
anakin87:
```suggestion
        # Cannot test creating ReasoningContent from StreamingChunk objects
```

integrations/anthropic/src/haystack_integrations/components/generators/anthropic/chat/chat_generator.py
anakin87: Ok I investigated a bit and agree with you.

Just for sharing... This would work, but I don't recommend this approach:
```python
def _is_tool_result_block(
    blk: Union[TextBlockParam, ToolUseBlockParam, ToolResultBlockParam, ImageBlockParam, ThinkingBlockParam]
) -> TypeGuard[ToolResultBlockParam]:
    """Type guard to check if a block is a ToolResultBlockParam."""
    return blk.get("type") == "tool_result"


...
if _is_tool_result_block(blk):
...
```

Amnah199: Hmm can you elaborate why shouldnt we use it?

anakin87: Yes, it's a matter of taste.

It's verbose and after a bit of investigation, I see that this issue comes from the Anthropic choice to use Typed Dictionaries, which don't allow checking `if isinstance(blk, ToolResultBlockParam)`.
I understand it is straightforward for users to provide dictionaries instead of objects, but ofc this comes with some limitations in type checking. 

So I am Ok with the `type: ignore`. However, do as you prefer.

integrations/anthropic/src/haystack_integrations/components/generators/anthropic/chat/chat_generator.py
```diff
@@ -575,6 +621,26 @@ def _process_response(
...
+            for streaming_chunk in chunks:
+                if (delta := streaming_chunk.meta.get("delta")) is not None:
+                    if delta.get("type") == "thinking_delta" and delta.get("thinking") is not None:
+                        reasoning_content += delta.get("thinking", "")
+                    if delta.get("type") == "signature_delta" and delta.get("signature") is not None:
+                        reasoning_signature += delta.get("signature", "")
+                if (content_block := streaming_chunk.meta.get("content_block")) is not None and content_block.get(
+                    "type"
+                ) == "redacted_thinking":
+                    reasoning_content += content_block.get("data", "")
```
anakin87: I believe that we should keep track of redacted thinking blocks, otherwise it becomes difficult to pass them back to the API when converting Haystack `ChatMessage` to the Anthropic format.

> When passing thinking and redacted_thinking blocks back to the API in a multi-turn conversation, you must include the complete unmodified block back to the API for the last assistant turn. This is critical for maintaining the model’s reasoning flow. 

from https://docs.anthropic.com/en/docs/build-with-claude/extended-thinking#thinking-redaction

Amnah199: That was a good catch. I separated the two so that redacted thinking can be passed as `RedactedThinkingBlockParam` to Anthropic. Also added a test.

integrations/anthropic/src/haystack_integrations/components/generators/anthropic/chat/chat_generator.py
```diff
@@ -413,12 +464,32 @@ def _convert_chat_completion_to_chat_message(
...
+                redacted_thinking += block.data
+
+        reasoning = (
+            ReasoningContent(
+                reasoning_text=reasoning_text,
+                extra={"signature": reasoning_signature, "redacted_thinking": redacted_thinking},
+            )
+            if reasoning_text or reasoning_signature or redacted_thinking
+            else None
+        )
```
anakin87: I see that we must use a string for `reasoning_text` (based on `ReasoningContent` types)
But for extra fields (signature and redacted thinking), I think that we should only include them if their length is >0.

integrations/anthropic/src/haystack_integrations/components/generators/anthropic/chat/chat_generator.py
```diff
@@ -413,12 +464,32 @@ def _convert_chat_completion_to_chat_message(
...
 
+        reasoning_text = ""
+        reasoning_signature = ""
+        redacted_thinking = ""
+        for block in anthropic_response.content:
+            if block.type == "thinking":
+                reasoning_text = block.thinking
+                reasoning_signature = block.signature
+            elif block.type == "redacted_thinking":
+                redacted_thinking += block.data
```
anakin87: my impression is that we expect a single thinking or redacted thinking block... (we don't want to support interleaved thinking beta feature for the moment). Right?

It's a nitpick but I find a bit confusing that we are doing `reasoning_signature = block.signature` and `redacted_thinking += block.data` (`=` vs `+=`)

integrations/anthropic/src/haystack_integrations/components/generators/anthropic/chat/chat_generator.py
```diff
@@ -575,6 +649,30 @@ def _process_response(
...
+                ) == "redacted_thinking":
+                    redacted_thinking += content_block.get("data", "")
+
+            reasoning = (
+                ReasoningContent(
+                    reasoning_text=reasoning_content,
+                    extra={"signature": reasoning_signature, "redacted_thinking": redacted_thinking},
+                )
+                if reasoning_content or reasoning_signature or redacted_thinking
+                else None
```
anakin87: refer to the comment above in `ReasoningContent` construction

integrations/anthropic/src/haystack_integrations/components/generators/anthropic/chat/chat_generator.py
```diff
@@ -575,6 +650,31 @@ def _process_response(
...
+            for streaming_chunk in chunks:
+                if (delta := streaming_chunk.meta.get("delta")) is not None:
+                    if delta.get("type") == "thinking_delta" and delta.get("thinking") is not None:
+                        reasoning_text += delta.get("thinking", "")
+                    if delta.get("type") == "signature_delta" and delta.get("signature") is not None:
+                        reasoning_signature += delta.get("signature", "")
+                if (content_block := streaming_chunk.meta.get("content_block")) is not None and content_block.get(
+                    "type"
+                ) == "redacted_thinking":
+                    redacted_thinking += content_block.get("data", "")
```
Amnah199: @anakin87 In streaming we get multiple chunks with redacted thinking. While converting it into a `ChatMessage` I combined them into a single string. This might be wrong.
Should I rather store them as a list of strings so that they can be passed as separate blocks to anthropic?

integrations/anthropic/src/haystack_integrations/components/generators/anthropic/chat/chat_generator.py
```diff
@@ -413,12 +464,33 @@ def _convert_chat_completion_to_chat_message(
...
 
+        reasoning_text = ""
+        reasoning_signature = ""
+        redacted_thinking = ""
+        for block in anthropic_response.content:
+            if block.type == "thinking":
+                reasoning_text = block.thinking
+                reasoning_signature = block.signature
+            elif block.type == "redacted_thinking":
+                redacted_thinking = block.data
```
sjrl: Hmm I'm a little worried about this set up. It's possible for Anthropic to return multiple thinking blocks. E.g. `[ThinkingBlock, RedactedThinking, ThinkingBlock]` 

And it seems like we are making the assumption that only one can be returned by the endpoint. I think it would be good to not make this assumption. Take a look at how I handled this in Bedrock here https://github.com/deepset-ai/haystack-core-integrations/blob/f8bd0bc5103bf9b55c8bc10e70e828453e33ae4d/integrations/amazon_bedrock/src/haystack_integrations/components/generators/amazon_bedrock/chat/utils.py#L311-L339

Amnah199: >  It's possible for Anthropic to return multiple thinking blocks. E.g. [ThinkingBlock, RedactedThinking, ThinkingBlock]

@sjrl me and Stefano were already discussing this, but I couldnt find anything in documentation confirming if single or multiple blocks are returned in non streaming. Can you share how did you conclude this?

integrations/anthropic/src/haystack_integrations/components/generators/anthropic/chat/chat_generator.py
sjrl: Huh interesting I didn't see this happen when using Bedrock. But yes I agree these would need to be stored separately since each redacted thinking block is a fully formed bytes string (at least it is in Bedrock). So doing a concat on these doesn't make sense I don't think.

integrations/anthropic/src/haystack_integrations/components/generators/anthropic/chat/chat_generator.py
anakin87: I think this is something different: @sjrl is pointing to thinking-related block with different types `[ThinkingBlock, RedactedThinking, ThinkingBlock]`. (Not multiple `ThinkingBlock`s).

So we might drop the `elif` and replace with an `if`

sjrl: @Amnah199 sure! I based it off this api docs page https://docs.aws.amazon.com/bedrock/latest/userguide/claude-messages-thinking-encryption.html

With this example they provided 
```
{
    "content": [
        {
            "type": "thinking",
            "thinking": "Let me analyze this step by step...",
            "signature":"WaUjzkypQ2mUEVM36O2TxuC06KN8xyfbJwyem2dw3URve/op91XWHOEBLLqIOMfFG/UvLEczmEsUjavL...."                },
        {
            "type": "redacted_thinking",
            "data":"EmwKAhgBEgy3va3pzix/LafPsn4aDFIT2Xlxh0L5L8rLVyIwxtE3rAFBa8cr3qpP..."
        },
        {
            "type": "text",
            "text": "Based on my analysis..."
        }
    ]
}
```

sjrl: > (Not multiple `ThinkingBlock`s).

No I do mean multiple ThinkingBlocks

Amnah199: > So we might drop the elif and replace with an if

@anakin87 this shouldnt be a problem as one block can be of only one type.

Amnah199: @sjrl Hmm I meant like how do we know we can have multiple blocks with same `type="thinking"`.  In the example we have one thinking and redacted_thinking block that will be handled separately. 🤔

sjrl: Ahh good point. I suppose there isn't a full proof way to guarantee this I would say. Probably safest to implement it in a way that allows for multiple interleaved thinking blocks but likely we will almost have one thinking block returned and perhaps two (one regular and one redacted) based on this example.

sjrl: Regardless one issue with the current implementation is that we lose which order the redacted and normal thinking blocks came in. Based on how the ReasoningContent dataclass is structured it's not possible to know if the redacted thinking came first or second in the order which could cause an issue when reconstructing the the Anthropic response.

anakin87: (Related to `ReasoningContent`: since we just introduced it, if we can a way to make it easier to use also in these use cases, I think we are still on time to change it slightly. I don't have a specific idea at the moment...)

sjrl: I was just giving this a think and perhaps ReasoningContent as it is fine, but rather than try and combine all reasoning content into one block we could embrace adding multiple ReasoningContents to one ChatMessage. Then in the example I link above we'd add two ReasoningContents, one for the normal and one for the redacted.

PR #2235, opened 2025-09-01
ci: download pre-built wheels for llama-cpp-python on macOS

anakin87:
### Related Issues
Nightly tests failing on macOS: https://github.com/deepset-ai/haystack-core-integrations/actions/runs/17364348515/job/49288820099
CMake error: OpenMP not found

This likely happens because [new GitHub macOS runners](https://github.blog/changelog/2025-07-11-upcoming-changes-to-macos-hosted-runners-macos-latest-migration-and-xcode-support-policy-updates/) are not well configured or lack the dependency.

### Proposed Changes:
- Use pre-built wheels on macOS by adding a custom index (llama-cpp-python provides these wheels)
- on macOS, test with Python 3.10 and 3.12 because wheels are only available for python>=3.10, <=3.12

### How did you test it?
CI

### Notes for the reviewer
This would not change things for end users. I have not seen many users reporting issues on this specific topic on llama-cpp-python and I would expect users to be able to install/configure missing system/compilation dependencies.

### Checklist

### DIFF
--- a/integrations/llama_cpp/pyproject.toml
+++ b/integrations/llama_cpp/pyproject.toml
@@ -28,6 +28,12 @@ classifiers = [
 ]
 dependencies = ["haystack-ai>=2.16.1", "llama-cpp-python>=0.2.87"]
 
+[[tool.uv.index]]
+# Optional name for the index.
+name = "llama-cpp-python"
+# Required URL for the index.
+url = "https://abetlen.github.io/llama-cpp-python/whl/metal/"
+
 [project.urls]
 Documentation = "https://github.com/deepset-ai/haystack-core-integrations/tree/main/integrations/llama_cpp#readme"
 Issues = "https://github.com/deepset-ai/haystack-core-integrations/issues"
--- a/integrations/llama_cpp/pyproject.toml
+++ b/integrations/llama_cpp/pyproject.toml
@@ -29,7 +29,7 @@ classifiers = [
 dependencies = ["haystack-ai>=2.16.1", "llama-cpp-python>=0.2.87"]
 
 [[tool.uv]]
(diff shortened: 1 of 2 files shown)


integrations/llama_cpp/pyproject.toml
```diff
@@ -28,6 +28,14 @@ classifiers = [
 ]
 dependencies = ["haystack-ai>=2.16.1", "llama-cpp-python>=0.2.87"]
 
+# On macOS GitHub runners, we use a custom index to download pre-built wheels.
+# Installing from source might fail due to missing dependencies (CMake fails with "OpenMP not found")
+[tool.uv]
+index-strategy = "unsafe-best-match"  # this ensures that packages are not only searched in the below index but also in PyPI
+[[tool.uv.index]]
+name = "llama-cpp-python-macos"
+url = "https://abetlen.github.io/llama-cpp-python/whl/metal/"
```
sjrl, line 37: Just for my own knowledge could you explain how this roughly works? Does this only get applied when installing this integration on a mac?

anakin87: This only happens on the CI, where the installer is uv. AFAIK, there is not an easy way to do this in Hatch.

In short, defining an index + "unsafe-best-match" makes sure the package is searched on the provided index + PyPI. Then the most appropriate version is installed.

This means that for macOS, the pre-built metal wheels will be installed. For Linux and Windows, the packages available on PyPI will be used.

More info: https://docs.astral.sh/uv/concepts/indexes/

sjrl: Thanks!

sjrl (approved):
looks good!

PR #2248, opened 2025-09-10
fix: fix mcp types + add py.typed

### LINKED ISSUE #1771: Investigate better typing configuration for core integrations

vblagoje:
### Related Issues

- part of #1771

### Proposed Changes:
- run `hatch run test:types` and fix type errors
  (this means testing with all dependencies installed, differently from `hatch run lint:typing` that did not install dependencies)
- introduce stricter configuration in pyproject.toml (and fix small errors)
  - as shown in #1910, these files should be put in submodules of the namespace packages
  - the namespace package is `haystack_integrations`
  - in this case, I put py.typed in 
    - `integrations/mcp/src/haystack_integrations/tools/mcp/py.typed`
    - as a rule of thumb, py.typed files should be placed at the same level as the inner folder that corresponds to the integration name.

### How did you test it?
CI

### Notes for the reviewer
This PR can be seen as a complete example of the work we need to do for each integration to fix #1771.

### Checklist

### DIFF
--- a/integrations/mcp/pyproject.toml
+++ b/integrations/mcp/pyproject.toml
@@ -78,18 +78,24 @@ unit = 'pytest -m "not integration" {args:tests}'
 integration = 'pytest -m "integration" {args:tests}'
 all = 'pytest {args:tests}'
 cov-retry = 'all --cov=haystack_integrations --reruns 3 --reruns-delay 30 -x'
-types = "mypy --install-types --non-interactive --explicit-package-bases {args:src/ tests}"
+types = """mypy -p haystack_integrations.tools.mcp {args}"""
 
-# TODO: remove lint environment once this integration is properly typed
-# test environment should be used instead
-# https://github.com/deepset-ai/haystack-core-integrations/issues/1771
-[tool.hatch.envs.lint]
-installer = "uv"
-detached = true
-dependencies = ["pip", "mypy>=1.0.0", "ruff>=0.0.243"]
+[tool.mypy]
+install_types = true
+non_interactive = true
+check_untyped_defs = true
+disallow_incomplete_defs = true
 
-[tool.hatch.envs.lint.scripts]
-typing = "mypy --install-types --non-interactive --explicit-package-bases {args:src/ tests}"
+[[tool.mypy.overrides]]
+# MCP package types
+module = [
+    "mcp.*",
+    "mcp",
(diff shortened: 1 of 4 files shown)


integrations/mcp/pyproject.toml
```diff
@@ -78,18 +78,24 @@ unit = 'pytest -m "not integration" {args:tests}'
...
+check_untyped_defs = true
+disallow_incomplete_defs = true
 
-[tool.hatch.envs.lint.scripts]
-typing = "mypy --install-types --non-interactive --explicit-package-bases {args:src/ tests}"
+[[tool.mypy.overrides]]
+# MCP package types
+module = [
+    "mcp.*",
+    "mcp",
```
anakin87: Can we avoid adding `mcp` to `ignore_missing_imports`?
Otherwise, we'd end up not verifying MCP types at all

integrations/mcp/pyproject.toml
```diff
@@ -78,18 +78,22 @@ unit = 'pytest -m "not integration" {args:tests}'
...
-[tool.hatch.envs.lint.scripts]
-typing = "mypy --install-types --non-interactive --explicit-package-bases {args:src/ tests}"
+[[tool.mypy.overrides]]
+# MCP package types
+module = [
+    "httpx",
+    "exceptiongroup",
+    "anyio.*",
+]
+ignore_missing_imports = true
```
anakin87: Have you tried removing this block altogether?
If possible, it would be the best thing to do. But sometimes, libraries do not provide valid types, so it's not always doable.

anakin87 (approved):
👏

PR #2252, opened 2025-09-11
chore: Fix linting cohere

sjrl:
### Related Issues

- fixes #issue-number

### Checklist

### DIFF
--- a/integrations/cohere/tests/test_document_image_embedder.py
+++ b/integrations/cohere/tests/test_document_image_embedder.py
@@ -178,7 +178,7 @@ class TestCohereDocumentImageEmbedder:
             Document(content="Image 1", meta={"file_path": "image_1.jpg"}),
         ]
 
-        with pytest.raises(RuntimeError, match=r"Conversion failed for some documents."):
+        with pytest.raises(RuntimeError, match=r"Conversion failed for some documents\."):
             embedder._extract_images_to_embed(documents)
 
     @patch(f"{IMPORT_PATH}._extract_image_sources_info")


integrations/cohere/tests/test_chat_generator.py
```diff
@@ -108,7 +108,7 @@ def test_format_message_image_in_non_user_message(self):
         message = ChatMessage.from_assistant(text="Here's an image.")
         message._content.append(image_content)  # Add image to assistant message
 
-        with pytest.raises(ValueError, match="`ImageContent` is only supported for user messages."):
+        with pytest.raises(ValueError, match=r"`ImageContent` is only supported for user messages."):
```
julian-risch:
```suggestion
        with pytest.raises(ValueError, match=r"`ImageContent` is only supported for user messages\."):
```
The dot at the end of this sentence was meant to match only a literal dot character. With this PR, it is treated as a regex special character. I suggest we escape it so it just matches the literal dot character. Alternative is to remove the dot character.

integrations/cohere/tests/test_document_image_embedder.py
```diff
@@ -178,7 +178,7 @@ def test_extract_images_to_embed_none_images(
             Document(content="Image 1", meta={"file_path": "image_1.jpg"}),
         ]
 
-        with pytest.raises(RuntimeError, match="Conversion failed for some documents."):
+        with pytest.raises(RuntimeError, match=r"Conversion failed for some documents."):
```
julian-risch:
```suggestion
        with pytest.raises(RuntimeError, match=r"Conversion failed for some documents\."):
```
same as above

integrations/cohere/tests/test_chat_generator.py
julian-risch: Line of code where the message is created: https://github.com/deepset-ai/haystack-core-integrations/blob/main/integrations/cohere/src/haystack_integrations/components/generators/cohere/chat/chat_generator.py#L98

sjrl: good catch! will do

PR #2212, opened 2025-08-22
feat: Add Gemini "Thinking" Support to GoogleGenAIChatGenerator

vblagoje:
## Why:
Introduces support for Gemini  "thinking/reasoning" .Models can now show their reasoning process, maintain thought context across multi-turn conversations, and use configurable thinking budgets.

## What:
- Added `thinking_budget` configuration via `generation_kwargs` with model-specific validation
- Implemented parsing of thought parts into `ReasoningContent` objects
- Extended streaming aggregation to handle reasoning deltas while reusing generic aggregator
- Implemented fail-fast error handling for unsupported models
- Added integration tests for thinking feature and error scenarios

## How can it be used:
```python
from haystack_integrations.components.generators.google_genai import GoogleGenAIChatGenerator
from haystack.dataclasses import ChatMessage

# Enable thinking with dynamic or explicit budget
generator = GoogleGenAIChatGenerator(
    model="gemini-2.5-pro",
    generation_kwargs={"thinking_budget": -1}  # -1 for dynamic, or specific token count
)

messages = [ChatMessage.from_user("Explain why the sky is blue")]
response = generator.run(messages=messages)

# Access reasoning content
message = response["replies"][0]
if message.reasonings:
    for reasoning in message.reasonings:
        print(f"Thought: {reasoning.reasoning_text}")

# Check thinking tokens used
print(f"Thinking tokens: {message.meta['usage'].get('thoughts_token_count', 0)}")
```

## How did you test it:
- Added integration tests for thinking feature with gemini-2.5-pro
- Added fail-fast tests for unsupported models  

## Notes for the reviewer:
- `_aggregate_streaming_chunks_with_reasoning` extends the generic aggregator (40 lines vs 78 if duplicated)
- Model detection may need updates as new models are released

PR #2266, opened 2025-09-12
feat: Add reasoning support for Cohere chat generator

### LINKED ISSUE #2179: Cohere - reasoning support

Relevant docs: https://docs.cohere.com/reference/chat

ChinmayBansal:
### Related Issues

 - fixes #2179
 - fixes parent issue in haystack repo: https://github.com/deepset-ai/haystack/issues/9700 

 ### Proposed Changes:

This PR implements reasoning support for Cohere's chat generator following the parent issue #9700 requirements. The implementation adds the ability to extract and store model reasoning content in the `ReasoningContent` field of `ChatMessage`.

  **Key Features Added:**
  - **Reasoning extraction function** (`_extract_reasoning_from_response`) that
  supports multiple formats:
    - XML-style tags: `<thinking>`, `<reasoning>`
    - Section headers: `## Reasoning`, `## Thinking`, `## My reasoning`
    - Natural language patterns: "Let me think", "Step by step", "First,"
    - Intelligent separator detection to determine reasoning boundaries
  - **Dual-path response parsing** that handles both regular responses and tool
  call scenarios
  - **Streaming support** with custom
  `_convert_streaming_chunks_to_chat_message_with_reasoning` function
  - **Tool call compatibility** ensuring reasoning works alongside tool calls
  without conflicts
  - **Content cleaning** that removes reasoning markers from the main response
  text
  - **Length filtering** with 30-character minimum threshold to avoid false
  positives

  The implementation follows the established patterns from Ollama (#2200) and
  Google GenAI (#2212) PRs, ensuring consistency across providers.

  ### How did you test it?

  - **Unit tests**: Added comprehensive test suite (`TestReasoningExtraction`)
  covering:
    - Various reasoning pattern formats (`<thinking>`, `<reasoning>`,
  step-by-step)
    - Edge cases (no reasoning, short reasoning ignored)
    - Natural language reasoning detection
  - **Integration tests**: Added `TestCohereChatGeneratorReasoning` class with:
    - End-to-end reasoning extraction with Command A Reasoning model
  (lightweight)
    - Mock response testing for consistent behavior
    - Tool call compatibility verification
  - **Manual verification**: All unit tests pass, integration tests show expected
   rate limiting (API working correctly)
  - **Code quality**: All linting (`hatch run fmt`) and type checking (`hatch run
   test:types`) pass

  ### Notes for the reviewer

  - The reasoning extraction logic is in `_extract_reasoning_from_response()`
  function (lines ~355-435)
  - Key integration points are in `_parse_response()` for both regular and tool
  call paths (lines ~194-220)
  - Streaming support uses a post-processing approach since Cohere embeds
  reasoning in response text
  - Test cases use the lightweight `command-a-reasoning-111b-2024-10-03` model as
   recommended
  - Implementation handles the Cohere-specific `thinking` parameter via
  `generation_kwargs={"thinking": True}`

  ### Checklist

  - [x] I've used one of the conventional commit types for my PR title: `fix:`,`feat:`, `build:`, `chore:`, `ci:`, `docs:`,`style:`, `refactor:`, `perf:`,`test:`.

### DIFF
--- a/integrations/cohere/src/haystack_integrations/components/generators/cohere/chat/chat_generator.py
+++ b/integrations/cohere/src/haystack_integrations/components/generators/cohere/chat/chat_generator.py
@@ -282,11 +282,19 @@ def _convert_cohere_chunk_to_streaming_chunk(
     start = False
     finish_reason = None
     tool_calls = None
+    reasoning = None
     meta: dict[str, Any] = {"model": model}
 
     if chunk.type == "content-delta" and chunk.delta and chunk.delta.message:
-        if chunk.delta.message and chunk.delta.message.content and chunk.delta.message.content.text is not None:
-            content = chunk.delta.message.content.text
+        if chunk.delta.message and chunk.delta.message.content:
+            # Handle thinking/reasoning content (prioritize over text if both exist)
+            # Note: StreamingChunk can only have ONE of content/tool_calls/reasoning set
+            thinking_text = getattr(chunk.delta.message.content, "thinking", None)
+            if thinking_text is not None and isinstance(thinking_text, str):
+                reasoning = ReasoningContent(reasoning_text=thinking_text)
+            # Handle text content (only if no reasoning)
+            elif chunk.delta.message.content.text is not None:
+                content = chunk.delta.message.content.text
 
     elif chunk.type == "tool-plan-delta" and chunk.delta and chunk.delta.message:
         if chunk.delta.message and chunk.delta.message.tool_plan is not None:
@@ -362,6 +370,7 @@ def _convert_cohere_chunk_to_streaming_chunk(
         component_info=component_info,
         index=index,
         tool_calls=tool_calls,
+        reasoning=reasoning,
         start=start,
         finish_reason=finish_reason,
         meta=meta,
--- a/integrations/cohere/tests/test_chat_generator.py
+++ b/integrations/cohere/tests/test_chat_generator.py
@@ -728,12 +728,6 @@ class TestCohereChatGeneratorInference:
         assert isinstance(results["replies"][0], ChatMessage)
         assert len(results["replies"][0].text) > 0
 
-
-class TestCohereChatGeneratorReasoning:
-    """Integration tests for reasoning functionality in CohereChatGenerator."""
-
-    @pytest.mark.skipif(not os.environ.get("COHERE_API_KEY"), reason="COHERE_API_KEY not set")
-    @pytest.mark.integration
     def test_reasoning_with_command_a_reasoning_model(self):
         """Test reasoning extraction with Command A Reasoning model."""
         generator = CohereChatGenerator(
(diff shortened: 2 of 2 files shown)


integrations/cohere/tests/test_chat_generator.py
```diff
@@ -656,3 +657,254 @@ def test_live_run_multimodal(self):
...
+
+    def test_extract_reasoning_with_thinking_tags(self):
+        """Test extraction of reasoning from <thinking> tags."""
+        response_text = """<thinking>
+I need to calculate the area of a circle.
+The formula is π * r².
+Given radius is 5, so area = π * 25 = 78.54
+</thinking>
+
+The area of a circle with radius 5 is approximately 78.54 square units."""
```
sjrl: Hey @ChinmayBansal is it really the case that Cohere only returns reasoning content as part of the text response wrapped in these tags?

sjrl: Based on the code snippet here from Cohere https://docs.cohere.com/docs/reasoning#getting-started it looks like that reasoning content should come separately 

```python

for content in response.message.content:
    if content.type == "thinking":
        print("Thinking:", content.thinking)
 
    if content.type == "text":
        print("Response:", content.text)
```

integrations/cohere/src/haystack_integrations/components/generators/cohere/chat/chat_generator.py
```diff
@@ -202,11 +203,20 @@ def _parse_response(chat_response: ChatResponse, model: str) -> ChatMessage:
                     )
                 )
 
+        # Extract reasoning from content if present, even with tool calls
+        reasoning_content = None
+        if chat_response.message.content and hasattr(chat_response.message.content[0], "text"):
+            raw_content = chat_response.message.content[0].text
+            reasoning_content, _ = _extract_reasoning_from_response(raw_content)
```
sjrl: Re my comment [here](https://github.com/deepset-ai/haystack-core-integrations/pull/2266/files#r2526491898) it seems like we should be accessing `chat_response.message.content[0].thinking` to access thinking content.

ChinmayBansal:
@sjrl I have addressed your feedback. The changes I have made are using the native method while keeping the fallback method if the native method does not apply reasoning content. I have also updated the test cases.

Let me know if more changes are needed!

integrations/cohere/src/haystack_integrations/components/generators/cohere/chat/chat_generator.py
```diff
@@ -192,6 +194,29 @@ def _parse_response(chat_response: ChatResponse, model: str) -> ChatMessage:
...
+                if hasattr(content_item, "text") and content_item.text:
+                    text_content = content_item.text
+
+    # Fallback: If reasoning wasn't found via native API but text contains reasoning markers,
+    # extract it from text (for backward compatibility)
+    if reasoning_content is None and text_content:
+        fallback_reasoning, cleaned_text = _extract_reasoning_from_text(text_content)
+        if fallback_reasoning is not None:
+            reasoning_content = fallback_reasoning
+            text_content = cleaned_text
```
sjrl: To be consistent with our other chat generator implementations I think we'd prefer not to have this fallback especially if it's not clear how we could trigger this in a real integration test. Do you know of a case where Cohere can do this?

integrations/cohere/tests/test_chat_generator.py
```diff
@@ -727,6 +728,159 @@ def test_live_run_multimodal(self):
         assert isinstance(results["replies"][0], ChatMessage)
         assert len(results["replies"][0].text) > 0
 
+
+class TestCohereChatGeneratorReasoning:
+    """Integration tests for reasoning functionality in CohereChatGenerator."""
```
sjrl: Hey could we just make the new tests you added as member functions to the class `TestCohereChatGeneratorInference`. I don't think we need a separate `TestCohereChatGeneratorReasoning`. 

Also the entire class `TestCohereChatGeneratorInference` is marked as an integration test with 
```python
@pytest.mark.skipif(
    not os.environ.get("COHERE_API_KEY", None) and not os.environ.get("CO_API_KEY", None),
    reason="Export an env var called COHERE_API_KEY/CO_API_KEY containing the Cohere API key to run this test.",
)
@pytest.mark.integration
class TestCohereChatGeneratorInference:
```

this means we don't need to mark each individual test with the `skipif` or the `integration` label.

ChinmayBansal:
@sjrl I had included it in the text-based fallback extraction thinking it was needed for streaming. After doing some more research, I found that streaming API does support thinking with `chunk.delta.message.content.thinking` in content delta events similar to the non streaming structure. I did remove the fallback logic. The current implementation now uses only the native API for non streaming responses. For streaming, it would have to come through content delta events but Haystack's StreamingChunk does not currently have a reasoning field to handle this. This could be added as a followup if needed.

PR #2272, opened 2025-09-15
refactor!: Weave - update patching support and unpin gql

anakin87:
### Related Issues

Nightly tests failing for Weave: https://github.com/deepset-ai/haystack-core-integrations/actions/runs/17718657209/job/50347178781

This happens because the management of autopatch settings has changed:
  - `AutopatchSettings` class no longer contains the `disable_autopatch` attribute : https://github.com/wandb/weave/blob/2a4106bba9605e8e9a427b91702c5845e0628ed6/weave/trace/autopatch.py#L33
  - `AutopatchSettings` class has also been deprecated
  - `autopatch_settings=AutopatchSettings(disable_autopatch=True)` -> ` settings={"implicitly_patch_integrations": False}`
  - more information in their docs: https://weave-docs.wandb.ai/guides/integrations/

### Proposed Changes:
- adopt the new simpler approach and stop serializing/deserializing `AutopatchSettings`
- unrelated: unpin `gql` because the new versions of Weave are compatible with `gql>4` and handle this internally. This basically reverts #2198

### How did you test it?
CI, adapted tests

### Notes for the reviewer
This is a breaking change and I would release a new major version.

I have also tried to think on how to avoid a breaking change but Weave changes are quite extensive to do that reliably.

### Checklist

### DIFF
--- a/integrations/weights_and_biases_weave/src/haystack_integrations/components/connectors/weave/weave_connector.py
+++ b/integrations/weights_and_biases_weave/src/haystack_integrations/components/connectors/weave/weave_connector.py
@@ -3,6 +3,7 @@ from typing import Any, Optional
 from haystack import component, default_from_dict, default_to_dict, logging, tracing
 
 from haystack_integrations.tracing.weave import WeaveTracer
+from weave.trace.settings import UserSettings
 
 logger = logging.getLogger(__name__)
 
@@ -99,7 +100,12 @@ class WeaveConnector:
         :returns:
             Dictionary with all the necessary information to recreate this component.
         """
-        return default_to_dict(self, pipeline_name=self.pipeline_name, weave_init_kwargs=self.weave_init_kwargs)
+        weave_init_kwargs = self.weave_init_kwargs.copy()
+        settings = weave_init_kwargs.get("settings", None)
+        if isinstance(settings, UserSettings):
+            weave_init_kwargs["settings"] = settings.model_dump(mode="json", exclude_defaults=True)
+
+        return default_to_dict(self, pipeline_name=self.pipeline_name, weave_init_kwargs=weave_init_kwargs)
 
     @classmethod
     def from_dict(cls, data: dict[str, Any]) -> "WeaveConnector":
@@ -110,4 +116,7 @@ class WeaveConnector:
         :returns:
             Deserialized component.
         """
+        # Weave's API accepts both dict and UserSettings objects for `settings`,
+        # so we can pass the raw settings dict directly without deserializing it
+
         return default_from_dict(cls, data)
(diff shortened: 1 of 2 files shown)


integrations/weights_and_biases_weave/src/haystack_integrations/components/connectors/weave/weave_connector.py
```diff
@@ -100,12 +99,7 @@ def to_dict(self) -> dict[str, Any]:
         :returns:
             Dictionary with all the necessary information to recreate this component.
         """
-        weave_init_kwargs = self.weave_init_kwargs.copy()
-        autopatch_settings = weave_init_kwargs.get("autopatch_settings", None)
-        if isinstance(autopatch_settings, AutopatchSettings):
-            weave_init_kwargs["autopatch_settings"] = autopatch_settings.model_dump(exclude_defaults=True)
-
-        return default_to_dict(self, pipeline_name=self.pipeline_name, weave_init_kwargs=weave_init_kwargs)
+        return default_to_dict(self, pipeline_name=self.pipeline_name, weave_init_kwargs=self.weave_init_kwargs)
```
sjrl: I think we need to handle serialization of [`UserSettings`](https://github.com/wandb/weave/blob/7f85a83ab8d714d0b15ed0427ba4d7b9f212ac9d/weave/trace/settings.py#L28) here which can be passed as an init kwarg. For example,
```python
        connector = WeaveConnector(
            pipeline_name="test_pipeline",
            weave_init_kwargs={
                "settings": UserSettings(print_call_link=False),
        )
```
and this is available through the `weave.init` call we make in the `WeaveTracer`
```python
def init(
    project_name: str,
    *,
    settings: UserSettings | dict[str, Any] | None = None,
    autopatch_settings: AutopatchSettings | None = None,
    global_postprocess_inputs: PostprocessInputsFunc | None = None,
    global_postprocess_output: PostprocessOutputFunc | None = None,
    global_attributes: dict[str, Any] | None = None,
) -> weave_client.WeaveClient:
```

anakin87: Thanks. Done in https://github.com/deepset-ai/haystack-core-integrations/pull/2272/commits/56d842cf44c37fbac486c4d3b70753762c04de04

integrations/weights_and_biases_weave/src/haystack_integrations/components/connectors/weave/weave_connector.py
```diff
@@ -110,4 +116,7 @@ def from_dict(cls, data: dict[str, Any]) -> "WeaveConnector":
         :returns:
             Deserialized component.
         """
+        # Weave's API accepts both dict and UserSettings objects for `settings`,
+        # so we can pass the raw settings dict directly without deserializing it
```
sjrl, line 120: Good find!

sjrl (approved):

PR #2292, opened 2025-09-19
feat!: adding `api_token` and `apit_token_id` support to `ElasticSearchDocumentStore`

### LINKED ISSUE #2245: ElasticsearchDocStore API-Key loading
We want to use our own environment variable names passed through Kube to get the ElasticsearchDocStore running. 

For this it would be necessary to pass the api_keys upon creating the objects in the pipeline, or having the component support the haystack-native secret functionality.

-

**Additional context**
- potential implementation / PR can be found here #2224

davidsbatista:
### Related Issues

- fixes #2245 

### Proposed Changes:

- Adds API token authentication support to the `ElasticsearchDocumentStore` 
- Supports multiple authentication methods:
   - Environment Variables: Support for ELASTIC_API_KEY and ELASTIC_API_KEY_ID environment variables
   - Secret Objects: Direct parameter support for api_key and api_key_id as Secret objects
   - Base64 Encoded Keys: Support for single base64-encoded API keys containing both ID and secret
   - No Authentication: Graceful fallback to unauthenticated connections when no credentials are provided

### How did you test it?

- unit tests and manual verification
- tested Serialization and Deserialization with env vars
- tes auth with SecretToken string
- test with single API key validation
- error handling, when only API_ID is given and no API_TOKEN

### Notes for the reviewer

- I did test connecting to locally Docker ES setup with auth. We can add it as an integration test, but this involves setting up an ES docker instance and generating API_ID and API_TOKEN on the fly and passing those to the tests. Happy to discuss wether we should setup this or not.

### Checklist

### DIFF
--- a/integrations/elasticsearch/src/haystack_integrations/document_stores/elasticsearch/document_store.py
+++ b/integrations/elasticsearch/src/haystack_integrations/document_stores/elasticsearch/document_store.py
@@ -87,8 +87,9 @@ class ElasticsearchDocumentStore:
         For the full list of supported kwargs, see the official Elasticsearch
         [reference](https://elasticsearch-py.readthedocs.io/en/stable/api.html#module-elasticsearch)
 
-        Authentication is provided via Secret objects, which by default are loaded from environment variables. 
-        You can either provide both `api_key_id` and `api_key`, or just `api_key` containing a base64-encoded string of `id:secret`. Secret instances can also be loaded from a token using the `Secret.from_token()` method.
+        Authentication is provided via Secret objects, which by default are loaded from environment variables.
+        You can either provide both `api_key_id` and `api_key`, or just `api_key` containing a base64-encoded string
+        of `id:secret`. Secret instances can also be loaded from a token using the `Secret.from_token()` method.
 
         :param hosts: List of hosts running the Elasticsearch client.
         :param custom_mapping: Custom mapping for the index. If not provided, a default mapping will be used.
--- a/integrations/elasticsearch/src/haystack_integrations/document_stores/elasticsearch/document_store.py
+++ b/integrations/elasticsearch/src/haystack_integrations/document_stores/elasticsearch/document_store.py
@@ -12,7 +12,6 @@ from haystack.dataclasses import Document
 from haystack.document_stores.errors import DocumentStoreError, DuplicateDocumentError
 from haystack.document_stores.types import DuplicatePolicy
 from haystack.utils import Secret, deserialize_secrets_inplace
-from haystack.utils.auth import TokenSecret
 from haystack.version import __version__ as haystack_version
 
 from elasticsearch import AsyncElasticsearch, Elasticsearch, helpers
@@ -42,7 +41,7 @@ class ElasticsearchDocumentStore:
     Usage example (Elastic Cloud):
     ```python
     from haystack_integrations.document_stores.elasticsearch import ElasticsearchDocumentStore
-    document_store = ElasticsearchDocumentStore(cloud_id="YOUR_CLOUD_ID", api_key="YOUR_API_KEY")
+    document_store = ElasticsearchDocumentStore(api_key_id="YOUR_CLOUD_ID", api_key="YOUR_API_KEY")
     ```
 
     Usage example (self-hosted Elasticsearch instance):
@@ -258,10 +257,8 @@ class ElasticsearchDocumentStore:
             hosts=self._hosts,
             custom_mapping=self._custom_mapping,
             index=self._index,
-            api_key=self._api_key.to_dict() if self._api_key and not isinstance(self._api_key, TokenSecret) else None,
-            api_key_id=self._api_key_id.to_dict()
-            if self._api_key_id and not isinstance(self._api_key, TokenSecret)
-            else None,
+            api_key=self._api_key.to_dict(),
+            api_key_id=self._api_key_id.to_dict(),
             embedding_similarity_function=self._embedding_similarity_function,
             **self._kwargs,
         )
@@ -276,7 +273,7 @@ class ElasticsearchDocumentStore:
         :returns:
             Deserialized component.
         """
-        deserialize_secrets_inplace(data, keys=["ELASTIC_API_KEY", "ELASTIC_API_KEY_ID"])
+        deserialize_secrets_inplace(data, keys=["api_key", "api_key_id"])
         return default_from_dict(cls, data)
 
     def count_documents(self) -> int:
--- a/integrations/elasticsearch/src/haystack_integrations/document_stores/elasticsearch/document_store.py
+++ b/integrations/elasticsearch/src/haystack_integrations/document_stores/elasticsearch/document_store.py
@@ -16,6 +16,7 @@ from haystack.version import __version__ as haystack_version
 
 from elasticsearch import AsyncElasticsearch, Elasticsearch, helpers
 
+from haystack.utils.auth import TokenSecret
 from .filters import _normalize_filters
 
 logger = logging.getLogger(__name__)
@@ -84,13 +85,22 @@ class ElasticsearchDocumentStore:
         For the full list of supported kwargs, see the official Elasticsearch
         [reference](https://elasticsearch-py.readthedocs.io/en/stable/api.html#module-elasticsearch)
 
+        Authentication can be provided in 2 ways. As environment variables, both ELASTIC_API_KEY_ID and ELASTIC_API_KEY
+        must be defined or alternatively, only ELASTIC_API_KEY needs to be defined, containing a base64-encoded string
+        that encodes both id and secret (separated by “:”).
+
+        Alternatively, authentication can be provided as a Secret. The api_key_id and api_key parameters are
+        provided as a Secret, or only api_key is provided as a Secret, which is a base64-encoded string that encodes
+        both id and secret (separated by “:”).
+
         :param hosts: List of hosts running the Elasticsearch client.
         :param custom_mapping: Custom mapping for the index. If not provided, a default mapping will be used.
         :param index: Name of index in Elasticsearch.
-        :param api_key: API key for authenticating with Elasticsearch. If not provided, the client will attempt to
-                        connect without authentication.
-        :param api_key_id: API key ID for authenticating with Elasticsearch. If not provided, the client will attempt to
-                        connect without authentication.
+
+        :param api_key: A Secret object containing the API key for authenticating or base64-encoded with the
+                        concatenated id and secret for authenticating with Elasticsearch (separated by “:”).
+        :param api_key_id: A Secret object containing the API key ID for authenticating with Elasticsearch.
+
         :param embedding_similarity_function: The similarity function used to compare Documents embeddings.
             This parameter only takes effect if the index does not yet exist and is created.
             To choose the most appropriate function, look for information about your embedding model.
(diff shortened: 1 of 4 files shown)


integrations/elasticsearch/src/haystack_integrations/document_stores/elasticsearch/document_store.py
```diff
@@ -38,13 +41,13 @@ class ElasticsearchDocumentStore:
 
     Usage example (Elastic Cloud):
     ```python
-    from haystack.document_store.elasticsearch import ElasticsearchDocumentStore
+    from haystack_integrations.document_stores.elasticsearch import ElasticsearchDocumentStore
```
sjrl, line 43: good catch!

integrations/elasticsearch/src/haystack_integrations/document_stores/elasticsearch/document_store.py
```diff
@@ -63,6 +66,8 @@ def __init__(
         hosts: Optional[Hosts] = None,
         custom_mapping: Optional[Dict[str, Any]] = None,
         index: str = "default",
+        api_key: Optional[Secret] = None,
+        api_key_id: Optional[Secret] = None,
```
sjrl: Could we make the default value of `api_key` be `Secret.from_env_var("ELASTIC_API_KEY")`? It seems like from your docstrings that no matter which auth method you choose that env variable always needs to be set right?

integrations/elasticsearch/src/haystack_integrations/document_stores/elasticsearch/document_store.py
```diff
@@ -94,6 +110,8 @@ def __init__(
         self._client: Optional[Elasticsearch] = None
         self._async_client: Optional[AsyncElasticsearch] = None
         self._index = index
+        self.api_key = api_key
+        self.api_key_id = api_key_id
```
sjrl: I think it's best if we stay consistent and keep these as private variables like the rest.

integrations/elasticsearch/src/haystack_integrations/document_stores/elasticsearch/document_store.py
davidsbatista: If the ES is not configured with auth, there's no need to set those; that's the current case - I think I can add them as a default with no strict (Secret).  If they are defined, they will be used to authenticate; if not, it assumes it's connecting to an ES instance without auth configured.

integrations/elasticsearch/src/haystack_integrations/document_stores/elasticsearch/document_store.py
```diff
@@ -38,13 +41,13 @@ class ElasticsearchDocumentStore:
 
     Usage example (Elastic Cloud):
     ```python
-    from haystack.document_store.elasticsearch import ElasticsearchDocumentStore
+    from haystack_integrations.document_stores.elasticsearch import ElasticsearchDocumentStore
     document_store = ElasticsearchDocumentStore(cloud_id="YOUR_CLOUD_ID", api_key="YOUR_API_KEY")
```
sjrl: We should probably update the example to use the new Secrets right?

integrations/elasticsearch/src/haystack_integrations/document_stores/elasticsearch/document_store.py
```diff
@@ -191,6 +258,10 @@ def to_dict(self) -> Dict[str, Any]:
             hosts=self._hosts,
             custom_mapping=self._custom_mapping,
             index=self._index,
+            api_key=self._api_key.to_dict() if self._api_key and not isinstance(self._api_key, TokenSecret) else None,
```
sjrl: I don't think we need this check for `TokenSecret`. We've been following the policy that we expect `to_dict` to fail if someone tries to serialize a component or a doc store and passed in a Secret constructed from a token rather than an env variable.

integrations/elasticsearch/src/haystack_integrations/document_stores/elasticsearch/document_store.py
```diff
@@ -205,6 +276,7 @@ def from_dict(cls, data: Dict[str, Any]) -> "ElasticsearchDocumentStore":
         :returns:
             Deserialized component.
         """
+        deserialize_secrets_inplace(data, keys=["ELASTIC_API_KEY", "ELASTIC_API_KEY_ID"])
```
sjrl: The keys here are meant to be init parameter names not the env variables.

integrations/elasticsearch/src/haystack_integrations/document_stores/elasticsearch/document_store.py
davidsbatista: ok, I will change it so it fails on purpose

integrations/elasticsearch/src/haystack_integrations/document_stores/elasticsearch/document_store.py
```diff
@@ -38,13 +40,13 @@ class ElasticsearchDocumentStore:
 
     Usage example (Elastic Cloud):
     ```python
-    from haystack.document_store.elasticsearch import ElasticsearchDocumentStore
-    document_store = ElasticsearchDocumentStore(cloud_id="YOUR_CLOUD_ID", api_key="YOUR_API_KEY")
+    from haystack_integrations.document_stores.elasticsearch import ElasticsearchDocumentStore
+    document_store = ElasticsearchDocumentStore(api_key_id="YOUR_CLOUD_ID", api_key="YOUR_API_KEY")
```
sjrl: String arguments would no longer work here right?
```suggestion
    document_store = ElasticsearchDocumentStore(
        api_key_id=Secret.from_env_var("ELASTIC_API_KEY_ID", strict=False),
        api_key=Secret.from_env_var("ELASTIC_API_KEY", strict=False),
    )
```

integrations/elasticsearch/src/haystack_integrations/document_stores/elasticsearch/document_store.py
```diff
@@ -158,6 +179,51 @@ def _ensure_initialized(self):
...
+
+        3) There's no authentication, neither api_key nor api_key_id are provided as a Secret nor defined as
+           environment variables. In this case, the client will connect without authentication.
+
+        returns:
+            api_key: Optional[Union[str, Tuple[str, str]]]
+
+        """
+
+        api_key: Optional[Union[str, Tuple[str, str]]] = None  # make the type checker happy
```
sjrl: This is only for type checking so we don't actually need to set it to None here
```suggestion
        api_key: Optional[Union[str, Tuple[str, str]]]  # make the type checker happy
```

integrations/elasticsearch/src/haystack_integrations/document_stores/elasticsearch/document_store.py
```diff
@@ -158,6 +179,51 @@ def _ensure_initialized(self):
...
+
+        # Scenario 1: both are found, use them
+        if api_key_id_resolved and api_key_resolved:
+            api_key = (api_key_id_resolved, api_key_resolved)
+            return api_key
+
+        # Scenario 2: only api_key is set, must be a base64-encoded string that encodes id and secret (separated by “:”)
+        elif api_key_resolved and not api_key_id_resolved:
+            api_key = self._api_key.resolve_value()
+            return api_key
```
sjrl:
```suggestion
            return api_key_resolved
```

integrations/elasticsearch/src/haystack_integrations/document_stores/elasticsearch/document_store.py
davidsbatista: oh right!

integrations/elasticsearch/src/haystack_integrations/document_stores/elasticsearch/document_store.py
```diff
@@ -158,6 +179,51 @@ def _ensure_initialized(self):
 
             self._initialized = True
 
+    def handle_auth(self) -> Optional[Union[str, Tuple[str, str]]]:
```
sjrl: could we also make this method private?

integrations/elasticsearch/src/haystack_integrations/document_stores/elasticsearch/document_store.py
```diff
@@ -80,9 +84,20 @@ def __init__(
         For the full list of supported kwargs, see the official Elasticsearch
         [reference](https://elasticsearch-py.readthedocs.io/en/stable/api.html#module-elasticsearch)
 
+        Authentication can be provided in 2 ways. As environment variables, both ELASTIC_API_KEY_ID and ELASTIC_API_KEY
+        must be defined or alternatively, only ELASTIC_API_KEY needs to be defined, containing a base64-encoded string
+        that encodes both id and secret (separated by “:”).
+
+        Alternatively, authentication can be provided as a Secret. The api_key_id and api_key parameters are
+        provided as a Secret, or only api_key is provided as a Secret, which is a base64-encoded string that encodes
+        both id and secret (separated by “:”).
```
sjrl: How does updating the docs to something like this sound?
```suggestion
        Authentication is provided via Secret objects, which by default are loaded from environment variables. 
        You can either provide both `api_key_id` and `api_key`, or just `api_key` containing a base64-encoded string of `id:secret`. Secret instances can also be loaded from a token using the `Secret.from_token()` method.
```

integrations/elasticsearch/src/haystack_integrations/document_stores/elasticsearch/document_store.py
```diff
@@ -158,6 +179,51 @@ def _ensure_initialized(self):
...
+
+        1) As an environment variables. Both ELASTIC_API_KEY_ID and ELASTIC_API_KEY must be defined, alternatively
+         only ELASTIC_API_KEY can be defined, which is a base64-encoded string that encodes both the id and secret.
+         They can be provided as Secret, the api_key_id and api_key parameters are provided as a Secret.
+
+        2) If only api_key_id is provided, either as a Secret or as an environment variable, fail since api_key_id
+           alone is not enough for authentication.
+
+        3) There's no authentication, neither api_key nor api_key_id are provided as a Secret nor defined as
+           environment variables. In this case, the client will connect without authentication.
```
sjrl: Could we update the scenarios here to match the dev comments below? Since the values are always provided as Secrets but the scenarios are just about whether which ones are present or not right?

integrations/elasticsearch/src/haystack_integrations/document_stores/elasticsearch/document_store.py
```diff
@@ -158,6 +179,51 @@ def _ensure_initialized(self):
...
+
+        returns:
```
sjrl:
```suggestion
        :returns:
```

integrations/elasticsearch/tests/test_bm25_retriever.py
```diff
@@ -38,6 +39,20 @@ def test_to_dict(_mock_elasticsearch_client):
...
+                        "strict": False,
+                        "type": "env_var",
+                    },
+                    "api_key_id": {
+                        "env_vars": [
+                            "ELASTIC_API_KEY_ID",
+                        ],
+                        "strict": False,
+                        "type": "env_var",
+                    },
```
sjrl, line 55: Could we also update the `from_dict` test below to add `api_key` and `api_key_id` in the dict that is being deserialized?

sjrl: actually never mind I see this is for the retriever and that you do this already in the doc store tests

integrations/elasticsearch/src/haystack_integrations/document_stores/elasticsearch/document_store.py
sjrl: Probably worth listing this PR as a breaking change then I think. Not a big deal, but good to do and perhaps releasing a new major version of the package.

integrations/elasticsearch/src/haystack_integrations/document_stores/elasticsearch/document_store.py
davidsbatista: much better

integrations/elasticsearch/src/haystack_integrations/document_stores/elasticsearch/document_store.py
davidsbatista: agree - I will make a major release when it's merged.

sjrl: sounds good! I think the exclamation point in the title is meant to come after the tag so `feat!:` at least from what I can see in other older PRs

davidsbatista: upps - corrected it

PR #2321, opened 2025-09-25
Feat/adding delete all docs to open search document store

davidsbatista:
### Related Issues

- fixes #issue-number

### Checklist

### DIFF
--- a/integrations/opensearch/src/haystack_integrations/document_stores/opensearch/document_store.py
+++ b/integrations/opensearch/src/haystack_integrations/document_stores/opensearch/document_store.py
@@ -87,7 +87,7 @@ class OpenSearchDocumentStore:
         Creates a new OpenSearchDocumentStore instance.
 
         The ``embeddings_dim``, ``method``, ``mappings``, and ``settings`` arguments are only used if the index does not
-        exists and needs to be created. If the index already exists, its current configurations will be used.
+        exist and needs to be created. If the index already exists, its current configurations will be used.
 
         For more information on connection parameters, see the [official OpenSearch documentation](https://opensearch.org/docs/latest/clients/python-low-level/#connecting-to-opensearch)
 
@@ -107,7 +107,7 @@ class OpenSearchDocumentStore:
         :param settings: The settings of the index to be created. Please see the [official OpenSearch docs](https://opensearch.org/docs/latest/search-plugins/knn/knn-index/#index-settings)
             for more information. Defaults to {"index.knn": True}
         :param create_index: Whether to create the index if it doesn't exist. Defaults to True
-        :param http_auth: http_auth param passed to the underying connection class.
+        :param http_auth: http_auth param passed to the underlying connection class.
             For basic authentication with default connection class `Urllib3HttpConnection` this can be
             - a tuple of (username, password)
             - a list of [username, password]
@@ -319,7 +319,8 @@ class OpenSearchDocumentStore:
         assert self._async_client is not None
         return (await self._async_client.count(index=self._index))["count"]
 
-    def _deserialize_search_hits(self, hits: List[Dict[str, Any]]) -> List[Document]:
+    @staticmethod
+    def _deserialize_search_hits(hits: List[Dict[str, Any]]) -> List[Document]:
         out = []
         for hit in hits:
             data = hit["_source"]
@@ -344,12 +345,12 @@ class OpenSearchDocumentStore:
     def _search_documents(self, request_body: Dict[str, Any]) -> List[Document]:
         assert self._client is not None
         search_results = self._client.search(index=self._index, body=request_body)
-        return self._deserialize_search_hits(search_results["hits"]["hits"])
+        return OpenSearchDocumentStore._deserialize_search_hits(search_results["hits"]["hits"])
 
     async def _search_documents_async(self, request_body: Dict[str, Any]) -> List[Document]:
         assert self._async_client is not None
         search_results = await self._async_client.search(index=self._index, body=request_body)
-        return self._deserialize_search_hits(search_results["hits"]["hits"])
+        return OpenSearchDocumentStore._deserialize_search_hits(search_results["hits"]["hits"])
 
     def filter_documents(self, filters: Optional[Dict[str, Any]] = None) -> List[Document]:
         """
@@ -418,7 +419,8 @@ class OpenSearchDocumentStore:
             "stats_only": False,
         }
 
-    def _process_bulk_write_errors(self, errors: List[Dict[str, Any]], policy: DuplicatePolicy) -> None:
+    @staticmethod
+    def _process_bulk_write_errors(errors: List[Dict[str, Any]], policy: DuplicatePolicy) -> None:
         if len(errors) == 0:
             return
 
@@ -461,7 +463,7 @@ class OpenSearchDocumentStore:
 
         bulk_params = self._prepare_bulk_write_request(documents=documents, policy=policy, is_async=False)
         documents_written, errors = bulk(**bulk_params)
-        self._process_bulk_write_errors(errors, policy)
+        OpenSearchDocumentStore._process_bulk_write_errors(errors, policy)
         return documents_written
 
     async def write_documents_async(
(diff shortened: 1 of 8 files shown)


integrations/opensearch/src/haystack_integrations/document_stores/opensearch/document_store.py
```diff
@@ -525,6 +528,42 @@ async def delete_documents_async(self, document_ids: List[str]) -> None:
...
         await async_bulk(**self._prepare_bulk_delete_request(document_ids=document_ids, is_async=True))
 
+    def delete_all_documents(self) -> None:
+        """
+        Deletes all documents in the document store by deleting and recreating the index.
+
+        A fast way to clear all documents from the document store while preserving any index settings and mappings.
+        """
+        self._ensure_initialized()
+        assert self._client is not None
```
julian-risch, line 546: It's consistent with the other methods to have this assert here. Therefore, I agree it's good to have it here.

However, I believe the only case it covers is if self._client was set at some point and then for some reason is set to None again. We have the same assert in `_ensure_index_exists`, which get's called as part of `_ensure_initialized` when it is run for the first time.

anakin87:
Just a quick thought... I am not sure deleting and recreating the index is a good idea.

As a user, I'd expect `delete_all_documents` to only remove the data, not touch the index itself. And even though we recommend managing the DB via Haystack, some users might still have changed index attributes manually. In that case, recreating the index could cause surprises.

A bulk delete might be a safer option. Maybe it's also worth checking how this was done in [Haystack 1.x](https://github.com/deepset-ai/haystack/tree/v1.26.x/haystack/document_stores).

integrations/opensearch/src/haystack_integrations/document_stores/opensearch/document_store.py
```diff
@@ -525,6 +528,42 @@ async def delete_documents_async(self, document_ids: List[str]) -> None:
...
+
+        # Delete index
+        self._client.indices.delete(index=self._index)
+
+        # Recreate with mappings and settings
+        body = {"mappings": self._mappings, "settings": self._settings}
```
julian-risch: Maybe you talked to Tobi about this: do we always want to recreate the index when we delete all documents? I agree that the default should be to recreate the index.
I could imagine a non-breaking extension of delete_all_documents where we add a parameter `bool: recreate_index=True` and then users could choose to not recreate the index.

integrations/opensearch/src/haystack_integrations/document_stores/opensearch/document_store.py
```diff
@@ -525,6 +528,42 @@ async def delete_documents_async(self, document_ids: List[str]) -> None:
...
+        assert self._async_client is not None
+
+        try:
+            # Delete index
+            await self._async_client.indices.delete(index=self._index)
+
+            # Recreate with mappings and settings
+            body = {"mappings": self._mappings, "settings": self._settings}
+            await self._async_client.indices.create(index=self._index, body=body)
+        except Exception as e:
```
julian-risch, line 607: If we have a try except here with this error message, I'd say it makes sense to add the same to the sync implementation of delete_all_documents. It's a few extra lines of code but ensures we have more similar behavior in both implementations, delete_all_documents and delete_all_documents_async

davidsbatista:
I was investigating into it and for both Elastic and Open, specially for large volumes of data, this is the most efficient way to do it.

In both cases the indexes are recreated with the same mappings/settings.

davidsbatista:
What about something like: 

```python
def delete_all_documents(self, recreate_index=False)
```

By default we apply a usual delete operation, and if `recreate_index=True`, we just recreate the index using the mappings/settings defined by the user or the default ones.

anakin87:
> What about something like:
> 
> ```python
> def delete_all_documents(self, recreate_index=False)
> ```

I would prefer this idea!

integrations/opensearch/tests/test_document_store.py
```diff
@@ -468,3 +469,27 @@ def filter_documents_no_embedding_returned(
...
+            Document(id="3", content="Third document", meta={"category": "other"}),
+        ]
+        document_store.write_documents(docs)
+        assert document_store.count_documents() == 3
+
+        # delete all documents
+        document_store.delete_all_documents()
+        assert document_store.count_documents() == 0
+
+        # verify index still exists and can accept new documents and retrieve
```
julian-risch: We could extend the test by checking that the actual index configuration remains unchanged.

While I see the advantages of a test that checks that writing data works after deletion just as before, there is also a downside. If in future there is a bug in how we write or retrieve documents, the test `delete_all_documents` will also fail. Even if there is no bug at all in how we delete documents. I prefer limiting the unit tests as much as possible only to the functionality that we want to test.

```suggestion
        # Capture index structure before deletion
        assert document_store._client is not None
        index_info_before = document_store._client.indices.get(index=document_store._index)
        mappings_before = index_info_before[document_store._index]["mappings"]
        settings_before = index_info_before[document_store._index]["settings"]

        # delete all documents
        document_store.delete_all_documents()
        assert document_store.count_documents() == 0

        # Verify index structure is preserved
        index_info_after = document_store._client.indices.get(index=document_store._index)
        mappings_after = index_info_after[document_store._index]["mappings"]
        settings_after = index_info_after[document_store._index]["settings"]
        
        assert mappings_after == mappings_before, "delete_all_documents shpuld preserve index mappings"
        assert settings_after == settings_before, "delete_all_documents should preserve index settings"

        # verify index can accept new documents and retrieve
```

integrations/opensearch/tests/test_document_store_async.py
```diff
@@ -241,3 +245,28 @@ async def test_delete_documents(self, document_store: OpenSearchDocumentStore):
...
+            Document(id="3", content="Third document", meta={"category": "other"}),
+        ]
+        await document_store.write_documents_async(docs)
+        assert await document_store.count_documents_async() == 3
+
+        # delete all documents
+        await document_store.delete_all_documents_async()
+        assert await document_store.count_documents_async() == 0
+
+        # verify index still exists and can accept new documents and retrieve
```
julian-risch:
```suggestion
        # Capture index structure before deletion
        assert document_store._async_client is not None
        index_info_before = await document_store._async_client.indices.get(index=document_store._index)
        mappings_before = index_info_before[document_store._index]["mappings"]
        settings_before = index_info_before[document_store._index]["settings"]

        # delete all documents
        await document_store.delete_all_documents_async()
        assert await document_store.count_documents_async() == 0

        # Verify index structure is preserved
        index_info_after = await document_store._async_client.indices.get(index=document_store._index)
        mappings_after = index_info_after[document_store._index]["mappings"]
        settings_after = index_info_after[document_store._index]["settings"]
        
        assert mappings_after == mappings_before, "delete_all_documents_async should preserve index mappings"
        assert settings_after == settings_before, "delete_all_documents_async should preserve index settings"

        # verify index can accept new documents and retrieve
```

similar suggestion as in `test_delete_all_documents`

julian-risch (changes requested):
Looks quite good to me already. My main suggestion is that if we have a try except in delete_all_documents_async, it makes sense to add the same to the sync implementation of delete_all_documents to ensures consistent behavior.
Other than that, I have a suggestion for the tests to make sure they only test one thing. You could create a separate test, for example `test_index_functionality_after_delete_all_documents` for testing writing and retrieving of documents after deleting all documents. Something like:
```python
def test_index_functionality_after_delete_all_documents(self, document_store):
    """Test that documents can be written and retrieved after delete_all_documents"""
    document_store.write_documents([Document(id="1", content="Test")])  # or add more documents
    document_store.delete_all_documents()
    
    new_doc = Document(id="2", content="New test")
    document_store.write_documents([new_doc])  # or add more documents
    assert document_store.count_documents() == 1
    
    results = document_store.filter_documents()
    assert len(results) == 1
    assert results[0].content == "New test"
```

integrations/opensearch/src/haystack_integrations/document_stores/opensearch/document_store.py
davidsbatista: > I agree that the default should be to recreate the index.

I think the default should rather be to delete all the documents, and the optional is to recreate the index with the same settings/mappings as an efficient way to delete all the documents.

integrations/opensearch/tests/test_document_store.py
davidsbatista: good idea 👍🏽

julian-risch (approved):
Looks good to me! 👍  `_prepare_delete_all_request` in particular is a good alternative to recreating the index and the default of `delete_all_documents(recreate_index=False)` makes sense to me.

PR #2339, opened 2025-09-26
feat: add s3_key_meta_field

tstadel:
### Related Issues
In the simplest case (`file_name_meta_key == s3_key_meta_key`):
- you don't need to configure anything
- your file is expected to have the same key on your s3 bucket as its file_name (e.g. `my_file.txt`)
- you can add prefix and suffixes of course, so if your s3_key would be `org/workspace/my_file.txt` you can easily configure a prefix `org/workspace/`

In the platform case  (`file_name_meta_key != s3_key_meta_key`):
- `s3_key_meta_key='file_id'`
- `file_name_meta_key='file_name'`
- and we still need to set prefixes and suffixes

### Checklist

- I have read the [contributors guidelines](https://github.com/deepset-ai/haystack-core-integrations/blob/main/CONTRIBUTING.md) and the [code of conduct](https://github.com/deepset-ai/haystack-core-integrations/blob/main/CODE_OF_CONDUCT.md)
- I have updated the related issue with new insights and changes
- I added unit tests and updated the docstrings
- I've used one of the [conventional commit types](https://www.conventionalcommits.org/en/v1.0.0/) for my PR title: `fix:`, `feat:`, `build:`, `chore:`, `ci:`, `docs:`, `style:`, `refactor:`, `perf:`, `test:`.

PR #2336, opened 2025-09-26
feat: `S3Downloader` - add configurable suffix to `S3Storage`

anakin87:
### Related Issues

- fixes #issue-number

### Checklist

- I have read the [contributors guidelines](https://github.com/deepset-ai/haystack-core-integrations/blob/main/CONTRIBUTING.md) and the [code of conduct](https://github.com/deepset-ai/haystack-core-integrations/blob/main/CODE_OF_CONDUCT.md)
- I have updated the related issue with new insights and changes
- I added unit tests and updated the docstrings
- I've used one of the [conventional commit types](https://www.conventionalcommits.org/en/v1.0.0/) for my PR title: `fix:`, `feat:`, `build:`, `chore:`, `ci:`, `docs:`, `style:`, `refactor:`, `perf:`, `test:`.

PR #2343, opened 2025-09-26
feat: S3Downloader - add `s3_key_generation_function` param to customize S3 key generation

anakin87:
### Related Issues

This PR is meant to replace #2336 and #2339, to support customization without sacrificing UX 

### Proposed Changes:
- add `s3_key_generation_function` init param to customize S3 key generation
  - if provided, the function will be invoked to generate the S3 key (otherwise we'll keep using `Document.meta[file_name_meta_key]`)
  - if `S3_DOWNLOADER_PREFIX` env var is set, its value will be automatically prefixed to the generated S3 key

### How did you test it?
CI, new unit and integration tests

### Checklist

### DIFF
--- a/integrations/amazon_bedrock/src/haystack_integrations/components/downloaders/s3/s3_downloader.py
+++ b/integrations/amazon_bedrock/src/haystack_integrations/components/downloaders/s3/s3_downloader.py
@@ -66,7 +66,7 @@ class S3Downloader:
             By default, all file extensions are allowed.
         :param max_workers: The maximum number of workers to use for concurrent downloads.
         :param max_cache_size: The maximum number of files to cache.
-        :param file_name_meta_key: The name of the meta key that contains the file name to download.
+        :param file_name_meta_key: The name of the meta key that contains the file name to download. The file name will also be used to create local file path for download.
             By default, the `Document.meta["file_name"]` is used. If you want to use a
             different key in `Document.meta`, you can set it here.
         :param s3_key_generation_function: An optional function that generates the S3 key for the file to download.
--- a/integrations/amazon_bedrock/src/haystack_integrations/components/downloaders/s3/s3_downloader.py
+++ b/integrations/amazon_bedrock/src/haystack_integrations/components/downloaders/s3/s3_downloader.py
@@ -5,12 +5,13 @@
 import os
 from concurrent.futures import ThreadPoolExecutor
 from pathlib import Path
-from typing import Any, Dict, List, Optional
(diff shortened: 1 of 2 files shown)


tstadel (approved):
Thanks for adjusting!

sjrl (approved):
looks good!

integrations/amazon_bedrock/src/haystack_integrations/components/downloaders/s3/s3_downloader.py
```diff
@@ -186,8 +194,9 @@ def _download_file(self, document: Document) -> Optional[Document]:
             file_path.touch()
 
         else:
+            s3_key = self.s3_key_generation_function(document) if self.s3_key_generation_function else file_name
```
Amnah199, line 198:
```
file_name = document.meta.get(self.file_name_meta_key)
        if not file_name:
            logger.warning(
                f"Document missing required file name metadata key '{self.file_name_meta_key}'. Skipping download."
            )
            return None
```
@anakin87 We don't need `file_name` for this function right? With the above code segment, wouldnt it skip download even if the user provides a `s3_key_generation_function` and `file_name` is missing from Document?

anakin87: Yes, I did not alter much of the existing code. `file_name` is mandatory to create the local file path.

If `s3_key_generation_function` is not provided, `file_name` is used both as S3 key and local file name. If `s3_key_generation_function` is provided, `file_name` is only used for local file name.

Amnah199: Right! I think we can document this with more clarity.

integrations/amazon_bedrock/src/haystack_integrations/components/downloaders/s3/s3_downloader.py
```diff
@@ -67,6 +69,11 @@ def __init__(
         :param file_name_meta_key: The name of the meta key that contains the file name to download.
```
Amnah199: Something like
```suggestion
        :param file_name_meta_key: The name of the meta key that contains the file name to download. The file name will also be used to create local file path for download.
```

integrations/amazon_bedrock/src/haystack_integrations/components/downloaders/s3/s3_downloader.py
anakin87: accepted your suggestion

Issue #2349, opened 2025-09-29
AIMLAPI: missing tasks


anakin87:
AIMLAPI was introduced in #2267, but some tasks are missing. List below.

### Tasks
- [x] The code is documented with docstrings and was merged in the `main` branch
- [x] #2525
- [x] A new label named like `integration:<your integration name>` has been added to the list of labels for this [repository](https://github.com/deepset-ai/haystack-core-integrations/labels)
- [x] The [labeler.yml](https://github.com/deepset-ai/haystack-core-integrations/blob/main/.github/labeler.yml) file has been updated
- [x] missing workflow in https://github.com/deepset-ai/haystack-core-integrations/tree/main/.github and API key for integration test
- [x] The package has been released on PyPI
- [x] An integration tile with a usage example has been added to https://github.com/deepset-ai/haystack-integrations
- [x] The integration has been listed in the [Inventory section](https://github.com/deepset-ai/haystack-core-integrations#inventory) of this repo README
- [x] The feature was announced through social media

julian-risch:
Hi @D1m7asis , congrats on the merged PR #2267 . As a follow up, there are a couple smaller tasks listed in the issue above.
Could you support us please with the following?

- An integration tile with a usage example has been added to https://github.com/deepset-ai/haystack-integrations
- The integration has been listed in the [Inventory section](https://github.com/deepset-ai/haystack-core-integrations#inventory) of this repo README

sjrl:
Closing as done

PR #2352, opened 2025-09-30
ci: run integration tests with Haystack main for integrations inheriting from `OpenAIChatGenerator`

anakin87:
### Related Issues
(related to https://github.com/deepset-ai/haystack/issues/9769)

We encountered several incompatibilities when changing `OpenAIChatGenerator` in Haystack, since several integrations depend on it.
The unit tests (which already run nightly with Haystack main) are not enough to guarantee stability across integrations.

For this reason, I suggest we run nightly integration tests against Haystack main for the integrations that explicitly inherit from `OpenAIChatGenerator`.

### Proposed Changes:
- for integrations inheriting from `OpenAIChatGenerator`, nightly run all tests (unit + integration) with Haystack main
- other minor fixes

### How did you test it?
CI

### Notes for the reviewer
- I discovered that AIMLAPI does not have a workflow, so I added an item to #2349.
- Nvidia tests are failing for unrelated reasons

### Checklist

### DIFF
--- a/.github/workflows/together_ai.yml
+++ b/.github/workflows/together_ai.yml
@@ -32,7 +32,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-latest, windows-latest, macos-latest]
-        python-version: ["3.9", "3.13"]
+        python-version: ["3.10", "3.13"]
 
     steps:
       - name: Support longpaths
@@ -50,11 +50,11 @@ jobs:
       - name: Install Hatch
         run: pip install --upgrade hatch
       - name: Lint
-        if: matrix.python-version == '3.9' && runner.os == 'Linux'
+        if: matrix.python-version == '3.10' && runner.os == 'Linux'
         run: hatch run fmt-check && hatch run test:types
 
       - name: Generate docs
-        if: matrix.python-version == '3.9' && runner.os == 'Linux'
+        if: matrix.python-version == '3.10' && runner.os == 'Linux'
         run: hatch run docs
 
       - name: Run tests


.github/workflows/llama_stack.yml
```diff
@@ -43,7 +43,7 @@ jobs:
           max_attempts: 3
           command: |
             curl -fsSL https://ollama.com/install.sh | sh
-            ollama serve &
+            nohup ollama serve > ollama.log 2>&1 &
```
anakin87, line 46: aligned with https://github.com/deepset-ai/haystack-core-integrations/pull/2322

.github/workflows/together_ai.yml
```diff
@@ -32,7 +32,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-latest, windows-latest, macos-latest]
-        python-version: ["3.10", "3.13"]
+        python-version: ["3.9", "3.13"]
```
anakin87: we usually test with the lowest and highest supported versions (unless there are specific reasons)

.github/labeler.yml
```diff
@@ -174,7 +174,7 @@ integration:stackit:
       - any-glob-to-any-file: "integrations/stackit/**/*"
       - any-glob-to-any-file: ".github/workflows/stackit.yml"
 
-integration:together_ai:
+integration:together-ai:
```
sjrl, line 177: I think the label is indeed `integration:together_ai` right?

.github/workflows/together_ai.yml
sjrl: @Amnah199 was there a reason to not run together-ai with python 3.9?

.github/labeler.yml
anakin87: yes but it's wrong. I hope that this will be fixed with this PR.

sjrl: sorry I think I'm misunderstanding. Should the label actually be `integration:together-ai`?

anakin87: yes, it would be aligned with others
https://github.com/deepset-ai/haystack-core-integrations/blob/d13eeee77cdac774f563cc92544e39c4a7b75769/.github/labeler.yml#L67

sjrl: sounds good!

.github/workflows/together_ai.yml
sjrl: If this integration does work with `3.9` we should also update the pyproject.toml to reflect that as well

anakin87: Ok I reverted this change. I'll create an issue for that. -> https://github.com/deepset-ai/haystack-core-integrations/issues/2353

PR #2354, opened 2025-09-30
feat: implement `delete_all_documents` for weaviate integration

### LINKED ISSUE #2309: add `delete_all_documents()` operation to `WeaviateDocumentStore`
```python
client.batch.delete_objects(
	class_name=class_name, 
	where={},  # Empty filter = delete all
	output="minimal" # Return minimal response (faster))
)
```

phigep:
### Related Issues

- fixes #2309 

### Proposed Changes
- Add `delete_all_documents(self, recreate_index: bool = False, deletion_batch_size: int = 1000) -> None`.
- **Two strategies:**
  - `recreate_index=True`: drop the collection and faithfully recreate it from the live config (`config.get().to_dict()`), preserving schema/vector/sharding settings. This is by far the most performant version.
  - `recreate_index=False`: keep the collection and delete objects in **batches** using `collection.iterator(...)+ delete_many(Filter.by_id().contains_any(...))` to avoid offset pagination limits and reduce memory pressure.
- Expose `deletion_batch_size` (default `1000`) to control delete chunk size. (could just not expose this)
- Add integration tests:
  - `test_delete_all_documents`
  - `test_delete_all_documents_recreate`

I did this in analogy to #2321. The version suggested in the initial #2309 issue post did not work with the batch wrapper of the weaviate package. Another issue with this is that weaviate seems to set some sort of max. deleted objects, which means we have to iterate either way.  (https://docs.weaviate.io/weaviate/manage-objects/delete)

Honestly, I would just always use the recreate_index strategy, especially because we get the up-to-date config on deletion from the server. The recreate strategy doesn't scale with the size of the index which is so much nicer. What exactly is the fear regarding just using this strategy?  

### Usage example
```
# slower but doesnt touch the collection/index itself
document_store.delete_all_documents()

# much faster, drops and recreates collection
document_store.delete_all_documents(recreate_index=True)
```
### Checklist

### DIFF
--- a/integrations/weaviate/src/haystack_integrations/document_stores/weaviate/document_store.py
+++ b/integrations/weaviate/src/haystack_integrations/document_stores/weaviate/document_store.py
@@ -543,7 +543,6 @@ class WeaviateDocumentStore:
                     logger.warning(
                         "Not all documents in the batch have been deleted. "
                         "Make sure to specify a deletion `batch_size` which is less than `QUERY_MAXIMUM_RESULTS`.",
-                        stacklevel=2,
                     )
                 uuids.clear()
 
@@ -553,7 +552,6 @@ class WeaviateDocumentStore:
                 logger.warning(
                     "Not all documents have been deleted. "
                     "Make sure to specify a deletion `batch_size` which is less than `QUERY_MAXIMUM_RESULTS`.",
-                    stacklevel=2,
                 )
 
     def _bm25_retrieval(


integrations/weaviate/tests/test_document_store.py
```diff
@@ -794,3 +794,17 @@ def test_connect_to_local(self):
...
+        assert document_store.count_documents() == 2
+        document_store.delete_all_documents()
+        assert document_store.count_documents() == 0
+
+    def test_delete_all_documents_recreate(self, document_store):
+        docs = [Document(content="test doc 1"), Document(content="test doc 2")]
+        assert document_store.write_documents(docs) == 2
+        assert document_store.count_documents() == 2
+        document_store.delete_all_documents(recreate_index=True)
+        assert document_store.count_documents() == 0
```
anakin87, line 816: here I would also test that the recreated collection configuration matches the original one

integrations/weaviate/tests/test_document_store.py
```diff
@@ -794,3 +794,17 @@ def test_connect_to_local(self):
     def test_connect_to_embedded(self):
         document_store = WeaviateDocumentStore(embedded_options=EmbeddedOptions())
         assert document_store.client
+
```
anakin87, line 798: let's also add a test with `deletion_batch_size`

We set it to a low value (e.g. 2), we write 5 docs, we check that everything is deleted

anakin87 (changes requested):
Thanks for the contribution.
I added two comments about tests.

I'll ask @davidsbatista to take a look as well

integrations/weaviate/tests/test_document_store.py
phigep: Good point. Actually, will also have to check whether that is larger than the environment variable of max queries by weaviate I mentioned before.

integrations/weaviate/src/haystack_integrations/document_stores/weaviate/document_store.py
```diff
@@ -505,6 +505,38 @@ def delete_documents(self, document_ids: List[str]) -> None:
         weaviate_ids = [generate_uuid5(doc_id) for doc_id in document_ids]
         self.collection.data.delete_many(where=weaviate.classes.query.Filter.by_id().contains_any(weaviate_ids))
 
+    def delete_all_documents(self, *, recreate_index: bool = False, deletion_batch_size: int = 1000) -> None:
+        """
+        Deletes all documents in a collection.
+        If recreate_index is False, it keeps the collection but deletes documents iteratively.
```
davidsbatista, line 512: Add an empty line between the first sentence and the rest of the docs. That's how it's on haystack, we want at some point to also add the same linting rules for docstrings in integrations.

davidsbatista:
Hi @phigep - thanks for your contribution! Appreciate it!

I just have a small comment regarding the docstrings, other than that I agree with the suggestions that @anakin87 did regarding the tests - afterwards we can merge.

Thanks once gain 👍🏽

phigep:
Alright, I hope its fine now! :smile: 
I had to add some stuff to fix the batch size issue. Specifically, the environment variable QUERY_MAXIMUM_RESULTS limits the max results that we can delete in one pass but its not accessible directly. 

Thanks for your reviews and time!

integrations/weaviate/src/haystack_integrations/document_stores/weaviate/document_store.py
```diff
@@ -505,6 +505,53 @@ def delete_documents(self, document_ids: List[str]) -> None:
...
+        # small helper function because we currently cant access the QUERY_MAXIMUM_RESULTS env variable at runtime.
+        def _delete_with_replay(ids: List[str], batch_size: int) -> int:
+            res = self.collection.data.delete_many(where=weaviate.classes.query.Filter.by_id().contains_any(ids))
+            # if deletion_batch_size was bigger than the max allowed -> shrink it
+            if getattr(res, "successful", None) is not None and res.successful < len(ids):
+                batch_size = max(res.successful, 1)
+                for i in range(0, len(ids), batch_size):
+                    chunk = ids[i : i + batch_size]
+                    self.collection.data.delete_many(where=weaviate.classes.query.Filter.by_id().contains_any(chunk))
+            return batch_size
```
anakin87: This approach seems a bit opaque for the user. 

I would simplify the code, only deleting objects using the given `deletion_batch_size`. If the final documents are not 0, we can log a warning saying something like: "Not all documents have been deleted. Make sure to specify a `deletion_batch_size` which is less than `QUERY_MAXIMUM_RESULTS`."

In short, I would remove `_delete_with_replay`.

integrations/weaviate/src/haystack_integrations/document_stores/weaviate/document_store.py
```diff
@@ -505,6 +505,53 @@ def delete_documents(self, document_ids: List[str]) -> None:
         weaviate_ids = [generate_uuid5(doc_id) for doc_id in document_ids]
         self.collection.data.delete_many(where=weaviate.classes.query.Filter.by_id().contains_any(weaviate_ids))
 
+    def delete_all_documents(self, *, recreate_index: bool = False, deletion_batch_size: int = 1000) -> None:
```
anakin87:
```suggestion
    def delete_all_documents(self, *, recreate_index: bool = False, batch_size: int = 1000) -> None:
```

nit: in this context, `batch_size` is enough

integrations/weaviate/src/haystack_integrations/document_stores/weaviate/document_store.py
```diff
@@ -505,6 +505,53 @@ def delete_documents(self, document_ids: List[str]) -> None:
...
+        """
+        Deletes all documents in a collection.
+
+        If recreate_index is False, it keeps the collection but deletes documents iteratively.
+        If recreate_index is True, the collection is dropped and faithfully recreated.
+        This is recommended for performance reasons.
+
+        :param recreate_index: Use drop and recreate strategy. (recommended for performance)
+        :param deletion_batch_size: Only relevant if recreate_index is false. Defines the deletion batch size.
```
anakin87: let's mention that this parameter must be < `QUERY_MAXIMUM_RESULTS ` and add a link to https://docs.weaviate.io/weaviate/manage-objects/delete#delete-all-objects

integrations/weaviate/tests/test_document_store.py
```diff
@@ -794,3 +794,38 @@ def test_connect_to_local(self):
...
+        new_config = document_store.client.collections.get(cls).config.get().to_dict()
+        assert previous_config == new_config
+
+    def test_delete_all_documents_batch_size_large(self, document_store):
+        # assume QUERY_MAXIMUM_RESULTS == 10000 with standard deployment
+        docs = [Document(content=str(i)) for i in range(0, 10005)]
+        assert document_store.write_documents(docs) == 10005
+        document_store.delete_all_documents(deletion_batch_size=20000)
+        assert document_store.count_documents() == 0
```
anakin87: Once you implement the simplified approach proposed in https://github.com/deepset-ai/haystack-core-integrations/pull/2354/files#r2397441473, let's transform this test to check that the warning is emitted

integrations/weaviate/src/haystack_integrations/document_stores/weaviate/document_store.py
phigep: That helper hurt my soul, I am glad I can put the burden on the user. :laughing:

anakin87 (approved):
I pushed some refinements, but looks good.
Thanks for the contribution!

PR #2357, opened 2025-10-02
chore: Add Agent pipeline deserialization test

vblagoje:
### Related Issues

- related to https://github.com/deepset-ai/haystack/issues/9807

### Proposed Changes:
Demo how MCPTool/set connect on init and they should delay connect to first tool use/warm_up stage to allow pipelines to set all the required secrets 
### Checklist

### DIFF
--- a/integrations/mcp/tests/test_mcp_toolset.py
+++ b/integrations/mcp/tests/test_mcp_toolset.py
@@ -7,9 +7,13 @@ import tempfile
 import time
 from unittest.mock import patch
 
+import haystack
+from haystack.dataclasses import ChatMessage
 import pytest
 import pytest_asyncio
 from haystack import logging
+from haystack.core.errors import DeserializationError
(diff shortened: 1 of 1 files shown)


integrations/mcp/tests/test_mcp_toolset.py
```diff
@@ -413,3 +416,83 @@ def subtract(a: int, b: int) -> int:
...
+                data={
+                    "agent": {
+                        "messages": [
+                            ChatMessage.from_user(
+                                text="Get comments on https://github.com/deepset-ai/haystack/pull/9849"
+                            )
+                        ]
+                    }
+                }
+            )
```
sjrl: I think to demonstrate this test fails when it does we should remove the `pipe.run` call. We want to explicitly see if `Pipeline.loads` causes the error.

sjrl (approved):
looks good!

PR #2358, opened 2025-10-02
feat: add new integration for `FallbackChatGenerator`

vblagoje:
### Why:
Introduces a robust fallback mechanism for chat generators that automatically switches between multiple generators when primary services fail, ensuring continuous service availability during API outages or rate limiting.

- fixes https://github.com/deepset-ai/haystack/issues/9786

### What:
- Added new `FallbackChatGenerator` component with sequential fallback logic
- Implemented per-generator timeout handling and comprehensive error handling (429, 401, 400, 408, 500+ errors)
- Added both sync/async execution support with streaming callback handling
- Added test suite  and three practical usage examples
- Set up full integrations structure with proper licensing and documentation

### How can it be used:

```python
from haystack_integrations.components.generators.fallback_chat import FallbackChatGenerator

primary = OpenAIChatGenerator(model="gpt-4o-mini")
backup = AnthropicChatGenerator(model="claude-3-5-sonnet-20241022")

fallback = FallbackChatGenerator(generators=[primary, backup], timeout=10.0)
result = fallback.run([ChatMessage.from_user("Hello!")])
print(result["replies"][0].text)
```

### How did you test it:
- Added comprehensive unit tests covering success/failure scenarios and timeout behavior
- Tested error handling for all specified HTTP status codes and streaming functionality
- Created integration examples demonstrating real-world usage patterns
- Validated serialization/deserialization and async/sync compatibility

### Notes for the reviewer:
Focus on the timeout logic in `_get_effective_timeout()` and error handling in `_run_generator_with_timeout()`. The streaming callback forwarding maintains proper async/sync compatibility.

### DIFF
--- a/integrations/fallback_chat/src/haystack_integrations/components/generators/fallback_chat/chat_generator.py
+++ b/integrations/fallback_chat/src/haystack_integrations/components/generators/fallback_chat/chat_generator.py
@@ -6,10 +6,6 @@ from __future__ import annotations
 
 import asyncio
 import logging
-import time
-from concurrent.futures import ThreadPoolExecutor
-from concurrent.futures import TimeoutError as FuturesTimeoutError
-from threading import Event as ThreadEvent
 from typing import Any
 
 from haystack import component, default_from_dict, default_to_dict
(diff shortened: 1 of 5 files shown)


vblagoje:
Review from anyone else interested in this area is welcome cc @julian-risch @sjrl

integrations/fallback_chat/src/haystack_integrations/components/generators/fallback_chat/chat_generator.py
```diff
@@ -0,0 +1,404 @@
...
+            raise ValueError(msg)
+        if timeout is None or timeout <= 0:
+            msg = "'timeout' must be a positive number of seconds"
+            raise ValueError(msg)
+
+        # Validation via duck-typing: require a callable 'run' method
+        for gen in generators:
+            if not hasattr(gen, "run") or not callable(gen.run):
+                msg = "All items in 'generators' must expose a callable 'run' method (duck-typed ChatGenerator)"
+                raise TypeError(msg)
```
sjrl, line 66: I'm not sure this check is needed. At least in our other components that take in components in their init we don't strictly double check that they are a Haystack component.

integrations/fallback_chat/src/haystack_integrations/components/generators/fallback_chat/chat_generator.py
```diff
@@ -0,0 +1,404 @@
...
+        # Completion mode: when no streaming callback is provided
+        if streaming_callback is None:
+
+            def _call_completion() -> dict[str, Any]:
+                return gen.run(
+                    messages=messages,
+                    generation_kwargs=generation_kwargs,
+                    tools=tools,
+                    streaming_callback=streaming_callback,
+                )
```
sjrl: It's possible we should only forward params to the generator's run method if it accepts it. E.g. I don't think generators have `tools` supported. or at the very least we should enforce in the init method what the run signature of each chat generator should be.

integrations/fallback_chat/src/haystack_integrations/components/generators/fallback_chat/chat_generator.py
```diff
@@ -0,0 +1,404 @@
...
+        return min(self.timeout, float(gen_timeout))
+
+    def _run_single_sync(
+        self,
+        gen: Any,
+        messages: list[ChatMessage],
+        generation_kwargs: dict[str, Any] | None,
+        tools: (list[Tool] | Toolset) | None,
+        streaming_callback: StreamingCallbackT | None,
+    ) -> dict[str, Any]:
```
sjrl, line 100: @vblagoje I'm pretty confused as to what this is function is doing. Why are we running the calls to the generator within a ThreadPoolExecutor with a single worker?

integrations/fallback_chat/src/haystack_integrations/components/generators/fallback_chat/chat_generator.py
```diff
@@ -0,0 +1,404 @@
...
+                        "failed_generators": failed,
+                        "timeout_used": effective_timeout,
+                        "execution_time": time.time() - start,
+                    }
+                )
+                return {"replies": replies, "meta": meta}
+            except asyncio.TimeoutError as e:
+                logger.warning("Generator %s timed out after %.2fs", gen_name, effective_timeout)
+                failed.append(gen_name)
+                last_error = e
```
sjrl: A general comment. I'm not sure I like how we've implemented our own additional timeout management here. I think we should be asking users to set timeouts on each individual ChatGenerator since that is normally specifiable as a timeout param to the ChatGenerator when making it. I think that is more clear and understandable than creating our own mechanism on top

sjrl: We can rewrite your example like this
```python
from haystack_integrations.components.generators.fallback_chat import FallbackChatGenerator

primary = OpenAIChatGenerator(model="gpt-4o-mini", timeout=10.0)  # <-- Added timeout here
backup = AnthropicChatGenerator(model="claude-3-5-sonnet-20241022", timeout=10.0)  # <-- Added timeout here

fallback = FallbackChatGenerator(generators=[primary, backup])
result = fallback.run([ChatMessage.from_user("Hello!")])
print(result["replies"][0].text)
```

sjrl:
@vblagoje I also wanted to ask, what was your reasoning for making this a core integration?

integrations/fallback_chat/src/haystack_integrations/components/generators/fallback_chat/chat_generator.py
vblagoje: Yes, ok we can do that. I'll drop the timeout on the FallbackChatGenerator and the complexities with what takes precendence: list of generators, outer etc

sjrl:
> > @vblagoje I also wanted to ask, what was your reasoning for making this a core integration?
> 
> My reasoning was that this was an optional sidecar not a core feature as we tend to reserve for truly necessary building blocks to core. I'm not hard pressed for integration - we can make it core feature as well. Perhaps via experimental?

I think this would be suitable as a core feature but unsure if it should be in experimental or not first. What do you think @julian-risch

PR #2360, opened 2025-10-06
test: Pinecone - relax flaky test

anakin87:
### Related Issues

These tests have been failing repeatedly: https://github.com/deepset-ai/haystack-core-integrations/actions/runs/18266359971/job/52001249975

I tried a lot with different combinations of vectors but I realized that while the best vectors are returned, order is not guaranteed ([Pinecone forum](https://community.pinecone.io/t/pinecone-return-results-consistency/5389/2))

### Proposed Changes:
- relax the test to only check that the correct docs are returned (not the order)

### How did you test it?
CI

### Checklist


sjrl (approved):
Thanks for fixing this! I guess another route might have been changing from ANN to KNN if pinecone allows it but I think this also fine

anakin87:
> Thanks for fixing this! I guess another route might have been changing from ANN to KNN if pinecone allows it but I think this also fine

Yes, unfortunately Pinecone does not seem to support KNN

PR #2361, opened 2025-10-06
feat: allow `OpenSearch` embedders to query a different `DocumentStore` at runtime

davidsbatista:
### Related Issues

- fixes #issue-number

### Checklist

### DIFF
--- a/integrations/opensearch/src/haystack_integrations/components/retrievers/opensearch/bm25_retriever.py
+++ b/integrations/opensearch/src/haystack_integrations/components/retrievers/opensearch/bm25_retriever.py
@@ -200,22 +200,23 @@ class OpenSearchBM25Retriever:
         fuzziness: Optional[Union[int, str]] = None,
         scale_score: Optional[bool] = None,
         custom_query: Optional[Dict[str, Any]] = None,
+        document_store: Optional[OpenSearchDocumentStore] = None,
     ) -> Dict[str, List[Document]]:
         """
         Retrieve documents using BM25 retrieval.
 
         :param query: The query string.
         :param filters: Filters applied to the retrieved documents. The way runtime filters are applied depends on
-                        the `filter_policy` specified at Retriever's initialization.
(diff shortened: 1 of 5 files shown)


integrations/opensearch/docker-compose.yml
```diff
@@ -1,15 +1,46 @@
...
-        retries: 10
\ No newline at end of file
+      test: curl --fail https://localhost:9200/_cat/health -ku admin:admin || exit 1
+      interval: 10s
+      timeout: 1s
+      retries: 10
+    networks:
+      - opensearch-net
+
+  opensearch2:
```
anakin87: in general, for OpenSearch, Haystack Document Store == OpenSearch Index

For this reason, I have the impression that this feature can be tested using a single OpenSearch Docker container with 2 indexes.

If possible, having a single container would make the test suite lighter and more maintainable in the long run.

integrations/opensearch/tests/conftest.py
```diff
@@ -117,55 +140,90 @@ def test_documents():
             content="Haskell is a functional programming language",
             meta={"likes": 100000, "language_type": "functional"},
             id="1",
+            embedding=[0.1, 0.2, 0.3] + [0.0] * 765,
```
julian-risch: If possible, I'd prefer keeping the `test_documents` fixture unchanged. The fixture is also used in other tests, for example in https://github.com/deepset-ai/haystack-core-integrations/blob/43d8a2a503ee8e35e3515e36b3df465a23a5634b/integrations/opensearch/tests/test_document_store.py and if we set embeddings in the fixture, there would be no reason to define documents with embeddings separately here: https://github.com/deepset-ai/haystack-core-integrations/blob/43d8a2a503ee8e35e3515e36b3df465a23a5634b/integrations/opensearch/tests/test_document_store.py#L261
I'd say a cleaner way would be leaving this fixture unchanged and adding another fixture `test_documents_with_embeddings`, which could be used in the newly added tests and optionally also be reused in other tests.

PR #2362, opened 2025-10-06
feat: add support for delete_all_documents for astra client

### LINKED ISSUE #2308: add `delete_all_documents()` operation to `AstraDocumentStore`
The operation should delete all the documents in the document store.

There seems to exist already in the API [a native operation](https://docs.datastax.com/en/astra-db-serverless/api-reference/document-methods/delete-many.html#:~:text=To%20delete%20all%20documents%2C%20use,in%20the%20collection%20was%20deleted.) to achieve this.

srini047:
### Related Issues

- fixes #2308 

### Proposed Changes:
- Added support for `delete_all_documents()` in document_stort and client
- Removed the `delete_all` from the `delete()` to decouple
- Some generic cleanup

### How did you test it?

- Run complete integration test for Astra integration
```bash
cd integrations/astra
hatch run test:integration
```
- Make use of `delete_all_documents()` where `delete(delete_all=True)` was used earlier

### Notes for the reviewer

A quick question: This is a breaking change do we still need the backward compatibility?

### Checklist


integrations/astra/src/haystack_integrations/document_stores/astra/document_store.py
```diff
@@ -413,12 +411,22 @@ def delete_documents(
...
+        """
+        Deletes all documents from the document store.
+        """
+        deletion_counter = 0
+        deletion_counter = self.index.delete_all_documents()
+
+        if deletion_counter == -1:
+            logger.info("All documents deleted")
+        else:
+            logger.info("Could not delete all documents")
```
mpangrazzi: This logging logic can be improved. What about removing this in favour of something like:

```python
logger.info(f"{deletion_counter} documents deleted")
```

integrations/astra/src/haystack_integrations/document_stores/astra/document_store.py
```diff
@@ -398,8 +398,6 @@ def search(
     def delete_documents(
         self,
         document_ids: Optional[List[str]] = None,
```
mpangrazzi: Not related to this PR, but this was not compliant with our [protocol](https://github.com/deepset-ai/haystack/blob/2a27e0d13193fd3910600ef2cbda21cabc2da659/haystack/document_stores/types/protocol.py).

We may want to remove `Optional` here.

integrations/astra/src/haystack_integrations/document_stores/astra/document_store.py
srini047: @mpangrazzi The reason for this if-else was `-1` which is the response provided by Astra client for successful delete all documents. But in general for a user mindset [`-1`](https://docs.datastax.com/en/astra-db-serverless/api-reference/document-methods/delete-many.html#delete-all-documents) would mean a error code. Hence I wanted to give a message instead of a counter value returned by Astra.

mpangrazzi: @srini047 Got it! So in this case I would do a better error handling: logging it with proper level (i.e. `error`), get the message (is it available somewhere?). WDYT?

srini047: @mpangrazzi This is the response from the Astra [API](https://docs.datastax.com/en/astra-api-docs/_attachments/python-client/astrapy/results.html#astrapy.results.CollectionDeleteResult). There's something called `raw_result` attribute but don't think it has a definite string/msg that we can parse and notify the user with actual error.

We should be handling it from our end but anyways don't have enough data to share with users. So I think this should be fine.

srini047: Just changed the verbosity level.

Issue #2341, opened 2025-09-26
Build MistralOCRDocumentConverter similar to AzureOCRDocumentConverter


Hansehart:
**Is your feature request related to a problem? Please describe.**
I've always been building my custom component to use the ocr capabilities of Mistral API. Mistral has great capabilities especially extracting images from .pdf files. See [Mistral API Docs](https://docs.mistral.ai/capabilities/document_ai/basic_ocr/)

**Describe the solution you'd like**
A wrapper for the Mistral API OCR that works seamless with haystack components and pipelines

**Describe alternatives you've considered**
Building always my custom MistralOCRDocumentConverter. Idk if its maybe possible to have a abstract class that wraps multiple providers. Because the list is not ending with Azure, Mistral etc. However I want to show, that I am ready if you need it.

**Additional context**
I am available to implement this.

anakin87:
@Hansehart there are still a few task to be done before closing this issue.

I would appreciate it if you could open a PR to update the integration page, basically including a short mention/example of the new component in https://github.com/deepset-ai/haystack-integrations/blob/main/integrations/mistral.md. LMK...

Hansehart:
I will do this by the end of the week. However I recognized in [#9922](https://github.com/deepset-ai/haystack/issues/9922) that metadata in the run method is missing. I am adressing this in the new [Draft](https://github.com/deepset-ai/haystack-core-integrations/pull/2457) (fixed metadata ASAP because I rely on it). Docs will follow very soon and then I open the draft for review.

PR #2376, opened 2025-10-13
feat: mistral ocr converter

Hansehart:
### Related Issues

- part of #2341

### Proposed Changes:

This PR adds a new OCR document converter component for the Mistral integration in Haystack.
The MistralOCRDocumentConverter uses Mistral’s Document AI / OCR API to extract text and structured annotations from documents and images.

Key features:
- Automatically converts PDFs output into a single Haystack Document object using OCR for hard parsable formats. 
- Supports multiple document input types (DocumentURLChunk, FileChunk (La Plateforme), ImageURLChunk).
- Optionally enriches image regions with structured annotations via Pydantic schemas.
- Supports document-level annotations (e.g. language detection, extracted URLs, any other definied metadata).

### How did you test it?

- Manual verification using real-world PDFs and images (e.g. public arXiv PDFs).
- Confirmed Mistral OCR API call returns structured responses with both bbox and document-level annotations.
- Verified Document creation and metadata enrichment in Haystack.

### Notes for the reviewer

There are definetly some things missing. Its works great on my structure, however needs adjustments for haystack. Including:
- Tests
- Docs
- Package Integration e.g in the example i use `haystack_integrations.components.converters.mistral.ocr_document_converter` which can not be resolved
- Due to using pipe union syntax (`|` instead `Optional[]`) its for python>=3.10 but can be adjusted
- Mistral API key required
- The other components are not using mistral sdk. I am unsure if its required to rely only on openai or can I use mistral SDK?
- I developed this component for my own purpose but want to share it. When you provide me some feedback on whats needs to be done, I will tackle that. Functionality is working, for the integrations steps i am going proactively into the pr nowing it need adjustments.

### Checklist

~~- I added unit tests and updated the docstrings~~

### DIFF
--- a/.github/utils/validate_version.py
+++ b/.github/utils/validate_version.py
@@ -2,7 +2,7 @@ import re
 import requests
 
 # * integrations/<INTEGRATION_FOLDER_NAME>-v1.0.0
-INTEGRATION_VERSION_REGEX = r"integrations/([a-zA-Z-]+)-v([0-9]\.[0-9]+\.[0-9]+)"
+INTEGRATION_VERSION_REGEX = r"integrations/([a-zA-Z_]+)-v([0-9]\.[0-9]+\.[0-9]+)"
 
 
 def validate_version_number(tag: str):
@@ -18,7 +18,8 @@ def validate_version_number(tag: str):
     print(f"Integration name: {integration_name}")
     print(f"Integration version to release: {version_to_release}")
 
-    integration_package = f"{integration_name}-haystack"
+    # Replace underscores with hyphens to look for the package on PyPi
+    integration_package = f"{integration_name.replace('_', '-')}-haystack"
     print(f"Integration PyPi package: {integration_package}")
 
     # connect to PyPi and check the latest version
--- a/.github/workflows/CI_docusaurus_sync.yml
+++ b/.github/workflows/CI_docusaurus_sync.yml
@@ -1,0 +1,122 @@
+name: Core / Sync API reference with Docusaurus
+
+on:
+  push:
+    tags:
+      - "**-v[0-9].[0-9]+.[0-9]+"
+
+  workflow_dispatch: # Activate this workflow manually
+    inputs:
+      tag:
+        description: "Tag with this format: integrations/<INTEGRATION_FOLDER_NAME>-v1.0.0. When running this workflow manually, version is irrelevant so you can use any value."
+        required: true
+        type: string
+        default: integrations/<INTEGRATION_FOLDER_NAME>-v1.0.0
+
+env:
+  TAG: ${{ inputs.tag || github.ref_name }}
+
+jobs:
+  generate-api-reference:
+    runs-on: ubuntu-latest
+    outputs:
+      integration_name: ${{ steps.pathfinder.outputs.integration_name }}
+
+    steps:
+      - name: Checkout this repo
+        uses: actions/checkout@v5
+
+      - name: Set up Python
+        uses: actions/setup-python@v6
+        with:
+          python-version: "3.10"
+
+      - name: Install dependencies
+        run: |
+          python -m pip install --upgrade pip
+          pip install -U haystack-pydoc-tools
+
+      - name: Get project folder
+        id: pathfinder
+        shell: python
+        run: |
+          import os
+          project_path = os.environ["TAG"].rsplit("-", maxsplit=1)[0]
+          integration_name = project_path.split("/")[-1]
+          with open(os.environ['GITHUB_OUTPUT'], 'a') as f:
+            print(f'project_path={project_path}', file=f)
+            print(f'integration_name={integration_name}', file=f)
+
+      - name: Generate API reference
+        working-directory: ${{ steps.pathfinder.outputs.project_path }}
+
+        run: pydoc-markdown pydoc/config_docusaurus.yml
+
+      - name: Upload API reference artifact
+        uses: actions/upload-artifact@v4
+        with:
+          name: ${{ steps.pathfinder.outputs.integration_name }}
+          path: ${{ steps.pathfinder.outputs.project_path }}/${{ steps.pathfinder.outputs.integration_name }}.md
+          if-no-files-found: error
+          retention-days: 1
+          overwrite: true
+
+
+  sync-api-reference:
+    runs-on: ubuntu-latest
+    needs: generate-api-reference
+    
+    steps:
+      - name: Checkout Haystack repo
+        uses: actions/checkout@v5
+        with:
+          repository: deepset-ai/haystack
+          ref: main
+          token: ${{ secrets.HAYSTACK_BOT_TOKEN }}
+
+      - name: Set up Python
+        uses: actions/setup-python@v6
+        with:
+          python-version: "3.10"          
+
+      - name: Download API reference artifact
+        uses: actions/download-artifact@v4
+        with:
+          name: ${{ needs.generate-api-reference.outputs.integration_name }}
+
+      - name: Sync API reference
+        shell: python
+        env:
+          INTEGRATION_NAME: ${{ needs.generate-api-reference.outputs.integration_name }}
+        run: |
+          import os
+          import shutil
+
+          artifact_filename = os.environ['INTEGRATION_NAME']+'.md'
+
+          # Copy to main API reference
+          shutil.copy(artifact_filename, "docs-website/reference/integrations-api/")
+
+          # Copy to versioned API reference
+          for version_dir in os.scandir("docs-website/reference_versioned_docs"):
+              if version_dir.is_dir():
+                  # example: docs-website/reference_versioned_docs/version-2.17/integrations-api
+                  integrations_api_ref_dir = os.path.join(version_dir.path, "integrations-api")
+                  shutil.copy(artifact_filename, integrations_api_ref_dir)
+
+          os.remove(artifact_filename)
+
+      - name: Create Pull Request
+        uses: peter-evans/create-pull-request@v7
+        env:
+          INTEGRATION_NAME: ${{ needs.generate-api-reference.outputs.integration_name }}
+        with:
+          token: ${{ secrets.HAYSTACK_BOT_TOKEN }}
+          commit-message: "Sync Core Integrations API reference (${{ env.INTEGRATION_NAME }}) on Docusaurus"
+          branch: sync-docusaurus-api-reference-${{ env.INTEGRATION_NAME }}
+          base: main
+          title: "docs: sync Core Integrations API reference (${{ env.INTEGRATION_NAME }}) on Docusaurus"
+          add-paths: |
+            docs-website
+          body: |
+            This PR syncs the Core Integrations API reference (${{ env.INTEGRATION_NAME }}) on Docusaurus. Just approve and merge it.
--- a/integrations/amazon_bedrock/CHANGELOG.md
+++ b/integrations/amazon_bedrock/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## [integrations/amazon_bedrock-v5.2.0] - 2025-10-22
+
+### 🚀 Features
+
+- `AmazonBedrockChatGenerator` update tools param to ToolsType (#2415)
+
 ## [integrations/amazon_bedrock-v5.1.0] - 2025-09-29
 
 ### 🚀 Features
--- a/integrations/anthropic/CHANGELOG.md
+++ b/integrations/anthropic/CHANGELOG.md
@@ -1,9 +1,28 @@
 # Changelog
+
+## [integrations/anthropic-v4.5.0] - 2025-10-22
+
+### 🚀 Features
+
+- `AnthropicChatGenerator` update tools param to ToolsType  (#2417)
+
+### 📚 Documentation
+
+- Add pydoc configurations for Docusaurus (#2411)
+
 ## [integrations/anthropic-v4.4.0] - 2025-09-15
 
 ### 🚀 Features
 
-- Enable reasoning support in AnthropicChatGenerator (#2230)
+- `AnthropicChatGenerator`- refine reasoning support (#2230)
+
+
+### 🧹 Chores
+
+- Standardize readmes - part 1 (#2202)
+- Standardize readmes - part 2 (#2205)
+- Fix linting in Anthropic (#2251)
+
 ## [integrations/anthropic-v4.3.0] - 2025-08-18
 
 ### 🚀 Features
@@ -15,20 +34,25 @@
 
 ### 🐛 Bug Fixes
 
-- Add `input_tokens` in `meta["usage"]` of Anthropic messages (#2173)
+- Add `input_tokens` in usage of Anthropic messages (#2173)
 
 
 ## [integrations/anthropic-v4.2.0] - 2025-08-07
 
 ### 🐛 Bug Fixes
 
-- Fixed a bug to enable parallel tool calling (#2154)
+- Error in `ToolCallDelta.index` for parallel tool calling (#2154)
+
 
 ## [integrations/anthropic-v4.1.0] - 2025-08-05
 
 ### 🚀 Features
 
 - Adopt new `StreamingChunk` in Anthropic (#2096)
+
+### 🌀 Miscellaneous
+
+- Fix/prompt caching support (#2051)
 
 ## [integrations/anthropic-v3.1.0] - 2025-07-04
 
--- a/integrations/anthropic/pyproject.toml
+++ b/integrations/anthropic/pyproject.toml
@@ -23,7 +23,7 @@ classifiers = [
   "Programming Language :: Python :: Implementation :: CPython",
   "Programming Language :: Python :: Implementation :: PyPy",
 ]
-dependencies = ["haystack-ai>=2.17.1", "anthropic>=0.47.0"]
+dependencies = ["haystack-ai>=2.19.0", "anthropic>=0.47.0"]
 
 [project.urls]
 Documentation = "https://github.com/deepset-ai/haystack-core-integrations/tree/main/integrations/anthropic#readme"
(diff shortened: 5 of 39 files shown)


integrations/mistral/src/haystack_integrations/components/converters/mistral/ocr_document_converter.py
```diff
@@ -0,0 +1,221 @@
+"""Mistral OCR Document Converter for Haystack.
```
anakin87: 1. This text should be included in the docstring below `class MistralOCRDocumentConverter`
2. Let's make it clear what happens if annotations are provided: a Vision LLM is called after the OCR model. https://docs.mistral.ai/capabilities/document_ai/annotations/#how-it-works

integrations/mistral/src/haystack_integrations/components/converters/mistral/ocr_document_converter.py
```diff
@@ -0,0 +1,221 @@
...
+    def __init__(
+        self,
+        api_key: Secret = Secret.from_env_var("MISTRAL_API_KEY"),
+        model: str = "mistral-ocr-2505",
+        include_image_base64: bool = False,
+        pages: list[int] | None = None,
+        image_limit: int | None = None,
+        image_min_size: int | None = None,
+        bbox_annotation_schema: Type[BaseModel] | None = None,
+        document_annotation_schema: Type[BaseModel] | None = None,
```
anakin87: Including `bbox_annotation_schema`  and `document_annotation_schema` in `__init__` would be problematic for serialization/deserialization: while a `BaseModel` can be easily serialized into JSON, in order to deserialize it we need to store the original `BaseModel`. 

As a first iteration, I would suggest to only expose these parameters in `run`, so we would not need to care about serialization.

integrations/mistral/src/haystack_integrations/components/converters/mistral/ocr_document_converter.py
```diff
@@ -0,0 +1,221 @@
...
+    ):
+        """
+        Initialize the MistralOCRDocumentConverter.
+
+        Args:
```
anakin87: Let's stick to the docstring format we generally use: https://github.com/deepset-ai/haystack/blob/18b6482e2e5199d711e61c114800ceb19301db48/haystack/components/converters/azure.py#L69

integrations/mistral/src/haystack_integrations/components/converters/mistral/ocr_document_converter.py
```diff
@@ -0,0 +1,221 @@
...
+        )
+        self.document_annotation_format = (
+            response_format_from_pydantic_model(document_annotation_schema) if document_annotation_schema else None
+        )
+
+        # Initialize Mistral client
+        self.client = Mistral(api_key=self.api_key.resolve_value())
+
+    @component.output_types(documents=list[Document], raw_mistral_response=OCRResponse)
+    def run(self, source: DocumentURLChunk | FileChunk | ImageURLChunk) -> dict:
```
anakin87: 1. I would try to support passing multiple sources (see `AzureOCRDocumentConverter`)
2. I'd like to make `source` (or `sources`) compatible with similar Haystack converters: we should support at least `Union[str, Path, ByteStream]` and internally handle the conversion to `FileChunk`.

integrations/mistral/pyproject.toml
```diff
@@ -58,7 +58,8 @@ dependencies = [
     "pytest-rerunfailures",
     "mypy",
     "pip",
-    "pytz"
+    "pytz",
+    "mistralai"
```
anakin87:
```suggestion
```

This is not needed since the `test` environment inherits from the `default` environments, where the dependency is already included.

integrations/mistral/src/haystack_integrations/components/converters/mistral/ocr_document_converter.py
```diff
@@ -0,0 +1,312 @@
...
+        self.image_limit = image_limit
+        self.image_min_size = image_min_size
+
+        # Initialize Mistral client
+        self.client = Mistral(api_key=self.api_key.resolve_value())
+
+    @component.output_types(documents=List[Document], raw_mistral_response=List[Dict[str, Any]])
+    def run(
+        self,
+        sources: List[Union[str, Path, ByteStream, DocumentURLChunk, FileChunk, ImageURLChunk]],
```
anakin87, line 175: @Hansehart Thanks for adding `Union[str, Path, ByteStream]`.

@sjrl do you think that supporting Mistral types in `sources` could be problematic? I am thinking of Breakpoints and pipeline inputs serialization...

integrations/mistral/src/haystack_integrations/components/converters/mistral/ocr_document_converter.py
```diff
@@ -0,0 +1,312 @@
...
+        self.api_key = api_key
+        self.model = model
+        self.include_image_base64 = include_image_base64
+        self.pages = pages
+        self.image_limit = image_limit
+        self.image_min_size = image_min_size
+
+        # Initialize Mistral client
+        self.client = Mistral(api_key=self.api_key.resolve_value())
+
```
anakin87, line 140: `to_dict`/`from_dict` methods are missing. See https://github.com/deepset-ai/haystack/blob/main/haystack/components/converters/azure.py#L155

integrations/mistral/src/haystack_integrations/components/converters/mistral/ocr_document_converter.py
sjrl: Yeah I think I'd prefer to stick with our normal support types of just `Union[str, Path, ByteStream]`. I suppose we could add se/de support for their types as well as an alternative. Since I at least see utility in `DocumentURLChunk` and `ImageURLChunk` (just guessing off the names) which would allow new functionality of converting urls pointing to text or images. 

What do you think?

anakin87: Yes, I think that they can be useful.

BTW, they are Pydantic `BaseModel`s. (https://github.com/mistralai/client-python/blob/992b0cbd3f86cef54e74a03cfd488f123cec4203/src/mistralai/models/documenturlchunk.py#L20). In this case, serialization on our side is automatic or should we do something?

sjrl: > In this case, serialization on our side is automatic or should we do something?

It's not automatic currently. We would need to add explicit support for pydantic models similar to what we recently did for python Enums in [here](https://github.com/deepset-ai/haystack/pull/9869#discussion_r2428363541) and [here](https://github.com/deepset-ai/haystack/pull/9869#discussion_r2428364809)

Hansehart: So I will wait until the issue [#9877](https://github.com/deepset-ai/haystack/issues/9877) is resolved

anakin87: I think we can move forward, just knowing that in case of errors, breakpoints might not work well.

integrations/mistral/src/haystack_integrations/components/converters/mistral/ocr_document_converter.py
```diff
@@ -0,0 +1,367 @@
...
+    - Annotations: https://docs.mistral.ai/capabilities/document_ai/annotations/
+
+    **Usage Example:**
+    ```python
+    from haystack.utils import Secret
+    from haystack_integrations.mistral import MistralOCRDocumentConverter
+    from mistralai.models import DocumentURLChunk, ImageURLChunk, FileChunk
+
+    converter = MistralOCRDocumentConverter(
+        api_key=Secret.from_env_var("MISTRAL_API_KEY"),
```
anakin87, line 49:
```suggestion
```

this is the default, so it can be removed from the example

integrations/mistral/src/haystack_integrations/components/converters/mistral/ocr_document_converter.py
```diff
@@ -0,0 +1,367 @@
...
+        summary: str = Field(..., description="Detailed summary of the image content")
+
+    # Define schema for structured document annotations
+    class DocumentAnnotation(BaseModel):
+        language: str = Field(..., description="Primary language of the document")
+        chapter_titles: List[str] = Field(..., description="Detected chapter or section titles")
+        urls: List[str] = Field(..., description="URLs found in the text")
+
+    converter = MistralOCRDocumentConverter(
+        api_key=Secret.from_env_var("MISTRAL_API_KEY"),
```
anakin87:
```suggestion
```

integrations/mistral/src/haystack_integrations/components/converters/mistral/ocr_document_converter.py
```diff
@@ -0,0 +1,367 @@
...
+        image_min_size: Optional[int] = None,
+        cleanup_uploaded_files: bool = True,
+    ):
+        """
+        Creates a MistralOCRDocumentConverter component.
+
+        :param api_key:
+            The Mistral API key. Defaults to the MISTRAL_API_KEY environment variable.
+        :param model:
+            The OCR model to use. Default is "mistral-ocr-2505".
```
anakin87, line 114: Where the user can find the list of available models? If there's a link, let's add it here.

integrations/mistral/src/haystack_integrations/components/converters/mistral/ocr_document_converter.py
```diff
@@ -0,0 +1,367 @@
...
+
+    **How Annotations Work:**
+    When annotation schemas (`bbox_annotation_schema` or `document_annotation_schema`) are provided,
+    the OCR model first extracts text and structure from the document. Then, a Vision LLM is called
+    to analyze the content and generate structured annotations according to your defined schemas.
+    For more details, see: https://docs.mistral.ai/capabilities/document_ai/annotations/#how-it-works
+
+    **API Reference:**
+    - Basic OCR: https://docs.mistral.ai/capabilities/document_ai/basic_ocr/
+    - Annotations: https://docs.mistral.ai/capabilities/document_ai/annotations/
```
anakin87: Since we abstract this logic, we can remove those links

integrations/mistral/src/haystack_integrations/components/converters/mistral/ocr_document_converter.py
```diff
@@ -0,0 +1,367 @@
...
+        document_annotation_format = (
+            response_format_from_pydantic_model(document_annotation_schema) if document_annotation_schema else None
+        )
+
+        # Process each source
+        documents = []
+        raw_responses = []
+        uploaded_file_ids = []  # Track files we uploaded for cleanup
+
+        try:
```
anakin87, line 271: If possible, I'd avoid nested try/except

integrations/mistral/src/haystack_integrations/components/converters/mistral/ocr_document_converter.py
```diff
@@ -0,0 +1,367 @@
...
+                except Exception as e:
+                    logger.warning(
+                        "Could not process source {source}. Skipping it. Error: {error}",
+                        source=source,
+                        error=e,
+                    )
+                    continue
+        finally:
+            # Cleanup uploaded files if enabled
+            if self.cleanup_uploaded_files and uploaded_file_ids:
```
anakin87: I see that you introduced this option to delete uploaded files. Could you please explain why you think it's necessary/helpful for a user?

integrations/mistral/src/haystack_integrations/components/converters/mistral/ocr_document_converter.py
```diff
@@ -0,0 +1,367 @@
...
+                        )
+
+        return {"documents": documents, "raw_mistral_response": raw_responses}
+
+    def _convert_source_to_chunk(
+        self,
+        source: Union[str, Path, ByteStream, DocumentURLChunk, FileChunk, ImageURLChunk],
+    ) -> Union[DocumentURLChunk, FileChunk, ImageURLChunk]:
+        """
+        Convert various source types to Mistral-compatible chunk format.
```
anakin87, line 324: let's also mention that this method upload files if needed

integrations/mistral/tests/test_ocr_document_converter.py
```diff
@@ -0,0 +1,849 @@
...
+            model="mistral-ocr-custom",
+            include_image_base64=True,
+        )
+
+        assert converter.api_key == Secret.from_token("test-api-key")
+        assert converter.model == "mistral-ocr-custom"
+        assert converter.include_image_base64 is True
+        assert converter.pages is None
+        assert converter.image_limit is None
+        assert converter.image_min_size is None
```
anakin87:
```suggestion
```

let's remove this. Already covered in other tests

integrations/mistral/tests/test_ocr_document_converter.py
```diff
@@ -0,0 +1,849 @@
...
+        mock_response.model_dump.return_value = {
+            "pages": [
+                {"markdown": "# Page 1", "images": []},
+                {"markdown": "# Page 2", "images": []},
+            ],
+            "document_annotation": None,
+        }
+        return mock_response
+
+    def test_run_with_document_url_chunk(self, mock_ocr_response):
```
anakin87: the following tests are almost identical, so we can use `@pytest.mark.parametrize` to avoid duplication

integrations/mistral/tests/test_ocr_document_converter.py
```diff
@@ -0,0 +1,849 @@
...
+        with patch.object(
+            converter.client.ocr, "process", return_value=mock_ocr_response
+        ):
+            sources = [ImageURLChunk(image_url="https://example.com/image.jpg")]
+            result = converter.run(sources=sources)
+
+            assert len(result["documents"]) == 1
+            assert isinstance(result["documents"][0], Document)
+
+    def test_run_with_file_path(self, mock_ocr_response, tmp_path):
```
anakin87: `@pytest.mark.parametrize` can be used in the following 3 tests as well

integrations/mistral/tests/test_ocr_document_converter.py
```diff
@@ -0,0 +1,849 @@
...
+        sources = [
+            DocumentURLChunk(document_url="https://arxiv.org/pdf/1706.03762"),
+            DocumentURLChunk(document_url="https://arxiv.org/pdf/2309.06180"),
+        ]
+        result = converter.run(sources=sources)
+
+        assert len(result["documents"]) == 2
+        assert all(isinstance(doc, Document) for doc in result["documents"])
+        assert all(len(doc.content) > 0 for doc in result["documents"])
+        assert len(result["raw_mistral_response"]) == 2
```
anakin87:
```suggestion
```

integrations/mistral/tests/test_ocr_document_converter.py
```diff
@@ -0,0 +1,849 @@
...
+from haystack.dataclasses import ByteStream
+from haystack.utils import Secret
+from mistralai.models import DocumentURLChunk, FileChunk, ImageURLChunk
+from pydantic import BaseModel, Field
+
+from haystack_integrations.components.converters.mistral import (
+    MistralOCRDocumentConverter,
+)
+
+
```
anakin87, line 19: Some of the following tests can be removed without altering coverage. If possible, let's do that.

integrations/mistral/tests/test_ocr_document_converter.py
```diff
@@ -0,0 +1,849 @@
...
+    def test_to_dict(self, monkeypatch):
+        monkeypatch.setenv("MISTRAL_API_KEY", "test-api-key")
+        converter = MistralOCRDocumentConverter()
+        converter_dict = converter.to_dict()
+
+        assert converter_dict == {
+            "type": (
+                "haystack_integrations.components.converters.mistral."
+                "ocr_document_converter.MistralOCRDocumentConverter"
```
anakin87: I think we can define the type as a constant in this module. This should help with format.
Similarly to Bedrock:
https://github.com/deepset-ai/haystack-core-integrations/blob/1867cdd3d6b03ba5b191bd16cb397d434e269b53/integrations/amazon_bedrock/tests/test_chat_generator.py#L13

integrations/mistral/src/haystack_integrations/components/converters/mistral/ocr_document_converter.py
Hansehart: Files uploaded to Mistral's storage persist for an unknown amount of time. When sources are provided as local paths/ByteStreams (not file IDs), we upload them temporarily for OCR processing. Auto-cleanup prevents accumulation of orphaned files that users don't need to manage long-term, as they already have the original files locally. I am also in fear that users are not aware of uploading files to a third-party. However I could add a toggle that defaults to deletion.

https://docs.mistral.ai/api/#tag/files

integrations/mistral/tests/test_ocr_document_converter.py
Hansehart: I removed three tests. Do you want more?

integrations/mistral/src/haystack_integrations/components/converters/mistral/ocr_document_converter.py
anakin87: Ok let's add a init parameter to configure this behavior

integrations/mistral/pydoc/config.yml
```diff
@@ -5,6 +5,7 @@ loaders:
       "haystack_integrations.components.embedders.mistral.document_embedder",
       "haystack_integrations.components.embedders.mistral.text_embedder",
       "haystack_integrations.components.generators.mistral.chat.chat_generator",
+      "haystack_integrations.components.converters.mistral.ocr_document_converter",
```
anakin87, line 8: If you merge main, you'll see another pydoc config file called `config_docusaurus`. Please also update it.

integrations/mistral/src/haystack_integrations/components/converters/mistral/ocr_document_converter.py
anakin87: Ah, you already did this. Great!

anakin87:
Please update https://github.com/deepset-ai/haystack-core-integrations/blob/main/README.md
"Embedder, Generator" -> "Converter, Embedder, Generator"

---

In general, we are close to merging this PR. 👍

anakin87 (approved):
I ran integration tests locally (they don't run on PRs from forks) and all works well.

Thank you!

PR #2384, opened 2025-10-14
Add MCPTool/MCPToolset warm_up

vblagoje:
## Why:
Separates pipeline validation from MCP server connection. Allows pipelines to validate successfully without requiring server connectivity, then connect during pipeline warm_up.

Keep in ~draft stage until https://github.com/deepset-ai/haystack/pull/9856 is merge and Haystack 2.19 released.~ 

## What:
- Added `eager_connect` parameter to `MCPTool` and `MCPToolset` (defaults to `False`)
- Implemented `_ensure_connected()` with thread-safe lazy initialization
- Added `warm_up()` methods for connecting during pipeline warm_up phase

## How can it be used:
```python
# Validate pipeline without connecting to MCP server
tool = MCPTool(name="add", server_info=info, eager_connect=False)
pipeline.add_component("invoker", ToolInvoker(tools=[tool]))  # ✓ Validates

# Connect during pipeline warm_up
pipeline.run()  # triggers tool.warm_up() → connects to server
```

## How did you test it:
- Test verifies permissive schema → strict schema transition on first use
- Confirmed backward compatibility (eager_connect=True by default)
- Tested with itinerary agent

## Notes for the reviewer:
- Placeholder schema allows validation; strict schema loaded post-connection
- `warm_up()` is idempotent

### DIFF
--- a/integrations/mcp/src/haystack_integrations/tools/mcp/mcp_tool.py
+++ b/integrations/mcp/src/haystack_integrations/tools/mcp/mcp_tool.py
@@ -853,35 +853,10 @@ class MCPTool(Tool):
         logger.debug(f"TOOL: Initializing MCPTool '{name}'")
 
         try:
-            # Create client and spin up a long-lived worker that keeps the
-            # connect/close lifecycle inside one coroutine.
-            self._client = server_info.create_client()
-            logger.debug(f"TOOL: Created client for MCPTool '{name}'")
-
-            # The worker starts immediately and blocks here until the connection
-            # is established (or fails), returning the tool list.
-            self._worker = _MCPClientSessionManager(self._client, timeout=connection_timeout)
-
-            tools = self._worker.tools()
-            # Handle no tools case
-            if not tools:
-                logger.debug(f"TOOL: No tools found for '{name}'")
-                message = "No tools available on server"
-                raise MCPToolNotFoundError(message, tool_name=name)
-
-            # Find the specified tool
-            tool_dict = {t.name: t for t in tools}
-            logger.debug(f"TOOL: Available tools: {list(tool_dict.keys())}")
-
-            tool_info: types.Tool | None = tool_dict.get(name)
-
-            if not tool_info:
-                available = list(tool_dict.keys())
-                logger.debug(f"TOOL: Tool '{name}' not found in available tools")
-                message = f"Tool '{name}' not found on server. Available tools: {', '.join(available)}"
-                raise MCPToolNotFoundError(message, tool_name=name, available_tools=available)
-
+            logger.debug(f"TOOL: Connecting to MCP server for '{name}'")
+            tool_info = self._connect_and_initialize(name)
             logger.debug(f"TOOL: Found tool '{name}', initializing Tool parent class")
+
             # Initialize the parent class
             super().__init__(
                 name=name,
@@ -914,6 +889,36 @@ class MCPTool(Tool):
             message = f"Failed to initialize MCPTool '{name}': {error_message}"
             raise MCPConnectionError(message=message, server_info=server_info, operation="initialize") from e
 
+    def _connect_and_initialize(self, tool_name: str) -> types.Tool:
+        """
+        Connect to the MCP server and retrieve the tool schema.
+
+        :param tool_name: Name of the tool to look for
+        :returns: The tool schema for this tool
+        :raises MCPToolNotFoundError: If the tool is not found on the server
+        """
+        client = self._server_info.create_client()
+        worker = _MCPClientSessionManager(client, timeout=self._connection_timeout)
+        tools = worker.tools()
+
+        # Handle no tools case
+        if not tools:
+            message = "No tools available on server"
+            raise MCPToolNotFoundError(message, tool_name=tool_name)
+
+        # Find the specified tool
+        tool = next((t for t in tools if t.name == tool_name), None)
+        if tool is None:
+            available = [t.name for t in tools]
+            msg = f"Tool '{tool_name}' not found on server. Available tools: {', '.join(available)}"
+            raise MCPToolNotFoundError(msg, tool_name=tool_name, available_tools=available)
+
+        # Publish connection
+        self._client = client
+        self._worker = worker
+
+        return tool
+
     def _invoke_tool(self, **kwargs: Any) -> str:
         """
         Synchronous tool invocation.
@@ -928,12 +933,8 @@ class MCPTool(Tool):
 
             async def invoke():
                 logger.debug(f"TOOL: Inside invoke coroutine for '{self.name}'")
-                # This should never happen, and mypy doesn't know that
-                if self._client is None:
-                    raise MCPConnectionError(message="Not connected to an MCP server", operation="call_tool")
-                result = await asyncio.wait_for(
-                    self._client.call_tool(self.name, kwargs), timeout=self._invocation_timeout
-                )
+                client = cast(MCPClient, self._client)
+                result = await asyncio.wait_for(client.call_tool(self.name, kwargs), timeout=self._invocation_timeout)
                 logger.debug(f"TOOL: Invoke successful for '{self.name}'")
                 return result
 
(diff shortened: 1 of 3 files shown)


integrations/mcp/src/haystack_integrations/tools/mcp/mcp_tool.py
```diff
@@ -906,9 +923,15 @@ def _invoke_tool(self, **kwargs: Any) -> str:
         """
         logger.debug(f"TOOL: Invoking tool '{self.name}' with args: {kwargs}")
         try:
+            # Connect on first use if eager_connect is turned off
+            if not self._eager_connect:
+                self._ensure_connected()
```
sjrl: We don't need this right since we would expect a user to connect via warm_up?

sjrl: At the very least if we do want this behavior I think we should just do 
```suggestion
            # Connect on first use if eager_connect is turned off
            self.warm_up()
```
and let warm_up handle this if logic

integrations/mcp/src/haystack_integrations/tools/mcp/mcp_tool.py
```diff
@@ -939,7 +962,10 @@ async def ainvoke(self, **kwargs: Any) -> str:
         :raises TimeoutError: If the operation times out
         """
         try:
-            return await asyncio.wait_for(self._client.call_tool(self.name, kwargs), timeout=self._invocation_timeout)
+            if not self._eager_connect:
+                self._ensure_connected()
```
sjrl: Same here if we want this behavior just call `warm_up`
```suggestion
            self.warm_up()
```

integrations/mcp/src/haystack_integrations/tools/mcp/mcp_tool.py
```diff
@@ -949,6 +975,33 @@ async def ainvoke(self, **kwargs: Any) -> str:
             message = f"Failed to invoke tool '{self.name}' with args: {kwargs} , got error: {e!s}"
             raise MCPInvocationError(message, self.name, kwargs) from e
 
+    def warm_up(self) -> None:
+        """Connect and fetch the tool schema eager_connect is turned off."""
+        if self._eager_connect:
+            return
+        self._ensure_connected()
+
+    def _ensure_connected(self) -> None:
```
sjrl: If we end up just calling `warm_up` like I suggest in my above comments let's remove this function `_ensure_connected` and just directly put all its logic into `warm_up`

vblagoje: Ok yes, i think it was a remnant of previous work where I also wanted to reconnect when broken connection was detected. Let me double check, thanks for raising it

integrations/mcp/src/haystack_integrations/tools/mcp/mcp_tool.py
```diff
@@ -906,9 +923,14 @@ def _invoke_tool(self, **kwargs: Any) -> str:
...
         logger.debug(f"TOOL: Invoking tool '{self.name}' with args: {kwargs}")
         try:
+            # Connect on first use if eager_connect is turned off
+            self.warm_up()
 
             async def invoke():
                 logger.debug(f"TOOL: Inside invoke coroutine for '{self.name}'")
+                # This should never happen, and mypy doesn't know that
+                if self._client is None:
+                    raise MCPConnectionError(message="Not connected to an MCP server", operation="call_tool")
```
sjrl: Would it be better to use an ignore or an assert statement since we know this error can never be raised?

integrations/mcp/src/haystack_integrations/tools/mcp/mcp_tool.py
```diff
@@ -939,7 +961,9 @@ async def ainvoke(self, **kwargs: Any) -> str:
         :raises TimeoutError: If the operation times out
         """
         try:
-            return await asyncio.wait_for(self._client.call_tool(self.name, kwargs), timeout=self._invocation_timeout)
+            self.warm_up()
+            client = cast(MCPClient, self._client)
```
sjrl, line 966: Seems like you are using a different approach here to get around the mypy error than [here](https://github.com/deepset-ai/haystack-core-integrations/pull/2384/files#r2451182465). Let's make it consistent.

integrations/mcp/src/haystack_integrations/tools/mcp/mcp_tool.py
```diff
@@ -949,6 +973,29 @@ async def ainvoke(self, **kwargs: Any) -> str:
             message = f"Failed to invoke tool '{self.name}' with args: {kwargs} , got error: {e!s}"
             raise MCPInvocationError(message, self.name, kwargs) from e
 
+    def warm_up(self) -> None:
+        """Connect and fetch the tool schema if eager_connect is turned off."""
+        if self._eager_connect:
+            return
```
sjrl: Not a big deal but this isn't needed really since the check below
```python
            if self._client is not None:
                return
```
will also just return early right?

integrations/mcp/src/haystack_integrations/tools/mcp/mcp_tool.py
```diff
@@ -949,6 +973,29 @@ async def ainvoke(self, **kwargs: Any) -> str:
...
+                available = [t.name for t in tools]
+                msg = f"Tool '{self.name}' not found on server. Available tools: {', '.join(available)}"
+                raise MCPToolNotFoundError(msg, tool_name=self.name, available_tools=available)
+            # Publish connection and tighten parameters for better prompting
+            self._client = client
+            self._worker = worker
+            try:
+                self.parameters = tool.inputSchema
+            except Exception as e:
+                logger.debug(f"TOOL: Could not update strict parameters after connect: {e!s}")
```
sjrl: Hey I'm noticing that similar code to this lives in the init method starting at current line 855. Could we make a shared method that get's re-used in both places rather than having duplicate code?

sjrl: If possible, one way could be to just call `self.warm_up` in the init method if `self._eager_connect` is `True`

integrations/mcp/src/haystack_integrations/tools/mcp/mcp_toolset.py
```diff
@@ -129,15 +130,46 @@ def __init__(
...
+                             If False (default), defer connection to warm_up.
         :raises MCPToolNotFoundError: If any of the specified tool names are not found on the server
         """
         # Store configuration
         self.server_info = server_info
         self.tool_names = tool_names
         self.connection_timeout = connection_timeout
         self.invocation_timeout = invocation_timeout
+        self.eager_connect = eager_connect
+        self.warmup_called = False
```
sjrl: I think I'd make the `warmup_called` variable private

integrations/mcp/src/haystack_integrations/tools/mcp/mcp_toolset.py
```diff
@@ -129,15 +130,46 @@ def __init__(
...
+            placeholder_tool = Tool(
+                name=f"mcp_not_connected_placeholder_{id(self)}",
+                description="Placeholder tool initialised when eager_connect is turned off",
+                parameters={"type": "object", "properties": {}, "additionalProperties": True},
+                function=lambda: None,
+            )
+            super().__init__(tools=[placeholder_tool])
+        else:
+            tools = self._connect_and_load_tools()
+            super().__init__(tools=tools)
```
sjrl, line 156: We could also set `warmup_called=True` here right?  
```suggestion
        else:
            tools = self._connect_and_load_tools()
            super().__init__(tools=tools)
            self.warmup_called = True
```
This would allow us to simplify `warm_up` slightly

integrations/mcp/src/haystack_integrations/tools/mcp/mcp_toolset.py
```diff
@@ -129,15 +130,46 @@ def __init__(
...
+
+    def warm_up(self) -> None:
+        """Connect and load tools when running eager_connect is turned off.
+
+        Call this before handing the toolset to components like ``ToolInvoker`` so that
+        each tool's schema is available without performing a real invocation.
```
sjrl: This is a bit misleading since users could also:
- Call `ToolInvoker.warm_up()` instead of `MCPToolset.warm_up()`
- Call `Pipeline.warm_up()` if the `ToolInvoker` is part of a pipeline

integrations/mcp/src/haystack_integrations/tools/mcp/mcp_toolset.py
```diff
@@ -129,15 +130,46 @@ def __init__(
...
+        """
+        if self.eager_connect or self.warmup_called:
+            return
```
sjrl: If we make the change in this comment https://github.com/deepset-ai/haystack-core-integrations/pull/2384/files#r2451228527 then we can update this to 
```suggestion
        if self.warmup_called:
            return
```

vblagoje:
Great feedback @sjrl - I also tested this latest commit against itinerary agent

PR #2386, opened 2025-10-14
ci: block the release if the version number is invalid

### LINKED ISSUE #1998: Improve the release process of core-integrations
Currently, we manually release core integrations, as described in the [README](https://github.com/deepset-ai/haystack-core-integrations?tab=readme-ov-file#releasing).

This process is prone to human error, such as releasing the wrong integration or using an incorrect version number.

### Ideas for improvement

We discussed some possible improvements with @julian-risch:
- add a mechanism to validate version numbers to the release workflow, preventing cases like accidentally jumping from 3.4.1 to 5.2.0. This should not be hard to implement.
- we would also like to check if the integration being released corresponds to the latest commit on main. If not, that is legitimate but rare and should require explicit confirmation from the maintainer. This might involve building a CLI utility and would be more complex than the version check.

anakin87:
### Related Issues

- fixes #1998

### Proposed Changes:
- introduce a step in the `CI_pypi_release workflow` to validate version number: if the version number expressed via tag is incorrect, the release is blocked

### How did you test it?
I tested extensively the `validate_version` script locally.
I tested a simplified version of the workflow on a test repo: https://github.com/anakin87/test-repo/actions

### Checklist


.github/utils/validate_version.py
```diff
@@ -0,0 +1,58 @@
...
+    # connect to PyPi and check the latest version
+    try:
+        response = requests.get(f"https://pypi.org/pypi/{integration_package}/json")
+        response.raise_for_status()
+        latest_version = response.json()["info"]["version"]
+        print(f"Latest version on PyPI: {latest_version}")
+    except requests.exceptions.HTTPError as e:
+        if e.response.status_code == 404 and version_to_release in ["0.0.1", "0.1.0", "1.0.0"]:
+            print("Package not found on PyPI. Assuming this is the first release and skipping version check.")
+            return
```
sjrl, line 33: Just clarifying, but with this we are requiring new packages to use one of these versions when adding a new core-integration right?

anakin87: yes

PR #2400, opened 2025-10-18
fix: handle empty GitHub tokens in authentication headers

### LINKED ISSUE #2387: GitHub - Authorization failures due to empty Secret env var
**Describe the bug**
Accessing public repositories doesn't require a GitHub token.

When we set a `Secret` corresponding to an env var in `github_token`:
- If the env var is not present, we fail with an expressive error message.
- If the env var is set but empty (as it can happen on some platforms), we end up passing `headers["Authorization"] = "Bearer"`, causing the request to fail.

I would not say it's a proper bug, but we can make the implementation more robust.
 
**To Reproduce**
```python
from haystack_integrations.components.connectors.github import GitHubIssueViewer
import os

os.environ["GITHUB_TOKEN"] = ""

viewer = GitHubIssueViewer(github_token=Secret.from_env_var("GITHUB_TOKEN"))

result = viewer.run(url="https://github.com/deepset-ai/haystack-core-integrations/issues/2373")

# requests.exceptions.HTTPError: 401 Client Error: Unauthorized for url: ...
```

yashv6655:
### Related Issues

- fixes #2387

### Proposed Changes:

Fixed authentication header handling when GitHub token environment variable is empty.

**Problem**: Empty token values (`GITHUB_TOKEN=""`) caused `Authorization: Bearer ` headers to be sent, resulting in `401 Unauthorized` errors for public repositories.

**Solution**: Added truthy check before adding Authorization header. Now only non-empty tokens include the Authorization header, allowing anonymous access to public repos when token is empty or not provided.

**Components affected**: All 6 GitHub connector components (GitHubFileEditor, GitHubIssueCommenter, GitHubIssueViewer, GitHubPRCreator, GitHubRepoForker, GitHubRepoViewer)

### How did you test it?

Added unit tests for all 6 components verifying:
- Empty tokens don't add Authorization header
- Valid tokens are properly included
- Base headers remain present

### Notes for the reviewer

Consistent fix applied across all 6 GitHub connector components. Each `_get_request_headers()` method now resolves the token, checks if truthy, then conditionally adds the Authorization header.

### Checklist

### DIFF
--- a/integrations/github/src/haystack_integrations/components/connectors/github/file_editor.py
+++ b/integrations/github/src/haystack_integrations/components/connectors/github/file_editor.py
@@ -115,7 +115,9 @@ class GitHubFileEditor:
         """
         headers = self.base_headers.copy()
         if self.github_token is not None:
-            headers["Authorization"] = f"Bearer {self.github_token.resolve_value()}"
+            token_value = self.github_token.resolve_value()
+            if token_value:
+                headers["Authorization"] = f"Bearer {token_value}"
         return headers
 
     def _get_file_content(self, owner: str, repo: str, path: str, branch: str) -> tuple[str, str]:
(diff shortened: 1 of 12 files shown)


anakin87 (approved):
Good work. Thank you!

PR #2405, opened 2025-10-19
fest: add structured output support in `NvidiaChatGenerator`

Amnah199:
### Related Issues

- partly fixes https://github.com/deepset-ai/haystack/issues/9761

### Proposed Changes:

- Updated docstrings to include information about structured outputs.
- Add an example

### How did you test it?

Added a new test to test structured ouputs
### Notes for the reviewer

Nvidia doesn't support  json schema through `response_format` parameter like OpenAI. Pydantic models are also not supported.
Support for structured outputs in Nvidia is described here: https://docs.nvidia.com/nim/large-language-models/latest/structured-generation.html
### Checklist

### DIFF
--- a/integrations/nvidia/pyproject.toml
+++ b/integrations/nvidia/pyproject.toml
@@ -23,7 +23,7 @@ classifiers = [
   "Programming Language :: Python :: Implementation :: CPython",
   "Programming Language :: Python :: Implementation :: PyPy",
 ]
-dependencies = ["haystack-ai>=2.13.0", "requests>=2.25.0", "tqdm>=4.21.0"]
+dependencies = ["haystack-ai>=2.19.0", "requests>=2.25.0", "tqdm>=4.21.0"]
 
 [project.urls]
 Documentation = "https://github.com/deepset-ai/haystack-core-integrations/tree/main/integrations/nvidia#readme"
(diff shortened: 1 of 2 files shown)


integrations/openrouter/pyproject.toml
```diff
@@ -23,7 +23,7 @@ classifiers = [
   "Programming Language :: Python :: Implementation :: CPython",
   "Programming Language :: Python :: Implementation :: PyPy",
 ]
-dependencies = ["haystack-ai>=2.13.1"]
+dependencies = ["haystack-ai>=2.18.0"]
```
sjrl: I think the change here is unintentional for this PR right?

integrations/nvidia/pyproject.toml
```diff
@@ -160,4 +160,4 @@ addopts = "--strict-markers"
 markers = [
   "integration: integration tests",
 ]
-log_cli = true
+log_cli = true
```
sjrl, line 163: We are missing the bump to haystack-ai version in this pyproject.toml file

integrations/openrouter/pyproject.toml
Amnah199: The min haystack version should be where we added support for OpenAI structured outputs. That might 2.16, but do you think its better to support a lower version?

Amnah199: Ahhh I get it sorry, the wrong file. Let me fix it.

PR #2305, opened 2025-09-23
feat: add MFA authentication support to snowflake integration

ChinmayBansal:
### Related Issues

- fixes #2265

### Proposed Changes:
This PR implements multi-factor authentication (MFA) support for the Snowflake
  integration to address Snowflake's enhanced security requirements that mandate
  MFA for connections. The implementation adds non-interactive authentication
  methods while maintaining full backward compatibility.

  Key Features Added:
  - Key-pair JWT Authentication (SNOWFLAKE_JWT) that supports:
    - Pre-configured private key file paths via private_key_file parameter
    - Private key passphrases via private_key_file_pwd parameter
    - JWT-based authentication without user interaction prompts
  - OAuth 2.0 Authentication (OAUTH) supporting:
    - Client Credentials flow with oauth_client_id and oauth_client_secret
    - Configurable token request URLs via oauth_token_request_url
    - Authorization URLs via oauth_authorization_url for future extensibility
  - Enhanced Security Features:
    - Credential masking in URI logging via _create_masked_uri() method
    - Secure handling of all sensitive data using Haystack's Secret class
    - Comprehensive parameter validation with clear error messages
  - Non-interactive Design: All authentication credentials sourced from
  environment variables, eliminating user prompts as specifically requested
  - Backward Compatibility: Existing password-based authentication (SNOWFLAKE)
  remains unchanged with no breaking changes

### How did you test it?

- Unit Tests: Added comprehensive test suite with 11 new test cases (36 total
  tests passing)
  - Code Quality: hatch run fmt and hatch run test:types both pass
  - Security Testing: Verified credential masking in logs
  - Manual verification: Tested all authentication methods and parameter
  validation

### Notes for the reviewer

- The core authentication logic is in _snowflake_uri_constructor() method
  (lines ~257-301)
  - Parameter validation logic is centralized in _validate_auth_params() method
  (lines ~152-185)
  - Security enhancement via _create_masked_uri() method (lines ~303-329) ensures
   sensitive data never appears in logs
  - The implementation specifically addresses the "no user authentication
  prompts" requirement

### Checklist

- I have read the [contributors guidelines](https://github.com/deepset-ai/haystack-core-integrations/blob/main/CONTRIBUTING.md) and the [code of conduct](https://github.com/deepset-ai/haystack-core-integrations/blob/main/CODE_OF_CONDUCT.md)
- I have updated the related issue with new insights and changes
- I added unit tests and updated the docstrings
[...]

PR #2414, opened 2025-10-21
feat!: Add `warm_up` method to `SnowflakeTableRetriever` to delegate auth and database connection testing 

sjrl:
### Related Issues

- fixes https://github.com/deepset-ai/dc-pipeline-templates/issues/305
- This is a follow up to https://github.com/deepset-ai/haystack-core-integrations/pull/2305 which included a breaking change that was missed when creating the changelog and new version release. Namely that we added a new required positional parameter to the init of `SnowflakeTableRetriever`

### Proposed Changes:

 
 

- Reverts the breaking change of https://github.com/deepset-ai/haystack-core-integrations/pull/2305 by setting a default value for `authenticator` which matches the previous default behavior. We also set defaults for all Secrets instead of setting them to None by default.
- Move auth validation and database connection testing to the `warm_up` method of `SnowflakeTableRetriever`. This allows for the component to be initialized without needing all secrets present. This is useful if you want to perform Pipeline validation without needing all secrets present. 
  - This is a breaking change because it requires calling `warm_up` on the component before it can be used. This is only relevant if you are using this component in isolation. If being used in a Pipeline then the Pipeline will automatically call warm_up on it so usage of this component does not change. 

### How did you test it?

- Updated tests
- Locally tested to make sure https://github.com/deepset-ai/dc-pipeline-templates/issues/305 is resolved

### Checklist

### DIFF
--- a/integrations/snowflake/src/haystack_integrations/components/retrievers/snowflake/snowflake_table_retriever.py
+++ b/integrations/snowflake/src/haystack_integrations/components/retrievers/snowflake/snowflake_table_retriever.py
@@ -247,6 +247,7 @@ class SnowflakeTableRetriever:
         encoded_user = quote_plus(self.user)
         encoded_account = quote_plus(self.account)
 
+        # We ignore the mypy error since it doesn't know that self.authenticator_handler has been set at this point
         password = self.authenticator_handler.get_password_for_uri()  # type: ignore[union-attr]
         if password:
             # Traditional password authentication - encode password
@@ -273,6 +274,7 @@ class SnowflakeTableRetriever:
         params.append(f"login_timeout={self.login_timeout}")
 
         # Add authentication-specific parameters (pass user for JWT ADBC support)
+        # We ignore the mypy error since it doesn't know that self.authenticator_handler has been set at this point
         auth_params = self.authenticator_handler.build_auth_params(user=self.user)  # type: ignore[union-attr]
         params.extend(auth_params)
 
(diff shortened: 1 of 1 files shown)


integrations/snowflake/src/haystack_integrations/components/retrievers/snowflake/snowflake_table_retriever.py
```diff
@@ -244,7 +247,7 @@ def _snowflake_uri_constructor(self) -> str:
         encoded_user = quote_plus(self.user)
         encoded_account = quote_plus(self.account)
 
-        password = self.authenticator_handler.get_password_for_uri()
+        password = self.authenticator_handler.get_password_for_uri()  # type: ignore[union-attr]
```
anakin87, line 251: Why `type: ignore` is needed? In case we cannot avoid it, please add a comment

anakin87 (approved):
I am OK with this change, also given the low usage of this integration.

Let's remember to:
- check if the [example](https://github.com/deepset-ai/haystack-core-integrations/blob/main/integrations/snowflake/example/text2sql_example.py) needs an update
- release a new major version
- updated [docs](https://docs.haystack.deepset.ai/docs/snowflaketableretriever) and [integration page](https://haystack.deepset.ai/integrations/snowflake)

integrations/snowflake/src/haystack_integrations/components/retrievers/snowflake/snowflake_table_retriever.py
sjrl: Needed because the authenticator handler is only not None if the component is warmed up. This function is only callable if the component is already warmed up which is why I opted to ignore the error. I can add a comment

Issue #2409, opened 2025-10-20
Update Chat Generators to support mixed Tools and Toolsets


vblagoje:
### Background

In Haystack 2.19 we enhanced the `tools` parameter across all chat generators to accept `Optional[Union[list[Union[Tool, Toolset]], Toolset]]` (defined as `ToolsType`).

This enhancement allows users to:
- Pass a mixed list of `Tool` and `Toolset` objects
- Pass a single `Toolset` object directly
- Combine multiple toolsets with standalone tools in the same list

**Previous limitation:** Components required either a list of `Tool` objects OR a single `Toolset`, but not both in the same list.

**New capability:** Users can now organize tools into logical Toolsets while also including standalone Tool objects, providing greater flexibility. For example:
```python
generator = OpenAIChatGenerator(
    tools=[math_toolset, weather_toolset, standalone_tool]
)
```

This change is fully backward compatible and preserves structure during serialization/deserialization.

### Required Changes

All chat generators in `haystack-core-integrations` that support tools need to be updated to use the new `ToolsType` format. Based on the current inventory, the following integrations need updates:

#### Generators with Tool Support
- [x] `amazon-bedrock-haystack` - `AmazonBedrockChatGenerator`
- [x] `anthropic-haystack` - `AnthropicChatGenerator`
- [x] `cohere-haystack` - `CohereChatGenerator`
- [x] `google-genai-haystack` - Chat generators  
- [x] `llama-cpp-haystack` - `LlamaCppChatGenerator`
- [x] `llama-stack-haystack` - Chat generators
- [x] `meta-llama-haystack` - Chat generators
- [x] `mistral-haystack` - `MistralChatGenerator`
- [x] `nvidia-haystack` - Chat generators
- [x] `ollama-haystack` - `OllamaChatGenerator`
- [x] `openrouter-haystack` - Chat generators
- ~[ ] `stackit-haystack` - Chat generators~ tools not supported
- [x] `togetherai-haystack` - Chat generators
- ~[ ] `watsonx-haystack` - Chat generators~ tools not supported

### Reference Implementation

See the Haystack core implementations for complete examples:
- [`haystack/components/generators/chat/openai.py`](https://github.com/deepset-ai/haystack/blob/main/haystack/components/generators/chat/openai.py)

### Testing Considerations

Each integration should add tests to verify:
1. Accepting a list of `Tool` objects (existing behavior)
2. Accepting a single `Toolset` object (existing behavior)
[...]

PR #2422, opened 2025-10-21
feat: `NvidiaChatGenerator` add integration tests for mixing Tool/Toolset

vblagoje:
## Why:
Add integration tests for `ToolsType` Haystack 2.19 updates

part of:
- https://github.com/deepset-ai/haystack-core-integrations/issues/2409

## What:
- Added tests for mixed tool collection initialization and request parameter formatting

## How did you test it:
- Integration tests verify mixed Tool/Toolset usage
- Backward compatibility validated for existing patterns (list of tools, single toolset)

### DIFF
--- a/integrations/nvidia/tests/test_nvidia_chat_generator.py
+++ b/integrations/nvidia/tests/test_nvidia_chat_generator.py
@@ -339,53 +339,6 @@ class TestNvidiaChatGenerator:
         reason="Export an env var called NVIDIA_API_KEY containing the NVIDIA API key to run this test.",
     )
     @pytest.mark.integration
-    def test_integration_mixing_init_and_runtime_tools(self):
-        """Test mixing tools from init and runtime by passing tools to both __init__ and run()."""
-
-        def weather_function(city: str) -> str:
-            """Get weather information for a city."""
-            return f"Weather in {city}: 22°C, sunny"
-
-        def time_function(city: str) -> str:
-            """Get current time in a city."""
-            return f"Current time in {city}: 14:30"
-
-        # Create tools
-        weather_tool = Tool(
-            name="weather",
-            description="Get weather information for a city",
-            parameters={"type": "object", "properties": {"city": {"type": "string"}}, "required": ["city"]},
-            function=weather_function,
-        )
-
-        time_tool = Tool(
-            name="time",
-            description="Get current time in a city",
-            parameters={"type": "object", "properties": {"city": {"type": "string"}}, "required": ["city"]},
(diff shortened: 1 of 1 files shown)


integrations/nvidia/tests/test_nvidia_chat_generator.py
```diff
@@ -329,6 +329,114 @@ def test_live_run_with_json_object(self):
...
+
+        # Create Toolset with weather and time tools
+        toolset = Toolset([weather_tool, time_tool])
+
+        # Initialize with toolset
+        component = NvidiaChatGenerator(tools=toolset)
+
+        # Pass echo_tool as a list at runtime - runtime tools should take precedence
+        messages = [ChatMessage.from_user("Echo this: Hello World")]
+        results = component.run(messages, tools=[echo_tool])
```
Amnah199: Same comment about mixing tools and toolsets at runtime.

integrations/nvidia/tests/test_nvidia_chat_generator.py
```diff
@@ -329,6 +329,114 @@ def test_live_run_with_json_object(self):
         assert isinstance(output["rating"], int)
         assert "Inception" in output["title"]
 
+    @pytest.mark.skipif(
+        not os.environ.get("NVIDIA_API_KEY", None),
+        reason="Export an env var called NVIDIA_API_KEY containing the NVIDIA API key to run this test.",
+    )
+    @pytest.mark.integration
+    def test_integration_mixing_init_and_runtime_tools(self):
+        """Test mixing tools from init and runtime by passing tools to both __init__ and run()."""
```
Amnah199: I am not sure if this test was intended. We can rather have a test for ser/deser of mixed tool lists like in GoogleGenAiChatGenerator.

integrations/nvidia/tests/test_nvidia_chat_generator.py
vblagoje: but this one is different: there is no mention of tools and we need to wrap single tool into a list

integrations/nvidia/tests/test_nvidia_chat_generator.py
Amnah199: @vblagoje do you still think we need this test?

Amnah199: @vblagoje lets remove this test. I still dont see the need of keeping it. The rest looks good.

Amnah199 (approved):

PR #2436, opened 2025-10-22
feat: `LlamaStackChatGenerator` update tools param to ToolsType

vblagoje:
## Why:
Adopts Haystack's `ToolsType` to enable flexible tool composition, developers can now mix individual `Tool` objects and `Toolset` instances in a single list

part of:
- https://github.com/deepset-ai/haystack-core-integrations/issues/2409

## What:
- Updated `tools` parameter from `Union[List[Tool], Toolset]` to `ToolsType` in constructor
- Replaced manual tool handling with Haystack's built-in `ToolsType` support
- Haystack dependency to `2.19.0` for `ToolsType` support
- Added tests for mixed tool collection initialization and request parameter formatting

## How can it be used:

```python
from haystack.tools import Tool, Toolset
from haystack_integrations.components.generators.llama_stack import LlamaStackChatGenerator

weather_tool = Tool(name="weather", ...)
population_toolset = Toolset([...])

# NEW: Mix Tool and Toolset freely
generator = LlamaStackChatGenerator(tools=[weather_tool, population_toolset])
```

## How did you test it:
- Unit tests verify mixed tool initialization and correct parameter handling
- Backward compatibility validated for existing patterns (list of tools, single toolset)

## Notes for the reviewer:
- Centralizes tool normalization logic with Haystack's `ToolsType`—eliminates duplicate handling code
- Fully backward compatible; no migration needed
- Added integration test `test_live_run_with_mixed_tools()` demonstrates end-to-end LLM tool invocation from mixed collections

### DIFF
--- a/integrations/llama_stack/examples/llama-stack-with-tools.py
+++ b/integrations/llama_stack/examples/llama-stack-with-tools.py
@@ -34,7 +34,7 @@ weather_tool = Tool(
 tool_invoker = ToolInvoker(tools=[weather_tool])
 
 client = LlamaStackChatGenerator(
-    model="ollama/llama3.2:3b",  # model name varies depending on the inference provider used for the Llama Stack Server.
+    model="ollama/llama3.2:3b",  # model depends on the inference provider used for the Llama Stack Server.
     api_base_url="http://localhost:8321/v1/openai/v1",
 )
 messages = [ChatMessage.from_user("What's the weather in Tokyo?")]


.github/workflows/llama_stack.yml
```diff
@@ -82,13 +82,30 @@ jobs:
...
         shell: bash
         run: |
-          pip install uv
-          uv run --with llama-stack llama stack build --distro starter --image-type venv --run < /dev/null > server.log 2>&1 &
-          sleep 120
-          # Verify it's running
-          curl -f http://localhost:8321/v1/models || { cat server.log; exit 1; }
-          
-          echo "Llama Stack Server started successfully."
+          set -euo pipefail
```
anakin87, line 85: @Amnah199 please take a look

integrations/llama_stack/tests/test_llama_stack_chat_generator.py
```diff
@@ -66,27 +90,27 @@ def mock_chat_completion():
 
 class TestLlamaStackChatGenerator:
     def test_init_default(self):
-        component = LlamaStackChatGenerator(model="llama3.2:3b")
-        assert component.model == "llama3.2:3b"
+        component = LlamaStackChatGenerator(model="ollama/llama3.2:3b")
```
anakin87, line 93: Why are we changing the way we pass the model name? Is it necessary?

vblagoje: It seems to be - it was failing for the "llama3.2:3b". I looked at the CI run, the model was booted and ready. Then I tried this naming schema and it worked

vblagoje:
> Looks good from the `ToolsType` point of view.
> 
> However, I'd appreciate it if @Amnah199 could take a look at other changes, since she is more familiar with this integration.

Yes makes sense. @Amnah199 I spent half the day yesterday to get this integration  up and running because the old llama stack build command doesn't work anymore. Then I somehow made it run with lamma stack run but the model wasn't recognized without full name. And then with full name it finally started to work.

vblagoje:
Ah sorry @anakin87 I clicked by accident thinking it was Amna

.github/workflows/llama_stack.yml
vblagoje: So what was happening is [this](https://github.com/deepset-ai/haystack-core-integrations/actions/runs/18715590595/job/53374087764) look at the date stamp - 2 days ago. And it seems like llama stack doesn't have build command any more. So I switched to run command because that one made sense. Using run made a [difference](https://github.com/deepset-ai/haystack-core-integrations/actions/runs/18753973194/job/53500871804#step:6:290) and the server started properly.

Then I faced this issue of names. Tests would fail saying there is no such model. I then switched to full naming schema and then everything started to work. cc @anakin87 @Amnah199

anakin87 (approved):
👍 

Please also open a PR to update https://haystack.deepset.ai/integrations/llama_stack

For some reason, [docs](https://docs.haystack.deepset.ai/docs/llamastackchatgenerator) are already correct.

vblagoje:
> 👍
> 
> Please also open a PR to update https://haystack.deepset.ai/integrations/llama_stack
> 
> For some reason, [docs](https://docs.haystack.deepset.ai/docs/llamastackchatgenerator) are already correct.

That's crazy, premonition of a change! We were correct about their naming scheme even before them

vblagoje:
We need to verify and merge https://github.com/deepset-ai/haystack-integrations/pull/368 cc @bilgeyucel

PR #2463, opened 2025-10-26
feat/fix(langfuse): usage and cost tracking for mistral components

### LINKED ISSUE #2464: Langfuse SDK alignment & Support for embedders
When using MistralChatGenerator, MistralTextEmbedder, or MistralDocumentEmbedder with the Langfuse integration, usage metrics and cost details are not tracked properly. This is because these components are
  created as SPAN observation types instead of GENERATION and EMBEDDING types respectively.

  The Langfuse SDK only supports usage_details and cost_details parameters for ObservationTypeGenerationLike types (i.e., generation and embedding), but not for SPAN types. Without proper observation typing, it's
  impossible to track LLM costs and token usage for Mistral components, which is critical for production monitoring and cost management.

  Add Mistral components to the appropriate whitelists in the Langfuse tracer:
  1. Add MistralChatGenerator to the _SUPPORTED_CHAT_GENERATORS list so it's created as a generation observation type
  2. Add MistralTextEmbedder and MistralDocumentEmbedder to a new _SUPPORTED_EMBEDDERS list and update the tracer logic to create them as embedding observation types

  1. Manual workaround: Directly monkey-patching the tracer or manually logging usage - this is brittle and defeats the purpose of the integration
  2. Using only supported generators: Switch from Mistral to OpenAI/Anthropic - not ideal when Mistral is the preferred provider

**Additional context**
  - Mistral components inherit from OpenAI components (MistralChatGenerator(OpenAIChatGenerator), MistralTextEmbedder(OpenAITextEmbedder)), so they have the same interface and return usage data in the same format
  - Langfuse Python SDK @ 3.8.0

Hansehart:
### Related Issues

Fixes #2464 

### Proposed Changes:

Bug description:
In DefaultSpanHandler.handle, the Langfuse span update call incorrectly used the parameter usage= instead of usage_details=, which is silently ignored for SPAN types in the Langfuse SDK. This caused token usage metrics to be dropped from Langfuse traces.

On top, MistralChatGenerator, MistralTextEmbedder, and MistralDocumentEmbedder were not in the supported component lists, causing them to be created as generic SPAN observation types. The Langfuse SDK only supports usage_details and cost_details for ObservationTypeGenerationLike types (generation and embedding), not for SPAN types. This prevented cost and usage tracking for all Mistral components.
  
Fix implemented:
Replaced the incorrect usage= argument with the correct usage_details= when updating Langfuse spans for generator and chat generator components.

In addition:
  1. Added MistralChatGenerator to _SUPPORTED_CHAT_GENERATORS list → creates generation observation type
  2. Created new _SUPPORTED_EMBEDDERS list with MistralTextEmbedder and MistralDocumentEmbedder → creates embedding observation type

### How did you test it?

- unit tests
- manual verification

### Notes for the reviewer

https://python.reference.langfuse.com/langfuse
Langfuse@3.8.0

### Checklist

- I have read the contributors guidelines and the code of conduct
- I added unit tests and updated the docstrings (WIP)
- I've used one of the conventional commit types for my PR title: fix:, feat:, build:, chore:, ci:, docs:, style:, refactor:, perf:, test:.

### DIFF
--- a/integrations/langfuse/tests/test_tracer.py
+++ b/integrations/langfuse/tests/test_tracer.py
@@ -415,7 +415,7 @@ class TestLangfuseTracer:
             }
             with tracer.trace(operation_name="operation_name", tags=tags) as span:
                 ...
-            assert span.raw_span()._data["usage"] is None
+            assert span.raw_span()._data["usage_details"] is None
             assert span.raw_span()._data["model"] == "test_model"
             assert span.raw_span()._data["completion_start_time"] == datetime.datetime(2021, 7, 27, 16, 2, 8, 12345)  # noqa: DTZ001
 
@@ -482,7 +482,7 @@ class TestLangfuseTracer:
             }
             with tracer.trace(operation_name="operation_name", tags=tags) as span:
                 ...
-            assert span.raw_span()._data["usage"] is None
+            assert span.raw_span()._data["usage_details"] is None
             assert span.raw_span()._data["model"] == "test_model"
             assert span.raw_span()._data["completion_start_time"] is None
 


Hansehart:
@vblagoje Thanks for your answer.

I have to admit that I’m addressing multiple things at once — sorry about that. The first problem is:

1. Langfuse Python SDK expects usage_details instead of usage. I double checked it in the source code and you can see it here: https://python.reference.langfuse.com/langfuse. So that is patch number 1: changing usage to usage_details. IDK if Langfuse changed it recently?

2. The LangfuseConnector only supports costs and usage for traces of the type "generation" which is decided on the generators whitelist. Any other component is created as Langfuse type "span". The problem here is, that Langfuse currently only supports costs and usage details for generators and embeddings. So my second patch is, that we create a second whitelist for embedders. If they are recognized, they are created as type "embedders" (instead span) which will prevent dropping my cost and usage details by Langfuse. You can see it here:
```
elif context.component_type in _SUPPORTED_EMBEDDERS:
    return LangfuseSpan(self.tracer.start_as_current_observation(name=context.name, as_type="embedding"))
else:
    return LangfuseSpan(self.tracer.start_as_current_span(name=context.name))
```
Would you mind tracing a pipeline that uses embedders? You should see it will be shown as type "span" so any costs and usage details will be dropped when you insert it manually (you can hover in the GUI over the icons and you will see on your example trace, multiple spans and some generators).


2.5 As mentioned Langfuse has a list of types they support. There are few like Chain, Events, Tools etc. It should be considered inclduing them. That came in my mind as I saw on your trace that the tools are displayed as span (should be type "tool" maybe?)

3. The last problem that remains is, but it is depends on Langfuse and not haystack, costs and usage are only supported for type generators and embedders. I currently have more components like OCR and external API (market data, news data) with usage_details and cost_details too that will be dropped by Langfuse. That is something that remains unresolved till Langfuse adjusts it.

So finally the PR is about migrating to usage_details and support for any embedders (my PR only mistral)

PR #2475, opened 2025-10-28
ci: Change pytest command

sjrl:
### Related Issues

- fixes #issue-number

### Proposed Changes:

 
 

I recently noticed in some failing test logs that it looked like we weren't rerunning failing tests despite having `pytest-rerunfailures` set up. Additionally it appeared like the `-x` command was being ignored as well. 

Here is an example failing [test log](https://github.com/deepset-ai/haystack-core-integrations/actions/runs/18860087991/job/53816483824#step:10:60) for meta-llama where we don't appear to rerun the failed test (missing the RERUN log from pytest) and we still continue to run tests after that one failed which goes against the `-x` option. 

It appears something goes wrong when using `hatch run test:cov-retry` with the configuration
```
all = 'pytest {args:tests}'
cov-retry = 'all --cov=haystack_integrations --reruns 3 --reruns-delay 30 -x'
```
when I changed it to (same command as the one we use in haystack-experimental)
```
cov-retry = 'pytest --cov="haystack_integrations" --reruns 3 --reruns-delay 30 -x {args:tests}'
```
then I was able to see RERUN logs and early exits occur. Here is an [example](https://github.com/deepset-ai/haystack-core-integrations/actions/runs/18867981549/job/53839445653#step:8:44) for meta-llama. 

### Checklist

### DIFF
--- a/integrations/aimlapi/pyproject.toml
+++ b/integrations/aimlapi/pyproject.toml
@@ -64,7 +64,7 @@ dependencies = [
 unit = 'pytest -m "not integration" {args:tests}'
 integration = 'pytest -m "integration" {args:tests}'
 all = 'pytest {args:tests}'
-cov-retry = 'all --cov=haystack_integrations --reruns 3 --reruns-delay 30 -x'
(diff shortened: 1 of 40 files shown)


anakin87 (approved):
Thanks for the fix!

PR #2491, opened 2025-11-06
fix: Flatten usage_details dict

vblagoje:
## Why:
Fixes serialization failures when tracing AnthropicChatGenerator usage data. Langfuse requires flat dictionaries for `usage_details`, but Anthropic returns nested dicts with non-numeric values.

- fixes Anthropic Langfuse [failures](https://github.com/deepset-ai/haystack-core-integrations/actions/runs/18958744566/job/54177114465)

## What:
- Added `_sanitize_usage_data()` to flatten nested usage dictionaries using dot notation
- Filters out None values (`server_tool_use`), strings (`service_tier`), and booleans
- Applied to both `_SUPPORTED_GENERATORS` and `_SUPPORTED_CHAT_GENERATORS` handlers

## How can it be used:
Transparent to users. AnthropicChatGenerator usage now traces successfully:

```python
# Before: FAILED with nested dict
{'cache_creation': {'ephemeral_1h_input_tokens': 0}, 'server_tool_use': None, 'service_tier': 'standard'}

# After: PASSES with flat dict
{'cache_creation.ephemeral_1h_input_tokens': 0, 'prompt_tokens': 25, 'completion_tokens': 449}
```

## How did you test it:
- Tested with AnthropicChatGenerator (nested dicts, None, strings) 
- Verified OpenAI and Cohere generators still work 
- Confirmed root cause: nested dicts fail even with only numeric values

## Notes for the reviewer:
Key finding: Langfuse `usage_details` signature is `Optional[Dict[str, int]]` but actually requires flat structure. Nested dicts cause serialization to fail regardless of value types.

### DIFF
--- a/integrations/langfuse/tests/test_tracer.py
+++ b/integrations/langfuse/tests/test_tracer.py
@@ -18,6 +18,7 @@ from haystack_integrations.tracing.langfuse.tracer import (
     LangfuseSpan,
     LangfuseTracer,
     SpanContext,
+    _sanitize_usage_data,
 )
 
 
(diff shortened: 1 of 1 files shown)


integrations/langfuse/src/haystack_integrations/tracing/langfuse/tracer.py
```diff
@@ -256,6 +256,46 @@ def to_dict(self) -> Dict[str, Any]:
         return default_to_dict(self)
 
 
+def _sanitize_usage_data(usage: Dict[str, Any]) -> Dict[str, Any]:
+    """
+    Sanitize usage data for Langfuse by flattening to a single-level dictionary.
+
+    Langfuse's usage_details must be a flat dictionary with only numeric values. This function:
```
sjrl, line 263: Is there a docstring we could provide here that explains this limitation in langfuse?

vblagoje:
> @vblagoje thanks for the fix! Do you know what changed on Anthropics side to cause this issue? Also this change impacts other generators too right? I think OpenAI usage is also a nested dict but for some reason that wasn't causing issues?

Nothing changed on their side @sjrl , they always had this type of usage in meta:
```
'usage': {'cache_creation': {'ephemeral_1h_input_tokens': 0, 'ephemeral_5m_input_tokens': 0}, 'cache_creation_input_tokens': 0, 'cache_read_input_tokens': 0, 'server_tool_use': None, 'service_tier': 'standard', 'prompt_tokens': 25, 'completion_tokens': 449}}
```

The problem arose when we started using this new param in Langfuse usage_details from https://github.com/deepset-ai/haystack-core-integrations/pull/2481 as Langfuse barfs on multi-level dict. Most likely they'll fix it but for now this one does it. Intentionally as simple as possible

integrations/langfuse/src/haystack_integrations/tracing/langfuse/tracer.py
vblagoje: We can leave this PR notes I just added as the motivation for this change and then later, if Langfuse fixes this, we can remove it, wdyt?

sjrl: Ahh okay I gotcha. And yeah sounds good

vblagoje:
@sjrl yes, incoming.

PR #2501, opened 2025-11-10
feat: Add Agent state-mapping parameters to MCPTool

vblagoje:
## Why:
Enables MCPTool to support Agent state-mapping parameters (`outputs_to_string`, `inputs_from_state`, `outputs_to_state`), allowing easy integration with Agent state for automatic parameter injection and output handling without requiring wrapper functions.

## What:
- Added three state-mapping parameters to `MCPTool.__init__` signature with proper type hints
- Implemented parameter storage and propagation to `Tool` superclass in both eager and lazy connection paths
- Added schema adjustment logic to remove `inputs_from_state` keys from parameters, matching `ComponentTool` behavior
- Extended serialization/deserialization in `to_dict`/`from_dict` to preserve state-mapping configuration
- Added two comprehensive tests: `test_mcp_tool_state_mapping_parameters` and `test_mcp_tool_serde_with_state_mapping`

## How can it be used:
```python
from haystack_integrations.tools.mcp import MCPTool, StreamableHttpServerInfo

# Before: Required wrapper function to use inputs_from_state
server_info = StreamableHttpServerInfo(url="http://localhost:1417/mcp")
mcp_tool = MCPTool(name="mcp_tool", server_info=server_info)

def rag_via_mcp(query: str, filter: dict) -> str:
    return mcp_tool.invoke(query=query, filter=filter)

get_latest_version_tool = Tool(
    name="get_latest_version",
    description="Get the latest Haystack version.",
    parameters={"type": "object", "properties": {"query": {"type": "string"}}, "required": ["query"]},
    function=rag_via_mcp,
    inputs_from_state={"filter": "filter"}
)

# After: Direct MCPTool usage with state-mapping
mcp_tool = MCPTool(
    name="mcp_tool",
    server_info=StreamableHttpServerInfo(url="http://localhost:1417/mcp"),
    inputs_from_state={"filter": "filter"}  # Automatically inject 'filter' from Agent state
)

# Use directly in Agent
agent = Agent(tools=[mcp_tool], ...)
```

## How did you test it:
- Added `test_mcp_tool_state_mapping_parameters` to verify parameter initialization and storage
- Added `test_mcp_tool_serde_with_state_mapping` to validate serialization/deserialization preserves configuration
- Confirmed backward compatibility with existing MCPTool usage patterns

## Notes for the reviewer:
Pay attention to the schema adjustment logic in lines 985-991 (eager connect) and 1106-1112 (lazy connect via `warm_up`), which removes `inputs_from_state` keys from the JSON schema to match `ComponentTool` behavior. The state-mapping parameters are stored with underscore prefixes (`_outputs_to_string`, etc.) for serialization while being passed directly to the parent `Tool` class.

### DIFF
--- a/integrations/mcp/src/haystack_integrations/tools/mcp/mcp_tool.py
+++ b/integrations/mcp/src/haystack_integrations/tools/mcp/mcp_tool.py
@@ -925,11 +925,14 @@ class MCPTool(Tool):
         :param outputs_to_string: Optional dictionary defining how tool outputs should be converted into a string.
                                  If the source is provided only the specified output key is sent to the handler.
                                  If the source is omitted the whole tool result is sent to the handler.
+                                 Example: `{"source": "docs", "handler": my_custom_function}`
         :param inputs_from_state: Optional dictionary mapping state keys to tool parameter names.
                                  Example: `{"repository": "repo"}` maps state's "repository" to tool's "repo" parameter.
         :param outputs_to_state: Optional dictionary defining how tool outputs map to keys within state as well as
                                 optional handlers. If the source is provided only the specified output key is sent
                                 to the handler.
+                                Example with source: `{"documents": {"source": "docs", "handler": custom_handler}}`
+                                Example without source: `{"documents": {"handler": custom_handler}}`
         :raises MCPConnectionError: If connection to the server fails
         :raises MCPToolNotFoundError: If no tools are available or the requested tool is not found
         :raises TimeoutError: If connection times out
--- a/integrations/mcp/tests/test_mcp_tool.py
+++ b/integrations/mcp/tests/test_mcp_tool.py
@@ -138,20 +138,31 @@ class TestMCPTool:
         server_info = InMemoryServerInfo(server=calculator_mcp._mcp_server)
 
         # Create tool with state-mapping parameters
+        # Map state key "state_a" to tool parameter "a"
         tool = MCPTool(
             name="add",
             server_info=server_info,
             eager_connect=False,
             outputs_to_string={"source": "result", "handler": str},
-            inputs_from_state={"filter": "query_filter"},
+            inputs_from_state={"state_a": "a"},
             outputs_to_state={"result": {"source": "output", "handler": str}},
         )
         mcp_tool_cleanup(tool)
 
         # Verify the parameters are stored correctly
         assert tool._outputs_to_string == {"source": "result", "handler": str}
-        assert tool._inputs_from_state == {"filter": "query_filter"}
+        assert tool._inputs_from_state == {"state_a": "a"}
         assert tool._outputs_to_state == {"result": {"source": "output", "handler": str}}
+
+        # Warm up the tool to trigger schema adjustment
+        tool.warm_up()
+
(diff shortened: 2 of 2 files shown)


integrations/mcp/src/haystack_integrations/tools/mcp/mcp_tool.py
```diff
@@ -914,6 +922,14 @@ def __init__(
         :param eager_connect: If True, connect to server during initialization.
                              If False (default), defer connection until warm_up or first tool use,
                              whichever comes first.
+        :param outputs_to_string: Optional dictionary defining how tool outputs should be converted into a string.
+                                 If the source is provided only the specified output key is sent to the handler.
+                                 If the source is omitted the whole tool result is sent to the handler.
```
sjrl, line 927: Let's add an example of how the dict looks 

Example:
```python
{
    "source": "docs", "handler": my_custom_function
}
```

integrations/mcp/src/haystack_integrations/tools/mcp/mcp_tool.py
```diff
@@ -914,6 +922,14 @@ def __init__(
...
+        :param inputs_from_state: Optional dictionary mapping state keys to tool parameter names.
+                                 Example: `{"repository": "repo"}` maps state's "repository" to tool's "repo" parameter.
+        :param outputs_to_state: Optional dictionary defining how tool outputs map to keys within state as well as
+                                optional handlers. If the source is provided only the specified output key is sent
+                                to the handler.
```
sjrl, line 933: Same here let's add an example of how the dict looks 

        Example:
        ```python
        {
            "documents": {"source": "docs", "handler": custom_handler}
        }
        ```
        If the source is omitted the whole tool result is sent to the handler.
        Example:
        ```python
        {
            "documents": {"handler": custom_handler}
        }
        ```

integrations/mcp/tests/test_mcp_tool.py
```diff
@@ -133,6 +133,58 @@ def test_mcp_tool_serde(self, mcp_tool_cleanup):
...
+            outputs_to_string={"source": "result", "handler": str},
+            inputs_from_state={"filter": "query_filter"},
+            outputs_to_state={"result": {"source": "output", "handler": str}},
+        )
+        mcp_tool_cleanup(tool)
+
+        # Verify the parameters are stored correctly
+        assert tool._outputs_to_string == {"source": "result", "handler": str}
+        assert tool._inputs_from_state == {"filter": "query_filter"}
+        assert tool._outputs_to_state == {"result": {"source": "output", "handler": str}}
```
sjrl, line 155: would it also be possible to verify the updated parameter schema to see that `query_filter` has been removed from the parameter schema?

sjrl:
@vblagoje looking good! I also wanted to ask do we have any integration test where we run an MCPTool with a live Agent? If so it'd be nice to add an integration test that uses these new variables so we can see that the returned State contains the outputs in `outputs_to_state` and everything else runs as expected.

vblagoje:
@sjrl please see test_agent_with_state_mapping test method, it should be a nice example of how state schema is used and mapping from state as tool parameter resolved, right?

PR #2504, opened 2025-11-11
Add missing GA workflow for `aimlapi` integration

mpangrazzi:
As title, this is simply adding the missing GA workflow for `aimlapi` integration.

Note that we need to have a `AIMLAPI_API_KEY` env var for running integrations tests correctly.


anakin87 (approved):
Thanks!

PR #2534, opened 2025-11-19
feat: support structured outputs in `TogetherAIChatGenerator`

Amnah199:
### Related Issues

- partially fixes https://github.com/deepset-ai/haystack/issues/9761

### Proposed Changes:

- Update the docstrings
- support serialization of pydantic models
- Add an example

### How did you test it?

- Updated a unit test
- Added integration tests
### Checklist

### DIFF
--- a/integrations/togetherai/examples/chatgenerator_with_structured_outputs.py
+++ b/integrations/togetherai/examples/chatgenerator_with_structured_outputs.py
@@ -32,4 +32,4 @@ chat_messages = [
 component = TogetherAIChatGenerator(generation_kwargs={"response_format": NobelPrizeInfo})
 results = component.run(chat_messages)
 
-#print(results)
+# print(results)
--- a/integrations/togetherai/pyproject.toml
+++ b/integrations/togetherai/pyproject.toml
@@ -65,7 +65,7 @@ dependencies = [
 [tool.hatch.envs.test.scripts]
 unit = 'pytest -m "not integration" {args:tests}'
 integration = 'pytest -m "integration" {args:tests}'
-all = 'pytest --reruns-delay 30 -x {args:tests}'
+all = 'pytest --reruns 3 --reruns-delay 30 -x {args:tests}'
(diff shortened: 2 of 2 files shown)


integrations/togetherai/src/haystack_integrations/components/generators/togetherai/chat/chat_generator.py
```diff
@@ -131,6 +140,21 @@ def to_dict(self) -> dict[str, Any]:
...
+        response_format = generation_kwargs.get("response_format")
+        # If the response format is a Pydantic model, it's converted to openai's json schema format
+        # If it's already a json schema, it's left as is
+        if response_format and isinstance(response_format, type) and issubclass(response_format, BaseModel):
+            json_schema = {
+                "type": "json_schema",
+                "json_schema": {
+                    "name": response_format.__name__,
+                    "strict": True,
+                    "schema": to_strict_json_schema(response_format),
```
mpangrazzi, line 153: I was wondering if instead of using a private helper (which works but might be subject to change), we could create our own wrapper which outputs a strict JSON schema from a Pydantic model. WDYT?

mpangrazzi (commented):
Looks already good! ~~I've left a comment~~.

integrations/togetherai/src/haystack_integrations/components/generators/togetherai/chat/chat_generator.py
mpangrazzi: I've investigated a bit, tried to create another model using `create_model` and overriding the config (forbidding extra fields as required), but it doesn't seem to work for nested models.

So I guess we can use that private helper for now!

mpangrazzi (approved):

integrations/togetherai/src/haystack_integrations/components/generators/togetherai/chat/chat_generator.py
Amnah199: Yeah its a simpler fix for the moment.

Issue #2326, opened 2025-09-26
add `update_by_filter()` and `delete_by_filter()` operation to `ElasticsearchDocumentStore`


davidsbatista:
```python
def delete_by_filter(self, filters: Dict[str, Any]) -> int
```

```python
def update_by_filter(self, filters: Dict[str, Any], meta: Dict[str, Any]) -> int
```

davidsbatista:
yes, title is wrong - I will correct it

PR #2582, opened 2025-12-02
feat: adding update and delete by filter to `ElasticsearchDocumentStore`

davidsbatista:
### Related Issues

- fixes https://github.com/deepset-ai/haystack-core-integrations/issues/2326

### Proposed Changes:

Added four new methods to `ElasticsearchDocumentStore`

- `delete_by_filter(filters: dict[str, Any]) -> int`
- `delete_by_filter_async(filters: dict[str, Any]) -> int`
- `update_by_filter(filters: dict[str, Any], meta: dict[str, Any]) -> int`
- `update_by_filter_async(filters: dict[str, Any], meta: dict[str, Any]) -> int`

### How did you test it?

Four new tests added
- `test_delete_by_filter()` - Tests synchronous deletion by filter
-  `test_update_by_filter()` - Tests synchronous metadata updates
-  `test_delete_by_filter_async()` - Tests async deletion by filter
-  `test_update_by_filter_async()` - Tests async metadata updates

### Checklist

### DIFF
--- a/integrations/elasticsearch/src/haystack_integrations/document_stores/elasticsearch/document_store.py
+++ b/integrations/elasticsearch/src/haystack_integrations/document_stores/elasticsearch/document_store.py
@@ -43,6 +43,7 @@ UPDATE_SCRIPT = """
             }
             """
 
+
 class ElasticsearchDocumentStore:
     """
     An ElasticsearchDocumentStore instance that works with Elastic Cloud or your own


integrations/elasticsearch/src/haystack_integrations/document_stores/elasticsearch/document_store.py
```diff
@@ -677,6 +677,128 @@ async def delete_all_documents_async(self, recreate_index: bool = False) -> None
...
+        self._ensure_initialized()
+
+        try:
+            normalized_filters = _normalize_filters(filters)
+            # Build the update script to modify metadata fields
+            # Documents are stored with flattened metadata, so update fields directly in ctx._source
+            update_script_lines = []
+            for key in meta.keys():
+                update_script_lines.append(f"ctx._source.{key} = params.{key};")
+            update_script = " ".join(update_script_lines)
```
mpangrazzi: I think this can allow Painless script injection. Like if I have:

```python
meta = {"foo; ctx._source.clear(); ctx._source.foo":  "bar"}
```

(I didn't verify it).

Maybe we can avoid interpolation and pass directly params to the script?

```python
script_source = """
for (entry in params. updates.entrySet()) {
    ctx._source[entry.getKey()] = entry.getValue();
}
"""
body = {
    "query": {"bool":  {"filter": normalized_filters}},
    "script": {"source":  script_source, "params": {"updates": meta}, "lang": "painless"},
}
```

WDYT?

integrations/elasticsearch/tests/test_document_store.py
```diff
@@ -760,3 +810,59 @@ async def test_delete_all_documents_async_no_index_recreation(self, document_sto
...
+            Document(content="Doc 3", meta={"category": "A"}),
+        ]
+        await document_store.write_documents_async(docs)
+        assert await document_store.count_documents_async() == 3
+
+        # Delete documents with category="A"
+        deleted_count = await document_store.delete_by_filter_async(
+            filters={"field": "category", "operator": "==", "value": "A"}
+        )
+        time.sleep(2)  # wait for deletion to be reflected
```
mpangrazzi: Not really as issue, but I would use async sleep here: `await asyncio.sleep(2)`.

Maybe we can add `refresh` support to avoid waiting? This way we could have immediate consistency.

integrations/elasticsearch/tests/test_document_store.py
```diff
@@ -760,3 +810,59 @@ async def test_delete_all_documents_async_no_index_recreation(self, document_sto
...
+            Document(content="Doc 3", meta={"category": "A", "status": "draft"}),
+        ]
+        await document_store.write_documents_async(docs)
+        assert await document_store.count_documents_async() == 3
+
+        # Update status for category="A" documents
+        updated_count = await document_store.update_by_filter_async(
+            filters={"field": "category", "operator": "==", "value": "A"}, meta={"status": "published"}
+        )
+        time.sleep(2)  # wait for update to be reflected
```
mpangrazzi: same as above ;)

integrations/elasticsearch/src/haystack_integrations/document_stores/elasticsearch/document_store.py
davidsbatista: good point! I will updated it

integrations/elasticsearch/src/haystack_integrations/document_stores/elasticsearch/document_store.py
```diff
@@ -677,6 +677,131 @@ async def delete_all_documents_async(self, recreate_index: bool = False) -> None
...
```
mpangrazzi: I think we can update also this as we did with sync `update_by_filter`

integrations/elasticsearch/tests/test_document_store.py
mpangrazzi: This still need to be updated?

mpangrazzi (approved):
Looks good! One last nit:

```python
update_script = """
...
```

is exactly the same on sync / async methods and can be extracted as a constant.

---

Also, I would like to open another issue about `refresh` usage. IMHO we should consider exposing it as a param on any method which updates the index.

PR #2590, opened 2025-12-08
chore: Harden OpenRouter integration tests

vblagoje:
- more test fixes to reduce occasional OpenRouter integration failures

### DIFF
--- a/integrations/openrouter/tests/test_openrouter_chat_generator.py
+++ b/integrations/openrouter/tests/test_openrouter_chat_generator.py
@@ -494,7 +494,9 @@ class TestOpenRouterChatGenerator:
                 "You must respond with a JSON object containing 'city' and 'country' fields."
             )
         ]
-        comp = OpenRouterChatGenerator(generation_kwargs={"response_format": response_schema})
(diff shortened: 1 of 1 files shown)


integrations/openrouter/tests/test_openrouter_chat_generator.py
```diff
@@ -486,15 +486,33 @@ def test_live_run_with_response_format_json_schema(self):
             },
         }
 
-        chat_messages = [ChatMessage.from_user("What's the capital of France?")]
+        # Use a more explicit prompt that emphasizes JSON structure requirement
+        # gpt-5-mini is very rarely flaky but to harden CI we'll be more explicit
```
sjrl, line 490: Since we mention gpt-5-mini in the dev comment here perhaps we should explicitly specify gpt-mini in the init params of OpenRouterChatGenerator?

vblagoje: I know, I know - having experienced pain of changing the default model across the codebase I was reluctant to add it here 😄

Issue #2587, opened 2025-12-04
Bug (Langfuse): _sanitize_usage_data unexpectedly flattens OpenAI-style usage format


LastRemote:
**Describe the bug**
In langfuse-haystack>=3.0.0 we now use _sanitize_usage_data for better feeding into Langfuse specification. However, this unexpectedly flattens the usage data format in OpenAI style, e.g.:
```json
{
    "completion_tokens": 1080,
    "completion_tokens_details": {
        "accepted_prediction_tokens": 0,
        "audio_tokens": 0,
        "reasoning_tokens": 0,
        "rejected_prediction_tokens": 0
    },
    "prompt_tokens": 4126,
    "prompt_tokens_details": {
        "audio_tokens": 0,
        "cached_tokens": 4096
    },
    "total_tokens": 5206
}
```
After the flattening, it seems like the data is no longer recognized by Langfuse, and it falled back to the default behavior, showing 0->0 (14508) instead.

> Langfuse summarizes all usage types that include the string input as input usage types, similarly output as output usage types. If no total usage type is ingested, Langfuse sums up all usage type units to a total.

<img width="257" height="467" alt="Image" src="https://github.com/user-attachments/assets/59deac26-c4ae-4079-bc63-ba068b337bb1" />

Digging deeper inside Langfuse handling, it seems like it expects a yaml structure as follows:
```yaml
OpenAICompletionUsageSchema:
  docs: OpenAI Usage schema from (Chat-)Completion APIs
  properties:
    prompt_tokens: integer
    completion_tokens: integer
    total_tokens: integer
    prompt_tokens_details: optional<map<string, optional<integer>>>
    completion_tokens_details: optional<map<string, optional<integer>>>
```

**To Reproduce**
Run an OpenAI model in a pipeline and observe the trace.

**Suggestion on Fixing**
The easiest solution should be skip sanitizing if we found OpenAI-style usage pattern.

LastRemote:
@vblagoje No, I have to use a customized one due to company rerouting. But thanks for letting me know that `GoogleGenAIChatGenerator` already converts the usage format. I might need to do the same.

PR #2594, opened 2025-12-09
fix: Correct token usage accounting, fix trace polling in tests

vblagoje:
## Why:
Fixes token usage accounting in Langfuse to properly display exact input/output token values by standardizing usage data format. 

- fixes https://github.com/deepset-ai/haystack-core-integrations/issues/2587

## What:
- Simplified `_sanitize_usage_data` to only convert provider-specific keys (`prompt_tokens`, `completion_tokens`) to Langfuse standard keys (`input_tokens`, `output_tokens`, `total_tokens`) instead of flattening nested structures
- Increased trace polling timeout from 10s to 30s in integration tests to allow Langfuse API sufficient time to process traces
- ~Updated langfuse dependency to 3.6.2 (latest version supporting Python 3.9)~
- ~Relaxed test assertions to check for key presence and reasonable values instead of exact token counts~

## How can it be used:
Usage data is now automatically standardized when sent to Langfuse:

```python
# OpenAI format (prompt_tokens, completion_tokens) → converted to input_tokens, output_tokens
# Cohere format (input_tokens already present) → preserved as-is
# All formats → total_tokens preserved
```

Token values are now accurately displayed in the Langfuse dashboard with proper input/output token accounting.

## How did you test it:
- Updated unit tests to verify usage standardization logic
- Relaxed brittle assertions to check for key presence and values > 0
- Verified integration tests pass with increased polling timeout
- Confirmed langfuse 3.6.2 compatibility

## Notes for the reviewer:
The sanitization change removes nested dictionary flattening - it now only handles key conversion. This ensures exact token values are preserved and displayed correctly in Langfuse. The 30s polling timeout accounts for Langfuse API processing delays in test environments.

### DIFF
--- a/integrations/langfuse/pyproject.toml
+++ b/integrations/langfuse/pyproject.toml
@@ -22,7 +22,7 @@ classifiers = [
   "Programming Language :: Python :: Implementation :: CPython",
   "Programming Language :: Python :: Implementation :: PyPy",
 ]
-dependencies = ["haystack-ai>=2.17.1", "langfuse>=3.6.2, <4.0.0"]
+dependencies = ["haystack-ai>=2.17.1", "langfuse>=3.3.1, <4.0.0"]
 
 [project.urls]
 Documentation = "https://github.com/deepset-ai/haystack-core-integrations/tree/main/integrations/langfuse#readme"


vblagoje:
cc @LastRemote @Hansehart as well please check for regressions

integrations/langfuse/pyproject.toml
```diff
@@ -22,7 +22,7 @@ classifiers = [
   "Programming Language :: Python :: Implementation :: CPython",
   "Programming Language :: Python :: Implementation :: PyPy",
 ]
-dependencies = ["haystack-ai>=2.17.1", "langfuse>=3.3.1, <4.0.0"]
+dependencies = ["haystack-ai>=2.17.1", "langfuse>=3.6.2, <4.0.0"]
```
anakin87: Do we actually need to pin a newer version? Or the fixes in this PR are expected to work in any case?

integrations/langfuse/tests/test_tracer.py
```diff
@@ -204,20 +204,16 @@ def test_anthropic_usage_flattens_and_filters(self):
...
-            "completion_tokens": 449,
-        }
+        assert "input_tokens" in result and result["input_tokens"] > 0
+        assert "output_tokens" in result and result["output_tokens"] > 0
 
     def test_openai_usage_preserved(self):
         """Test OpenAI/Cohere flat dict with only numeric values works unchanged"""
         usage = {"prompt_tokens": 29, "completion_tokens": 267, "total_tokens": 296}
         result = _sanitize_usage_data(usage)
-        assert result == {"prompt_tokens": 29, "completion_tokens": 267, "total_tokens": 296}
```
anakin87, line 220: I'd prefer to check exact numbers since `_sanitize_usage_data` is deterministic. Am I missing some details?

integrations/langfuse/pyproject.toml
vblagoje: Latest version is 3.10.5, 3.6.2 and up is from October and the old one is from August. I'll find a lib version that works well and is most relaxed.

integrations/langfuse/tests/test_tracer.py
vblagoje: It is but I'm thinking some of these generations are a moving target. As long as they are positive values greater than say some low number - it is likely that they will be immune to model drift and all other sorts of things that make tests brittle...

anakin87: Correct me if I am wrong, but this and others are unit tests that only focus on `_sanitize_usage_data`. How could the values change over time (unless we change the function)?

vblagoje: Ah sorry, I was thinking of something else. You are right!

anakin87: Thanks for fixing this

PR #2613, opened 2025-12-16
feat: expose refresh parameter in ElasticsearchDocumentStore and OpenSearchDocumentStore

GunaPalanivel:
## Related Issues
Closes #2065

## Description
This PR exposes the `refresh` parameter to relevant methods in `ElasticsearchDocumentStore` and `OpenSearchDocumentStore`, allowing users to control when index changes become visible to search operations.

## Motivation
Both Elasticsearch and OpenSearch are "near real-time" search engines where changes are not immediately visible after write/update/delete operations. Previously, users had to use `time.sleep()` workarounds to ensure consistency. This change exposes the underlying `refresh` parameter, giving users explicit control over this behavior.

## Changes
New `RefreshType` parameter added to the following methods in both document stores:

| Method | Default Value |
|--------|---------------|

**Parameter values:**

## Test Updates

## Additional Fixes
- Fixed `delete_all_documents_async` in OpenSearch to set `wait_for_completion=True` (matching Elasticsearch behavior), ensuring the async delete operation completes before returning
- Added file-level `# ruff: noqa: FBT001, FBT002` comment to OpenSearch to match Elasticsearch's linting configuration
- Removed unused `asyncio` import from Elasticsearch tests

## How was this tested?

## Checklist
- [x] I added tests that prove my fix is effective or that my feature works
- [x] I added the docstrings to document new functions and parameters

### DIFF
--- a/integrations/elasticsearch/src/haystack_integrations/document_stores/elasticsearch/document_store.py
+++ b/integrations/elasticsearch/src/haystack_integrations/document_stores/elasticsearch/document_store.py
@@ -25,6 +25,12 @@ from .filters import _normalize_filters
 
 logger = logging.getLogger(__name__)
 
+# Type alias for the refresh parameter
+# - True: Refresh immediately (force refresh)
+# - False: Don't refresh (best for bulk performance)
+# - "wait_for": Wait for next refresh cycle to make docs visible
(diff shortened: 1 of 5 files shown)


anakin87:
Ok, thank you!

Issue #2605, opened 2025-12-11
Expose `refresh` param to relevant methods in `ElasticsearchDocumentStore`


mpangrazzi:
**Is your feature request related to a problem? Please describe.**
In some use cases (also tests) it may be needed to have the index updated (i.e. refreshed) immediately after an update operation.  

**Describe the solution you'd like**
Exposing `refresh` as a param with default value can ensure the read-your-writes consistency.

**NOTE**: this feature will also allow us to avoid usage of `time.sleep` in tests (no need to wait for the index to be refreshed if we use the `refresh` param)

sjrl:
Hey @GauravBansal29 thanks for offering to help! You are welcome to work on this. We only assign issues internally to our team, so no need to wait for an assignment.

PR #2622, opened 2025-12-18
feat: expose refresh parameter in ElasticsearchDocumentStore

GunaPalanivel:
## Related Issues
Part of #2605

## Description
This PR exposes the `refresh` parameter to relevant methods in `ElasticsearchDocumentStore`, allowing users to control when index changes become visible to search operations.

### Motivation
Elasticsearch is a "near real-time" search engine where changes are not immediately visible after write/update/delete operations. Previously, users had to use `time.sleep()` workarounds to ensure consistency. This change exposes the underlying `refresh` parameter, giving users explicit control over this behavior.

### Changes
**New `RefreshType` parameter** added to the following methods:

| Method | Default Value |
|--------|---------------|

**Parameter values:**

### Test Updates

## How was this tested?

### DIFF
--- a/integrations/amazon_bedrock/CHANGELOG.md
+++ b/integrations/amazon_bedrock/CHANGELOG.md
@@ -1,5 +1,12 @@
 # Changelog
 
+## [integrations/amazon_bedrock-v5.3.1] - 2025-12-19
+
+### 🐛 Bug Fixes
+
+- Relax model name validation for Bedrock Embedders (#2625)
+
+
 ## [integrations/amazon_bedrock-v5.3.0] - 2025-12-17
 
 ### 🚀 Features
--- a/integrations/amazon_bedrock/src/haystack_integrations/components/embedders/amazon_bedrock/document_embedder.py
+++ b/integrations/amazon_bedrock/src/haystack_integrations/components/embedders/amazon_bedrock/document_embedder.py
@@ -1,5 +1,5 @@
 import json
-from typing import Any, Literal, Optional
+from typing import Any, Optional
 
 from botocore.config import Config
 from botocore.exceptions import ClientError
@@ -15,15 +15,6 @@ from haystack_integrations.common.amazon_bedrock.errors import (
 from haystack_integrations.common.amazon_bedrock.utils import get_aws_session
 
 logger = logging.getLogger(__name__)
-
-SUPPORTED_EMBEDDING_MODELS = [
-    "amazon.titan-embed-text-v1",
-    "amazon.titan-embed-text-v2:0",
-    "amazon.titan-embed-image-v1",
-    "cohere.embed-english-v3",
-    "cohere.embed-multilingual-v3",
-    "cohere.embed-v4:0",
-]
 
 
 @component
@@ -58,14 +49,7 @@ class AmazonBedrockDocumentEmbedder:
 
     def __init__(
         self,
-        model: Literal[
-            "amazon.titan-embed-text-v1",
-            "amazon.titan-embed-text-v2:0",
-            "amazon.titan-embed-image-v1",
-            "cohere.embed-english-v3",
-            "cohere.embed-multilingual-v3",
-            "cohere.embed-v4:0",
-        ],
+        model: str,
         aws_access_key_id: Optional[Secret] = Secret.from_env_var("AWS_ACCESS_KEY_ID", strict=False),  # noqa: B008
         aws_secret_access_key: Optional[Secret] = Secret.from_env_var(  # noqa: B008
             "AWS_SECRET_ACCESS_KEY", strict=False
@@ -90,8 +74,13 @@ class AmazonBedrockDocumentEmbedder:
         constructor. Aside from model, three required parameters are `aws_access_key_id`, `aws_secret_access_key`,
          and `aws_region_name`.
 
-        :param model: The embedding model to use. The model has to be specified in the format outlined in the Amazon
-            Bedrock [documentation](https://docs.aws.amazon.com/bedrock/latest/userguide/model-ids.html).
+        :param model: The embedding model to use.
+            Amazon Titan and Cohere embedding models are supported, for example:
+            "amazon.titan-embed-text-v1", "amazon.titan-embed-text-v2:0", "amazon.titan-embed-image-v1",
+            "cohere.embed-english-v3", "cohere.embed-multilingual-v3", "cohere.embed-v4:0".
+            To find all supported models, refer to the Amazon Bedrock
+            [documentation](https://docs.aws.amazon.com/bedrock/latest/userguide/models-supported.html) and
+            filter for "embedding", then select models from the Amazon Titan and Cohere series.
         :param aws_access_key_id: AWS access key ID.
         :param aws_secret_access_key: AWS secret access key.
         :param aws_session_token: AWS session token.
@@ -109,11 +98,8 @@ class AmazonBedrockDocumentEmbedder:
         :raises ValueError: If the model is not supported.
         :raises AmazonBedrockConfigurationError: If the AWS environment is not configured correctly.
         """
-
-        if not model or model not in SUPPORTED_EMBEDDING_MODELS:
-            msg = "Please provide a valid model from the list of supported models: " + ", ".join(
-                SUPPORTED_EMBEDDING_MODELS
-            )
+        if "titan" not in model and "cohere" not in model:
+            msg = f"Model {model} is not supported. Only Amazon Titan and Cohere embedding models are supported."
             raise ValueError(msg)
 
         self.model = model
@@ -254,7 +240,7 @@ class AmazonBedrockDocumentEmbedder:
         elif "titan" in self.model:
             documents_with_embeddings = self._embed_titan(documents=documents)
         else:
-            msg = f"Model {self.model} is not supported. Supported models are: {', '.join(SUPPORTED_EMBEDDING_MODELS)}."
+            msg = f"Model {self.model} is not supported. Only Amazon Titan and Cohere embedding models are supported."
             raise ValueError(msg)
 
         return {"documents": documents_with_embeddings}
--- a/integrations/amazon_bedrock/src/haystack_integrations/components/embedders/amazon_bedrock/document_image_embedder.py
+++ b/integrations/amazon_bedrock/src/haystack_integrations/components/embedders/amazon_bedrock/document_image_embedder.py
@@ -4,7 +4,7 @@
 
 import json
 from dataclasses import replace
-from typing import Any, Literal, Optional
+from typing import Any, Optional
 
 from botocore.config import Config
 from botocore.exceptions import ClientError
@@ -26,13 +26,6 @@ from haystack_integrations.common.amazon_bedrock.errors import (
 from haystack_integrations.common.amazon_bedrock.utils import get_aws_session
 
 logger = logging.getLogger(__name__)
-
-SUPPORTED_EMBEDDING_MODELS = [
-    "amazon.titan-embed-image-v1",
-    "cohere.embed-english-v3",
-    "cohere.embed-multilingual-v3",
-    "cohere.embed-v4:0",
-]
 
 
 @component
@@ -74,12 +67,7 @@ class AmazonBedrockDocumentImageEmbedder:
     def __init__(
         self,
         *,
-        model: Literal[
-            "amazon.titan-embed-image-v1",
-            "cohere.embed-english-v3",
-            "cohere.embed-multilingual-v3",
-            "cohere.embed-v4:0",
-        ],
+        model: str,
         aws_access_key_id: Optional[Secret] = Secret.from_env_var("AWS_ACCESS_KEY_ID", strict=False),  # noqa: B008
         aws_secret_access_key: Optional[Secret] = Secret.from_env_var(  # noqa: B008
             "AWS_SECRET_ACCESS_KEY", strict=False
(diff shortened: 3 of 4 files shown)


integrations/elasticsearch/src/haystack_integrations/document_stores/elasticsearch/document_store.py
```diff
@@ -410,12 +416,21 @@ def _deserialize_document(hit: dict[str, Any]) -> Document:
...
+    ) -> int:
         """
         Writes `Document`s to Elasticsearch.
 
         :param documents: List of Documents to write to the document store.
         :param policy: DuplicatePolicy to apply when a document with the same ID already exists in the document store.
+        :param refresh: Controls when changes are made visible to search operations.
+            - `True`: Force refresh immediately after the operation.
+            - `False`: Do not refresh (better performance for bulk operations).
+            - `"wait_for"`: Wait for the next refresh cycle (default, ensures read-your-writes consistency).
```
anakin87, line 428: Let's also link https://www.elastic.co/docs/reference/elasticsearch/rest-apis/refresh-parameter so users can better understand this parameter

integrations/elasticsearch/src/haystack_integrations/document_stores/elasticsearch/document_store.py
```diff
@@ -410,12 +416,21 @@ def _deserialize_document(hit: dict[str, Any]) -> Document:
 
         return Document.from_dict(data)
 
-    def write_documents(self, documents: list[Document], policy: DuplicatePolicy = DuplicatePolicy.NONE) -> int:
+    def write_documents(
+        self,
+        documents: list[Document],
+        policy: DuplicatePolicy = DuplicatePolicy.NONE,
+        refresh: RefreshType = "wait_for",
```
anakin87: I think that `refresh=False` would be a better default, but I'm also interested in @mpangrazzi's opinion

integrations/elasticsearch/src/haystack_integrations/document_stores/elasticsearch/document_store.py
```diff
@@ -630,20 +664,24 @@ def delete_all_documents(self, recreate_index: bool = False) -> None:
             self._client.indices.create(index=self._index, mappings=mappings)  # type: ignore
 
         else:
-            result = self._client.delete_by_query(**self._prepare_delete_all_request(is_async=False))  # type: ignore
+            result = self._client.delete_by_query(**self._prepare_delete_all_request(is_async=False, refresh=refresh))  # type: ignore
```
GauravBansal29, line 667: Hi from what I checked , delete by query doesn't support wait_for Refresh Type ; this should be handled differently

integrations/elasticsearch/src/haystack_integrations/document_stores/elasticsearch/document_store.py
GunaPalanivel: I've added the documentation link to all the refresh parameter docstrings.

integrations/elasticsearch/src/haystack_integrations/document_stores/elasticsearch/document_store.py
GunaPalanivel: I'll wait for @mpangrazzi's opinion on this, as you said. Once we have a consensus, I'll update the default values accordingly. 

Happy to make any other changes needed!

integrations/elasticsearch/src/haystack_integrations/document_stores/elasticsearch/document_store.py
GunaPalanivel: You're absolutely right. According to the [Elasticsearch documentation delete by query](https://www.elastic.co/guide/en/elasticsearch/reference/current/docs-delete-by-query.html):

>refresh BOOLEAN

>
>If true, Elasticsearch refreshes all shards involved in the delete by query after the request completes. This is different than the delete API's `refresh` parameter, which causes just the shard that received the delete request to be refreshed. **Unlike the delete API, it does not support `wait_for.`**

This means `delete_by_query` and `update_by_query` APIs only accept `True` or `False`, not `"wait_for"`.

**Affected methods in our implementation:**
- `delete_all_documents` / `delete_all_documents_async` (uses `delete_by_query`)
- `delete_by_filter` / `delete_by_filter_async` (uses `delete_by_query`)
- `update_by_filter` / `update_by_filter_async` (uses `update_by_query`)

**Proposed approach:**

I suggest converting `"wait_for"` → `True` internally for these specific methods. This way:
- Maintains a consistent API surface for users across all methods
- Handles the limitation transparently without breaking user code
- Semantically similar behavior (both ensure changes are visible after the operation)
- No need to introduce different type signatures for different methods

```python
# For methods using delete_by_query/update_by_query
effective_refresh = True if refresh == "wait_for" else refresh
```

What do you think @anakin87 @mpangrazzi? Should I implement this approach?

mpangrazzi: Even if generally I wouldn't recommend a `True` default value for refresh, in this case we could do that to maintain the backward compatibility. Also, since this is an infrequent operation, it makes sense. 

IMHO, since `wait_for` is not supported here as correctly pointed out, I would convert the `refresh` type to `bool` for consistency, so skipping the runtime validation.

integrations/elasticsearch/src/haystack_integrations/document_stores/elasticsearch/document_store.py
anakin87: Thinking better about this point, it was not a good idea

integrations/elasticsearch/src/haystack_integrations/document_stores/elasticsearch/document_store.py
```diff
@@ -25,6 +25,12 @@
 
 logger = logging.getLogger(__name__)
 
+# Type alias for the refresh parameter
+# - True: Refresh immediately (force refresh)
+# - False: Don't refresh (best for bulk performance)
+# - "wait_for": Wait for next refresh cycle to make docs visible
+RefreshType = Union[bool, Literal["wait_for"]]
```
anakin87:
```suggestion
```

Since we found out that in some methods we cannot use this type alias, let's just remove it (and list the possible values in the method signature)

integrations/elasticsearch/src/haystack_integrations/document_stores/elasticsearch/document_store.py
```diff
@@ -591,20 +624,23 @@ async def delete_documents_async(self, document_ids: list[str]) -> None:
...
         """
         Deletes all documents in the document store.
 
         A fast way to clear all documents from the document store while preserving any index settings and mappings.
 
         :param recreate_index: If True, the index will be deleted and recreated with the original mappings and
             settings. If False, all documents will be deleted using the `delete_by_query` API.
+        :param refresh: If True, Elasticsearch refreshes all shards involved in the delete by query after the request
+            completes. If False, no refresh is performed. For more details, see the
+            [Elasticsearch refresh documentation](https://www.elastic.co/docs/reference/elasticsearch/rest-apis/refresh-parameter).
```
anakin87: let's link the right docs here: https://www.elastic.co/docs/api/doc/elasticsearch/operation/operation-delete-by-query#operation-delete-by-query-refresh

integrations/elasticsearch/src/haystack_integrations/document_stores/elasticsearch/document_store.py
```diff
@@ -630,20 +666,23 @@ def delete_all_documents(self, recreate_index: bool = False) -> None:
...
+    async def delete_all_documents_async(self, recreate_index: bool = False, refresh: bool = True) -> None:
         """
         Asynchronously deletes all documents in the document store.
 
         A fast way to clear all documents from the document store while preserving any index settings and mappings.
         :param recreate_index: If True, the index will be deleted and recreated with the original mappings and
             settings. If False, all documents will be deleted using the `delete_by_query` API.
+        :param refresh: If True, Elasticsearch refreshes all shards involved in the delete by query after the request
+            completes. If False, no refresh is performed. For more details, see the
+            [Elasticsearch refresh documentation](https://www.elastic.co/docs/reference/elasticsearch/rest-apis/refresh-parameter).
```
anakin87: https://www.elastic.co/docs/api/doc/elasticsearch/operation/operation-delete-by-query#operation-delete-by-query-refresh

integrations/elasticsearch/src/haystack_integrations/document_stores/elasticsearch/document_store.py
```diff
@@ -683,20 +722,23 @@ async def delete_all_documents_async(self, recreate_index: bool = False) -> None
...
-    def delete_by_filter(self, filters: dict[str, Any]) -> int:
+    def delete_by_filter(self, filters: dict[str, Any], refresh: bool = True) -> int:
         """
         Deletes all documents that match the provided filters.
 
         :param filters: The filters to apply to select documents for deletion.
             For filter syntax, see [Haystack metadata filtering](https://docs.haystack.deepset.ai/docs/metadata-filtering)
+        :param refresh: If True, Elasticsearch refreshes all shards involved in the delete by query after the request
+            completes. If False, no refresh is performed. For more details, see the
+            [Elasticsearch refresh documentation](https://www.elastic.co/docs/reference/elasticsearch/rest-apis/refresh-parameter).
```
anakin87: https://www.elastic.co/docs/api/doc/elasticsearch/operation/operation-delete-by-query#operation-delete-by-query-refresh

integrations/elasticsearch/src/haystack_integrations/document_stores/elasticsearch/document_store.py
```diff
@@ -708,20 +750,23 @@ def delete_by_filter(self, filters: dict[str, Any]) -> int:
...
-    async def delete_by_filter_async(self, filters: dict[str, Any]) -> int:
+    async def delete_by_filter_async(self, filters: dict[str, Any], refresh: bool = True) -> int:
         """
         Asynchronously deletes all documents that match the provided filters.
 
         :param filters: The filters to apply to select documents for deletion.
             For filter syntax, see [Haystack metadata filtering](https://docs.haystack.deepset.ai/docs/metadata-filtering)
+        :param refresh: If True, Elasticsearch refreshes all shards involved in the delete by query after the request
+            completes. If False, no refresh is performed. For more details, see the
+            [Elasticsearch refresh documentation](https://www.elastic.co/docs/reference/elasticsearch/rest-apis/refresh-parameter).
```
anakin87, line 759: https://www.elastic.co/docs/api/doc/elasticsearch/operation/operation-delete-by-query#operation-delete-by-query-refresh

integrations/elasticsearch/src/haystack_integrations/document_stores/elasticsearch/document_store.py
```diff
@@ -733,13 +778,16 @@ async def delete_by_filter_async(self, filters: dict[str, Any]) -> int:
...
+    def update_by_filter(self, filters: dict[str, Any], meta: dict[str, Any], refresh: bool = True) -> int:
         """
         Updates the metadata of all documents that match the provided filters.
 
         :param filters: The filters to apply to select documents for updating.
             For filter syntax, see [Haystack metadata filtering](https://docs.haystack.deepset.ai/docs/metadata-filtering)
         :param meta: The metadata fields to update.
+        :param refresh: If True, Elasticsearch refreshes all shards involved in the update by query after the request
+            completes. If False, no refresh is performed. For more details, see the
+            [Elasticsearch refresh documentation](https://www.elastic.co/docs/reference/elasticsearch/rest-apis/refresh-parameter).
```
anakin87: https://www.elastic.co/docs/api/doc/elasticsearch/operation/operation-update-by-query#operation-update-by-query-refresh

integrations/elasticsearch/src/haystack_integrations/document_stores/elasticsearch/document_store.py
```diff
@@ -765,13 +813,18 @@ def update_by_filter(self, filters: dict[str, Any], meta: dict[str, Any]) -> int
             msg = f"Failed to update documents by filter in Elasticsearch: {e!s}"
             raise DocumentStoreError(msg) from e
 
-    async def update_by_filter_async(self, filters: dict[str, Any], meta: dict[str, Any]) -> int:
+    async def update_by_filter_async(
+        self, filters: dict[str, Any], meta: dict[str, Any], refresh: bool = True
+    ) -> int:
```
anakin87: https://www.elastic.co/docs/api/doc/elasticsearch/operation/operation-update-by-query#operation-update-by-query-refresh

integrations/elasticsearch/src/haystack_integrations/document_stores/elasticsearch/document_store.py
```diff
@@ -25,6 +25,11 @@
 
 logger = logging.getLogger(__name__)
 
+# Type alias for the refresh parameter
+# - True: Refresh immediately (force refresh)
+# - False: Don't refresh (best for bulk performance)
+# - "wait_for": Wait for next refresh cycle to make docs visible
+
```
anakin87:
```suggestion
```

integrations/elasticsearch/src/haystack_integrations/document_stores/elasticsearch/document_store.py
```diff
@@ -556,33 +575,43 @@ async def write_documents_async(
             msg = f"Failed to write documents to Elasticsearch: {e!s}"
             raise DocumentStoreError(msg) from e
 
-    def delete_documents(self, document_ids: list[str]) -> None:
+    def delete_documents(self, document_ids: list[str], refresh: bool = True) -> None:
```
anakin87: adjust the type hint and set `wait_for` as default value

integrations/elasticsearch/src/haystack_integrations/document_stores/elasticsearch/document_store.py
```diff
@@ -556,33 +575,43 @@ async def write_documents_async(
...
         return {
             "index": self._index,
             "body": {"query": {"match_all": {}}},  # Delete all documents
             "wait_for_completion": False if is_async else True,  # block until done (set False for async)
-            "refresh": True,  # Ensure changes are visible immediately
+            "refresh": refresh,
         }
 
-    async def delete_documents_async(self, document_ids: list[str]) -> None:
+    async def delete_documents_async(self, document_ids: list[str], refresh: bool = True) -> None:
```
anakin87: adjust the type hint and set `wait_for` as default value (breaking but makes sense)

GunaPalanivel:
@anakin87 I see what you mean about the breaking change now.

Looking at the current code, I noticed that:
- `delete_by_query` and `update_by_query` don't currently have any `refresh` parameter at all
- These methods only accept `bool` (not `"wait_for"`), as @GauravBansal29 pointed out

So I'm thinking:
1. For `delete_by_filter` / `update_by_filter` → use `refresh=True` as default to match the current behavior (no breaking change since these are already blocking operations)
2. For `write_documents` / `delete_documents` → use `refresh="wait_for"` as default (slight breaking change, but better consistency)

But I'm not sure if this is the best approach. @mpangrazzi, would this align with Haystack's philosophy on consistency vs. backward compatibility?

Happy to adjust based on your feedback!

GunaPalanivel:
@mpangrazzi The table makes it crystal clear - backward compatibility first, consistency second. 

I'll wait for @anakin87's thoughts and then update the PR with refresh=False for update_by_filter / delete_by_filter methods to keep everything non-breaking.

Thanks for the thorough analysis!

PR #2623, opened 2025-12-18
feat!: expose refresh parameter in `OpenSearchDocumentStore`

GunaPalanivel:
## Related Issues
Part of #2605

## Description
This PR exposes the `refresh` parameter to relevant methods in `OpenSearchDocumentStore`, allowing users to control when index changes become visible to search operations.

### Motivation
OpenSearch is a "near real-time" search engine where changes are not immediately visible after write/update/delete operations. Previously, users had to use `time.sleep()` workarounds to ensure consistency. This change exposes the underlying `refresh` parameter, giving users explicit control over this behavior.

### Changes
**New `RefreshType` parameter** added to the following methods:

| Method | Default Value |
|--------|---------------|

**Parameter values:**

### Test Updates

### Additional Fixes
- Fixed `delete_all_documents_async` to set `wait_for_completion=True`, ensuring the async delete operation completes before returning

## How was this tested?

### DIFF
--- a/.github/workflows/pinecone.yml
+++ b/.github/workflows/pinecone.yml
@@ -36,8 +36,8 @@ jobs:
         python-version: ["3.10", "3.13"]
         # the INDEX_NAME is used in test_serverless_index_creation_from_scratch
         include:
-          - python-version: "3.9"
-            INDEX_NAME: "index-39"
+          - python-version: "3.10"
+            INDEX_NAME: "index-310"
           - python-version: "3.13"
             INDEX_NAME: "index-313"
 
--- a/integrations/amazon_bedrock/src/haystack_integrations/components/generators/amazon_bedrock/chat/chat_generator.py
+++ b/integrations/amazon_bedrock/src/haystack_integrations/components/generators/amazon_bedrock/chat/chat_generator.py
@@ -140,7 +140,7 @@ class AmazonBedrockChatGenerator:
     automatically from the environment or the AWS configuration file.
     If the AWS environment is not configured, set `aws_access_key_id`, `aws_secret_access_key`,
       and `aws_region_name` as environment variables or pass them as
-     [Secret](https://docs.haystack.deepset.ai/v2.0/docs/secret-management) arguments. Make sure the region you set
+     [Secret](https://docs.haystack.deepset.ai/docs/secret-management) arguments. Make sure the region you set
     supports Amazon Bedrock.
     """
 
--- a/integrations/amazon_bedrock/src/haystack_integrations/components/generators/amazon_bedrock/generator.py
+++ b/integrations/amazon_bedrock/src/haystack_integrations/components/generators/amazon_bedrock/generator.py
@@ -58,7 +58,7 @@ class AmazonBedrockGenerator:
     automatically from the environment or the AWS configuration file.
     If the AWS environment is not configured, set `aws_access_key_id`, `aws_secret_access_key`,
     `aws_session_token`, and `aws_region_name` as environment variables or pass them as
-     [Secret](https://docs.haystack.deepset.ai/v2.0/docs/secret-management) arguments. Make sure the region you set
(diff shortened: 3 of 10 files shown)


anakin87:
Thanks for opening a separate PR. Let's put this on hold and first agree on #2622

anakin87:
@GunaPalanivel when you have time, feel free to adapt this PR based on #2622 🙂

integrations/opensearch/src/haystack_integrations/document_stores/opensearch/document_store.py
```diff
@@ -624,18 +679,25 @@ async def delete_all_documents_async(self, recreate_index: bool = False) -> None
                 await self._async_client.indices.delete(index=self._index)
                 await self._async_client.indices.create(index=self._index, body=body)
             else:
-                await self._async_client.delete_by_query(**self._prepare_delete_all_request(is_async=True))
+                # use delete_by_query for more efficient deletion without index recreation
+                # For async, we need to wait for completion to ensure documents are deleted
+                delete_request = self._prepare_delete_all_request(is_async=True, refresh=refresh)
+                delete_request["wait_for_completion"] = True  # Override to wait for completion in async
```
anakin87: modifying the `wait_for_completion` value here is unexpected. If we want to change it, let's do it in `_prepare_delete_all_request`.

I'm not an expert here but maybe we can set `wait_for_completion` to True if `async=True and wait_for=True`?
Also curious to hear @davidsbatista's opinion

GunaPalanivel: @anakin87,

I've fixed the `wait_for_completion` logic! 

I found an issue with my initial implementation. The problem was:
- When `wait_for_completion=False`, `delete_by_query` returns a task object instead of a result dict
- This caused a `KeyError: 'deleted'` when trying to access the deleted count

**The correct fix:**
- **Sync calls** (`is_async=False`): `wait_for_completion=True` ALWAYS → ensures we get the result dict with "deleted" count
- **Async calls** (`is_async=True`): `wait_for_completion=True` only when `refresh=True` → allows async operation to complete if refresh is requested

**The expression:** `wait_for_completion = (not is_async) or refresh`

I'm also curious to hear @davidsbatista's opinion on this approach — specifically whether setting `wait_for_completion=True` when `async=True and refresh=True` makes sense semantically.

davidsbatista: I had a look at the `delete_all_documents_async()` in both `ElasticSearch` and in this PR:

- `ElasticSearch`: always `True` ([overridden in the async method](https://github.com/deepset-ai/haystack-core-integrations/blob/main/integrations/elasticsearch/src/haystack_integrations/document_stores/elasticsearch/document_store.py#L711)) - because it's waiting for the result, how many deleted docs, for logging.
- `OpenSearch`: `wait_for_completion` depends on refresh (True if refresh=True, False if refresh=False)

I suggest aligning both. We have the following options:

1) replicate `ElasticSearch` behaviour, set it to `True`, and also log the result

2) both the `wait_for_completion` on `ElastichSearch` and `OpenSearch` dependent on the `is_async` parameter and remove the logging, such that `wait_for_completion` is only True for sync calls, e.g.:

```python
def _prepare_delete_all_request(self, *, is_async: bool, refresh: bool) -> dict[str, Any]:
    return {
        "index": self._index,
        "body": {"query": {"match_all": {}}},  
        "wait_for_completion": not is_async,  # `True` for sync calls
        "refresh": refresh,
    }
```

GunaPalanivel: Got it

anakin87: Thanks David!
@GunaPalanivel went for option 1, which seems OK to me.

PR #2631, opened 2025-12-29
feat: adding `delete_by_filter` and `update_by_filter`

davidsbatista:
### Related Issues

- fixes https://github.com/deepset-ai/haystack-core-integrations/issues/2329

### Proposed Changes:

- `update_by_filter`: updates documents matching filters using dot notation for nested meta fields (e.g., meta.status). New fields are created if missing via Astra's $set operator.
- `delete_by_filter`: deletes documents matching the provided filters.
- `update_many` to AstraClient: Supports bulk updates with proper return value handling.
- added copyright headers to source files.
- removed unused variable and outdated docstring parameter.

### How did you test it?

- added test_update_by_filter and test_delete_by_filter integration tests.

### Checklist

### DIFF
--- a/integrations/astra/src/haystack_integrations/document_stores/astra/document_store.py
+++ b/integrations/astra/src/haystack_integrations/document_stores/astra/document_store.py
@@ -482,7 +482,7 @@ class AstraDocumentStore:
         # use dot notation to update nested fields in the meta-object - ensures fields are created if they don't exist
         update_fields = {f"meta.{key}": value for key, value in meta.items()}
         update_operation = {"$set": update_fields}
-        update_count = self.index.update(filter_condition=converted_filters, update=update_operation)  # type: ignore
+        update_count = self.index.update(filters=converted_filters, update=update_operation)  # type: ignore
 
         logger.info(f"{update_count} documents updated by filter")
 


integrations/astra/src/haystack_integrations/document_stores/astra/astra_client.py
```diff
@@ -356,3 +360,22 @@ def count_documents(self, upper_bound: int = 10000) -> int:
         :returns: the number of documents in the index
         """
         return self._astra_db_collection.count_documents({}, upper_bound=upper_bound)
+
+    def update_many(
```
sjrl: Apologies if this is a silly question but why call this `update_many` vs `update` or `update_many_documents`?

The naming convention for the `delete` functions seems like the bulk delete is just called `delete`

integrations/astra/src/haystack_integrations/document_stores/astra/astra_client.py
```diff
@@ -356,3 +360,22 @@ def count_documents(self, upper_bound: int = 10000) -> int:
...
+        self,
+        *,
+        filter_condition: dict[str, Union[str, float, int, bool, list, dict]],
```
sjrl: Also it seems like we are usually calling this `filters` in some of our other methods like in `delete`. Do you think we should rename this to `filters`?

integrations/astra/src/haystack_integrations/document_stores/astra/astra_client.py
davidsbatista: was just following the underneath API name - `self._astra_db_collection.update_many` - you suggestion makes sense :

integrations/astra/src/haystack_integrations/document_stores/astra/astra_client.py
davidsbatista: I had it as `filter` but it's python reserved keyword - `filters` seems a good choice 👍🏽

integrations/astra/src/haystack_integrations/document_stores/astra/astra_client.py
sjrl: Okay no worries! I think I was mistaking these functions as also being part of the document store but I see this is a separate class that is meant to serve as a convenience wrapper to the underlying client. So I see why it's fine to be looser with the naming.

PR #2645, opened 2026-01-01
fix: OpenSearch async client initialization

Nitro963:
### Proposed Changes:

In async environment `_ensure_initialized` sync method is being called on the main thread, and as it may take time depending on network it will cause the entire event loop to block and the application will freeze until all retries finished if any.

### How did you test it?

Manual testing within FastAPI async server

### DIFF
--- a/integrations/meta_llama/tests/test_llama_chat_generator.py
+++ b/integrations/meta_llama/tests/test_llama_chat_generator.py
@@ -10,7 +10,7 @@ import pytz
 from haystack import Pipeline
 from haystack.components.generators.utils import print_streaming_chunk
 from haystack.components.tools import ToolInvoker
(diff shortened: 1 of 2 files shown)


davidsbatista:
Thanks for your proposal @Nitro963 

This makes sense, but I would rather do it differently, since this might introduce lots of overhead each time an operation is performed.

Ideally, creating a `async def _ensure_initialized_async(self)` check how it's done for the [Chroma document store.](https://github.com/deepset-ai/haystack-core-integrations/blob/main/integrations/chroma/src/haystack_integrations/document_stores/chroma/document_store.py#L142)

davidsbatista:
thanks @Nitro963 LGTM

davidsbatista (approved):
LGTM!

PR #2650, opened 2026-01-02
feat: adding `delete_by_filter()` and `update_by_filter()` to `QdrantDocumentStore`

### LINKED ISSUE #2334: add `update_by_filter()` and `delete_by_filter()` operations to `QdrantDocumentStore`
```python
def delete_by_filter(self, filters: Dict[str, Any]) -> int
```

```python
def update_by_filter(self, filters: Dict[str, Any], meta: Dict[str, Any]) -> int
```

davidsbatista:
### Related Issues

- fixes #2334 

### Proposed Changes:

- adding updated and delete by filter operations, both syn and async versions

### How did you test it?

- add new tests

### Checklist

### DIFF
--- a/integrations/qdrant/tests/test_document_store_async.py
+++ b/integrations/qdrant/tests/test_document_store_async.py
@@ -274,9 +274,7 @@ class TestQdrantDocumentStore:
         assert await document_store.count_documents_async() == 3
 
         # Delete documents with category="A"
-        await document_store.delete_by_filter_async(
-            filters={"field": "meta.category", "operator": "==", "value": "A"}
-        )
+        await document_store.delete_by_filter_async(filters={"field": "meta.category", "operator": "==", "value": "A"})
         assert await document_store.count_documents_async() == 1
 
         # Verify only category B remains
@@ -287,9 +285,7 @@ class TestQdrantDocumentStore:
         assert remaining_docs[0].meta["category"] == "B"
 
         # Delete remaining document by year
-        await document_store.delete_by_filter_async(
-            filters={"field": "meta.year", "operator": "==", "value": 2023}
-        )
+        await document_store.delete_by_filter_async(filters={"field": "meta.year", "operator": "==", "value": 2023})
         assert await document_store.count_documents_async() == 0
 
     @pytest.mark.asyncio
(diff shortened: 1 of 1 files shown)


integrations/qdrant/src/haystack_integrations/document_stores/qdrant/document_store.py
```diff
@@ -517,6 +517,261 @@ async def delete_documents_async(self, document_ids: list[str]) -> None:
...
+            qdrant_filter = convert_filters_to_qdrant(filters)
+
+            if qdrant_filter is None:
+                return 0
+
+            matching_docs = list(self.filter_documents(filters))
+            count = len(matching_docs)
+
+            if count == 0:
+                return 0
```
sjrl: This seems a little inefficient, to run a `filter_documents` before calling the real `self._client.delete` below. Why don't we remove this and just use the `self._client.delete` call? 

This would also help avoid a situation where the filters match a huge number of documents which could be a slow operation using `self.filter_documents`

integrations/qdrant/src/haystack_integrations/document_stores/qdrant/document_store.py
```diff
@@ -517,6 +517,261 @@ async def delete_documents_async(self, document_ids: list[str]) -> None:
...
+            if qdrant_filter is None:
+                return 0
+
+            matching_docs = []
+            async for doc in self._get_documents_generator_async(filters):
+                matching_docs.append(doc)
+            count = len(matching_docs)
+
+            if count == 0:
+                return 0
```
sjrl: Same comment as [here](https://github.com/deepset-ai/haystack-core-integrations/pull/2650/files#r2667603728).

Also if we really need to perform this `count == 0` check before calling `self._async_client.delete` then we should add a `break` into the for loop as soon as one matching doc is found/returned since all we care about is if it's 0 or not. So no need to loop over all matching docs right?

integrations/qdrant/src/haystack_integrations/document_stores/qdrant/document_store.py
sjrl: Ahh okay I also see that you use the `count` in the logger message and return it. I'm guessing this means Qdrant doesn't return a count when calling `self._client.delete`? 

If so, for performance reasons I'd prefer to remove the `count` and explain in the logger message that the count is not available and as a return probably just always return 0 and document why that's the case.

davidsbatista: removing the logging or performances reasons

integrations/qdrant/src/haystack_integrations/document_stores/qdrant/document_store.py
davidsbatista: I'm removing the count anyways, but you are right - no need to consume the whole generator.

PR #2651, opened 2026-01-04
feat: add `client_settings` to `ChromaDocumentStore`

### LINKED ISSUE #2570: Offer passing Settings to Chroma Client in __init__
https://github.com/deepset-ai/haystack-core-integrations/blob/ed73460d663256f270c8817501195100990d4066/integrations/chroma/src/haystack_integrations/document_stores/chroma/document_store.py#L113

The Chroma backend Client can be provided with Settings when initializing. (See https://docs.trychroma.com/reference/python/client#client)
If we want to change these using the haystack integration we have to illegally access the private _client like so:
```
db = ChromaDocumentStore()
db._client.get_settings().anonymized_telemetry = False
```

It would be preferable to offer a way to pass Client Settings to the `ChromaDocumentStore` constructor.

SaraCalla:
### Related Issues

- fixes #2570

### Proposed Changes:

Added `client_settings` parameter to `__init__` to allow users to configure the underlying Chroma client's Settings without accessing private attributes.

### How did you test it?

- Added unit tests for `client_settings` initialization
- Added integration test for async client with settings

### Notes for the reviewer

I also took the liberty of fixing the async test class marking. Previously, running `hatch run test:unit` would fail because tests required a Chroma server instance. The entire `TestDocumentStoreAsync` class is now marked as `@pytest.mark.integration`.

### Checklist


integrations/chroma/src/haystack_integrations/document_stores/chroma/document_store.py
```diff
@@ -67,6 +69,8 @@ def __init__(
         :param metadata: a dictionary of chromadb collection parameters passed directly to chromadb's client
             method `create_collection`. If it contains the key `"hnsw:space"`, the value will take precedence over the
             `distance_function` parameter above.
+        :param client_settings: a dictionary of Chroma Settings configuration options passed to
+            `chromadb.config.Settings`. These settings configure the underlying Chroma client behavior.
```
anakin87, line 73: - Let's add a link to the supported settings. I haven't found docs, so we could reference this file: https://github.com/chroma-core/chroma/blob/main/chromadb/config.py
- Let's also add that specifying these settings may interfere with the standard client initialization, and it's intended for advanced customization

integrations/chroma/src/haystack_integrations/document_stores/chroma/document_store.py
```diff
@@ -102,18 +107,28 @@ def _ensure_initialized(self):
                     "You cannot specify both options."
                 )
                 raise ValueError(error_message)
+
+            client_kwargs: dict[str, Any] = {}
+            if self._client_settings:
+                try:
+                    client_kwargs["settings"] = Settings(**self._client_settings)
+                except Exception as e:
```
anakin87: Let's catch the specific error we get - if possible

integrations/chroma/src/haystack_integrations/document_stores/chroma/document_store.py
```diff
@@ -102,18 +107,28 @@ def _ensure_initialized(self):
                     "You cannot specify both options."
                 )
                 raise ValueError(error_message)
+
+            client_kwargs: dict[str, Any] = {}
```
anakin87, line 115: nit: instead of a dictionary with a single key, we can define define a `client_settings` object with a `None` default value.

`client_settings_obj = None`
and later
`client_settings_obj = Settings(**self._client_settings)`

integrations/chroma/tests/test_document_store.py
```diff
@@ -83,9 +88,41 @@ def test_invalid_initialization_both_host_and_persist_path(self):
             store = ChromaDocumentStore(persist_path="./path/to/local/store", host="localhost")
             store._ensure_initialized()
 
+    def test_client_settings_applied(self):
+        """
+        Chroma's in-memory client uses a singleton pattern with an internal cache.
+        Once a client is created with certain settings, Chroma rejects creating another
+        with different settings in the same process. We clear the cache before and after
+        this test to avoid conflicts with other tests that use default settings.
```
anakin87, line 109: Ah, this is a good catch, but I'd prefer to extract a fixture to do the same. Something like this (not tested)
```python
@pytest.fixture
def clear_chroma_system_cache():
    SharedSystemClient.clear_system_cache()
    yield
    SharedSystemClient.clear_system_cache() 
```

anakin87 (commented):
Thank you again. I left some comments.

Have you investigated how `Settings` interfere with other client parameters?
```python
HttpClient(host: str = "localhost",
               port: int = 8000,
               settings: my_custom_settings)
```

integrations/chroma/src/haystack_integrations/document_stores/chroma/document_store.py
SaraCalla: Invalid settings raise `pydantic.v1.ValidationError` because Chroma uses pydantic's v1 compatibility layer internally and importing directly from `pydantic.v1` felt a bit awkward so I initially went with a generic `Exception`.
After your comment I dug a bit deeper and I found that `pydantic.v1.ValidationError` inherits from `ValueError` so catching `ValueError` gives us a nice middle ground. WDYT?

anakin87: Ok for me!

integrations/chroma/src/haystack_integrations/document_stores/chroma/document_store.py
SaraCalla: I tried this before but passing `settings=None` to Chroma's clients raises an error because it expects either a valid Settings object or the argument to be omitted entirely.

anakin87: Ok let's add a small comment explaining this aspect

integrations/chroma/src/haystack_integrations/document_stores/chroma/document_store.py
```diff
@@ -102,18 +110,28 @@ def _ensure_initialized(self):
...
+            if self._client_settings:
+                try:
+                    client_kwargs["settings"] = Settings(**self._client_settings)
+                except ValueError as e:
+                    msg = f"Invalid client_settings: {e}"
```
anakin87: I would report `self._client_settings` in the error message

PR #2653, opened 2026-01-05
feat: adding  count with filtering operations to `OpenSearchDocumentStore`

### LINKED ISSUE #2635: add the following operations to `OpenSearchDocumentStore`
`count_documents_by_filter(filter: dict) -> int` applies a filter and counts the documents that matched it.

`count_distinct_values_by_filter(filter: dict) -> dict[str, int]` applies a filter selecting documents and counts the unique values for each meta field of the matched documents

`get_fields_info() -> dict[str, dict]` returns the meta fields and the corresponding types

`get_field_min_max(metadata_field: str) -> dict[str, Any]` for a given metadata field, find its max and min value

`get_field_unique_values(field_name: str, search_term: str | None, from_: int, size: int) -> tuple[list[str], int]` - retrieves unique values for a field matching a search_term or all possible values if no search term is given

`query_sql(query: str)` - optional - issue an SQL query for the doc stores that support it

davidsbatista:
### Related Issues

- fixes #2635 

### Proposed Changes:

- `count_documents_by_filter()` - count documents matching filter criteria
- `count_distinct_values_by_filter()`- get distinct value counts for metadata fields with optional filtering
- `get_fields_info()` - retrieve field type information from index mapping
- `get_field_min_max()` - get min/max values for numeric metadata fields
- `get_field_unique_values()` -  get unique values for a field with pagination and content-based filtering
- `query_sql()` - execute SQL queries against OpenSearch with support for multiple response formats (JSON, CSV, JDBC, RAW)

### How did you test it?

- added integrations tests covering the new methods both or sync and async versions

### Notes for the reviewer

- added `httpx>=0.28.1` dependency
- the query_sql() method performs a raw http request (based on httpx) if the specified response format is not JSON

### Checklist

### DIFF
--- a/integrations/opensearch/src/haystack_integrations/document_stores/opensearch/document_store.py
+++ b/integrations/opensearch/src/haystack_integrations/document_stores/opensearch/document_store.py
@@ -23,8 +23,6 @@ logger = logging.getLogger(__name__)
 
 SPECIAL_FIELDS = {"content", "embedding", "id", "score", "sparse_embedding", "blob"}
 
-ResponseFormat = Literal["json", "jdbc", "csv", "raw"]
-
 Hosts = str | list[str | Mapping[str, str | int]]
 
 # document scores are essentially unbounded and will be scaled to values between 0 and 1 if scale_score is set to


integrations/opensearch/pyproject.toml
```diff
@@ -25,7 +25,9 @@ classifiers = [
 ]
 dependencies = [
   "haystack-ai>=2.14.0",
-  "opensearch-py[async]>=2.4.0,<3"]
+  "opensearch-py[async]>=2.4.0,<3",
```
sjrl: @davidsbatista an unrelated question. Do you happen to know why we are pinning to less than 3? I see that the latest release for `opensearch-py`  is 3.1.

integrations/opensearch/src/haystack_integrations/document_stores/opensearch/document_store.py
```diff
@@ -344,10 +349,31 @@ async def count_documents_async(self) -> int:
...
+                data.pop(key, None)
+
+            if meta:
+                data["meta"] = meta
+
             if "highlight" in hit:
-                data["metadata"]["highlighted"] = hit["highlight"]
+                if "meta" not in data:
+                    data["meta"] = {}
+                data["meta"]["highlighted"] = hit["highlight"]
```
sjrl: Could you explain what was happening before making these changes? Before this were we throwing away all meta information when reconstructing the Document?

sjrl: Also could we add some integration tests in `test_bm25_retriever.py` and `test_embedding_retriever.py` to do a full check of all fields of a returned Document? It seems we are missing some tests to confirm that returned Docs are reconstructed properly.

integrations/opensearch/src/haystack_integrations/document_stores/opensearch/document_store.py
```diff
@@ -1098,3 +1124,511 @@ def _render_custom_query(self, custom_query: Any, substitutions: dict[str, Any])
...
+        normalized_filters = normalize_filters(filters)
+        body = {"query": {"bool": {"filter": normalized_filters}}}
+        return (await self._async_client.count(index=self._index, body=body))["count"]
+
+    @staticmethod
+    def _build_cardinality_aggregations(index_mapping: dict[str, Any]) -> dict[str, Any]:
+        """
+        Builds cardinality aggregations for all metadata fields in the index mapping.
+        """
+        special_fields = {"content", "embedding", "id", "score", "blob", "sparse_embedding"}
```
sjrl: Seems like this set of fields is reused a few times. Perhaps we could make it a global variable at the top of this file (or a class attribute) so we can have one source of truth?

integrations/opensearch/src/haystack_integrations/document_stores/opensearch/document_store.py
```diff
@@ -1098,3 +1124,511 @@ def _render_custom_query(self, custom_query: Any, substitutions: dict[str, Any])
...
+        assert self._async_client is not None
+
+        normalized_filters = normalize_filters(filters)
+        body = {"query": {"bool": {"filter": normalized_filters}}}
+        return (await self._async_client.count(index=self._index, body=body))["count"]
+
+    @staticmethod
+    def _build_cardinality_aggregations(index_mapping: dict[str, Any]) -> dict[str, Any]:
+        """
+        Builds cardinality aggregations for all metadata fields in the index mapping.
```
sjrl: I think it could be helpful to link to the OpenSearch docs on cardinality aggregations https://docs.opensearch.org/latest/aggregations/metric/cardinality/ in the docstring

integrations/opensearch/src/haystack_integrations/document_stores/opensearch/document_store.py
sjrl: Also it seems like there is another function called [_deserialize_document](https://github.com/deepset-ai/haystack-core-integrations/blob/d96cc4cf85c57b02e8b7535c62c528258a63e683/integrations/opensearch/src/haystack_integrations/document_stores/opensearch/document_store.py#L564C9-L564C30) that contained the same logic here but doesn't seem to be used anywhere. Could we remove it?

davidsbatista: This is not needed and was over-engineered. 

I've added extensive tests to ensure that both BM25 and Embedding retrievers can store and retrieve documents with "complex" metadata. It's working with and without these changes. I will revert it.

Thanks for spotting this!

integrations/opensearch/pyproject.toml
davidsbatista: Running all the tests in the latest version breaks only one operation: `delete_all_documents()`, which can be easily fixed. 

I would nevertheless do this in a different PR. WDYT?

sjrl: Yeah different PR sounds good!

integrations/opensearch/src/haystack_integrations/document_stores/opensearch/document_store.py
```diff
@@ -1141,3 +1147,511 @@ def _render_custom_query(self, custom_query: Any, substitutions: dict[str, Any])
             return substitutions.get(custom_query, custom_query)
 
         return custom_query
+
+    def count_documents_by_filter(self, filters: dict) -> int:
```
sjrl: small change to keep the type hint consistent from other function defs
```suggestion
    def count_documents_by_filter(self, filters: dict[str, Any]) -> int:
```

integrations/opensearch/src/haystack_integrations/document_stores/opensearch/document_store.py
```diff
@@ -1141,3 +1147,511 @@ def _render_custom_query(self, custom_query: Any, substitutions: dict[str, Any])
...
+        :returns: The number of documents that match the filters.
+        """
+        self._ensure_initialized()
+        assert self._client is not None
+
+        normalized_filters = normalize_filters(filters)
+        body = {"query": {"bool": {"filter": normalized_filters}}}
+        return self._client.count(index=self._index, body=body)["count"]
+
+    async def count_documents_by_filter_async(self, filters: dict) -> int:
```
sjrl:
```suggestion
    async def count_documents_by_filter_async(self, filters: dict[str, Any]) -> int:
```

integrations/opensearch/src/haystack_integrations/document_stores/opensearch/document_store.py
```diff
@@ -1141,3 +1147,511 @@ def _render_custom_query(self, custom_query: Any, substitutions: dict[str, Any])
...
+        See: https://docs.opensearch.org/latest/aggregations/metric/cardinality/
+        """
+        aggs = {}
+        for field_name in index_mapping.keys():
+            if field_name not in SPECIAL_FIELDS:
+                aggs[f"{field_name}_cardinality"] = {"cardinality": {"field": field_name}}
+        return aggs
+
+    @staticmethod
+    def _build_distinct_values_query_body(filters: dict, aggs: dict[str, Any]) -> dict[str, Any]:
```
sjrl: It seems like `filters` is optional based on the rest of the function
```suggestion
    def _build_distinct_values_query_body(filters: Optional[dict[str, Any]], aggs: dict[str, Any]) -> dict[str, Any]:
```

integrations/opensearch/src/haystack_integrations/document_stores/opensearch/document_store.py
```diff
@@ -1141,3 +1147,511 @@ def _render_custom_query(self, custom_query: Any, substitutions: dict[str, Any])
...
+        """
+        distinct_counts = {}
+        for field_name in index_mapping.keys():
+            if field_name not in SPECIAL_FIELDS:
+                agg_key = f"{field_name}_cardinality"
+                if agg_key in aggregations:
+                    distinct_counts[field_name] = aggregations[agg_key]["value"]
+        return distinct_counts
+
+    def count_distinct_values_by_filter(self, filters: dict) -> dict[str, int]:
```
sjrl: Maybe a suggestion for a different name of this function. Maybe something like  `count_unique_meta_by_filter` or `count_distinct_meta_by_filter`. I'm thinking adding the word `meta` or `metadata` to the function name would make it clearer that we are counting unique values for meta fields and not counting distinct documents, for example. WDYT?

integrations/opensearch/src/haystack_integrations/document_stores/opensearch/document_store.py
```diff
@@ -1141,3 +1147,511 @@ def _render_custom_query(self, custom_query: Any, substitutions: dict[str, Any])
...
+            return {}
+
+        # build and execute search query
+        body = self._build_distinct_values_query_body(filters, aggs)
+        result = await self._async_client.search(index=self._index, body=body)
+
+        # extract cardinality values from aggregations
+        return self._extract_distinct_counts_from_aggregations(result.get("aggregations", {}), index_mapping)
+
+    def get_fields_info(self) -> dict[str, dict]:
```
sjrl: Similar here, I'd probably rename this to `get_meta_fields_info` or some way incorporate the word `meta` into the name. WDYT?

integrations/opensearch/src/haystack_integrations/document_stores/opensearch/document_store.py
```diff
@@ -1141,3 +1147,511 @@ def _render_custom_query(self, custom_query: Any, substitutions: dict[str, Any])
...
+    @staticmethod
+    def _extract_min_max_from_stats(stats: dict[str, Any]) -> dict[str, Any]:
+        """
+        Extracts min and max values from stats aggregation results.
+        """
+        min_value = stats.get("min")
+        max_value = stats.get("max")
+        return {"min": min_value, "max": max_value}
+
+    def get_field_min_max(self, metadata_field: str) -> dict[str, Any]:
```
sjrl: Maybe rename to `get_meta_field_min_max` or `get_metadata_field_min_max`?

integrations/opensearch/src/haystack_integrations/document_stores/opensearch/document_store.py
```diff
@@ -1141,3 +1147,511 @@ def _render_custom_query(self, custom_query: Any, substitutions: dict[str, Any])
...
+        assert self._async_client is not None
+
+        field_name = self._normalize_metadata_field_name(metadata_field)
+        body = self._build_min_max_query_body(field_name)
+        result = await self._async_client.search(index=self._index, body=body)
+        stats = result.get("aggregations", {}).get("field_stats", {})
+
+        return self._extract_min_max_from_stats(stats)
+
+    def get_field_unique_values(
```
sjrl: Whatever decision is made for naming let's update this as well

integrations/opensearch/src/haystack_integrations/document_stores/opensearch/document_store.py
```diff
@@ -1141,3 +1147,511 @@ def _render_custom_query(self, custom_query: Any, substitutions: dict[str, Any])
...
+            # extract only the query results
+            if isinstance(response_data, dict) and "hits" in response_data:
+                hits = response_data.get("hits", {}).get("hits", [])
+                # extract _source from each hit, which contains the actual document data
+                return [hit.get("_source", {}) for hit in hits]
+            return response_data
+        else:
+            return response_data if isinstance(response_data, str) else str(response_data)
+
+    def query_sql(self, query: str, response_format: ResponseFormat = "json") -> Any:
```
sjrl: I feel like this doesn't quite fit into our current component pattern. E.g. for bm25 and embedding retrieval we have internal functions like `_bm25_retrieval` and then we have a separate component e.g. `OpenSearchBM25Retriever` which calls this internal method. 

I think it would make sense to follow something similarly here? At the very least we will need a separate component that can call this method to return it's result. This is b/c this method is used by Sol in a pipeline to fetch I believe metadata at runtime.

sjrl: Also I'm not sure I like that the return type of this function is Any. Is it possible to restrict this to certain return types?

integrations/opensearch/src/haystack_integrations/document_stores/opensearch/document_store.py
```diff
@@ -1281,7 +1281,7 @@ async def count_distinct_values_by_filter_async(self, filters: dict) -> dict[str
         # extract cardinality values from aggregations
         return self._extract_distinct_counts_from_aggregations(result.get("aggregations", {}), index_mapping)
 
-    def get_fields_info(self) -> dict[str, dict]:
+    def get_meta_fields_info(self) -> dict[str, dict]:
```
sjrl: Let's make sure to be consistent. I see that you went for `metadata` in the functions `get_metadata_field_min_max` and `get_metadata_field_unique_values`. So perhaps lets go with `metadata` everywhere?

integrations/opensearch/src/haystack_integrations/document_stores/opensearch/document_store.py
davidsbatista: went with `count_unique_metadata_by_filter()`

integrations/opensearch/src/haystack_integrations/document_stores/opensearch/document_store.py
davidsbatista: I would then propose a `OpenSearchSQLRetriever` as new retriever component.

integrations/opensearch/src/haystack_integrations/components/retrievers/opensearch/__init__.py
```diff
@@ -5,5 +5,11 @@
...
 from .open_search_hybrid_retriever import OpenSearchHybridRetriever
+from .sql_retriever import OpenSearchSQLRetriever
 
-__all__ = ["OpenSearchBM25Retriever", "OpenSearchEmbeddingRetriever", "OpenSearchHybridRetriever"]
+__all__ = [
+    "OpenSearchBM25Retriever",
+    "OpenSearchEmbeddingRetriever",
+    "OpenSearchHybridRetriever",
+    "OpenSearchSQLRetriever",
+]
```
sjrl: We also need to update the `pydoc/config_docusaurus.yml` file as well I believe

integrations/opensearch/src/haystack_integrations/document_stores/opensearch/document_store.py
```diff
@@ -8,22 +8,28 @@
...
 from haystack_integrations.document_stores.opensearch.auth import AsyncAWSAuth, AWSAuth
 from haystack_integrations.document_stores.opensearch.filters import normalize_filters
 
 logger = logging.getLogger(__name__)
 
+SPECIAL_FIELDS = {"content", "embedding", "id", "score", "sparse_embedding", "blob"}
 
 Hosts = Union[str, list[Union[str, Mapping[str, Union[str, int]]]]]
 
+ResponseFormat = Literal["json", "jdbc", "csv", "raw"]
```
sjrl: Could we reuse this inside of the sql retriever instead of redefining it?

integrations/opensearch/src/haystack_integrations/components/retrievers/opensearch/sql_retriever.py
```diff
@@ -0,0 +1,190 @@
...
+        )
+        return default_from_dict(cls, data)
+
+    @component.output_types(result=Any)
+    def run(
+        self,
+        query: str,
+        response_format: Optional[ResponseFormat] = None,
+        document_store: Optional[OpenSearchDocumentStore] = None,
+    ) -> dict[str, Any]:
```
sjrl: If the response format can change the output type then I think we need to follow a similar approach as we do in `OutputAdapter` and `ConditionalRouter` where we set the output type of the component in the init method. Otherwise this return type of `Any` will cause this component to not be compatible with any other component other than `OutputAdapter`, `PromptBuilder`, and `ChatPromptBuilder`. Basically only the few components that accept `Any` as input.

sjrl: Maybe as an alternative to supporting all response types we could find out from @ju-gu which one he is currently using and provide support for just that response type. Then we could also make the return dictionary more structured. 

E.g. if we only support `json` then we could return the top-level keys as part of our return dict
```
return {"hits": ..., "took": ..., etc.}
```

integrations/opensearch/src/haystack_integrations/document_stores/opensearch/document_store.py
```diff
@@ -1141,3 +1147,523 @@ def _render_custom_query(self, custom_query: Any, substitutions: dict[str, Any])
...
+        for field_name in index_mapping.keys():
+            if field_name not in SPECIAL_FIELDS:
+                agg_key = f"{field_name}_cardinality"
+                if agg_key in aggregations:
+                    distinct_counts[field_name] = aggregations[agg_key]["value"]
+        return distinct_counts
+
+    def count_unique_metadata_by_filter(self, filters: dict) -> dict[str, int]:
+        """
+        Returns the number of unique values for each meta field of the documents that match the provided filters.
```
sjrl:
```suggestion
        Returns the number of unique values for each metadata field of the documents that match the provided filters.
```

integrations/opensearch/src/haystack_integrations/document_stores/opensearch/document_store.py
```diff
@@ -1141,3 +1147,523 @@ def _render_custom_query(self, custom_query: Any, substitutions: dict[str, Any])
...
+
+        :param filters: The filters to apply to count documents.
+            For filter syntax, see [Haystack metadata filtering](https://docs.haystack.deepset.ai/docs/metadata-filtering)
+        :returns: The number of unique values for each meta field of the documents that match the filters.
```
sjrl:
```suggestion
        :returns: A dictionary mapping each metadata field name to the count of its unique values among the filtered documents.
```

integrations/opensearch/src/haystack_integrations/document_stores/opensearch/document_store.py
```diff
@@ -1141,3 +1147,523 @@ def _render_custom_query(self, custom_query: Any, substitutions: dict[str, Any])
...
+        """
+        distinct_counts = {}
+        for field_name in index_mapping.keys():
+            if field_name not in SPECIAL_FIELDS:
+                agg_key = f"{field_name}_cardinality"
+                if agg_key in aggregations:
+                    distinct_counts[field_name] = aggregations[agg_key]["value"]
+        return distinct_counts
+
+    def count_unique_metadata_by_filter(self, filters: dict) -> dict[str, int]:
```
sjrl:
```suggestion
    def count_unique_metadata_by_filter(self, filters: dict[str, Any]) -> dict[str, int]:
```

integrations/opensearch/src/haystack_integrations/document_stores/opensearch/document_store.py
```diff
@@ -1141,3 +1147,523 @@ def _render_custom_query(self, custom_query: Any, substitutions: dict[str, Any])
...
+        # build and execute search query
+        body = self._build_distinct_values_query_body(filters, aggs)
+        result = self._client.search(index=self._index, body=body)
+
+        # extract cardinality values from aggregations
+        return self._extract_distinct_counts_from_aggregations(result.get("aggregations", {}), index_mapping)
+
+    async def count_unique_metadata_by_filter_async(self, filters: dict) -> dict[str, int]:
+        """
+        Asynchronously returns the number of unique values for each meta field of the documents that match the
```
sjrl:
```suggestion
        Asynchronously returns the number of unique values for each metadata field of the documents that match the
```

integrations/opensearch/src/haystack_integrations/document_stores/opensearch/document_store.py
```diff
@@ -1141,3 +1147,523 @@ def _render_custom_query(self, custom_query: Any, substitutions: dict[str, Any])
...
+        provided filters.
+
+        :param filters: The filters to apply to count documents.
+            For filter syntax, see [Haystack metadata filtering](https://docs.haystack.deepset.ai/docs/metadata-filtering)
+        :returns: The number of unique values for each meta field of the documents that match the filters.
```
sjrl:
```suggestion
        :returns: A dictionary mapping each metadata field name to the count of its unique values among the filtered documents.
```

integrations/opensearch/src/haystack_integrations/document_stores/opensearch/document_store.py
```diff
@@ -1141,3 +1147,523 @@ def _render_custom_query(self, custom_query: Any, substitutions: dict[str, Any])
...
+            return {}
+
+        # build and execute search query
+        body = self._build_distinct_values_query_body(filters, aggs)
+        result = self._client.search(index=self._index, body=body)
+
+        # extract cardinality values from aggregations
+        return self._extract_distinct_counts_from_aggregations(result.get("aggregations", {}), index_mapping)
+
+    async def count_unique_metadata_by_filter_async(self, filters: dict) -> dict[str, int]:
```
sjrl:
```suggestion
    async def count_unique_metadata_by_filter_async(self, filters: dict[str, Any]) -> dict[str, int]:
```

integrations/opensearch/src/haystack_integrations/document_stores/opensearch/document_store.py
```diff
@@ -1141,3 +1147,523 @@ def _render_custom_query(self, custom_query: Any, substitutions: dict[str, Any])
...
+        result = await self._async_client.search(index=self._index, body=body)
+
+        # extract cardinality values from aggregations
+        return self._extract_distinct_counts_from_aggregations(result.get("aggregations", {}), index_mapping)
+
+    def get_metadata_fields_info(self) -> dict[str, dict]:
+        """
+        Returns the information about the fields in the index.
+
+        :returns: The information about the fields in the index.
```
sjrl: Could we a bit more descriptive here in the docs about what is returned? E.g. is there a docs page on OpenSearch that shows an example of what the response could look like?

integrations/opensearch/src/haystack_integrations/document_stores/opensearch/document_store.py
```diff
@@ -1141,3 +1147,523 @@ def _render_custom_query(self, custom_query: Any, substitutions: dict[str, Any])
...
+                    "stats": {
+                        "field": field_name,
+                    }
+                }
+            },
+            "size": 0,  # We only need aggregations, not documents
+        }
+
+    @staticmethod
+    def _extract_min_max_from_stats(stats: dict[str, Any]) -> dict[str, Any]:
```
sjrl:
```suggestion
    def _extract_min_max_from_stats(stats: dict[str, Any]) -> dict[str, int | None]:
```

integrations/opensearch/src/haystack_integrations/document_stores/opensearch/document_store.py
```diff
@@ -1141,3 +1147,523 @@ def _render_custom_query(self, custom_query: Any, substitutions: dict[str, Any])
...
+        """
+        Extracts min and max values from stats aggregation results.
+        """
+        min_value = stats.get("min")
+        max_value = stats.get("max")
+        return {"min": min_value, "max": max_value}
+
+    def get_metadata_field_min_max(self, metadata_field: str) -> dict[str, Any]:
```
sjrl:
```suggestion
    def get_metadata_field_min_max(self, metadata_field: str) -> dict[str, int | None]:
```

integrations/opensearch/src/haystack_integrations/document_stores/opensearch/document_store.py
```diff
@@ -1141,3 +1147,523 @@ def _render_custom_query(self, custom_query: Any, substitutions: dict[str, Any])
...
+        """
+        Returns the minimum and maximum values for the given metadata field.
+
+        :param metadata_field: The metadata field to get the minimum and maximum values for.
+        :returns: The minimum and maximum values for the given metadata field.
```
sjrl:
```suggestion
        :returns: A dictionary with the keys "min" and "max", where each value is the minimum or maximum value of the metadata field across all documents.
```

integrations/opensearch/src/haystack_integrations/document_stores/opensearch/document_store.py
```diff
@@ -1141,3 +1147,523 @@ def _render_custom_query(self, custom_query: Any, substitutions: dict[str, Any])
...
+        stats = result.get("aggregations", {}).get("field_stats", {})
+
+        return self._extract_min_max_from_stats(stats)
+
+    async def get_metadata_field_min_max_async(self, metadata_field: str) -> dict[str, Any]:
+        """
+        Asynchronously returns the minimum and maximum values for the given metadata field.
+
+        :param metadata_field: The metadata field to get the minimum and maximum values for.
+        :returns: The minimum and maximum values for the given metadata field.
```
sjrl:
```suggestion
        :returns: A dictionary with the keys "min" and "max", where each value is the minimum or maximum value of the metadata field across all documents.
```

integrations/opensearch/src/haystack_integrations/document_stores/opensearch/document_store.py
```diff
@@ -1141,3 +1147,523 @@ def _render_custom_query(self, custom_query: Any, substitutions: dict[str, Any])
...
+        assert self._client is not None
+
+        field_name = self._normalize_metadata_field_name(metadata_field)
+        body = self._build_min_max_query_body(field_name)
+        result = self._client.search(index=self._index, body=body)
+        stats = result.get("aggregations", {}).get("field_stats", {})
+
+        return self._extract_min_max_from_stats(stats)
+
+    async def get_metadata_field_min_max_async(self, metadata_field: str) -> dict[str, Any]:
```
sjrl:
```suggestion
    async def get_metadata_field_min_max_async(self, metadata_field: str) -> dict[str, int | None]:
```

integrations/opensearch/src/haystack_integrations/document_stores/opensearch/document_store.py
```diff
@@ -1141,3 +1147,523 @@ def _render_custom_query(self, custom_query: Any, substitutions: dict[str, Any])
...
+        # Extract unique values from terms aggregation buckets
+        unique_values_buckets = aggregations.get("unique_values", {}).get("buckets", [])
+        # Apply pagination by slicing the results
+        paginated_buckets = unique_values_buckets[from_ : from_ + size]
+        unique_values = [str(bucket["key"]) for bucket in paginated_buckets]
+
+        # Extract total count from cardinality aggregation
+        total_count = int(aggregations.get("total_count", {}).get("value", 0))
+
+        return unique_values, total_count
```
sjrl: Out of curiosity why do we return the list of unique values and the total count? I get the impression that the`len(unique_values)` should equal `total_count` but perhaps not?

cc @tstadel

integrations/opensearch/src/haystack_integrations/components/retrievers/opensearch/sql_retriever.py
davidsbatista: I would vote for this approach 👍🏽

davidsbatista: > If the response format can change the output type then I think we need to follow a similar approach as we do in `OutputAdapter` and `ConditionalRouter` where we set the output type of the component in the init method. Otherwise this return type of `Any` will cause this component to not be compatible with any other component other than `OutputAdapter`, `PromptBuilder`, and `ChatPromptBuilder`. Basically only the few components that accept `Any` as input.

Thanks for this detailed answer! I understand now the main issue behind the `Any` 👍🏽

integrations/opensearch/src/haystack_integrations/components/retrievers/opensearch/__init__.py
davidsbatista: good catch - thanks 👍🏽

integrations/opensearch/src/haystack_integrations/document_stores/opensearch/document_store.py
davidsbatista: from my understanding it seems we could maybe simple return the `list[unique_values]`

integrations/opensearch/src/haystack_integrations/document_stores/opensearch/document_store.py
```diff
@@ -1141,3 +1144,420 @@ def _render_custom_query(self, custom_query: Any, substitutions: dict[str, Any])
...
+
+    def _prepare_sql_http_request_params(
+        self, base_url: str, response_format: ResponseFormat
+    ) -> tuple[str, dict[str, str], Any]:
```
sjrl: Could we also move all sql related functions to your draft PR https://github.com/deepset-ai/haystack-core-integrations/pull/2746 and remove them from this one?

integrations/opensearch/src/haystack_integrations/document_stores/opensearch/document_store.py
sjrl: If @tstadel confirms I agree just returning `list[unique_values]` makes sense

integrations/opensearch/src/haystack_integrations/document_stores/opensearch/document_store.py
```diff
@@ -1141,3 +1144,420 @@ def _render_custom_query(self, custom_query: Any, substitutions: dict[str, Any])
...
+            return {}
+
+        # build and execute search query
+        body = self._build_distinct_values_query_body(filters, aggs)
+        result = await self._async_client.search(index=self._index, body=body)
+
+        # extract cardinality values from aggregations
+        return self._extract_distinct_counts_from_aggregations(result.get("aggregations", {}), index_mapping)
+
+    def get_metadata_fields_info(self) -> dict[str, dict]:
```
sjrl: Would it be safe to update the return type to 
```suggestion
    def get_metadata_fields_info(self) -> dict[str, dict[str, str]]:
```

integrations/opensearch/src/haystack_integrations/document_stores/opensearch/document_store.py
```diff
@@ -1141,3 +1144,420 @@ def _render_custom_query(self, custom_query: Any, substitutions: dict[str, Any])
...
+        self._ensure_initialized()
+        assert self._client is not None
+
+        mapping = self._client.indices.get_mapping(index=self._index)
+        index_mapping = mapping[self._index]["mappings"]["properties"]
+        # remove all fields that are not metadata fields
+        index_mapping = {k: v for k, v in index_mapping.items() if k not in SPECIAL_FIELDS}
+        return index_mapping
+
+    async def get_metadata_fields_info_async(self) -> dict[str, dict]:
```
sjrl: Also here
```suggestion
    async def get_metadata_fields_info_async(self) -> dict[str, dict[str, str]]:
```

integrations/opensearch/src/haystack_integrations/document_stores/opensearch/document_store.py
```diff
@@ -1141,3 +1144,420 @@ def _render_custom_query(self, custom_query: Any, substitutions: dict[str, Any])
...
+            Document(content="Doc 2", meta={"category": "B", "status": "inactive"})
+
+        This method would return:
+
+            {
+                'content': {'type': 'text'},
+                'category': {'type': 'keyword'},
+                'status': {'type': 'keyword'},
+                'priority': {'type': 'long'},
+            }
```
sjrl, line 1335: It seems like the return type of this function `get_metadata_fields_info` will be specific to OpenSearch. Whereas all other functions you've added that we are looking to add across all Document Stores have a format we created. 

Do you think it's possible or reasonable for us to try and standardize something like this?

integrations/opensearch/tests/test_document_store.py
```diff
@@ -593,6 +593,202 @@ def test_update_by_filter(self, document_store: OpenSearchDocumentStore):
         assert len(draft_docs) == 1
         assert draft_docs[0].meta["category"] == "B"
 
+    def test_count_documents_by_filter(self, document_store: OpenSearchDocumentStore):
```
sjrl, line 596: Probably out of scope for this PR, but could we open an issue to move these standardized tests into `haystack.testing.document_store` as testing modules? I think this would help reduce the duplicate test code we have in all of our document store integrations. 

WDYT?

integrations/opensearch/src/haystack_integrations/document_stores/opensearch/document_store.py
davidsbatista: We can add a transformation inside each DocumentStore `get_metadata_fields_info ` so that it returns in this format. When this PR is merged I will adapt the issues: noting this + renaming the function names and signature.

integrations/opensearch/tests/test_document_store.py
davidsbatista: good catch yes, that's a good idea 👍🏽

davidsbatista: https://github.com/deepset-ai/haystack-core-integrations/issues/2747

The only issue I'm seeing with this is the parameters of the tests when tied to a specific DocStore. We probably can extend the DocStore Protocol and add a few more operations, if all of our document stores already support them.

integrations/opensearch/src/haystack_integrations/document_stores/opensearch/document_store.py
```diff
@@ -1141,3 +1144,387 @@ def _render_custom_query(self, custom_query: Any, substitutions: dict[str, Any])
...
+
+    def count_unique_metadata_by_filter(self, filters: dict[str, Any]) -> dict[str, int]:
+        """
+        Returns the number of unique values for each metadata field of the documents that match the provided filters.
+
+        :param filters: The filters to apply to count documents.
+            For filter syntax, see [Haystack metadata filtering](https://docs.haystack.deepset.ai/docs/metadata-filtering)
+        :returns: A dictionary mapping each metadata field name to the count of its unique values among the filtered
+                  documents.
+        """
```
tstadel: Just a thought: for big indexes calculating all unique values of all fields can be a very expensive operation. So my thinking is: wouldn't it make more sense to 
1) restrict the unique values calculation to one field
2) optionally restrict the unique values calculation to a list of fields

I guess 2) would be easy to add without changing too much.

integrations/opensearch/src/haystack_integrations/document_stores/opensearch/document_store.py
tstadel: No need to return total_count here, we have `count_unique_metadata_by_filter` instead.

integrations/opensearch/src/haystack_integrations/document_stores/opensearch/document_store.py
```diff
@@ -1141,3 +1144,387 @@ def _render_custom_query(self, custom_query: Any, substitutions: dict[str, Any])
...
+        if search_term:
+            # Use match_phrase for exact phrase matching to avoid tokenization issues
+            query = {"match_phrase": {"content": search_term}}
+
+        # Build aggregations
+        # Terms aggregation for paginated unique values
+        # Note: Terms aggregation doesn't support 'from' parameter directly,
+        # so we fetch from_ + size results and slice them
+        # Cardinality aggregation for total count
+        terms_size = from_ + size if from_ > 0 else size
```
tstadel: Note that this only works till 10k. Overall the pagination is not ideal here. I'd suggest to switch over to [composite aggregations](https://docs.opensearch.org/latest/aggregations/bucket/composite/#paginating-composite-results) which support proper pagination. However you'd need to change the signature to support the after param instead of from.

tstadel: I'd suggest to remove the from param altogether as it gives incorrect impressions here. That would be the quickest measure.

For full pagination support I'd recommend:
1) support from and after on protocol level (as some document stores might support from-based and some after-based pagination)
2) raise NotSupportedError for OpenSearch when from is set
3) implement after-based pagination for OpenSearch using composite aggregations

integrations/opensearch/src/haystack_integrations/document_stores/opensearch/document_store.py
davidsbatista: done, update the parent issue

integrations/opensearch/src/haystack_integrations/document_stores/opensearch/document_store.py
davidsbatista: I would vote for option 2)

integrations/opensearch/src/haystack_integrations/document_stores/opensearch/document_store.py
davidsbatista: I went with option 3.

```python
        self,
        metadata_field: str,
        size: int | None = 10000,
        after: dict[str, Any] | None = None,
    )
```

Regarding the Protocol - I support your suggestion of adding both `from` and `after` - but let's see how this applies to other DocumentStores, and then we can see if it's easier to add to `Protocol`.

I think there are common operations that can already be added to the `Protocol`. I need to review it in detail and make use of it to avoid having so much duplicated code for tests. There's an issue with it, and I will probably take it soon.

integrations/opensearch/src/haystack_integrations/document_stores/opensearch/document_store.py
```diff
@@ -21,6 +21,9 @@
 
 logger = logging.getLogger(__name__)
 
+SPECIAL_FIELDS = {"content", "embedding", "id", "score", "sparse_embedding", "blob"}
+
+ResponseFormat = Literal["json", "jdbc", "csv", "raw"]
```
sjrl: I think this can be removed now since it's not being used

sjrl (approved):
One minor comment, otherwise looks good!

PR #2656, opened 2026-01-06
feat: adding `delete_by_filter` and `update_by_filter` to `WeaviateDocumentStore`

### LINKED ISSUE #2335: add `update_by_filter()` and `delete_by_filter()` operations to `WeaviateDocumentStore`
```python
def delete_by_filter(self, filters: Dict[str, Any]) -> int
```

```python
def update_by_filter(self, filters: Dict[str, Any], meta: Dict[str, Any]) -> int
```

davidsbatista:
### Related Issues

- fixes #2335 

### Proposed Changes:

- added delete_by_filter and update_by_filter methods to Weaviate document store
- Implemented sync and async versions of both methods

### How did you test it?

- added tests for sync (test_document_store.py) and async (test_document_store_async.py) versions, following the same pattern as other document stores (OpenSearch, Elasticsearch)

### Checklist

### DIFF
--- a/integrations/qdrant/CHANGELOG.md
+++ b/integrations/qdrant/CHANGELOG.md
@@ -1,5 +1,20 @@
 # Changelog
 
+## [integrations/qdrant-v9.5.0] - 2026-01-07
+
+### 🚀 Features
+
+- Adding `delete_by_filter()` and `update_by_filter()` to `QdrantDocumentStore` (#2650)
+
+### 🐛 Bug Fixes
+
+- Fix types for qdrant-client 1.16.1 (#2561)
+
+### 🧹 Chores
+
+- Remove Readme API CI workflow and configs (#2573)
+
+
 ## [integrations/qdrant-v9.4.0] - 2025-11-18
 
 ### 🧹 Chores
--- a/integrations/qdrant/src/haystack_integrations/document_stores/qdrant/document_store.py
+++ b/integrations/qdrant/src/haystack_integrations/document_stores/qdrant/document_store.py
@@ -1,7 +1,7 @@
 import inspect
 from collections.abc import AsyncGenerator, Generator
 from itertools import islice
-from typing import Any, ClassVar, Optional, Union
+from typing import Any, ClassVar, Optional, Union, cast
 
 import qdrant_client
 from haystack import default_from_dict, default_to_dict, logging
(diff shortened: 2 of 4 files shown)


integrations/weaviate/src/haystack_integrations/document_stores/weaviate/document_store.py
```diff
@@ -610,6 +609,255 @@ def delete_all_documents(self, *, recreate_index: bool = False, batch_size: int
...
+                    filters=weaviate_filter,
+                    include_vector=True,
+                    limit=DEFAULT_QUERY_LIMIT,
+                    offset=offset,
+                    return_properties=properties,
+                )
+                matching_objects.extend(partial_result.objects)
+                offset += DEFAULT_QUERY_LIMIT
+                if len(partial_result.objects) < DEFAULT_QUERY_LIMIT:
+                    break
```
sjrl: Could you explain why we need this break? The other place I see we use a similar pagination structure is in `_query_with_filters` and there we don't seem to use this break check. 

Actually could we be using `self._query_with_filters` here to help avoid duplicate code?

integrations/weaviate/src/haystack_integrations/document_stores/weaviate/document_store.py
```diff
@@ -610,6 +609,255 @@ def delete_all_documents(self, *, recreate_index: bool = False, batch_size: int
...
+
+            for obj in matching_objects:
+                try:
+                    # Get current properties
+                    current_properties = obj.properties.copy() if obj.properties else {}
+
+                    # Update with new metadata values
+                    # Note: metadata fields are stored directly in properties (flattened)
+                    for key, value in meta.items():
+                        current_properties[key] = value
```
sjrl, line 700: you mention here that metadata fields are stored directly in properties but in the method `_to_data_object` which converts a document to a data object I don't see us flattening the meta data. It seems we more or less pass `document.to_dict()` directly. 

Could you point me to where we normally flatten the metadata?

sjrl: I do see how your tests indicate that this is correct, so this is more for my own education

integrations/weaviate/src/haystack_integrations/document_stores/weaviate/document_store.py
davidsbatista: this break is redundant the same check is done at the head o the loop - I've update this to reuse the `query_with_filters` for the sync version.

integrations/weaviate/src/haystack_integrations/document_stores/weaviate/document_store.py
davidsbatista: It's on the default behaviour of the [`document.to_dict()`](https://github.com/deepset-ai/haystack/blob/main/haystack/dataclasses/document.py#L137C1-L139C36) in haystack, `flatten=True`

sjrl: ahh right thanks!

PR #2657, opened 2026-01-07
feat: Add 'Client-Platform' header for server processing for PaddleOCR

Bobholamovic:
This is a new specification introduced for the official PaddleOCR service. By identifying the source of client requests, we will be able to implement additional server-side extensions in the future, such as providing special handling for certain clients or enabling more fine-grained traffic control.


davidsbatista:
thanks for the contribution @Bobholamovic 👍🏽

davidsbatista (approved):

PR #2672, opened 2026-01-08
test: Add `pytest-xdist` to execute tests in parallel on CI (and/or locally)

mpangrazzi:
### Proposed Changes:

I've added `pytest-xdist` which distributes tests on N different workers for faster execution. On my Mac M3 Pro I can get tests executed ~75% faster. (35s vs 2m 15s on avg).

### How did you test it?

unit tests, integration tests

### Notes for the reviewer

By default, tests will be still run serially. To run them concurrently, you must add `-n` option. For example:

```bash
hatch run test:integration -n auto  # workers == available cores
hatch run test:all -n 4  # using 4 workers
```

### Checklist

### DIFF
--- a/integrations/opensearch/tests/conftest.py
+++ b/integrations/opensearch/tests/conftest.py
@@ -17,7 +17,7 @@ def _get_unique_index_name() -> str:
 
 
 @pytest.fixture
-def document_store(request):
+def document_store():
     """
     We use this document store for basic tests and for testing filters.
     `return_embedding` is set to True because in filters tests we compare embeddings.
@@ -45,7 +45,7 @@ def document_store(request):
 
 
 @pytest.fixture
-def document_store_2(request):
+def document_store_2():
     hosts = ["https://localhost:9200"]
     index = f"test_index_2_{_get_unique_index_name()}"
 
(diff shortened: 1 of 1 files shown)


integrations/opensearch/tests/test_document_store.py
```diff
@@ -443,28 +443,6 @@ def test_write_documents_max_chunk_bytes(self, mock_bulk, document_store):
 
         assert mock_bulk.call_args.kwargs["max_chunk_bytes"] == DEFAULT_MAX_CHUNK_BYTES
 
-    @pytest.fixture
```
anakin87, line 446: this was unsused?

integrations/opensearch/tests/conftest.py
```diff
@@ -1,11 +1,27 @@
 import asyncio
+import re
 
 import pytest
 from haystack import Document
 
 from haystack_integrations.document_stores.opensearch.document_store import OpenSearchDocumentStore
 
 
+def _get_unique_index_name(request) -> str:
```
anakin87: could we simplify this? I am thinking of generating a UUID or something similar.

integrations/opensearch/tests/test_document_store.py
mpangrazzi: I think it's already present in `conftest.py`

integrations/opensearch/tests/conftest.py
mpangrazzi: Done. I initially though that having the test name as the index name may help. But on a second though, we can also use an UUID without issues.

PR #2746, opened 2026-01-13
feat: add an `SQLRetriever` to `OpenSearchDocumentStore`

davidsbatista:
### Proposed Changes:

- part of #2635 - adding an `SQLRetriever`

### How did you test it?

- added integration tests
- manual verification

### Checklist

### DIFF
--- a/integrations/opensearch/src/haystack_integrations/components/retrievers/opensearch/sql_retriever.py
+++ b/integrations/opensearch/src/haystack_integrations/components/retrievers/opensearch/sql_retriever.py
@@ -84,7 +84,7 @@ class OpenSearchSQLRetriever:
         query: str,
         document_store: OpenSearchDocumentStore | None = None,
         fetch_size: int | None = None,
-    ) -> dict[str, dict[str, Any] | None]:
+    ) -> dict[str, dict[str, Any]]:
         """
         Execute a raw OpenSearch SQL query against the index.
 
@@ -129,17 +129,17 @@ class OpenSearchSQLRetriever:
                     error=str(e),
                     exc_info=True,
                 )
-                result = None
+                result = {}
 
         return {"result": result}
 
-    @component.output_types(result=Any)
+    @component.output_types(result=dict[str, Any])
     async def run_async(
         self,
         query: str,
         document_store: OpenSearchDocumentStore | None = None,
         fetch_size: int | None = None,
-    ) -> dict[str, dict[str, Any] | None]:
+    ) -> dict[str, dict[str, Any]]:
         """
         Asynchronously execute a raw OpenSearch SQL query against the index.
 
@@ -184,6 +184,6 @@ class OpenSearchSQLRetriever:
                     error=str(e),
                     exc_info=True,
                 )
-                result = None
+                result = {}
 
         return {"result": result}
--- a/integrations/opensearch/tests/test_sql_retriever.py
+++ b/integrations/opensearch/tests/test_sql_retriever.py
@@ -235,7 +235,7 @@ def test_sql_retriever_error_handling(document_store: OpenSearchDocumentStore):
     # Test with raise_on_failure=False
     retriever_no_raise = OpenSearchSQLRetriever(document_store=document_store, raise_on_failure=False)
     result = retriever_no_raise.run(query=invalid_query)
-    assert result["result"] is None
+    assert result["result"] == {}
 
 
 @pytest.mark.integration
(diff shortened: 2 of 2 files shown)


davidsbatista:
@tstadel @sjrl regarding the `ResponseFormat` for the SQLRetriever, as discussed before, I think it's better to stick with one simple format - I suggest JSON - and drop the others. Makes the code simpler.

WDYT?

integrations/opensearch/src/haystack_integrations/components/retrievers/opensearch/sql_retriever.py
```diff
@@ -0,0 +1,185 @@
...
+        )
+        return default_from_dict(cls, data)
+
+    @component.output_types(result=Any)
+    def run(
+        self,
+        query: str,
+        document_store: OpenSearchDocumentStore | None = None,
+        fetch_size: int | None = None,
+    ) -> dict[str, Any]:
```
sjrl: Is it possible to make the type of `result` more specific than `Any`? I think we previously discussed how this makes it difficult to connect the output of this component to any other component in our library.

sjrl: Based on your docstrings it looks like we can already update it to `list[dict]` and perhaps we can further type it based on what the dict structure looks like.

integrations/opensearch/src/haystack_integrations/components/retrievers/opensearch/sql_retriever.py
```diff
@@ -0,0 +1,185 @@
...
+        except Exception as e:
+            if self._raise_on_failure:
+                raise e
+            else:
+                logger.warning(
+                    "An error during SQL query execution occurred and will be ignored by returning None: {error}",
+                    error=str(e),
+                    exc_info=True,
+                )
+                result = None
```
sjrl: Instead of returning `None`, I think we should do the same as we do for doc retrievers which is return an empty `list` here so it still follows the `list[dict]` type of `result`. If we were to make `result`'s type `list[dict] | None` this also makes it hard to connect to other components in our library since then the receiving component we need to support both `list[dict]` and `None`.

integrations/opensearch/src/haystack_integrations/components/retrievers/opensearch/sql_retriever.py
davidsbatista: We can update it to 

```python
dict[str, list[dict[str, Any]]]
```

or 

```python
dict[str, list[dict]]
```

The outer dict is the typical component return type, with the key "result" and value being a list of dicts.

where each dict in the list represents a SQL query row:

- keys: column names from the SQL SELECT clause (strings)
- values: Any type (str, int, float, bool, None, list, dict, etc.)

integrations/opensearch/src/haystack_integrations/components/retrievers/opensearch/sql_retriever.py
davidsbatista: I completely removed all the `None` from the return type.

integrations/opensearch/src/haystack_integrations/components/retrievers/opensearch/sql_retriever.py
```diff
@@ -0,0 +1,189 @@
...
+
+        :returns:
+            Deserialized component.
+        """
+        data["init_parameters"]["document_store"] = OpenSearchDocumentStore.from_dict(
+            data["init_parameters"]["document_store"]
+        )
+        return default_from_dict(cls, data)
+
+    @component.output_types(result=Any)
```
sjrl: I think since we can let's be a little more specific here
```suggestion
    @component.output_types(result=dict[str, Any])
```

integrations/opensearch/src/haystack_integrations/components/retrievers/opensearch/sql_retriever.py
```diff
@@ -0,0 +1,189 @@
...
+                logger.warning(
+                    "An error during SQL query execution occurred and will be ignored by returning None: {error}",
+                    error=str(e),
+                    exc_info=True,
+                )
+                result = None
+
+        return {"result": result}
+
+    @component.output_types(result=Any)
```
sjrl: Also update the return type here

integrations/opensearch/src/haystack_integrations/components/retrievers/opensearch/sql_retriever.py
sjrl: Also let's keep the change that we return an empty dict instead of `None` in case something goes wrong.

Issue #2592, opened 2025-12-09
Instead of throwing a `RuntimeError` if a Component has not been warmed up call `warm_up` at runtime


sjrl:
As a follow up to the issue https://github.com/deepset-ai/haystack/issues/9966 we should also update the components in our core integrations to also automatically call `warm_up` at runtime instead of throwing an error. More detailed motivation can be found in the linked issue. 

Core integrations that should be checked (based off a simple `grep -RIi "def warm_up" integrations` search):
- [x] optimum
- [x] ~~opensearch~~ No update needed
- [x] fastembed
- [x] snowflake
- [x] nvidia
- [x] amazon_bedrock
- [x] llama_cpp
- [x] weave
- [x] hanlp

PR #2748, opened 2026-01-14
feat: Update Llama CPP components to auto call run `warm_up`

sjrl:
### Related Issues

- part of https://github.com/deepset-ai/haystack-core-integrations/issues/2592
- part of https://github.com/deepset-ai/haystack-private/issues/215

### Proposed Changes:

 
 

- Update Generator and ChatGenerator to auto call run `warm_up`
- Add missing license headers

### How did you test it?

Existing tests

### Checklist


integrations/llama_cpp/tests/test_chat_generator.py
julian-risch: I suggest we remove the `warm_up()` call here as it is no longer needed. While we could test that calling warm_up explicitly doesn't raise an error, more importantly we should test that not calling it explicitly raises no error.

julian-risch: we can remove this warm_up() call as it is no longer needed

PR #2768, opened 2026-01-20
feat:  adding count, filtering and metadata related operations to `PGVectorDocumentStore`

### LINKED ISSUE #2639: add the following operations to `PgvectorDocumentStore`
```python
def count_documents_by_filter(filter: dict) -> int
```
- applies a filter and counts the documents that matched it.

---

```python
def count_unique_metadata_by_filter(filters: dict[str, Any], metadata_fields: list[str]) -> dict[str, int]` 
```
- applies a filter selecting documents and counts the unique values for each meta field of the matched documents

---

```python
def get_metadata_fields_info() -> dict[str, dict]
```
- returns the meta fields and the corresponding types, it should return the same format as in the `OpenSearchDocumentStore`

```python
{
    'content': {'type': 'text'},
    'category': {'type': 'keyword'},
    'status': {'type': 'keyword'},
    'priority': {'type': 'long'},
}
```

---

```python
def get_metadata_field_min_max(metadata_field: str) -> dict[str, Any]
```
- for a given metadata field, find its max and min value

---

```python
def get_metadata_field_unique_values(metadata_field: str, search_term: str | None, from_: int, size: int) -> tuple[list[str], int]:
```
- retrieves unique values for a field matching a search_term or all possible values if no search term is given

---

```python
def query_sql(query: str)
```
- (optional) issue an SQL query for the doc stores that support it

davidsbatista:
### Related Issues

- fixes #2639 

### Proposed Changes:

```
def count_documents_by_filter(filter: dict) -> int

def count_unique_metadata_by_filter(filters: dict[str, Any], metadata_fields: list[str]) -> dict[str, int]`

def get_metadata_fields_info() -> dict[str, dict]

def get_metadata_field_min_max(metadata_field: str) -> dict[str, Any]

def get_metadata_field_unique_values(metadata_field: str, search_term: str | None, from_: int, size: int) -> tuple[list[str]
```

### How did you test it?

- new integrations tests for both the sync and async version

### NOTES

- SQLRetriever or similar will be handled in a different issue/PR

### Checklist

### DIFF
--- a/integrations/pgvector/src/haystack_integrations/document_stores/pgvector/document_store.py
+++ b/integrations/pgvector/src/haystack_integrations/document_stores/pgvector/document_store.py
@@ -1512,8 +1512,10 @@ class PgvectorDocumentStore:
         # Validate field name to prevent SQL injection
         # Only allow alphanumeric characters, underscores, and hyphens
         if not all(c.isalnum() or c in ("_", "-") for c in field_name):
-            msg = (f"Invalid metadata field name: '{field_name}'. Field names can only contain alphanumeric "
-                   f"characters, underscores, and hyphens.")
+            msg = (
+                f"Invalid metadata field name: '{field_name}'. Field names can only contain alphanumeric "
+                f"characters, underscores, and hyphens."
+            )
             raise ValueError(msg)
 
         return field_name
@@ -1552,8 +1554,7 @@ class PgvectorDocumentStore:
 
     @staticmethod
     def _process_count_unique_metadata_result(
-        result: dict[str, Any] | None,
-        normalized_fields: list[str]
+        result: dict[str, Any] | None, normalized_fields: list[str]
     ) -> dict[str, int]:
         """
         Processes the result from counting unique metadata values.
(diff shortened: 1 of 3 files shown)


integrations/pgvector/src/haystack_integrations/document_stores/pgvector/document_store.py
```diff
@@ -1438,3 +1438,597 @@ async def _embedding_retrieval_async(
...
+        :param records: The list of records from the select query.
+        :returns: A tuple containing (unique_values, total_count).
+        """
+        total_count = count_result.get("total", 0) if count_result else 0
+        unique_values = [str(record.get("value", "")) for record in records if record.get("value") is not None]
+        return unique_values, total_count
+
+    def get_metadata_field_unique_values(
+        self, metadata_field: str, search_term: str | None, from_: int, size: int
+    ) -> tuple[list[str], int]:
```
sjrl, line 1959: I noticed that this doesn't have the same signature as the method for `OpenSearchDcoumentStore`. Should we make those consistent in case we ever want to update the document store protocol?

davidsbatista: This is because the pagination on `PGVectorDocumentStore` is being done on an offset base, whereas in `OpenSearchDcoumentStore` is on cursor-based.

I would suggest merging as is, and we can make all the operations part of the Protocol, except `get_metadata_field_unique_values()`

We need to investigate a way to have them all work the same way, or to add a wrapper around each underlying method.

sjrl: sounds good!

PR #2771, opened 2026-01-21
feat: adding a `MetadataRetriever` to `OpenSearchDocumentStore`

davidsbatista:
### Related Issues

- This is part of adding the `sql_search()` operation to the OpenSearchDocumentStore, which turned out to be an `SQLRetriever`. Later, we realised that we also need to add a `MetadataRetriever`.

### Proposed Changes:

- Adding a `MetadataRetriever`. This allows multi-field query searches over metadata, ranking the returned metadata fields using the Jaccard similarity. It also has two search methods, strict and fuzzy and option to boost exact matches scores.

### How did you test it?

- new integration tests + CI

### Checklist

### DIFF
--- a/integrations/opensearch/src/haystack_integrations/components/retrievers/opensearch/metadata_retriever.py
+++ b/integrations/opensearch/src/haystack_integrations/components/retrievers/opensearch/metadata_retriever.py
@@ -17,21 +17,18 @@ class OpenSearchMetadataRetriever:
     """
     Retrieves and ranks metadata from documents stored in an OpenSearchDocumentStore.
 
-    It searches specified metadata fields for matches to a given query, ranks the results
-    based on relevance using Jaccard similarity, and returns the top-k results containing
-    only the specified metadata fields.
+    It searches specified metadata fields for matches to a given query, ranks the results based on relevance using
+    Jaccard similarity, and returns the top-k results containing only the specified metadata fields. Additionally, it
+    adds a boost to the score of exact matches.
 
-    The search is designed for metadata fields whose values are **text** (strings). It uses
-    prefix, wildcard and fuzzy matching to find candidate documents; these query
-    types operate only on text/keyword fields in OpenSearch.
+    The search is designed for metadata fields whose values are **text** (strings). It uses prefix, wildcard and fuzzy
+    matching to find candidate documents; these query types operate only on text/keyword fields in OpenSearch.
 
-    Metadata fields with **non-string types** (integers, floats, booleans, lists of non-strings)
-    are indexed by OpenSearch as numeric, boolean, or array types. Those field types do not
-    support prefix, wildcard, or full-text match queries, so documents are typically not
-    found when you search only by such fields.
+    Metadata fields with **non-string types** (integers, floats, booleans, lists of non-strings) are indexed by
+    OpenSearch as numeric, boolean, or array types. Those field types do not support prefix, wildcard, or full-text
+    match queries, so documents are typically not found when you search only by such fields.
 
-    **Mixed types** in the same metadata field (e.g. a list containing both strings and
-    numbers) are not supported.
+    **Mixed types** in the same metadata field (e.g. a list containing both strings and numbers) are not supported.
 
     Must be connected to the OpenSearchDocumentStore to run.
 
@@ -92,6 +89,8 @@ class OpenSearchMetadataRetriever:
         fuzziness: int | Literal["AUTO"] = 2,
         prefix_length: int = 0,
         max_expansions: int = 200,
+        tie_breaker: float = 0.7,
+        jaccard_n: int = 3,
         raise_on_failure: bool = True,
     ):
         """
@@ -104,8 +103,8 @@ class OpenSearchMetadataRetriever:
             Default is 0.6. It's used on both "strict" and "fuzzy" modes and applied after the search executes.
         :param mode: Search mode. "strict" uses prefix and wildcard matching,
             "fuzzy" uses fuzzy matching with dis_max queries. Default is "fuzzy".
-            In both modes, results are scored using Jaccard similarity (n-gram based, n=3)
-            computed server-side via a Painless script.
+            In both modes, results are scored using Jaccard similarity (n-gram based)
+            computed server-side via a Painless script; n is controlled by jaccard_n.
         :param fuzziness: Maximum allowed Damerau-Levenshtein distance (edit distance) for fuzzy matching.
             Accepts an integer (e.g., 0, 1, 2) or "AUTO" which chooses based on term length.
             Default is 2. Only applies when mode is "fuzzy".
@@ -113,6 +112,9 @@ class OpenSearchMetadataRetriever:
             Default is 0 (no prefix requirement). Only applies when mode is "fuzzy".
         :param max_expansions: Maximum number of term variations the fuzzy query can generate.
             Default is 200. Only applies when mode is "fuzzy".
+        :param tie_breaker: Weight (0..1) for other matching clauses in the dis_max query.
+            Boosts documents that match multiple clauses. Default is 0.7. Only applies when mode is "fuzzy".
+        :param jaccard_n: N-gram size for Jaccard similarity scoring. Default 3; larger n favors longer token matches.
         :param raise_on_failure:
             If `True`, raises an exception if the API call fails.
             If `False`, logs a warning and returns an empty list.
@@ -135,6 +137,8 @@ class OpenSearchMetadataRetriever:
         self._fuzziness = fuzziness
         self._prefix_length = prefix_length
         self._max_expansions = max_expansions
+        self._tie_breaker = tie_breaker
+        self._jaccard_n = jaccard_n
         self._raise_on_failure = raise_on_failure
 
     def to_dict(self) -> dict[str, Any]:
@@ -154,6 +158,8 @@ class OpenSearchMetadataRetriever:
             fuzziness=self._fuzziness,
             prefix_length=self._prefix_length,
             max_expansions=self._max_expansions,
+            tie_breaker=self._tie_breaker,
+            jaccard_n=self._jaccard_n,
             raise_on_failure=self._raise_on_failure,
         )
 
@@ -186,6 +192,8 @@ class OpenSearchMetadataRetriever:
         fuzziness: int | Literal["AUTO"] | None = None,
         prefix_length: int | None = None,
         max_expansions: int | None = None,
+        tie_breaker: float | None = None,
+        jaccard_n: int | None = None,
         filters: dict[str, Any] | None = None,
     ) -> dict[str, list[dict[str, Any]]]:
         """
@@ -205,8 +213,7 @@ class OpenSearchMetadataRetriever:
             If not provided, the exact_match_weight provided in `__init__` is used.
         :param mode: Search mode. "strict" uses prefix and wildcard matching,
             "fuzzy" uses fuzzy matching with dis_max queries.
-            In both modes, results are scored using Jaccard similarity (n-gram based, n=3)
-            computed server-side via a Painless script.
+            In both modes, results are scored using Jaccard similarity (n-gram based) via a Painless script.
             If not provided, the mode provided in `__init__` is used.
         :param fuzziness: Maximum allowed Damerau-Levenshtein distance (edit distance) for fuzzy matching.
             Accepts an integer (e.g., 0, 1, 2) or "AUTO" which chooses based on term length.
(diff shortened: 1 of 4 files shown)


integrations/opensearch/src/haystack_integrations/components/retrievers/opensearch/metadata_retriever.py
```diff
@@ -0,0 +1,247 @@
...
+    only the specified metadata fields.
+
+    Must be connected to the OpenSearchDocumentStore to run.
+    """
+
+    def __init__(
+        self,
+        *,
+        document_store: OpenSearchDocumentStore,
+        fields: list[str],
```
sjrl: I think we have used the name `metadata_fields` elsewhere so let's reuse that here

integrations/opensearch/src/haystack_integrations/components/retrievers/opensearch/metadata_retriever.py
```diff
@@ -0,0 +1,247 @@
...
+        top_k: int = 20,
+        exact_match_weight: float = 0.6,
```
sjrl, line 87: Just based on the naming is this factor only used when the mode is set to `fuzzy`? If so I'd mention that in the doc strings or mention that it's always used if that's the case

integrations/opensearch/src/haystack_integrations/components/retrievers/opensearch/metadata_retriever.py
```diff
@@ -0,0 +1,247 @@
...
+        data["init_parameters"]["document_store"] = OpenSearchDocumentStore.from_dict(
+            data["init_parameters"]["document_store"]
+        )
+        return default_from_dict(cls, data)
+
+    @component.output_types(metadata=list[dict[str, Any]])
+    def run(
+        self,
+        query: str,
+        document_store: OpenSearchDocumentStore | None = None,
```
sjrl, line 187: Since there is quite a few variables perhaps we could enforce keyword args only after query?
```suggestion
    def run(
        self,
        query: str,
        *,
        document_store: OpenSearchDocumentStore | None = None,
```

integrations/opensearch/src/haystack_integrations/components/retrievers/opensearch/metadata_retriever.py
```diff
@@ -0,0 +1,247 @@
...
+
+        Example:
+            ```python
+            retriever = OpenSearchMetadataRetriever(
+                document_store=store,
+                fields=["category", "status", "priority"]
+            )
+            result = retriever.run(query="Python, active")
+            # Returns: {"metadata": [{"category": "Python", "status": "active", ...}, ...]}
+            ```
```
sjrl, line 253: Thanks for adding an example! Could you also provide the example of a document that would need to be present in the db for this result to be returned?

integrations/opensearch/src/haystack_integrations/document_stores/opensearch/document_store.py
```diff
@@ -1010,6 +1011,352 @@ async def _bm25_retrieval_async(
...
+        query_part: str, fields: list[str], mode: Literal["strict", "fuzzy"]
+    ) -> dict[str, Any]:
+        """
+        Build an OpenSearch query for metadata search.
+
+        :param query_part: The cleaned query part to search for.
+        :param fields: List of metadata field names to search within.
+        :param mode: Search mode. "strict" uses prefix and wildcard matching,
+            "fuzzy" uses fuzzy matching with dis_max queries.
+        :returns: OpenSearch query dictionary.
```
sjrl: Could we mention when the Jaccard similarity script is being used? 

I originally thought it was being used for the fuzzy matching but that doesn't seem to be the full story.

integrations/opensearch/src/haystack_integrations/components/retrievers/opensearch/metadata_retriever.py
sjrl: Yeah double checking elsewhere it seems for public methods we decided to use `metadata_fields` whereas for private methods I do see we use `fields`. So for all public methods let's reuse `metadata_fields`

integrations/opensearch/src/haystack_integrations/document_stores/opensearch/document_store.py
```diff
@@ -1010,6 +1011,352 @@ async def _bm25_retrieval_async(
...
+                            "filter": [{"exists": {"field": field}} for field in fields],
+                            "should": should_clauses,
+                            "minimum_should_match": 1,
+                        }
+                    },
+                    "script": {
+                        "lang": "painless",
+                        "source": METADATA_SEARCH_JACCARD_SCRIPT,
+                        "params": {"field": fields[0], "q": query_part, "n": 3},
+                    },
```
sjrl, line 1069: Re-reading the docstrings of the new component I see you mentioned that the Jaccard script is being used for ranking purposes. 

Do you know why we need the jaccard ranking for strict mode but not for fuzzy mode?

integrations/opensearch/src/haystack_integrations/document_stores/opensearch/document_store.py
```diff
@@ -1010,6 +1011,352 @@ async def _bm25_retrieval_async(
...
+            for field in fields:
+                dis_max_queries.append(
+                    {
+                        "match": {
+                            field: {
+                                "query": query_part,
+                                "fuzziness": 2,
+                                "prefix_length": 0,
+                                "max_expansions": 200,
+                            }
```
sjrl, line 1084: In our OpenSearch BM25Retriever we expose the fuzziness option explicitly. e.g. `fuzziness: int | str = "AUTO",` do you think it's worth exposing that option in this component as well instead of hard-coding the fuzziness param to have to be 2?

integrations/opensearch/src/haystack_integrations/document_stores/opensearch/document_store.py
```diff
@@ -1010,6 +1011,352 @@ async def _bm25_retrieval_async(
...
+                    self._boost_exact_matches(hits, query_part_clean, fields, exact_match_weight)
+
+                hit_list.extend(hits)
+            except Exception as e:
+                msg = f"Failed to execute metadata search in OpenSearch: {e!s}"
+                raise DocumentStoreError(msg) from e
+
+        # Add multi-field exact match boosting
+        if exact_match_weight > 0:
+            self._apply_multi_field_boosting(hit_list, query, fields, exact_match_weight)
```
sjrl: I'm a little confused why are boosting twice for exact matches? I'm not really sure I follow the logic or understand how the value of `exact_match_weight` works. Is it meant to be bounded between 0 and 1 or is it valid to be larger than 1?

integrations/opensearch/src/haystack_integrations/components/retrievers/opensearch/metadata_retriever.py
```diff
@@ -0,0 +1,247 @@
...
+            If not provided, the fields provided in `__init__` are used.
+        :param top_k: Maximum number of top results to return based on relevance.
+            If not provided, the top_k provided in `__init__` is used.
+        :param exact_match_weight: Weight to boost the score of exact matches in metadata fields.
+            If not provided, the exact_match_weight provided in `__init__` is used.
+        :param mode: Search mode. "strict" uses prefix and wildcard matching,
+            "fuzzy" uses fuzzy matching with dis_max queries.
+            If not provided, the mode provided in `__init__` is used.
+        :param filters: Additional filters to apply to the search query.
+        :returns: A dictionary containing the top-k retrieved metadata results.
```
sjrl: Could we better explain the structure of the metadata results? Does it provide both the metadata key and the metadata value?

integrations/opensearch/src/haystack_integrations/components/retrievers/opensearch/metadata_retriever.py
```diff
@@ -0,0 +1,247 @@
...
+        query: str,
+        document_store: OpenSearchDocumentStore | None = None,
+        fields: list[str] | None = None,
+        top_k: int | None = None,
+        exact_match_weight: float | None = None,
+        mode: Literal["strict", "fuzzy"] | None = None,
+        filters: dict[str, Any] | None = None,
+    ) -> dict[str, list[dict[str, Any]]]:
+        """
+        Execute a search query against the metadata fields of documents stored in the Document Store.
```
sjrl, line 200: Also a more general question does this search work only for metadata fields whose values are of type string or does this also work for other types like ints, lists, etc?

integrations/opensearch/src/haystack_integrations/components/retrievers/opensearch/metadata_retriever.py
davidsbatista: yes, agree 👍🏽

integrations/opensearch/src/haystack_integrations/components/retrievers/opensearch/metadata_retriever.py
davidsbatista: It's used on both "strict" and "fuzzy" modes, and it's only applied after the search executes, it's a post-processing step. I will add this info to the docstring.

integrations/opensearch/src/haystack_integrations/components/retrievers/opensearch/metadata_retriever.py
davidsbatista: I've added an example of such a Document.

integrations/opensearch/src/haystack_integrations/document_stores/opensearch/document_store.py
davidsbatista: So, the mode only influences how the query is built, prefix/wildcard or fuzzy matching.

The Jaccard similarity is used to compute the scoring in both modes. 

see: `_build_metadata_search_query()`, both modes use the Jaccard Similarity script that runs on the server.

integrations/opensearch/src/haystack_integrations/document_stores/opensearch/document_store.py
davidsbatista: it's being used on both, see my comment above

integrations/opensearch/src/haystack_integrations/document_stores/opensearch/document_store.py
sjrl: Ahh I see thanks for the explanation I missed the full construction of the fuzzy query

integrations/opensearch/src/haystack_integrations/document_stores/opensearch/document_store.py
davidsbatista: We are boosting twice yes, once for per-query-part and 

1. For each query part (seperated by comma) if it appears in any metadata field 
```python
(query_part.lower() in str(hit_source[field]).lower()) 
```
we add exact_match_weight to the score. It's applied for each query part

2. a second time, multi-field boosting. We count how many query parts match across all fields, and if matched > 1, adds: exact_match_weight * (matched - 1) to the score

### Example:

Given the query: "Python, active" and if we have a document with the following metadata: 

```{"category": "Python", "status": "active"}```

- per-part boost: exact_match_weight for "Python" + exact_match_weight for "active" = 2 * exact_match_weight
- multi-field boost: exact_match_weight * (2 - 1) = exact_match_weight
- total boost: 3 * exact_match_weight

The `exact_match_weight` is not bounded. I added the check to apply the boosting code only when the `exact_match_weight > 0` since I don't think penalizing exact matches is a use case. We can add a check that `exact_match_weight > 0` during init,.

integrations/opensearch/src/haystack_integrations/document_stores/opensearch/document_store.py
davidsbatista: we can expose that parameter and maybe also the other two which I think work together with `fuziness` (need to confirm)

- prefix_length=0, 
- max_expansions=200

others we can consider:

- n=3  n-gram  for jaccard similarity
- size=1000 max number of hits to retrieve
- tie_breaker=0.7 dis_max tie breaker

WDYT? (exposing them but keeping the default values)

integrations/opensearch/src/haystack_integrations/document_stores/opensearch/document_store.py
sjrl: Thanks for the explanation but I still find this unintuitive especially if I'm trying to judge what value to provide for `exact_match_weight`. 

Could we just do `multi-field boost` only with the updated equation to be `exact_match_weight * num_matched` to give a final total boost of `2* exact_match_weight`? 

This would change things slightly but I think it would already be easier to explain to end-users. Currently I'm unsure what value to provide to `exact_match_weight` to make sure that all results with exact matches are at the top of my final ranked list.

And ideally I would have liked to say something like set `exact_match_weight=0` to use only Jaccard Similarity and use `exact_match_weight=1` for only exact match ranking. But this would require a different rework of the boosting.

WDYT?

davidsbatista: solved offline

integrations/opensearch/src/haystack_integrations/components/retrievers/opensearch/metadata_retriever.py
davidsbatista: I've added a full example to the class docstring, as present in the other retrievers

integrations/opensearch/src/haystack_integrations/components/retrievers/opensearch/metadata_retriever.py
davidsbatista: It won't work, the queries are based on prefix/wildcard or fuzzy matching, and those are text/keyword query types, meant for string-like data and rely on text indexes.

I've added a more detailed description/disclaimer at the top of the MetadataRetriever class to make users aware of this.

sjrl (approved):
Thanks for exposing more params. Looks good!

PR #2772, opened 2026-01-22
feat: Add operations to PineConeDocumentStore

### LINKED ISSUE #2641: add the following operations to `PineConeDocumentStore`
```python
def count_documents_by_filter(filter: dict) -> int
```
- applies a filter and counts the documents that matched it.

---

```python
def count_unique_metadata_by_filter(filters: dict[str, Any], metadata_fields: list[str]) -> dict[str, int]` 
```
- applies a filter selecting documents and counts the unique values for each meta field of the matched documents

---

```python
def get_metadata_fields_info() -> dict[str, dict]
```
- returns the meta fields and the corresponding types, it should return the same format as in the `OpenSearchDocumentStore`

```python
{
    'content': {'type': 'text'},
    'category': {'type': 'keyword'},
    'status': {'type': 'keyword'},
    'priority': {'type': 'long'},
}
```

---

```python
def get_metadata_field_min_max(metadata_field: str) -> dict[str, Any]
```
- for a given metadata field, find its max and min value

---

```python
def get_metadata_field_unique_values(metadata_field: str, search_term: str | None, from_: int, size: int) -> tuple[list[str], int]:
```
- retrieves unique values for a field matching a search_term or all possible values if no search term is given

---

```python
def query_sql(query: str)
```
- (optional) issue an SQL query for the doc stores that support it

GauravBansal29:
### Related Issues

- fixes #2641 

### Proposed Changes:

Implemented five new metadata query and filtering operations for `PineconeDocumentStore`:

#### Methods Added:

1. **`count_documents_by_filter(filters)`** - Count documents matching filter criteria
2. **`count_unique_metadata_by_filter(filters, metadata_fields)`** - Count unique values across specified metadata fields
3. **`get_metadata_fields_info()`** - Infer metadata field types by sampling documents
4. **`get_metadata_field_min_max(metadata_field)`** - Get min/max values for numeric metadata fields
5. **`get_metadata_field_unique_values(metadata_field, search_term, from_, size)`** - Retrieve unique values with pagination and search support

All methods include both synchronous and asynchronous versions (e.g., `count_documents_by_filter_async`).

#### Implementation Details:

- **Client-side aggregation**: Pinecone lacks native aggregation APIs, so all operations fetch documents (up to `TOP_K_LIMIT` of 1000) and perform aggregations in Python
- **Schema inference**: `get_metadata_fields_info()` samples documents and infers types with the following mappings:
  - `bool` → `"boolean"`
  - `int`/`float` → `"long"`
  - `str` → `"keyword"`
  - `list` elements → inferred from first element type
  - Mixed types → defaults to `"keyword"` with warning
- **Pagination**: `get_metadata_field_unique_values()` supports `from_` and `size` parameters
- **Search filtering**: Case-insensitive substring matching for unique value searches
- **Warning logs**: Users are informed when hitting Pinecone's `TOP_K_LIMIT`, indicating potential data incompleteness

#### Design Decisions:

- Followed OpenSearch document store as reference for output format and type mappings
- Used existing `filter_documents()` method to leverage Pinecone's filtering capabilities
- All methods handle edge cases (empty documents, non-existent fields, non-numeric values)
- Type inference checks for `bool` first (before `int`) to avoid false positives since `bool` is a subclass of `int` in Python

#### Integration Tests:
- **9 sync tests** in `test_document_store.py`:
  - Basic filtering and counting operations
  - Empty document store handling
  - Mixed types detection in metadata
  - Numeric min/max with error handling for non-numeric fields
  - Unique values with pagination and list metadata handling
  
- **5 async tests** in `test_document_store_async.py`:
  - Async versions of all core operations
  - Pagination and search term filtering
  
All tests run against a real Pinecone instance (marked with `@pytest.mark.integration`) and passed successfully in CI.

#### Test Coverage:
- ✅ Happy path scenarios
- ✅ Edge cases (empty stores, missing fields, non-numeric values)
- ✅ Type inference with mixed types
- ✅ List metadata handling
- ✅ Pagination boundaries
- ✅ Search term filtering

### Notes for the reviewer

1. **Client-side limitations**: Due to Pinecone's `TOP_K_LIMIT` of 1000 documents, these methods may return incomplete results for large collections. Warning logs inform users when this limit is reached.

2. **Type inference approach**: `get_metadata_fields_info()` collects all observed types for each field across sampled documents and detects inconsistencies. If mixed types are found, it logs a warning and defaults to `"keyword"`.

3. **No `query_sql` method**: Unlike some document stores, Pinecone doesn't support SQL queries, so this optional method from the issue was not implemented.

4. **Consistency with codebase**: 
   - Followed existing Pinecone test patterns (integration tests with fixtures)
   - Used same logging conventions and error handling patterns
   - Matched type mappings used in OpenSearch integration

### Checklist

- [x] I've used one of the [conventional commit types](https://www.conventionalcommits.org/en/v1.0.0/) for my PR title: `feat:`, `feat:`, `build:`, `chore:`, `ci:`, `docs:`, `style:`, `refactor:`, `perf:`, `test:`.

### DIFF
--- a/integrations/amazon_bedrock/CHANGELOG.md
+++ b/integrations/amazon_bedrock/CHANGELOG.md
@@ -1,5 +1,18 @@
 # Changelog
 
+## [integrations/amazon_bedrock-v6.3.0] - 2026-01-28
+
+### 🌀 Miscellaneous
+
+- Feat: Bedrock - support images in tool results (#2783)
+
+## [integrations/amazon_bedrock-v6.2.1] - 2026-01-15
+
+### 🐛 Bug Fixes
+
+- None value handling of flattened generation kwargs for AmazonBedrockChatGenerator (#2752)
+
+
 ## [integrations/amazon_bedrock-v6.2.0] - 2026-01-13
 
 ### 🚀 Features


integrations/pinecone/src/haystack_integrations/document_stores/pinecone/document_store.py
```diff
@@ -740,3 +740,400 @@ def _prepare_documents_for_writing(
...
+        Asynchronously counts unique values for each specified metadata field in documents matching the filters.
+
+        :param filters: The filters to apply to select documents.
+        :param metadata_fields: List of metadata field names to count unique values for.
+        :returns: Dictionary mapping field names to counts of unique values.
+        """
+        documents = await self.filter_documents_async(filters=filters)
+
+        result = {}
+        for field in metadata_fields:
```
davidsbatista: We should refactor the duplicated logic in both sync and async functions, avoiding duplicate code. That logic can be extracted into helper functions, which can then be used by both synchronous and asynchronous methods.

integrations/pinecone/src/haystack_integrations/document_stores/pinecone/document_store.py
```diff
@@ -740,3 +740,400 @@ def _prepare_documents_for_writing(
...
+
+        field_types: dict[str, dict[str, str]] = {}
+
+        if any(doc.content is not None for doc in documents):
+            field_types["content"] = {"type": "text"}
+
+        # Collect all field values to infer types accurately
+        field_samples: dict[str, set[str]] = {}
+
+        for doc in documents:
```
davidsbatista, line 876: Same comment as above regarding duplicated logic.

integrations/pinecone/src/haystack_integrations/document_stores/pinecone/document_store.py
```diff
@@ -740,3 +740,400 @@ def _prepare_documents_for_writing(
...
+            logger.info(
+                f"Schema inference based on {TOP_K_LIMIT} documents (Pinecone's query limit). "
+                f"If you have more documents with different metadata fields, they won't be reflected here."
+            )
+
+        return field_types
+
+    def get_metadata_field_min_max(self, metadata_field: str) -> dict[str, Any]:
+        """
+        Returns the minimum and maximum values for a numeric metadata field.
```
davidsbatista: This should also work for other types: bool and keyword

integrations/pinecone/src/haystack_integrations/document_stores/pinecone/document_store.py
GauravBansal29: Thanks for pointing this out @davidsbatista ; I have made the relevant code changes to use helper methods to avoid code duplication

integrations/pinecone/src/haystack_integrations/document_stores/pinecone/document_store.py
GauravBansal29: Thanks for pointing out. Addressed this with addition of helper methods.

integrations/pinecone/src/haystack_integrations/document_stores/pinecone/document_store.py
GauravBansal29: I’ve added support for boolean and keyword types too. For booleans, I assumed true > false and implemented it that way - not sure of the expected behaviour though. Please let me know if this needs any changes. Thanks.

PR #2782, opened 2026-01-26
feat: Update for PaddleOCR-VL-1.5 interface

Bobholamovic:
### Proposed Changes:

Support PaddleOCR-VL-1.5 (which is backward compatible for PaddleOCR-VL v1 models).

### How did you test it?

unit tests and integration tests.

### DIFF
--- a/integrations/amazon_bedrock/CHANGELOG.md
+++ b/integrations/amazon_bedrock/CHANGELOG.md
@@ -1,5 +1,18 @@
 # Changelog
 
+## [integrations/amazon_bedrock-v6.3.0] - 2026-01-28
+
+### 🌀 Miscellaneous
+
+- Feat: Bedrock - support images in tool results (#2783)
+
+## [integrations/amazon_bedrock-v6.2.1] - 2026-01-15
+
+### 🐛 Bug Fixes
+
+- None value handling of flattened generation kwargs for AmazonBedrockChatGenerator (#2752)
+
+
 ## [integrations/amazon_bedrock-v6.2.0] - 2026-01-13
 
 ### 🚀 Features
--- a/integrations/amazon_bedrock/pyproject.toml
+++ b/integrations/amazon_bedrock/pyproject.toml
@@ -22,7 +22,7 @@ classifiers = [
   "Programming Language :: Python :: Implementation :: CPython",
   "Programming Language :: Python :: Implementation :: PyPy",
 ]
-dependencies = ["haystack-ai>=2.22.0", "boto3>=1.28.57", "aioboto3>=14.0.0"]
+dependencies = ["haystack-ai>=2.23.0", "boto3>=1.28.57", "aioboto3>=14.0.0"]
 
 [project.urls]
 Documentation = "https://github.com/deepset-ai/haystack-core-integrations/tree/main/integrations/amazon_bedrock#readme"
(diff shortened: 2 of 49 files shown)


integrations/paddleocr/src/haystack_integrations/components/converters/paddleocr/paddleocr_vl_document_converter.py
```diff
@@ -146,23 +146,33 @@ def __init__(
         api_url: str,
         access_token: Secret = Secret.from_env_var("AISTUDIO_ACCESS_TOKEN"),
         file_type: FileTypeInput = None,
-        use_doc_orientation_classify: bool | None = None,
-        use_doc_unwarping: bool | None = None,
+        use_doc_orientation_classify: bool | None = False,
```
anakin87, line 149: just a thought...

Would it make sense to use more informative type hints, like
`use_doc_orientation_classify: bool = False,`?

It seems that `None` and `False` would be equivalent in this case. Right?

I'd do the same for all these parameters, if it makes sense.

Happy to discuss, as you are more familiar with the API.

integrations/paddleocr/src/haystack_integrations/components/converters/paddleocr/paddleocr_vl_document_converter.py
```diff
@@ -204,9 +217,10 @@ def __init__(
             dict with page-specific values.
         :param layout_merge_bboxes_mode:
             Layout merge bounding boxes mode. Can be a string or a dict.
+        :param layout_shape_mode:
+            Layout shape mode.
         :param prompt_label:
-            Prompt type for the VLM. Possible values are "ocr", "formula",
-            "table", and "chart".
+            Prompt type for the VLM.
```
anakin87: I think that previous info was helpful. Is it outdated?

integrations/paddleocr/src/haystack_integrations/components/converters/paddleocr/paddleocr_vl_document_converter.py
Bobholamovic: This is a somewhat complex issue to explain, as it involves some historical design decisions in PaddleOCR.

In short, within our API semantics, `None` means "use the server-side default configuration". These defaults are usually tuned by us and are intended to work well for most use cases.

However, in this case, we previously made a poor design choice by enabling image preprocessing by default, including `use_doc_orientation_classify` and `use_doc_unwarping`. These features are mainly useful for handling distorted or rotated images, but for normal images they can actually hurt accuracy and also slow down inference.

Since such distorted or heavily rotated images are relatively rare, in order to ensure accuracy and performance in the majority of cases, we currently have an unfortunate requirement: if the client does not need image preprocessing (which is true for most scenarios), these two parameters should be explicitly set to `False`. This is far from ideal, but due to backward compatibility concerns, we have not yet found a better solution. A proper fix may be introduced in a future major release.

In the context of this code, `bool | None` represents the supported types of this parameter: `bool` indicates an explicit user preference, while `None` means "I don’t care; use the default best configuration" (even though, for these two parameters, the default is not actually the best). Therefore, in my personal opinion, although this design may seem a bit awkward, it is arguably reasonable given the current historical constraints.

integrations/paddleocr/src/haystack_integrations/components/converters/paddleocr/paddleocr_vl_document_converter.py
Bobholamovic: This was mistakenly deleted. I've updated the description.

integrations/paddleocr/src/haystack_integrations/components/converters/paddleocr/paddleocr_vl_document_converter.py
anakin87: Ok, I understand. Thanks for the deep explanation.
Let's keep things as they are.

anakin87 (approved):

Issue #2776, opened 2026-01-22
Add prompt caching for AmazonBedrockGenerator and AmazonBedrockChatGenerator


deep-rloebbert:
**Is your feature request related to a problem? Please describe.**
Yes. When using `AmazonBedrockGenerator` or `AmazonBedrockChatGenerator` with large, static contexts (such as long system prompts or few-shot examples), the model re-processes the input tokens for every request. This results in unnecessarily high costs and increased latency.

Amazon Bedrock supports **Prompt Caching** to reuse processed tokens, but currently, there is no mechanism in the Haystack components to pass the required caching parameters to the underlying Bedrock API.

**Describe the solution you'd like**
I would like the ability to specify prompt caching parameters within the Haystack Bedrock components. Best case a simple bool toggle.

Ideally, this would be handled via metadata in `ChatMessage` (for Chat Generators) or call parameters. When the component constructs the API payload, it should detect these flags and inject the specific cache control structure required by Bedrock into the request body (e.g., adding `cache_control` or `cachePoint` fields to the message blocks).

**Describe alternatives you've considered**
Spending too much money.

**Additional context**

* **AWS Documentation:** [[Amazon Bedrock Prompt Caching](https://docs.aws.amazon.com/bedrock/latest/userguide/prompt-caching.html#prompt-caching-simplified)](https://docs.aws.amazon.com/bedrock/latest/userguide/prompt-caching.html#prompt-caching-simplified)
* **Benefit:** caching can significantly reduce inference costs (up to 90% for cached tokens) and latency for supported models.

PR #2796, opened 2026-01-29
feat: Bedrock - support prompt caching

anakin87:
### Related Issues

- part of #2776 (to complete this, I also paln to add a small section about Prompt caching in docs)

### Proposed Changes:
- allow setting cache points in messages
`ChatMessage.from_user("long message..."), meta={"cachePoint": {"type": "default"}}`
  - aligns with [Bedrock docs ](https://docs.aws.amazon.com/bedrock/latest/userguide/prompt-caching.html#prompt-caching-get-started)
  - The documentation is ambiguous about multiple cache points per message. Implementing this would be hard/hacky within the current `ChatMessage` abstraction, so this PR only supports one cache point per message.
  - we also discussed in https://github.com/deepset-ai/haystack-core-integrations/issues/2776#issuecomment-3812028128 about adding a global parameter to the Chat Generator to cache all user/system messages. However, Bedrock allows a maximum of 4 cache points per request, so per-message cache points are more transparent and allow finer control.
- add a `tools_cachepoint_config` init parameter to the ChatGenerator to enable caching for tools
  - Bedrock may support more granular caching for tools, but it is hard to implement while respecting Haystack API, so I adopted this simpler solution.

### How did you test it?
CI, new unit and integration tests.

### Checklist

### DIFF
--- a/integrations/amazon_bedrock/src/haystack_integrations/components/generators/amazon_bedrock/chat/chat_generator.py
+++ b/integrations/amazon_bedrock/src/haystack_integrations/components/generators/amazon_bedrock/chat/chat_generator.py
@@ -143,6 +143,12 @@ class AmazonBedrockChatGenerator:
       and `aws_region_name` as environment variables or pass them as
      [Secret](https://docs.haystack.deepset.ai/docs/secret-management) arguments. Make sure the region you set
     supports Amazon Bedrock.
+
+    This component supports prompt caching. You can use the `tools_cachepoint_config` parameter to configure the cache
+    point for tools.
+    To cache messages, you can use the `cachePoint` key in `ChatMessage.meta` attribute.
+    Example: `ChatMessage.from_user("Long message...", meta={"cachePoint": {"type": "default"}})`
+    For more information, see the [Amazon Bedrock documentation](https://docs.aws.amazon.com/bedrock/latest/userguide/prompt-caching.html).
     """
 
     def __init__(
@@ -206,6 +212,7 @@ class AmazonBedrockChatGenerator:
         :param tools_cachepoint_config: Optional configuration to use prompt caching for tools.
             The dictionary must match the
             [CachePointBlock schema](https://docs.aws.amazon.com/bedrock/latest/APIReference/API_runtime_CachePointBlock.html).
+            Example: `{"type": "default", "ttl": "5m"}`
 
 
         :raises ValueError: If the model name is empty or None.
@@ -230,9 +237,9 @@ class AmazonBedrockChatGenerator:
         _validate_guardrail_config(guardrail_config=guardrail_config, streaming=streaming_callback is not None)
         self.guardrail_config = guardrail_config
 
-        if tools_cachepoint_config:
-            _validate_and_format_cache_point(tools_cachepoint_config)
-        self.tools_cachepoint_config = tools_cachepoint_config
+        self.tools_cachepoint_config = (
+            _validate_and_format_cache_point(tools_cachepoint_config) if tools_cachepoint_config else None
+        )
 
         def resolve_secret(secret: Secret | None) -> str | None:
             return secret.resolve_value() if secret else None
--- a/integrations/amazon_bedrock/src/haystack_integrations/components/generators/amazon_bedrock/chat/utils.py
+++ b/integrations/amazon_bedrock/src/haystack_integrations/components/generators/amazon_bedrock/chat/utils.py
@@ -41,7 +41,7 @@ FINISH_REASON_MAPPING: dict[str, FinishReason] = {
 
 # Haystack to Bedrock util methods
 def _format_tools(
-    tools: list[Tool] | None = None, tools_cachepoint_config: dict[str, str] | None = None
+    tools: list[Tool] | None = None, tools_cachepoint_config: dict[str, dict[str, str]] | None = None
 ) -> dict[str, Any] | None:
     """
     Format Haystack Tool(s) to Amazon Bedrock toolConfig format.
@@ -53,7 +53,7 @@ def _format_tools(
     if not tools:
         return None
 
-    tool_specs = []
+    tool_specs: list[dict[str, Any]] = []
     for tool in tools:
         tool_specs.append(
             {"toolSpec": {"name": tool.name, "description": tool.description, "inputSchema": {"json": tool.parameters}}}
@@ -279,7 +279,8 @@ def _validate_and_format_cache_point(cache_point: dict[str, str] | None) -> dict
         err_msg = "Cache point can only contain 'type' and 'ttl' keys."
         raise ValueError(err_msg)
     if "ttl" in cache_point and cache_point["ttl"] not in ("5m", "1h"):
-        raise ValueError("Cache point 'ttl' must be one of '5m', '1h'.")
+        err_msg = "Cache point 'ttl' must be one of '5m', '1h'."
+        raise ValueError(err_msg)
 
     return {"cachePoint": cache_point}
 
@@ -297,7 +298,7 @@ def _format_messages(messages: list[ChatMessage]) -> tuple[list[dict[str, Any]],
               non_system_messages is a list of properly formatted message dictionaries.
     """
     # Separate system messages, tool calls, and tool results
-    system_prompts = []
+    system_prompts: list[dict[str, Any]] = []
     bedrock_formatted_messages = []
     for msg in messages:
         cache_point = _validate_and_format_cache_point(msg.meta.get("cachePoint"))


anakin87:
@sjrl good question...

According to [docs](https://docs.aws.amazon.com/bedrock/latest/userguide/prompt-caching.html), prompt caching is applicable to long messages (>1K tokens, depending on the model).

In general, in agentic loops, I would expect we want to cache tools, system message and the first user message. Assistant messages with tool calls would vary, so they do not seem like good candidates for caching.

The current Agent implementation allows passing `messages` with `meta` and preserves them during the loop. So I'd say we cover the most common use case. Am I overlooking something?

integrations/amazon_bedrock/tests/test_chat_generator.py
```diff
@@ -945,6 +951,28 @@ def test_live_run_with_guardrail(self, streaming_callback):
         assert "trace" in results["replies"][0].meta
         assert "guardrail" in results["replies"][0].meta["trace"]
 
+    @pytest.mark.parametrize("streaming_callback", [None, print_streaming_chunk])
+    @pytest.mark.parametrize("model_name", MODELS_TO_TEST_WITH_PROMPT_CACHING)
+    def test_prompt_caching_live_run_with_user_message(self, model_name, streaming_callback):
+        generator = AmazonBedrockChatGenerator(model=model_name, streaming_callback=streaming_callback)
+
+        system_message = ChatMessage.from_system("Always respond with: 'Life is beatiful' (and nothing else).")
```
julian-risch:
```suggestion
        system_message = ChatMessage.from_system("Always respond with: 'Life is beautiful' (and nothing else).")
```

integrations/amazon_bedrock/src/haystack_integrations/components/generators/amazon_bedrock/chat/utils.py
```diff
@@ -251,6 +259,29 @@ def _format_text_image_message(message: ChatMessage) -> dict[str, Any]:
     return {"role": message.role.value, "content": bedrock_content_blocks}
 
 
+def _validate_and_format_cache_point(cache_point: dict[str, str] | None) -> dict[str, Any] | None:
```
julian-risch: I believe the return type annotation can be narrowed: 
```suggestion
def _validate_and_format_cache_point(cache_point: dict[str, str] | None) -> dict[str, dict[str, str]] | None:
```

integrations/amazon_bedrock/src/haystack_integrations/components/generators/amazon_bedrock/chat/utils.py
```diff
@@ -40,7 +40,9 @@
 
 
 # Haystack to Bedrock util methods
-def _format_tools(tools: list[Tool] | None = None) -> dict[str, Any] | None:
+def _format_tools(
+    tools: list[Tool] | None = None, tools_cachepoint_config: dict[str, Any] | None = None
```
julian-risch: I believe the type of tools_cachepoint_config can be narrowed:
```suggestion
    tools: list[Tool] | None = None, tools_cachepoint_config: dict[str, str] | None = None
```

integrations/amazon_bedrock/src/haystack_integrations/components/generators/amazon_bedrock/chat/utils.py
```diff
@@ -251,6 +259,29 @@ def _format_text_image_message(message: ChatMessage) -> dict[str, Any]:
...
+    if not cache_point:
+        return None
+
+    if "type" not in cache_point or cache_point["type"] != "default":
+        err_msg = "Cache point must have a 'type' key with value 'default'."
+        raise ValueError(err_msg)
+    if not set(cache_point).issubset({"type", "ttl"}):
+        err_msg = "Cache point can only contain 'type' and 'ttl' keys."
+        raise ValueError(err_msg)
+
```
julian-risch, line 284: We could extend validation to catch typos in the ttl value by adding the following (and updating :raises docstring)
```suggestion
    if "ttl" in cache_point and cache_point["ttl"] not in ("5m", "1h"):
        raise ValueError("Cache point 'ttl' must be one of '5m', '1h'.")

```

integrations/amazon_bedrock/src/haystack_integrations/components/generators/amazon_bedrock/chat/chat_generator.py
julian-risch, line 145: We could briefly mention the support of prompt caching here. For example:
```python
ChatMessage.from_user("...", meta={"cachePoint": {"type": "default"}}) (and optionally "ttl": "5m" or "1h").
```

integrations/amazon_bedrock/src/haystack_integrations/components/generators/amazon_bedrock/chat/chat_generator.py
```diff
@@ -201,6 +203,9 @@ def __init__(
             See the
             [Guardrails Streaming documentation](https://docs.aws.amazon.com/bedrock/latest/userguide/guardrails-streaming.html)
             for more information.
+        :param tools_cachepoint_config: Optional configuration to use prompt caching for tools.
+            The dictionary must match the
+            [CachePointBlock schema](https://docs.aws.amazon.com/bedrock/latest/APIReference/API_runtime_CachePointBlock.html).
```
julian-risch, line 214: I suggest we list an example here

anakin87:
> Is it possible to add this metadata via the ChatPromptBuilder? Otherwise in a Pipeline I'm not sure how we would pass this meta information.

Yes, the `ChatPromptBuilder` allows passing `meta`, either with the template as a list of `ChatMessage` or as a string.

> A side point is that I could see users wanting to put a cache point just after the system message which doesn't work currently since our Agent takes its system prompt in via an init param. I suppose there is a workaround for this by not using the init param and do it all in the ChatPromptBuilder but then we would need the meta specification to work.

Yes, I tried passing the system prompt as the first message in `Agent.run` and that works.

> Also I think for long running Agents (eg calling a retrieval tool 5 time) I think it does make sense to add cache points to earlier assistant and tool result messages. Basically to help alleviate the cost when your list messages starts to approach the full context window.

I understand we are not covering this use case that extends beyond the base behavior. I would be curious to know your ideas on how to implement this. (If possible, I'd try to let the user choose and not make assumptions about their preferences)

integrations/amazon_bedrock/src/haystack_integrations/components/generators/amazon_bedrock/chat/chat_generator.py
anakin87: Done

integrations/amazon_bedrock/src/haystack_integrations/components/generators/amazon_bedrock/chat/chat_generator.py
anakin87: Done

anakin87:
> My main question/concern is about usability and the "global parameter" that you refer to in the PR description. Even if we cannot set cache points in each message because of the maximum of 4 cache points per request, can we make it easier for users to use caching? Instead of requiring them to specify cachePoints in meta explicitly? Sebastian brought up the ChatPromptBuilder and Robin commented on the issue regarding this aspect too.

I evaluated the "global setting" idea but ended up discarding it for its opaqueness: the current approach requires a bit more effort from the user but it's explicit, leaving them control over caching behaviors.

If we want to explore making caching simpler via a global parameter, one possibility would be to automatically cache system and user messages until the 4-cache-point limit is reached. I'd not count tokens because it requires time and both tokenization and limits are dependent on the model.

Personally, I'd prefer to keep caching explicit rather than introducing automatic behavior, but let me know which approach you prefer...

anakin87:
> Not entirely sure how I'd implement this inside of Agent in a satisfying way that would also not be specific only to Bedrock and Anthropic. I'd have to give it more thought. But I do believe this is something Robin was asking for.

@deep-rloebbert your specific use case involves caching more than system, user messages and tools? E.g. assistant messages, tool call results

deep-rloebbert:
@anakin87 Great discussion so far! I’m trying to wrap my head around how the caching scales fundamentally, and I have two specific questions:

1. Is the limit of 4 cache breakpoints applied per request or is it a global limit for my entire subscription?
2. Is it possible to recycle or shift cache points? (e.g., Can I move the first checkpoint further down the context window as the conversation grows?)

Depending on the answers above, it would be valuable to eventually cache tool results as well, since those payloads can be quite large. However, enabling to set cache points on the previous messages (system prompt + chat history) via the ChatPromptBuilder, so that the Agent can do long processing on it quicker, is probably a good idea. Wdyt?

anakin87:
> 1. Is the limit of 4 cache breakpoints applied per request or is it a global limit for my entire subscription?

Per request. Unfortunately, AWS does not share details about global limits, etc.

> 2. Is it possible to recycle or shift cache points? (e.g., Can I move the first checkpoint further down the context window as the conversation grows?)

Each request defines its own cache points and cache reuse depends on how much of the prompt prefix matches previous requests. It's always feasible to create new cache points, but existing cache points can't be rewritten. I can't tell about Bedrock total limits or how effective it would be.
 
> Depending on the answers above, it would be valuable to eventually cache tool results as well, since those payloads can be quite large. However, enabling to set cache points on the previous messages (system prompt + chat history) via the ChatPromptBuilder, so that the Agent can do long processing on it quicker, is probably a good idea. Wdyt?

I haven't tried these complex use cases, so I can't say for sure. If you find that the current behavior is already sufficient to experiment and save some tokens, I'd start there. Then, based on your experimentation, we could evaluate more advanced approaches (for example, allowing the Agent or ToolInvoker to set specific `ChatMessage.meta` as a workaround for passing cache points, as discussed with Seb).

sjrl (approved):
Looks good! Looking forward to your feedback @deep-rloebbert

PR #2798, opened 2026-01-30
fix(langfuse): replace MagicMock with Mock to prevent RAM exhaustion (#2543)

### LINKED ISSUE #2543: bug(langfuse): test_create_new_span causes excessive RAM usage and hangs due to unspecced MagicMock
**Describe the bug**
  The issue is on line 537 where mock_raw_span = MagicMock() is created without a spec parameter. When this mock is passed through the tracer's handle() method, it triggers unbounded creation of child MagicMock objects because:
  - Every attribute access (span.get_data(), span.raw_span(), etc.) creates new MagicMock instances
  - MagicMock stores all calls and arguments in memory for assertion tracking
  - Without a spec, there are no boundaries preventing infinite mock object creation
  - The handle() method makes numerous such calls (lines 352-421 in tracer.py)

**To Reproduce**
```python
  cd integrations/langfuse
  hatch run test:unit tests/test_tracer.py::TestLangfuseTracer::test_create_new_span -v
```
 - OS: [Universal Dev Container](https://mcr.microsoft.com/en-us/artifact/mar/devcontainers/universal/tags)
 - Integration version: Commit 4ed293674e4f8b4698584e9e8da9e7ddcf5f7d2e

GunaPalanivel:
- Fixes #2543 

- test hangs on Linux Python 3.9/3.13 due to unspecced `MagicMock()`.

## Changes

Replaced `MagicMock()` with `Mock()` on lines `571` and `589` in `test_tracer.py`.

Unspecced `MagicMock` creates infinite child mocks on attribute access, which caused `handle()` to hang while iterating mock objects. A partial workaround existed (`get_data.return_value = {}`), but the root mocks were still unspecced.

## Testing

* `test_create_new_span` passes
* All other tracer tests pass (32 total)

## Note

Two pre-existing failures (`test_set_content_tag_updates_input_and_output*`) exist on `main` and are unrelated; should be tracked separately.

### DIFF
--- a/integrations/elasticsearch/CHANGELOG.md
+++ b/integrations/elasticsearch/CHANGELOG.md
@@ -1,5 +1,12 @@
 # Changelog
 
+## [integrations/elasticsearch-v5.2.0] - 2026-02-02
+
+### 🚀 Features
+
+- Add SQLRetriever to ElasticsearchDocumentStore (#2801)
+
+
 ## [integrations/elasticsearch-v5.1.1] - 2026-01-29
 
 ### 🐛 Bug Fixes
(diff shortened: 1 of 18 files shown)


davidsbatista (approved):
LGTM!

PR #2823, opened 2026-02-09
chore: Google GenAI - switch default embedding model to gemini-embedding-001

anakin87:
### Related Issues
Failing tests: https://github.com/deepset-ai/haystack-core-integrations/actions/runs/21808171728/job/62915074488

In our tests, we use Gemini API.

We were using `text-embedding-004` as a default
- removed from Gemini API ([source](https://ai.google.dev/gemini-api/docs/deprecations#embedding-models)); also `text-embedding-005` is not available on Gemini API, based on my tests (can't find official docs)
- still available on Vertex AI ([source](https://docs.cloud.google.com/vertex-ai/generative-ai/docs/learn/model-versions#embeddings-models))

### Proposed Changes:
- switch default embedding model to `gemini-embedding-001` (available both in Gemini and Vertex APIs)

### How did you test it?
CI

### Notes for the reviewer
The most straightforward way to use this integration is via Gemini API and should work out of the box.

For Vertex AI, there is a risk of breaking some use cases if users don't explicitly set the model. **Let me know if you prefer me to release this with a major version bump.**

### Checklist


sjrl (approved):
Looks good thanks!

Issue #2816, opened 2026-02-06
API reference sync sometimes break Docusaurus builds


anakin87:
Whenever a core integration is released, the API reference is generated and a PR targeting Haystack is automatically created.

### Problem
Sometimes the generated API reference causes the Docusaurus build to fail, even if the generated Markdown itself is valid. This happens with characters like `{` in docstrings and in other situations.

When this happens, the maintainer has to:
- Open another PR targeting core-integrations (and manually verify that Docusaurus can build successfully)
- Get the PR approved
- Manually re-run the "Core / Sync API reference with Docusaurus" workflow or re-release the integration
- Approve the new PR generated on Haystack

This process is manual and time-consuming, and mainly comes from the fact that core integrations API reference is hosted in the separate Haystack repo.

### Ideas
I don't expect it to be easy to solve, but it would already help to offer maintainers a way to check if Docusaurus can build the website correctly.

One possible idea could be to automatically open a draft PR on Haystack for each PR on core integrations that changes the API reference and inform contributors they can inspect the preview there. Needs investigation.

PR #2829, opened 2026-02-11
ci: check that API references do not cause Docusaurus build failures

anakin87:
### Related Issues

- fixes https://github.com/deepset-ai/haystack-core-integrations/issues/2816

### Proposed Changes:

I started from the idea that the simplest way to check whether generated Markdown API references are compatible with Docusaurus is to check out the Docs Website and try building it. Then I found better tools, and this idea evolved into something simpler.

- Created a new "Core / Check API reference changes" workflow
  - detect integrations with API reference changes (if pydoc config or docstring checksum have changed)
  - for each one of the changed integrations:
    - generates Markdown API reference
    - moves it to a fresh `website` folder for testing
  - runs `docusaurus-mdx-checker`, as suggested in [Docusaurus docs](https://docusaurus.io/blog/preparing-your-site-for-docusaurus-v3#using-the-mdx-checker-cli): this does not build the website but just checks generated Markdown files.
    - if it fails, the informative build error is available on CI logs

### How did you test it?
Temporarily added some changes:
- valid config change for Pgvector + invalid Docstring in Qdrant -> clear error: https://github.com/deepset-ai/haystack-core-integrations/actions/runs/21941003098/job/63366427299?pr=2829
- new integration (pgvector2) -> triggers the check, which is OK -> https://github.com/deepset-ai/haystack-core-integrations/actions/runs/21941146095/job/63366923997?pr=2829

### Notes for the reviewer
1. By running the checker but not building the website, we only get MD/MDX errors: for example, we don't get errors for duplicate page ids or broken links. In general, it seems OK since our previous errors were caused by invalid MD/MDX. Plus, the check is much faster than building the website.

2. With this change, we no longer need to run `hatch run docs` in each integration workflow:
https://github.com/deepset-ai/haystack-core-integrations/blob/d96420126ca9edf24787e0714af32eec12efbff0/.github/workflows/aimlapi.yml#L57-L59
I prefer to remove this step in another PR to keep this focused and minimal.

### Checklist

### DIFF
--- a/integrations/cohere/CHANGELOG.md
+++ b/integrations/cohere/CHANGELOG.md
@@ -1,5 +1,18 @@
 # Changelog
 
+## [integrations/cohere-v7.0.2] - 2026-02-12
+
+### 🐛 Bug Fixes
+
+- Cohere - use dictionaries for Chat Message representation (#2832)
+
+## [integrations/cohere-v7.0.1] - 2026-01-16
+
+### 🐛 Bug Fixes
+
+- *(cohere)* Use dataclass replace to avoid modifying input documents (#2755)
+
+
 ## [integrations/cohere-v7.0.0] - 2026-01-12
 
 ### 🧹 Chores
--- a/integrations/cohere/pyproject.toml
+++ b/integrations/cohere/pyproject.toml
@@ -56,6 +56,7 @@ dependencies = [
     "pytest-cov",
     "pytest-rerunfailures",
     "mypy",
+    "types-tqdm",
     "pip",
     "pillow",   # image support
     "pypdfium2" # image support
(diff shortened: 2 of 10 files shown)


.github/workflows/CI_check_api_ref.yml
```diff
@@ -0,0 +1,135 @@
...
+        with:
+          python-version: "3.13"
+
+      - name: Detect integrations with API reference changes
+        id: changed
+        shell: python
+        env:
+          BASE_SHA: ${{ github.event.pull_request.base.sha }}
+        run: |
+          import json
```
anakin87, line 26: 1. This is long but tightly coupled with the workflow, so I haven't extracted it to a utility script
2. While bash would have been more natural for it (invokes Git, ...), I chose to use Python because we are all more familiar with it, so this seems to be more maintainable in the long run

PR #2849, opened 2026-02-17
feat: use StreamingChunk.reasoning field for Anthropic thinking content

yaowubarbara:
### Related Issues

- fixes deepset-ai/haystack#10479

### Proposed Changes:

Right now the Anthropic integration dumps the entire stream event into `StreamingChunk.meta` via `chunk.model_dump()`, which means thinking/reasoning content ends up buried in `meta["delta"]["thinking"]` instead of being accessible through the dedicated `StreamingChunk.reasoning` field that was added in Haystack 2.18.0.

This PR updates `_convert_anthropic_chunk_to_streaming_chunk` to populate `StreamingChunk.reasoning` with a proper `ReasoningContent` object when it encounters thinking-related events:

- `content_block_start` with type `thinking` or `redacted_thinking`
- `content_block_delta` with type `thinking_delta` or `signature_delta`

Also updated `_process_reasoning_contents` to read from `chunk.reasoning` instead of digging through `chunk.meta`, which simplifies the logic quite a bit. Extracted the repeated group-finalization code into a small helper (`_finalize_reasoning_group`).

The raw event data is still available in `meta` for anyone who needs it -- this just makes reasoning content properly accessible through the structured field.

### How did you test it?

- Updated existing unit tests to assert that `StreamingChunk.reasoning` is populated correctly for thinking block start, thinking delta, and signature delta events
- All existing tests still pass

### Notes for the reviewer

The non-streaming path (`_convert_chat_completion_to_chat_message`) already uses `ReasoningContent` correctly, so no changes needed there. This only affects the streaming code path.

### Checklist

### DIFF
--- a/integrations/anthropic/src/haystack_integrations/components/generators/anthropic/chat/utils.py
+++ b/integrations/anthropic/src/haystack_integrations/components/generators/anthropic/chat/utils.py
@@ -404,12 +404,14 @@ def _process_reasoning_contents(chunks: list[StreamingChunk]) -> ReasoningConten
         if chunk.reasoning is None:
             continue
 
-        # Skip chunks with empty reasoning content
-        if (
-            not chunk.reasoning.reasoning_text
-            and not chunk.reasoning.extra.get("signature")
-            and not chunk.reasoning.extra.get("redacted_thinking")
-        ):
+        # Only thinking text or redacted thinking trigger grouping. Signatures are metadata only
+        has_thinking_text = bool(chunk.reasoning.reasoning_text)
+        has_redacted_thinking = chunk.reasoning.extra.get("redacted_thinking") is not None
+
+        if not (has_thinking_text or has_redacted_thinking):
+            # This is a signature-only chunk. Accumulate it but don't start a new group.
+            if chunk.reasoning.extra.get("signature") is not None:
+                content_block_signature = chunk.reasoning.extra["signature"]
             continue
 
         # Start new group when index changes


integrations/anthropic/src/haystack_integrations/components/generators/anthropic/chat/utils.py
```diff
@@ -377,79 +389,56 @@ def _process_reasoning_contents(chunks: list[StreamingChunk]) -> ReasoningConten
...
+            content_block_signature,
+            content_block_redacted_thinking,
+        )
 
-    # Combine all reasoning texts into a single string for the main reasoning_text field
+    # Combine all reasoning texts
     final_reasoning_text = ""
     for content in formatted_reasoning_contents:
         if "reasoning_text" in content["reasoning_content"]:
-            # mypy somehow thinks that content["reasoning_content"]["reasoning_text"]["text"] can be of type None
```
anakin87, line 452: this comment is still valid. Please restore it

integrations/anthropic/src/haystack_integrations/components/generators/anthropic/chat/utils.py
```diff
@@ -461,3 +450,24 @@ def _process_reasoning_contents(chunks: list[StreamingChunk]) -> ReasoningConten
         if formatted_reasoning_contents
         else None
     )
+
+
+def _finalize_reasoning_group(
```
anakin87, line 466: 1. Please include parameters descriptions in the docstring
2. Add one/more unit tests for this function

integrations/anthropic/src/haystack_integrations/components/generators/anthropic/chat/utils.py
```diff
@@ -377,79 +389,56 @@ def _process_reasoning_contents(chunks: list[StreamingChunk]) -> ReasoningConten
...
     for chunk in chunks:
-        if (delta := chunk.meta.get("delta")) is not None:
-            if delta.get("type") == "thinking_delta" and delta.get("thinking") is not None:
-                content_block_index = chunk.meta.get("index", None)
-        if (content_block := chunk.meta.get("content_block")) is not None and content_block.get(
-            "type"
-        ) == "redacted_thinking":
-            content_block_index = chunk.meta.get("index", None)
+        if chunk.reasoning is None:
+            continue
```
anakin87, line 405: Can you please modify this function to match the previous structure more closely?
In particular, add a check to skip chunks with empty reasoning content

anakin87 (commented):
I left some initial comments.

integrations/anthropic/src/haystack_integrations/components/generators/anthropic/chat/utils.py
yaowubarbara: done thanks！

integrations/anthropic/src/haystack_integrations/components/generators/anthropic/chat/utils.py
```diff
@@ -377,75 +389,61 @@ def _process_reasoning_contents(chunks: list[StreamingChunk]) -> ReasoningConten
...
-                formatted_reasoning_contents.append(
-                    {"reasoning_content": {"redacted_thinking": content_block_redacted_thinking}}
-                )
+        # Skip chunks with empty reasoning content
+        if (
+            not chunk.reasoning.reasoning_text
+            and not chunk.reasoning.extra.get("signature")
+            and not chunk.reasoning.extra.get("redacted_thinking")
+        ):
+            continue
```
anakin87:
```suggestion
        # Only thinking text or redacted thinking trigger grouping. Signatures are metadata only
        has_thinking_text = bool(chunk.reasoning.reasoning_text)
        has_redacted_thinking = chunk.reasoning.extra.get("redacted_thinking") is not None
        
        if not (has_thinking_text or has_redacted_thinking):
            # This is a signature-only chunk. Accumulate it but don't start a new group.
            if chunk.reasoning.extra.get("signature") is not None:
                content_block_signature = chunk.reasoning.extra["signature"]
            continue
```

This would better match previous behavior.
(Untested, format might be wrong)

yaowubarbara: Thanks for the suggestion, this makes a lot of sense! Signatures are really just metadata and shouldnt trigger new content groups. I have applied your suggested change and pushed it.

integrations/anthropic/src/haystack_integrations/components/generators/anthropic/chat/utils.py
yaowubarbara: Done, sorry I forgot to reply earlier. The mypy comment was restored in the previous commit already.

integrations/anthropic/src/haystack_integrations/components/generators/anthropic/chat/utils.py
yaowubarbara: Both done in the previous commit, added parameter descriptions to the docstring and 4 unit tests covering thinking text, redacted thinking, both combined, and empty inputs. Sorry for not replying sooner!

Issue #2804, opened 2026-02-02
Expose cached token count and other metadata in GoogleGenAIChatGenerator


kahkhang:
**Is your feature request related to a problem? Please describe.**

For `GoogleGenAIChatGenerator` and other google genai chat generators, the cached token count, and other metadata is not transparently passed to the output of the function, making it impossible for our use case of calculating api price since the cached token count is not exposed.

https://github.com/deepset-ai/haystack-core-integrations/blob/051ccb84f5a44edeb7882d346d507f7402005809/integrations/google_genai/src/haystack_integrations/components/generators/google_genai/chat/chat_generator.py#L311

**Describe the solution you'd like**

Can you do something similar to `OpenAIChatGenerator`, where `usage` dict from underlying provider response is transparently passed? I understand there are intricacies involved since you ended up mapping google token metadata output to a subset of OpenAI's response shape, but ultimately we just need a way to access the cached token output data. Thank you!

https://github.com/deepset-ai/haystack/blob/1d7bc772123b4d871a641683cce7d84c8bb73533/haystack/components/generators/chat/openai.py#L586

**Describe alternatives you've considered**

**Additional context**

PR #2851, opened 2026-02-17
feat: include token usage count in `GoogleGenAIChatGenerator` response metadata

davidsbatista:
### Related Issues

- fixes #[2804](https://github.com/deepset-ai/haystack-core-integrations/issues/2804)

### Proposed Changes:

- `GoogleGenAIChatGenerator` response now includes ["_meta"]["usage"]["cached_content_token_count"]
- Converted some methods to static 

### How did you test it?

Added two new tests:

- `test_live_run_implicit_cache_usage_returned` - an integration test that runs the chat generator with a big context and asks 10 questions sequentially, asserting that the big context will be cached and we can parse it

- `test_convert_google_genai_response_to_chatmessage_parses_cached_tokens` - a unit test, the response is mocked from the API

Note regarding the integration test :

- We want to test that the Component can parse the cached tokens if present in the response - we don't control when it's present or not, so maybe the unit test is enough 

- On the other side, having it as an integration test can make us aware if there are changes in the API

### Checklist

### DIFF
--- a/integrations/google_genai/src/haystack_integrations/components/generators/google_genai/chat/chat_generator.py
+++ b/integrations/google_genai/src/haystack_integrations/components/generators/google_genai/chat/chat_generator.py
@@ -299,7 +299,7 @@ def _convert_tools_to_google_genai_format(tools: ToolsType) -> list[types.Tool]:
 
 
 def _convert_usage_metadata_to_serializable(
-        usage_metadata: UsageMetadata|GenerateContentResponseUsageMetadata|None
+    usage_metadata: UsageMetadata | GenerateContentResponseUsageMetadata | None,
 ) -> dict[str, Any]:
     """Build a JSON-serializable usage dict from a UsageMetadata object.
 


davidsbatista:
> Before I review, I'd like to discuss some high-level aspects.
> 
> 1. My impression is that [Expose cached token count and other metadata in GoogleGenAIChatGenerator #2804](https://github.com/deepset-ai/haystack-core-integrations/issues/2804) asked exposing cache usage and other metadata (not support for explicit caching). According to [Gemini docs](https://ai.google.dev/gemini-api/docs/caching), implicit caching is enabled by default and should already cover most common use cases, without adding complexity to our integration. (@sjrl feel free to correct me based on your investigation)

> 2. If we decide to support [explicit caching](https://ai.google.dev/gemini-api/docs/caching), I'd prefer handling it via `ChatMessage.meta` as we do for Bedrock and Anthropic. If you've explored this approach, I'd be curious to hear what led you in a different direction.

I added an explicit mechanism to activate the cache since I wasn't seeing any cache hits in my testing. And agree, the usage info belongs in `ChatMessage.meta`, my mistake, will fix.

I will remove the explicit cache mechanism and put together a clearer example to demonstrate implicit cached tokens. I've already verified it works, so it's just a matter of finding a working example.

integrations/google_genai/src/haystack_integrations/components/generators/google_genai/chat/chat_generator.py
```diff
@@ -297,6 +297,25 @@ def _convert_tools_to_google_genai_format(tools: ToolsType) -> list[types.Tool]:
...
+    if hasattr(val, "modality") and hasattr(val, "token_count"):
+        mod = getattr(val, "modality", None)
+        mod_str = getattr(mod, "name", getattr(mod, "value", str(mod))) if mod is not None else None
+        return {"modality": mod_str, "token_count": getattr(val, "token_count", None)}
+    if hasattr(val, "name"):
+        return getattr(val, "value", val.name)
+    if hasattr(val, "__dict__"):
+        return {k: _usage_metadata_value_to_serializable(getattr(val, k)) for k in vars(val)}
+    return val
+
```
anakin87, line 345: 1. My impression is that, based on the [`UsageMetadata` object](https://ai.google.dev/api/generate-content#UsageMetadata), we might simplify this function to only handle the types of values supported by the object
2. Let's also add unit tests for this function

integrations/google_genai/src/haystack_integrations/components/generators/google_genai/chat/chat_generator.py
```diff
@@ -297,6 +297,25 @@ def _convert_tools_to_google_genai_format(tools: ToolsType) -> list[types.Tool]:
     return [types.Tool(function_declarations=function_declarations)]
 
 
+def _usage_metadata_value_to_serializable(val: Any) -> Any:
```
anakin87: I'd use the `UsageMetadata` type for `val`

integrations/google_genai/tests/test_chat_generator.py
```diff
@@ -798,6 +838,89 @@ def test_live_run(self) -> None:
...
+            replies.append(reply)
+            usage = reply.meta.get("usage") or {}
+            usages.append(usage)
+            assert "prompt_tokens" in usage or "total_tokens" in usage, f"Usage should contain token counts: {usage}"
+
+        # Implicit cache: same long system prefix across requests can yield cached_content_token_count in at least one
+        any_cached = any((u.get("cached_content_token_count") or 0) > 0 for u in usages)
+        assert any_cached, (
+            f"Expected cached_content_token_count > 0 in at least one reply (implicit cache). Usages: {usages}"
+        )
```
anakin87: I ran the test locally and it takes ~14s.
Initially, I tried to improve it, then I also noticed that it takes some time to build the cache: this means that at the beginning, the test would fail.

Unlike Anthropic/Bedrock, Google does not expose how many tokens were written to the cache. If it did, we could check tokens written or tokens read.

For these reasons, I propose removing the test.

integrations/google_genai/src/haystack_integrations/components/generators/google_genai/chat/chat_generator.py
```diff
@@ -361,7 +394,33 @@ def _convert_google_genai_response_to_chatmessage(response: types.GenerateConten
...
+            "cache_tokens_details",
+            "candidates_tokens_details",
+            "prompt_tokens_details",
+            "tool_use_prompt_token_count",
+            "tool_use_prompt_tokens_details",
+        )
+        for attr in _usage_attr_names:
+            val = getattr(usage_metadata, attr, None)
+            if val is not None:
+                usage[attr] = _usage_metadata_value_to_serializable(val)
```
anakin87: I realized that when I previously suggested using `UsageMetadata`, I was wrong, unless we include this code in the `_usage_metadata_value_to_serializable` (and maybe rename it to `_convert_usage_metadata_to_serializable`).

I'd suggest doing that.

integrations/google_genai/tests/test_chat_generator.py
```diff
@@ -384,6 +386,102 @@ def test_convert_google_chunk_to_streaming_chunk_real_example(self, monkeypatch)
...
+    obj_camel = ModalityTokenCountCamel(mod_enum, 200)
+    result_camel = _usage_metadata_value_to_serializable(obj_camel)
+    assert result_camel == {"modality": "TEXT", "token_count": 200}
+
+    # list of ModalityTokenCount
+    list_result = _usage_metadata_value_to_serializable([obj, obj_camel])
+    assert list_result == [
+        {"modality": "TEXT", "token_count": 100},
+        {"modality": "TEXT", "token_count": 200},
+    ]
```
anakin87: If possible, let's use actual Google types here (but see my suggestion above first)

integrations/google_genai/tests/test_chat_generator.py
```diff
@@ -425,61 +429,42 @@ def test_convert_google_genai_response_to_chatmessage_parses_cached_tokens(monke
...
+    """_convert_usage_metadata_to_serializable builds a serializable dict from a UsageMetadata object."""
+    assert _convert_usage_metadata_to_serializable(None) == {}
+    assert _convert_usage_metadata_to_serializable(False) == {}
 
-    # enum-like with only name
-    class EnumNameOnly:
-        name = "AUDIO"
-
-    assert _usage_metadata_value_to_serializable(EnumNameOnly) == "AUDIO"
+    mock_usage = Mock()
```
anakin87: Here I'd use the `UsageMetadata` object if possible

integrations/google_genai/src/haystack_integrations/components/generators/google_genai/chat/chat_generator.py
```diff
@@ -298,36 +297,48 @@ def _convert_tools_to_google_genai_format(tools: ToolsType) -> list[types.Tool]:
     return [types.Tool(function_declarations=function_declarations)]
 
 
-def _usage_metadata_value_to_serializable(val: UsageMetadata) -> Any:
-    """Convert a UsageMetadata field value to a JSON-serializable form.
+def _convert_usage_metadata_to_serializable(usage_metadata: Any) -> dict[str, Any]:
```
anakin87:
```suggestion
def _convert_usage_metadata_to_serializable(usage_metadata: UsageMetadata) -> dict[str, Any]:
```

should be OK to adopt `UsageMetadata` now

Issue #2751, opened 2026-01-14
Update integration pages to reflect that components auto `warm_up` on their first run


sjrl:
_Originally posted by @julian-risch in https://github.com/deepset-ai/haystack-core-integrations/pull/2749#pullrequestreview-3660516000_

> Looks good to me! 👍 Let's not forget to simplify the code example in the integration page: https://haystack.deepset.ai/integrations/hanlp  and the documentation page https://docs.haystack.deepset.ai/docs/chinesedocumentsplitter
Maybe we could have one GitHub issue as a reminder to go through all the integrations that you updates? As a separate follow up to https://github.com/deepset-ai/haystack-core-integrations/issues/2592 not as another step of that issue.

Core integrations whose docs pages should be updated to remove the explicit `warm_up` call
- [ ] optimum
- [ ] fastembed
- [ ] snowflake
- [ ] nvidia
- [ ] amazon_bedrock
- [ ] llama_cpp
- [ ] weave
- [ ] hanlp

PR #2852, opened 2026-02-17
docs(llama_cpp): remove explicit warm_up calls

GunaPalanivel:
## Context

- Partially addresses #2751

Llama.cpp components now auto-warm-up on first run, so explicit `warm_up()` calls in examples are redundant.

## Changes

* Removed explicit `warm_up()` call from docstring examples in:

  * `LlamaCppChatGenerator`

* Removed explicit `warm_up()` call from:

  * `examples/llama_cpp_generator_example.py`

## Testing

* Ran `hatch run test` (skipped locally due to missing build environment; relying on CI validation).


sjrl (approved):
Thanks!

Issue #2857, opened 2026-02-19
Raw OpenSearch SQL query does not support json format anymore for OpenSearch 3.0 and newer


julian-risch:
Unfortunately format json is not supported anymore for OpenSearch 3.0 and newer. See https://github.com/opensearch-project/sql/issues/5038#issuecomment-3748655214"

Current implementation of the Haystack integration does rely on it here:
[https://github.com/deepset-ai/haystack-core-integrations/blob/6c8c1bfdf7466c4b5882[…]stack_integrations/document_stores/opensearch/document_store.py](https://github.com/deepset-ai/haystack-core-integrations/blob/6c8c1bfdf7466c4b5882954fcd8864a4f87dedad/integrations/opensearch/src/haystack_integrations/document_stores/opensearch/document_store.py#L2006)

PR #2876, opened 2026-02-25
fix: OpenSearch-3.x SQL stopped supporting DSL-format SQL JSON

davidsbatista:
### Related Issues

- fixes #[2857](https://github.com/deepset-ai/haystack-core-integrations/issues/2857)

### Proposed Changes:

- The new OpenSearch changed since 3.x onwards.
- SQL responses are still JSON, but no longer the legacy DSL-shaped JSON (i.e.: with `hits`, `_shards`, `aggregations`) returned via `format=json` in 2.x.
- Updated the OpenSearch SQL integration tests to use a docker with OpenSearch 3.5.
- Updated tests to validate the tabular SQL JSON format (schema, datarows, size, status, optional cursor). 

### How did you test it?

-  updated all the SQL-related tests (both sync and async) to assert for a different output, still JSON, just no longer in a DSL-shaped format

### Notes for the reviewer

- Our `OpenSearchDocumentStore` will still work with OpenSearch 2.x, the only change was in the `sql_query` method where the `params = {"format": "json"}` was removed.
- Both `run()` and `run_async()` from the `OpenSearchSQLRetriever` just return the result from the SQL call, so there's no breaking changes introduced

### Checklist

### DIFF
--- a/.github/workflows/CI_check_api_ref.yml
+++ b/.github/workflows/CI_check_api_ref.yml
@@ -81,7 +81,7 @@ jobs:
 
       - name: Install Hatch
         if: steps.changed.outputs.integrations != '[]'
-        run: pip install --upgrade hatch
+        uses: pypa/hatch@install
 
       - name: Generate API references
         if: steps.changed.outputs.integrations != '[]'
--- a/.github/workflows/CI_docusaurus_sync.yml
+++ b/.github/workflows/CI_docusaurus_sync.yml
@@ -32,7 +32,7 @@ jobs:
           python-version: "3.10"
 
       - name: Install Hatch
-        run: pip install --upgrade hatch
+        uses: pypa/hatch@install
 
       - name: Get project folder
         id: pathfinder
--- a/.github/workflows/CI_pypi_release.yml
+++ b/.github/workflows/CI_pypi_release.yml
@@ -34,8 +34,11 @@ jobs:
         with:
           python-version: "3.12"
 
+      - name: Install Hatch
+        uses: pypa/hatch@install          
+
       - name: Install dependencies
-        run: pip install hatch requests
+        run: pip install requests
 
       - name: Validate version number
         run: python .github/utils/validate_version.py --tag ${{ github.ref_name }}
--- a/.github/workflows/aimlapi.yml
+++ b/.github/workflows/aimlapi.yml
@@ -48,7 +48,7 @@ jobs:
           python-version: ${{ matrix.python-version }}
 
       - name: Install Hatch
-        run: pip install --upgrade hatch
+        uses: pypa/hatch@install
 
       - name: Lint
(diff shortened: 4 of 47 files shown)


integrations/opensearch/tests/conftest.py
```diff
@@ -28,11 +28,11 @@ def document_store():
     store = OpenSearchDocumentStore(
         hosts=hosts,
         index=index,
-        http_auth=("admin", "admin"),
+        http_auth=("admin", "SecureHaystack!2026"),
         verify_certs=False,
         embedding_dim=768,
         return_embedding=True,
-        method={"space_type": "cosinesimil", "engine": "nmslib", "name": "hnsw"},
+        method={"space_type": "cosinesimil", "engine": "lucene", "name": "hnsw"},
```
anakin87, line 35: I am curious. What happened to nmslib?

integrations/opensearch/tests/test_document_store.py
```diff
@@ -164,12 +164,24 @@ def document_store(self, document_store):
...
+
+        We don't want to compare the score, so we set it to None before comparing.
+
+        Embeddings are not exactly the same when retrieved from OpenSearch (float round-trip),
+        so we compare them approximately and then set both to None for the final equality check.
         """
-        for doc in received:
-            doc.score = None
-        assert received == expected
+        assert len(received) == len(expected)
```
anakin87, line 174: nit: `assert len(received) == len(expected)` is unneded because `zip(..,  strict=True)` also checks length equality

integrations/opensearch/tests/conftest.py
davidsbatista: https://docs.opensearch.org/latest/breaking-changes/

`The NMSLIB engine is now deprecated. We recommend using the Faiss or Lucene engines instead.`

integrations/opensearch/tests/test_document_store.py
davidsbatista: ah good catch 👍🏽

davidsbatista: I went the check the details of how zip works and:

By default, [zip()](https://docs.python.org/3/library/functions.html#zip) stops when the shortest iterable is exhausted. It will ignore the remaining items in the longer iterables, cutting off the result to the length of the shortest iterable.

One could add the `strict=True` which will then raise a `Value Error` if one iterable is exhausted before the others.

I would rather leave this len comparision, if it fails it's clear why it's failing.

anakin87: Let's leave both (I think ruff enforces using `strict` parameter)

davidsbatista: it does, I will revert it

anakin87 (approved):
👍

PR #2877, opened 2026-02-25
feat: Add support for metadata that contains lists of supported types

### LINKED ISSUE #2214: Chroma - The meta in write_documents only supports str, int, float, bool, but not list. However, its search filter does support list.
#DUG Describe
When using ChromaDocumentStore from Haystack integrations, the write_documents method does not support storing list values in metadata fields, only supporting str, int, float, and bool types. However, during query operations, the system does support list values in filters, creating an inconsistency between write and read operations.

##Steps to reproduce the behavior:
Import ChromaDocumentStore: from haystack_integrations.document_stores.chroma import ChromaDocumentStore
Create a document with list metadata: doc.metadata = {"countries": ["UK", "US"]}
Attempt to write document: document_store.write_documents([doc])
Get validation error: ValueError: Expected metadata value to be a str, int, float or bool

##Expected behavior：
The write_documents method should support list values in metadata to maintain consistency with query operations that already support list filtering.

##Additional context
The query method search_embeddings supports list values in filters parameter.
This creates a mismatch between data ingestion and query capabilities.
Workarounds currently require serializing list data to strings before storage.

##Possible Solution
Suggest either:
Remove the type restriction in write_documents validation.
Add automatic serialization/deserialization for list values.
Extend supported types to include list and dict.

Environment
Haystack version: 0.42.0
chroma-haystack: 3.1.0
Python version: 3.10
OS: Windows

maxdswain:
### Related Issues

- fixes #2214 

### Proposed Changes:

 
 
Update validation logic to support lists of supported ChromaDB metadatas types.

### How did you test it?

Modified existing tests. I also ran the following script to test that the issue the user mentioned has been fixed:
```python
from haystack import Document

from haystack_integrations.document_stores.chroma import ChromaDocumentStore

doc = Document(content="test", meta={"countries": ["UK", "US"]})

document_store = ChromaDocumentStore()
number = document_store.write_documents([doc])
print(number)
```

### Checklist

### DIFF
--- a/integrations/chroma/pyproject.toml
+++ b/integrations/chroma/pyproject.toml
@@ -24,7 +24,7 @@ classifiers = [
 ]
 dependencies = [
   "haystack-ai>=2.24.0",
-  "chromadb>=1.0.2"
+  "chromadb>=1.5.0"
 ]
 
 [project.urls]


integrations/chroma/src/haystack_integrations/document_stores/chroma/document_store.py
```diff
@@ -22,6 +22,7 @@
 
 VALID_DISTANCE_FUNCTIONS = "l2", "cosine", "ip"
 SUPPORTED_TYPES_FOR_METADATA_VALUES = str, int, float, bool
```
sjrl, line 25: Could we add a dev comment above this line that provides the chroma db docs for what types of meta data value types are supported?

integrations/chroma/src/haystack_integrations/document_stores/chroma/document_store.py
```diff
@@ -533,12 +534,14 @@ def _convert_document_to_chroma(doc: Document) -> dict[str, Any] | None:
         data: dict[str, Any] = {"ids": [doc.id], "documents": [doc.content]}
 
         if doc.meta:
-            valid_meta = {}
+            valid_meta: dict[str, SUPPORTED_METADATA_VALUES_TYPE | list[SUPPORTED_METADATA_VALUES_TYPE]] = {}
```
sjrl, line 538: Just to double check is this type hint needed to get type checking to pass?

maxdswain: Yes, this type hint is needed for the mypy type checking to pass.

sjrl (approved):
Looks good thanks!

PR #2900, opened 2026-03-01
feat: use reasoning field in StreamingChunk for Google GenAI

Br1an67:
### Related Issues

- fixes deepset-ai/haystack#10478

### Proposed Changes:

Populate `StreamingChunk.reasoning` with `ReasoningContent` instead of storing reasoning deltas as dicts in `meta["reasoning_deltas"]`. This aligns the Google GenAI integration with the standard `StreamingChunk.reasoning` field, consistent with other integrations (e.g., Ollama in #2850).

Changes:
- `_convert_google_chunk_to_streaming_chunk()`: Collect reasoning text from thought parts into a `ReasoningContent` object and pass it via the `reasoning` kwarg instead of `meta["reasoning_deltas"]`
- `_aggregate_streaming_chunks_with_reasoning()`: Read from `chunk.reasoning.reasoning_text` instead of `chunk.meta["reasoning_deltas"]`
- Ensure `StreamingChunk` mutual exclusivity constraint is respected (only one of content/tool_calls/reasoning set per chunk)

### How did you test it?

- Added `test_convert_google_chunk_to_streaming_chunk_with_thought` to verify thought parts populate `StreamingChunk.reasoning` instead of meta
- Updated `test_aggregate_streaming_chunks_with_reasoning` to use `chunk.reasoning` instead of `meta["reasoning_deltas"]`
- All 19 non-image unit tests pass

### Notes for the reviewer

The `thought_signature_deltas` remain in `meta` since they are Google-specific metadata for multi-turn context preservation, not standard reasoning content.

### Checklist

### DIFF
--- a/integrations/google_genai/src/haystack_integrations/components/generators/google_genai/chat/utils.py
+++ b/integrations/google_genai/src/haystack_integrations/components/generators/google_genai/chat/utils.py
@@ -608,18 +608,8 @@ def _convert_google_chunk_to_streaming_chunk(
             elif hasattr(part, "thought") and part.thought:
                 reasoning_deltas.append(part.text if part.text else "")
 
-    # Combine reasoning deltas and thought signatures into a single ReasoningContent.
-    # Thought signature deltas are stored in ReasoningContent.extra when reasoning content
-    # is present (consistent with the Anthropic approach in PR #2849).
-    # When there's no reasoning content, signatures go in meta to avoid the StreamingChunk
-    # mutual exclusivity constraint (content and reasoning cannot both be set).
-    if reasoning_deltas:
-        reasoning_extra: dict[str, Any] = {}
-        if thought_signature_deltas:
-            reasoning_extra["thought_signature_deltas"] = thought_signature_deltas
-        reasoning = ReasoningContent(reasoning_text="".join(reasoning_deltas), extra=reasoning_extra)
-    else:
-        reasoning = None
+    # Combine reasoning deltas into a single ReasoningContent
+    reasoning = ReasoningContent(reasoning_text="".join(reasoning_deltas)) if reasoning_deltas else None
 
     # start is only used by print_streaming_chunk. We try to make a reasonable assumption here but it should not be
     # a problem if we change it in the future.
@@ -631,8 +621,9 @@ def _convert_google_chunk_to_streaming_chunk(
         "usage": usage,
     }
 
-    # Thought signatures go in meta when there's no reasoning content (e.g. text-only or tool-call chunks)
-    if thought_signature_deltas and not reasoning_deltas:
+    # Thought signatures can appear in both reasoning and non-reasoning response parts,
+    # so we always store them in meta for consistency.
+    if thought_signature_deltas:
         meta["thought_signature_deltas"] = thought_signature_deltas
 
     # StreamingChunk allows only one of content/tool_calls/reasoning to be set.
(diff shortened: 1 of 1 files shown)


integrations/google_genai/src/haystack_integrations/components/generators/google_genai/chat/utils.py
```diff
@@ -606,39 +606,39 @@ def _convert_google_chunk_to_streaming_chunk(
...
         "usage": usage,
     }
 
-    # Add reasoning deltas to meta if available
-    if reasoning_deltas:
-        meta["reasoning_deltas"] = reasoning_deltas
-
     # Add thought signature deltas to meta if available (for multi-turn context)
     if thought_signature_deltas:
         meta["thought_signature_deltas"] = thought_signature_deltas
```
anakin87, line 628: My impression is that we can use the `extra` field of `ReasoningContent` ([code](https://github.com/deepset-ai/haystack/blob/e2d41e6578594ccffeb41bd89e6a59027e621f86/haystack/dataclasses/chat_message.py#L182)) to store `thought_signature_deltas`.
In my opinion, we did something similar in #2849 for Anthropic redacted thinking and thinking signature.

Since all this info is related to reasoning, I'd like to have it grouped into `ReasoningContent`.

But I haven't tried, and I don't know if this is simple to implement or if it would make the code much more complex. So please try and let me know.

integrations/google_genai/tests/test_chat_generator_utils.py
```diff
@@ -489,6 +489,38 @@ def test_convert_google_chunk_to_streaming_chunk_real_example(self, monkeypatch)
         assert streaming_chunk.tool_calls[5].id is None
         assert streaming_chunk.tool_calls[5].index == 5
 
+    def test_convert_google_chunk_to_streaming_chunk_with_thought(self, monkeypatch):
```
anakin87, line 492: Could you try using actual objects from Google API in this test?
See `test_convert_google_chunk_to_streaming_chunk_real_example` for an example

integrations/google_genai/src/haystack_integrations/components/generators/google_genai/chat/utils.py
Br1an67: Thanks for the suggestion! I've moved `thought_signature_deltas` into `ReasoningContent.extra` for reasoning chunks, consistent with the Anthropic approach in #2849.

One caveat: `StreamingChunk` enforces mutual exclusivity between `content` and `reasoning` (raises `ValueError` in `__post_init__`), so for text/tool-call chunks that also carry thought signatures, the signatures still go in `meta`. The aggregation logic reads from both sources. This keeps all reasoning-related info grouped into `ReasoningContent` where possible without breaking the constraint.

integrations/google_genai/tests/test_chat_generator_utils.py
Br1an67: Good call — updated the test to use actual `types.Part`, `types.Content`, `types.Candidate`, and `types.GenerateContentResponse` objects, following the pattern in `test_convert_google_chunk_to_streaming_chunk_real_example`.

integrations/google_genai/src/haystack_integrations/components/generators/google_genai/chat/utils.py
anakin87: Seeing the code, I now realize that [Thought Signatures](https://ai.google.dev/gemini-api/docs/thought-signatures) can be included in non-reasoning response parts.

For this reason, I recommend going back to the previous version of your code where `thought_signature_deltas` are always stored in `meta`. This would be simpler and consistent.
I'd just add a comment on top of the code line explaining that this data can be part of non-reasoning content parts.

Makes sense?

Br1an67: Makes sense! Reverted in bd3d479c — `thought_signature_deltas` are now always stored in `meta` with a comment explaining that thought signatures can appear in both reasoning and non-reasoning response parts.

Issue #2070, opened 2025-07-08
feat: Update our ChatGenerators in integrations to use the new StreamingChunk fields


sjrl:
We have expanded `StreamingChunk` to include new fields such as 
```python
    content: str
    meta: Dict[str, Any] = field(default_factory=dict, hash=False)
    component_info: Optional[ComponentInfo] = field(default=None)
# All fields below here are new
    index: Optional[int] = field(default=None)
    tool_calls: Optional[List[ToolCallDelta]] = field(default=None)
    tool_call_result: Optional[ToolCallResult] = field(default=None)
    start: bool = field(default=False)
    finish_reason: Optional[FinishReason] = field(default=None)
```
so we should update our existing ChatGenerator integrations to use these new fields. This will standardize the streaming in Haystack making it easy to use one `streaming_callback` like our `print_streaming_chunk` for all ChatGenerators. 

This really emphasizes the interoperability of Haystack and will make it very straightforward to switch out to different ChatGenerators from different providers in a pipeline or when using a ChatGenerator within the Agent component. 

When working on this I'd suggest:
- Testing with `haystack.components.generators.utils.print_streaming_chunk` to make sure that the output looks clean. Ideally use the ChatGenerator in an Agent and try to get the Agent to output at least two tool calls in one round. You can refer to the PR description of this [PR](https://github.com/deepset-ai/haystack/pull/9525) you can see example outputs of `print_streaming_chunk`
- Refer to the OpenAIChatGenerator implementation to get inspiration: 
  - [`_handle_stream_response`](https://github.com/deepset-ai/haystack/blob/4b515433d9ba18de36f4d0649b3d7dfbfcd2b0da/haystack/components/generators/chat/openai.py#L426-L436) which loops through each stream object from OpenAI and then converts it into a `StreamingChunk`
  - [`_convert_chat_completion_chunk_to_streaming_chunk`](https://github.com/deepset-ai/haystack/blob/4b515433d9ba18de36f4d0649b3d7dfbfcd2b0da/haystack/components/generators/chat/openai.py#L509-L578) which converts each individual stream object from OpenAI and converts it into a singular `StreamingChunk`
[...]

PR #2902, opened 2026-03-01
feat: `WatsonXChatGenerator` - support Tools and use the `StreamingChunk` fields

### LINKED ISSUE #2217: feat: Update `WatsonXChatGenerator` to use the `StreamingChunk` fields
Refer to parent issue for more details

maxdswain:
### Related Issues

- fixes #2217 & #2070 

### Proposed Changes:

 
 
feat: Update `WatsonXChatGenerator` to use the `StreamingChunk` `index` and `finish_reason` fields.

### How did you test it?

Updated existing tests.

### Checklist

### DIFF
--- a/integrations/watsonx/src/haystack_integrations/components/generators/watsonx/chat/chat_generator.py
+++ b/integrations/watsonx/src/haystack_integrations/components/generators/watsonx/chat/chat_generator.py
@@ -522,7 +522,7 @@ class WatsonxChatGenerator:
         return self._process_response(response)
 
     @staticmethod
-    def _parse_tool_call_json(tool_call: object) -> dict[str, Any]:
+    def _parse_tool_call_json(tool_call: str | dict) -> dict[str, Any]:
         """Parse tool call json from Watsonx tool calls."""
         if isinstance(tool_call, dict):
             return tool_call


maxdswain:
> Thanks for the contribution!
> 
> I think that a little more adaptation work will be needed, especially in handling tool calls.
> 
> I suggest taking inspiration from #2078 and #2108 (but keeping in mind that each provider has its own particularities and code can be very different).
> 
> Also, I'd like to see a test with real chunks coming from WatsonX ([similar test for Google GenAI](https://github.com/deepset-ai/haystack-core-integrations/blob/c89680c06213aeb4b223142aeb30360bd2d189ba/integrations/google_genai/tests/test_chat_generator_utils.py#L299)).

Okay, I'll convert it to a draft for now and work on it a bit more to add support for tool calls as well as more tests. I'll let you know when it's ready.

integrations/watsonx/src/haystack_integrations/components/generators/watsonx/chat/chat_generator.py
```diff
@@ -266,24 +305,35 @@ async def run_async(
...
+        _check_duplicate_tool_names(flattened_tools)
+        tool_definitions = [{"type": "function", "function": {**tool.tool_spec}} for tool in flattened_tools]
+
         for msg in messages:
-            if msg.is_from("tool"):
-                logger.debug("Skipping tool message - tool calls are not currently supported")
+            # Watsonx tool call result messages are of the same format as OpenAI chat completions
+            if msg.tool_call_results:
+                watsonx_messages.append(
+                    msg._tool_result_message_to_openai({"role": msg.role.value}, require_tool_call_ids=True)
```
anakin87: this method is private and we might remove/rename it at some point without knowing it's used in this integration.

Let's instead use `to_openai_dict_format`, which is public, and we are committed to not breaking it.

integrations/watsonx/pyproject.toml
```diff
@@ -22,7 +22,7 @@ classifiers = [
     "Programming Language :: Python :: 3.12",
     "Programming Language :: Python :: 3.13",
 ]
-dependencies = ["haystack-ai>=2.17.1", "ibm-watsonx-ai>=1.3.26", "pandas>=2.2.3"]
+dependencies = ["haystack-ai>=2.24.1", "ibm-watsonx-ai>=1.3.26", "pandas>=2.2.3"]
```
anakin87, line 25: Do we really need to pin this recent version of Haystack, or can we choose a lower pin?
(Asking, haven't tried)

maxdswain: Yes as I'm using `ToolsType`from `haystack.tools` which was added in 2.24. I'm using `ToolsType` as I was taking inspiration from the Google GenAI integration which is also pinned at >=2.24.1.

anakin87: Makes sense. Thanks for explaining.

PR #2905, opened 2026-03-02
feat: Add scaffold script to ease addition of new integration

bogdankostic:
### Proposed Changes:

 
 

Added a Python script (`scripts/create_new_integration.py`) that automates the creation of a new integration in this repository. Previously, contributors had to use `hatch new -i` and then manually set up the CI workflow, labeler config, pyproject.toml, pydoc config, and root README entry. This was error-prone and resulted in an outdated integration scaffold.

The new script handles everything in one step:
* Scaffolds the integration folder under `integrations/` with the correct package structure (`pyproject.toml`, source package with `py.typed`, tests, pydoc config, LICENSE, and README).
* Creates a GitHub Actions CI workflow at `.github/workflows/<name>.yml`.
* Inserts label rules into `.github/labeler.yml` in alphabetical order.
* Adds the new integration to the table in the root README.md in alphabetical order.

The script supports both interactive prompts and CLI arguments (`--name`, `--type`) for non-interactive use.
The logic is split into focused utility modules under `scripts/utils/`:
* `naming.py` — name validation and string conversions (folder name to package name, label, module path).
* `templates.py` — template strings for all generated files.
* `scaffold.py` — file creation and config-file update operations. 

The changes are reflected in the "Create a new integration" section in CONTRIBUTING.md to document the new script, replacing the old `hatch new -i` instructions.

### How did you test it?

* Manual verification: ran the script with various integration names and component types

### Checklist

### DIFF
--- a/scripts/utils/templates/pyproject.toml
+++ b/scripts/utils/templates/pyproject.toml
@@ -107,8 +107,6 @@ ignore = [
     "B027",
     # Allow function calls in argument defaults (common Haystack pattern for Secret.from_env_var)
     "B008",
-    # zip() without strict= is OK when lengths are guaranteed equal
-    "B905",
     # Ignore checks for possible passwords
     "S105",
     "S106",
@@ -119,10 +117,6 @@ ignore = [
     "PLR0912",
     "PLR0913",
     "PLR0915",
-]
-unfixable = [
-    # Don't touch unused imports
-    "F401",
 ]
 
 [tool.ruff.lint.isort]


CONTRIBUTING.md
```diff
@@ -236,31 +236,31 @@ $ hatch run test:integration
...
-├── src
-│   └── deepset_haystack
-│       ├── __about__.py
-│       └── __init__.py
-├── tests
-│   └── __init__.py
-├── LICENSE.txt
-├── README.md
-└── pyproject.toml
+python scripts/create_new_integration.py --name deepset --type generators
```
anakin87:
```suggestion
python scripts/create_new_integration.py --name YOUR_INTEGRATION_NAME --type generators
```

scripts/create_new_integration.py
```diff
@@ -0,0 +1,142 @@
...
+            sys.exit(1)
+
+    component_type = args.component_type
+    if component_type is None:
+        component_type = prompt_component_type()
+    elif component_type not in COMPONENT_TYPES:
+        print(
+            f"  Error: Unknown component type: '{component_type}'. Must be one of: {', '.join(COMPONENT_TYPES)}.",
+            file=sys.stderr,
+        )
```
anakin87: 1. Translators do not seem to be covered
2. It might happen that we create a new integration with components we never introduced before, so I'd keep this part manually customizable

scripts/create_new_integration.py
```diff
@@ -0,0 +1,142 @@
...
+
+    readme_path = update_root_readme(name, component_type, TYPE_LABELS, repo_root=REPO_ROOT)
+    print(f"  Updated: {readme_path}")
+
+    print(f"\nDone! Next steps:")
+    print(f"  1. Add your component code in integrations/{name}/src/")
+    print(f"  2. Export the component classes from the __init__.py in your module")
+    print(f"  3. Add integration-specific dependencies to integrations/{name}/pyproject.toml")
+    print(f"  4. Add tests in integrations/{name}/tests/")
+    print(f"  5. If your integration tests need API keys, add the secret to the GitHub repo settings")
```
anakin87: Only a few people can add Secrets to GitHub, so I'd mention that you need to contact a maintainer for this step

scripts/utils/__init__.py
```diff
@@ -0,0 +1,20 @@
+from .naming import folder_to_label, folder_to_package, get_module_path, validate_name
```
anakin87: do we really need to re-export all these symbols?
It seems that we import them from their precise locations, so we might keep this file empty.
(Haven't tried)

scripts/utils/templates.py
```diff
@@ -0,0 +1,315 @@
...
+
+
+def pydoc_config(name: str, component_type: str) -> str:
+    """Return the pydoc docusaurus config YAML for the integration."""
+    mod = get_module_path(name, component_type)
+    title = name.replace("_", " ").title()
+    return textwrap.dedent(f"""\
+        loaders:
+          - modules:
+              - {mod}.{component_type[:-1]}
```
anakin87: For tracing (special case), this would produce 
`      - haystack_integrations.components.tracing.integration_name.tracin`

Maybe it's better to handle this with an explicit mapping or in any case consider this special case

scripts/utils/templates.py
```diff
@@ -0,0 +1,315 @@
+"""Template strings for scaffolded integration files."""
+
```
anakin87: Personal taste: I don't particularly like having templates as Python code. I'd prefer to have them as separate template files.
You could explore using https://docs.python.org/3/library/string.html#string.Template.safe_substitute.
I discarded Jinja2 because this would require additional dependencies.

This is just a suggestion, if easy to implement (not a requirement).

scripts/utils/templates.py
bogdankostic: Good catch!

scripts/utils/templates/pyproject.toml
```diff
@@ -0,0 +1,154 @@
...
+    "W",
+    "YTT",
+]
+ignore = [
+    # Allow non-abstract empty methods in abstract base classes
+    "B027",
+    # Allow function calls in argument defaults (common Haystack pattern for Secret.from_env_var)
+    "B008",
+    # zip() without strict= is OK when lengths are guaranteed equal
+    "B905",
```
anakin87:
```suggestion
```

we are now trying to adopt this rule, so let's not ignore it by default

scripts/utils/templates/pyproject.toml
```diff
@@ -0,0 +1,154 @@
...
+    "C901",
+    "PLR0911",
+    "PLR0912",
+    "PLR0913",
+    "PLR0915",
+]
+unfixable = [
+    # Don't touch unused imports
+    "F401",
+]
```
anakin87:
```suggestion
```

same for this rule

PR #2911, opened 2026-03-05
ci: send Slack notifications on CI failures

anakin87:
### Related Issues

- part of https://github.com/deepset-ai/haystack-private/issues/256

### Proposed Changes:

Instead of sending DD events for CI failures and then Slack notifications from DD monitors, skip DD

- uses the composite action https://github.com/deepset-ai/notify-slack-action
- adopt the composite action in all integrations workflows
  - to make sure we get a single notification when a matrix workflow fails (not one for each element in the matrix), I created a new lightweight job (uses ubuntu-slim)
- updated the template for new integrations to use this new pattern
- removed the old send_failure composite action (integrating with DD)
   
To make this work, before opening this PR, I completed the setup described in the [Slack documentation](https://docs.slack.dev/tools/slack-github-action/sending-techniques/sending-data-slack-incoming-webhook/#setup).

### How did you test it?
In this PR, I temporarily made Anthropic workflow fail and got the expected notification.

image

Since this was a test, I avoided tagging people. The actual action would tag `@here` in the Slack channel.

### Checklist

### DIFF
--- a/.github/workflows/anthropic.yml
+++ b/.github/workflows/anthropic.yml
@@ -70,7 +70,6 @@ jobs:
           hatch -e test env run -- uv pip install git+https://github.com/deepset-ai/haystack.git@main
           hatch run test:unit
 
-
   notify-slack-on-failure:
     needs: run
     if: failure() && github.event_name == 'schedule'


.github/workflows/CI_license_compliance.yml
```diff
@@ -90,11 +89,8 @@ jobs:
...
         if: failure() && github.event_name == 'schedule'
-        uses: ./.github/actions/send_failure
+        uses: deepset-ai/notify-slack-action@v1
         with:
-          title: |
-            Core integrations license compliance nightly failure: ${{ github.workflow }}
-          api-key: ${{ secrets.CORE_DATADOG_API_KEY }}
-          
\ No newline at end of file
+          slack-webhook-url: ${{ secrets.SLACK_WEBHOOK_URL_NOTIFICATIONS }}
```
bogdankostic, line 96: In all other workflow files we have the notification step as a separate job.

.github/workflows/anthropic.yml
```diff
@@ -70,10 +70,12 @@ jobs:
           hatch -e test env run -- uv pip install git+https://github.com/deepset-ai/haystack.git@main
           hatch run test:unit
 
-      - name: Send event to Datadog for nightly failures
-        if: failure() && github.event_name == 'schedule'
-        uses: ./.github/actions/send_failure
+
```
bogdankostic: One blank line should be enough.

```suggestion
```

.github/workflows/CI_license_compliance.yml
anakin87: Yes, in other workflows, we are using a separate job for this reason
> to make sure we get a single notification when a matrix workflow fails (not one for each element in the matrix), I created a new lightweight job (uses ubuntu-slim)

In this case, we don't need a matrix: the license check only needs to run on a single operating system and with one Python version. Where possible, it's better to have a step instead of a job because GitHub does not need to spin up a new container.

.github/workflows/anthropic.yml
anakin87: done in https://github.com/deepset-ai/haystack-core-integrations/pull/2911/commits/f77dfcbf723266fbc8572ce3e58aa7877bf1fb05

PR #2938, opened 2026-03-10
fix: `OpenSearchDocumentStore` AsyncAWSAuth method signature changed

### LINKED ISSUE #2937: OpenSearch: Async AWS Auth method signature changed
**Describe the bug**
### The Bug

**opensearch-haystack versions 6.4.0 and 6.5.0** have an `AsyncAWSAuth` wrapper class with an incompatible signature for **opensearch-py 3.1.0**.

### API Mismatch

**opensearch-py 3.1.0 calls auth handlers with:**
```python
self._http_auth(method=method, url=url, body=body, headers=req_headers)
```

**opensearch-haystack 6.4.0/6.5.0 AsyncAWSAuth expects:**
```python
def __call__(self, method: str, url: str, query_string: str, body: Any)
```

### The Problem

1. **opensearch-py 3.1.0** passes `headers` as a keyword argument (NEW!)  
2. **opensearch-py 3.1.0** does NOT pass `query_string` (REMOVED!)
3. **opensearch-haystack AsyncAWSAuth** expects `query_string`, not `headers`
4. Result: `TypeError: AsyncAWSAuth.__call__() got an unexpected keyword argument 'headers'`

---

## 📍 Exact Location of Bug

**File:** `haystack_integrations/document_stores/opensearch/document_store.py`  
**Class:** `AsyncAWSAuth`  
**Method:** `__call__`

**Buggy Code (6.4.0 and 6.5.0):**
```python
def __call__(self, method: str, url: str, query_string: str, body: Any) -> dict[str, str]:
    """Signs the request and returns headers."""
    return self._async_aws_v4_signer_auth(method, url, query_string, body)
```

---

## 🔧 Required Fix

opensearch-haystack needs to update `AsyncAWSAuth.__call__` to match opensearch-py 3.x API:

**Fixed Code:**
```python
def __call__(self, method: str, url: str, body: Any = None, headers: dict[str, str] | None = None) -> dict[str, str]:
    """Signs the request and returns headers."""
    # opensearch-py 3.x API: AWSV4SignerAsyncAuth expects (method, url, body, headers)
    # Note: query_string parameter was removed in 3.x
    return self._async_aws_v4_signer_auth(method, url, body, headers)
```

---

**To Reproduce**
✅ **opensearch-py 3.1.0:**
- `AWSV4SignerAsyncAuth.__call__` signature includes `headers` parameter
- `AsyncHttpConnection.perform_request` passes `headers=req_headers`

✅ **opensearch-haystack 6.4.0:**
- `AsyncAWSAuth.__call__` does NOT accept `headers` parameter
- Expects `query_string` which opensearch-py 3.x no longer provides

| Package | Version | Status |
|---------|---------|--------|
| opensearch-py | 3.1.0 | ✅ Has breaking change |
| opensearch-haystack | 6.4.0 | ❌ Incompatible |
| haystack-ai | 2.25.2 | ✅ OK |

davidsbatista:
### Related Issues

- fixes #2937 

### Proposed Changes:

- changed the signature of the AWSAuth __call__ in both sync and async due to upgrade of the `opensearch-py` dependency 

### How did you test it?

- updated the `test_call` assertion to expect the signer to be called with four arguments: ("GET", "http://some.url", "some body", None).
- New `test_call_with_headers` that calls the sync auth with `headers={"Host": "localhost"}` and checks the signer is invoked with those four arguments.

### Checklist

### DIFF
--- a/integrations/opensearch/tests/test_auth.py
+++ b/integrations/opensearch/tests/test_auth.py
@@ -153,9 +153,10 @@ class TestAWSAuth:
 
 class TestAWSAuthRealSigning:
     """
-    Single test that uses real boto3.Session and real signing (no mocks on Session or signer).
-    
-    Verifies AWSAuth.__call__(method, url, body, headers) produces valid SigV4 headers.
+    Tests that use real boto3.Session and real signing (no mocks on Session or signer).
+
+    Verifies both AWSAuth and AsyncAWSAuth produce valid SigV4 headers when called
+    with (method, url, body, headers) as opensearch-py 3.x does.
     """
 
     def test_sync_auth_real_signing_no_network(self, monkeypatch):
@@ -167,6 +168,27 @@ class TestAWSAuthRealSigning:
         aws_auth = AWSAuth()
 
         signed_headers = aws_auth(
+            method="GET",
+            url="https://example.com/_search?q=test",
+            body=b"",
+            headers={"Host": "example.com"},
+        )
+
+        assert "Authorization" in signed_headers
+        assert signed_headers["Authorization"].startswith("AWS4-HMAC-SHA256 ")
+        assert "X-Amz-Date" in signed_headers
+        assert "X-Amz-Content-SHA256" in signed_headers
+
+    def test_async_auth_real_signing_no_network(self, monkeypatch):
+        monkeypatch.setenv("AWS_EC2_METADATA_DISABLED", "true")
+        monkeypatch.setenv("AWS_ACCESS_KEY_ID", "AKIDEXAMPLE")
+        monkeypatch.setenv("AWS_SECRET_ACCESS_KEY", "wJalrXUtnFEMI/K7MDENG+bPxRfiCYEXAMPLEKEY")
+        monkeypatch.setenv("AWS_DEFAULT_REGION", "us-east-1")
+
+        aws_auth = AWSAuth()
+        async_aws_auth = AsyncAWSAuth(aws_auth)
+
+        signed_headers = async_aws_auth(
             method="GET",
             url="https://example.com/_search?q=test",
             body=b"",


integrations/opensearch/tests/test_auth.py
```diff
@@ -120,15 +120,25 @@ def test_call(self, _get_aws_v4_signer_auth):
...
+        signer_auth_mock.assert_called_once_with("GET", "http://some.url", "some body", None)
+
+    @patch("haystack_integrations.document_stores.opensearch.auth.AWSAuth._get_aws_v4_signer_auth")
+    def test_call_with_headers(self, _get_aws_v4_signer_auth):
+        """AWSAuth accepts optional headers for opensearch-py 3.x sync connection compatibility."""
+        signer_auth_mock = Mock()
+        _get_aws_v4_signer_auth.return_value = signer_auth_mock
+        aws_auth = AWSAuth()
+        aws_auth(method="GET", url="http://some.url", body="some body", headers={"Host": "localhost"})
+        signer_auth_mock.assert_called_once_with("GET", "http://some.url", "some body", {"Host": "localhost"})
```
anakin87: My impression is that the previous bug slipped in in because the existing tests are mocking too much, so in case AWS signatures change again in the future, we might not be notified.

I'd prefer to have a local test that actually calls AWS method.

Something like (asked a LLM, could be totally wrong, just take the spirit)
```python
def test_sync_auth_real_signing_no_network(monkeypatch):
    # Prevent any metadata/IMDS probing (which can cause network calls/timeouts)
    monkeypatch.setenv("AWS_EC2_METADATA_DISABLED", "true")

    # Provide deterministic local credentials + region
    monkeypatch.setenv("AWS_ACCESS_KEY_ID", "AKIDEXAMPLE")
    monkeypatch.setenv("AWS_SECRET_ACCESS_KEY", "wJalrXUtnFEMI/K7MDENG+bPxRfiCYEXAMPLEKEY")
    monkeypatch.setenv("AWS_DEFAULT_REGION", "us-east-1")

    from haystack_integrations.document_stores.opensearch.auth import AWSAuth

    aws_auth = AWSAuth()  # uses real _get_aws_v4_signer_auth and real boto session resolution

    signed_headers = aws_auth(
        method="GET",
        url="https://example.com/_search?q=test",
        body=b"",
        headers={"Host": "example.com"},
    )

    assert "Authorization" in signed_headers
    assert signed_headers["Authorization"].startswith("AWS4-HMAC-SHA256 ")
    assert "X-Amz-Date" in signed_headers
    assert "X-Amz-Content-SHA256" in signed_headers
```

davidsbatista: that's a good point - I will update the tests

integrations/opensearch/tests/test_auth.py
```diff
@@ -150,6 +151,34 @@ def test_async_aws_auth_init(self):
...
+            body=b"",
+            headers={"Host": "example.com"},
+        )
+
+        assert "Authorization" in signed_headers
+        assert signed_headers["Authorization"].startswith("AWS4-HMAC-SHA256 ")
+        assert "X-Amz-Date" in signed_headers
+        assert "X-Amz-Content-SHA256" in signed_headers
+
+
```
anakin87, line 203: to be extra-safe, let's also add a similar test for `async_aws_auth`

PR #2951, opened 2026-03-11
fix: fix wrong batching in Google GenAI Document Embedder

anakin87:
### Related Issues
We noticed that the Google GenAI Document Embedder behaved differently than when using the Google API directly, leading to non-meaningful results for text retrieval.

After investigation, it turned out that the batching mechanism was wrong, actually passing 1-character texts to the embedding API

### Proposed Changes:
- fix batching
- add unit and integration test to catch this behavior in the future

### How did you test it?
CI, new tests

### Checklist


sjrl (approved):
Thanks for the fix! Apologies for missing this on the original review of this contribution.

PR #2964, opened 2026-03-16
fix: PgVectorDocumentStore `_treat_meta_field` and comparison functions now return Composed - string escaping done by `psycopg`

davidsbatista:
### Related Issues

- fixes #2881 

### Proposed Changes:

- `_treat_meta_field` returns a `psycopg.sql.Composed` object and using `psycopg.sql.Literal` to embed the field name, so string escaping is now done by psycopg itself
- filtering helpers (_equal, _not_equal, _greater_than, etc.) have been updated to accept and return Composed objects

### How did you test it?

- filtering tests updated
- Added a test demonstrating the old string-interpolation approach was vulnerable and that the new approach safely escapes the injection attempt via `SQLLiteral`

### Checklist

### DIFF
--- a/integrations/mistral/tests/test_mistral_chat_generator.py
+++ b/integrations/mistral/tests/test_mistral_chat_generator.py
@@ -576,7 +576,6 @@ class TestMistralChatGenerator:
         results = component.run(chat_messages)
         assert len(results["replies"]) == 1
         message = results["replies"][0]
-        assert message.text == ""
 
         assert message.tool_calls
         tool_call = message.tool_call
--- a/integrations/weaviate/CHANGELOG.md
+++ b/integrations/weaviate/CHANGELOG.md
@@ -1,5 +1,12 @@
 # Changelog
 
+## [integrations/weaviate-v7.4.0] - 2026-03-17
+
+### 🐛 Bug Fixes
+
+- Weaviate - stop ignoring _split_overlap meta field (#2966)
+
+
 ## [integrations/weaviate-v7.3.0] - 2026-03-11
 
 ### 🚀 Features
--- a/integrations/weaviate/src/haystack_integrations/document_stores/weaviate/document_store.py
+++ b/integrations/weaviate/src/haystack_integrations/document_stores/weaviate/document_store.py
@@ -745,15 +745,6 @@ class WeaviateDocumentStore:
         # The embedding vector is stored separately from the rest of the data
         del data["embedding"]
 
-        # _split_overlap meta field is unsupported because of a bug
-        # https://github.com/deepset-ai/haystack-core-integrations/issues/1172
-        if "_split_overlap" in data:
-            data.pop("_split_overlap")
-            logger.warning(
-                "Document {id} has the unsupported `_split_overlap` meta field. It will be ignored.",
-                id=data["_original_id"],
-            )
-
         if "sparse_embedding" in data:
             sparse_embedding = data.pop("sparse_embedding", None)
             if sparse_embedding:
--- a/integrations/weaviate/tests/test_document_store.py
+++ b/integrations/weaviate/tests/test_document_store.py
@@ -464,29 +464,32 @@ class TestWeaviateDocumentStore(DocumentStoreBaseExtendedTests):
             ],
         )
 
-    def test_meta_split_overlap_is_skipped(self, document_store):
+    def test_split_overlap_preserved(self, document_store):
+        """Split overlap meta is written and read back correctly."""
+        overlap = [
+            {"range": [3.0, 13.0], "doc_id": "34326b7e6be489cb4c031152fc378cb50479ca5fcc3861e7e61dfb2e4e4e968b"},
+            {"range": [0.0, 13.0], "doc_id": "780f791c09d499c0bf01f87bce047b45c44224d36c79f0c9d8c1405a3197fc1a"},
+        ]
         doc = Document(
-            content="The moonlight shimmered ",
+            id="6edd24e8b01f3cd6e4b71fef7d57b52f17664e14db5ab01b8ef429f97add3620",
+            content="an eighth test. ",
             meta={
-                "source_id": "62049ba1d1e1d5ebb1f6230b0b00c5356b8706c56e0b9c36b1dfc86084cd75f0",
-                "page_number": 1,
-                "split_id": 0,
-                "split_idx_start": 0,
-                "_split_overlap": [
-                    {"doc_id": "68ed48ba830048c5d7815874ed2de794722e6d10866b6c55349a914fd9a0df65", "range": (0, 20)}
-                ],
+                "_split_overlap": overlap,
+                "page_number": 1.0,
+                "split_id": 33.0,
+                "split_idx_start": 159.0,
+                "source_id": "fdbde6d217f04d3dd60c01f36541794f3153a61f13b4ca669655f4c5610c1664",
             },
         )
         document_store.write_documents([doc])
-
         written_doc = document_store.filter_documents()[0]
-
-        assert written_doc.content == "The moonlight shimmered "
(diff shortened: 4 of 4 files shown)


integrations/pgvector/src/haystack_integrations/document_stores/pgvector/filters.py
```diff
@@ -84,17 +83,14 @@ def _parse_logical_condition(condition: dict[str, Any]) -> tuple[str, list[Any]]
...
     if operator == "AND":
-        sql_query = f"({' AND '.join(query_parts)})"
-    elif operator == "OR":
-        sql_query = f"({' OR '.join(query_parts)})"
-    else:
-        msg = f"Unknown logical operator '{operator}'"
-        raise FilterError(msg)
+        sql_query = SQL("(") + SQL(" AND ").join(query_parts) + SQL(")")
+    else:  # operator == "OR"
+        sql_query = SQL("(") + SQL(" OR ").join(query_parts) + SQL(")")
```
anakin87, line 88: I'd keep the previous behavior: if operator is not AND or OR, raise an error

integrations/pgvector/src/haystack_integrations/document_stores/pgvector/filters.py
```diff
@@ -110,34 +106,34 @@ def _parse_comparison_condition(condition: dict[str, Any]) -> tuple[str, list[An
...
     >>> _treat_meta_field(field="meta.number", value=9)
-    "(meta->>'number')::integer"
+    Composed([SQL('(meta->>'), Literal('number'), SQL(')::integer')])
 
     >>> _treat_meta_field(field="meta.name", value="my_name")
-    "meta->>'name'"
+    Composed([SQL('meta->>'), Literal('name')])
     """
-
-    # use the ->> operator to access keys in the meta JSONB field
```
anakin87, line 132: this comment is still valid: let's merge it with the comment below.
For a reader, it could be non-trivial to understand what "->>" does

integrations/pgvector/src/haystack_integrations/document_stores/pgvector/filters.py
```diff
@@ -110,34 +106,34 @@ def _parse_comparison_condition(condition: dict[str, Any]) -> tuple[str, list[An
     value: Any = condition["value"]
 
     if field.startswith("meta."):
-        field = _treat_meta_field(field, value)
+        sql_field: str | Composable = _treat_meta_field(field, value)
+    else:
+        sql_field = field
```
anakin87: 1. If we convert the non-meta field to a Composable, we can get rid of `str` everywhere
2. I'd use `Identifier(field)` (not `SQL(str)`). [psycopg docs](https://www.psycopg.org/psycopg3/docs/api/sql.html#psycopg.sql.Identifier).

integrations/pgvector/tests/test_filters.py
```diff
@@ -129,24 +134,38 @@ def test_complex_filter(self, document_store, filterable_docs):
...
+        "bool"
+    ) + SQL(")::boolean")
 
     # do not cast the field if its value is not one of the known types, an empty list or None
-    assert _treat_meta_field(field="meta.other", value={"a": 3, "b": "example"}) == "meta->>'other'"
-    assert _treat_meta_field(field="meta.empty_list", value=[]) == "meta->>'empty_list'"
-    assert _treat_meta_field(field="meta.name", value=None) == "meta->>'name'"
+    assert _treat_meta_field(field="meta.other", value={"a": 3, "b": "example"}) == SQL("meta->>") + SQLLiteral("other")
+    assert _treat_meta_field(field="meta.empty_list", value=[]) == SQL("meta->>") + SQLLiteral("empty_list")
+    assert _treat_meta_field(field="meta.name", value=None) == SQL("meta->>") + SQLLiteral("name")
```
anakin87: Let's discuss tests: I understand that the return type has changed, so we need to do something like this, which, unfortunately, is hard to read/maintain.

An alternative would be to build a psycopg `cursor` and use it to convert SQL to str, keeping the test similar to the previous one. An example for `test_treat_meta_field`:

```python
def test_treat_meta_field(document_store):
    document_store._ensure_db_setup()
    cursor = document_store._cursor

    expr = _treat_meta_field(field="meta.number", value=9)
    assert expr.as_string(cursor) == "(meta->>'number')::integer"
    ...
```

The only problem with this approach: the test is no longer a pure unit test since it needs a DB connection. But we keep readability. WDYT?

integrations/pgvector/src/haystack_integrations/document_stores/pgvector/filters.py
davidsbatista: This is already checked on line 65:

```python
    if operator not in ["AND", "OR"]:
        msg = f"Unknown logical operator '{operator}'. Valid operators are: 'AND', 'OR'"
        raise FilterError(msg)

```

anakin87: ah, right!

integrations/pgvector/src/haystack_integrations/document_stores/pgvector/filters.py
davidsbatista: nice - good catch 👍🏽

integrations/pgvector/src/haystack_integrations/document_stores/pgvector/filters.py
davidsbatista: adding it back, but being this an internal/private method, I wouldn't worry that much

integrations/pgvector/tests/test_filters.py
davidsbatista: What about something like this:

```python
def test_treat_meta_field():
      cast_integer = SQL("(") + SQL("meta->>") + SQLLiteral("number") + SQL(")::integer")
      no_cast_name  = SQL("meta->>") + SQLLiteral("name")
      cast_real     = SQL("(") + SQL("meta->>") + SQLLiteral("number") + SQL(")::real")
      cast_boolean  = SQL("(") + SQL("meta->>") + SQLLiteral("bool") + SQL(")::boolean")

      assert _treat_meta_field(field="meta.number", value=9) == cast_integer
      assert _treat_meta_field(field="meta.number", value=[1, 2, 3]) == cast_integer
      assert _treat_meta_field(field="meta.name", value="my_name") == no_cast_name
      assert _treat_meta_field(field="meta.name", value=["my_name"]) == no_cast_name
      assert _treat_meta_field(field="meta.number", value=1.1) == cast_real
      assert _treat_meta_field(field="meta.number", value=[1.1, 2.2]) == cast_real
      assert _treat_meta_field(field="meta.bool", value=True) == cast_boolean
      assert _treat_meta_field(field="meta.bool", value=[True, False]) == cast_boolean
```

it compares two `Composed` instances

anakin87: This test remains quite understandable in any case.
~`test_comparison_condition_missing_operator`~ `test_logical_condition_nested` is hard to read instead.

For this reason, I was advocating to resolve to string in the test.

What's your opinion?

davidsbatista: I'm confused.

To make sure I understand your point.

You agree with the approach for the test `test_treat_meta_field` as is, but instead of `repr()` you suggest having it as a string to compare the full output - is that it?

Regarding the `test_comparison_condition_missing_operator`- it's not changed by this PR - but I miss what you are suggesting.

anakin87: Sorry, I was pointing to the wrong test. It's `test_logical_condition_nested` that gets a bit hard to read.

If we want to resolve to a string for easier comparison, we can use a cursor as I proposed in https://github.com/deepset-ai/haystack-core-integrations/pull/2964#discussion_r2945270210

In any case, I'll leave this decision up to you.

PR #2934, opened 2026-03-10
feat: add str handling for ElasticsearchDocumentStore api_key

tstadel:
### Related Issues

- fixes ElasticSearchDocumentStore not deserializable without a env var secret

### Proposed Changes:
- document store credentials can be passed as str (similar to OpenSearchDocumentStore)

### How did you test it?
- added tests

### Notes for the reviewer
- we can't use `TokenSecret` as it is not deserializable by design

### Checklist

- I have read the [contributors guidelines](https://github.com/deepset-ai/haystack-core-integrations/blob/main/CONTRIBUTING.md) and the [code of conduct](https://github.com/deepset-ai/haystack-core-integrations/blob/main/CODE_OF_CONDUCT.md)
- I have updated the related issue with new insights and changes
- I added unit tests and updated the docstrings
- I've used one of the [conventional commit types](https://www.conventionalcommits.org/en/v1.0.0/) for my PR title: `fix:`, `feat:`, `build:`, `chore:`, `ci:`, `docs:`, `style:`, `refactor:`, `perf:`, `test:`.

PR #2967, opened 2026-03-17
fix: OpenSearch - do not serialize string Secrets + authentication refactoring

anakin87:
### Related Issues

- fixes https://github.com/deepset-ai/haystack-private/issues/268

### Proposed Changes:
- avoid serializing plain string Secrets
- simplify authentication handling, similarly to Elasticsearch:
  - do not resolve Secrets in `__init__` but in `_ensure_initialized`
  - fail if only one of username and password is provided
  - properly serialize env vars Secrets

### How did you test it?
CI, new unit tests

### Notes for the reviewer
To be on the safe side, I'd release a major version.
However, I saw this comment: https://github.com/deepset-ai/haystack-private/issues/268#issuecomment-4055392995

WDYT?

### Checklist

### DIFF
--- a/integrations/opensearch/src/haystack_integrations/document_stores/opensearch/document_store.py
+++ b/integrations/opensearch/src/haystack_integrations/document_stores/opensearch/document_store.py
@@ -82,7 +82,7 @@ class OpenSearchDocumentStore:
         mappings: dict[str, Any] | None = None,
         settings: dict[str, Any] | None = DEFAULT_SETTINGS,
         create_index: bool = True,
-        http_auth: Any = (
+        http_auth: tuple[Secret, Secret] | tuple[str, str] | list[str] | str | AWSAuth | None = (
             Secret.from_env_var("OPENSEARCH_USERNAME", strict=False),  # noqa: B008
             Secret.from_env_var("OPENSEARCH_PASSWORD", strict=False),  # noqa: B008
         ),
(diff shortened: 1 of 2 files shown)


integrations/opensearch/src/haystack_integrations/document_stores/opensearch/document_store.py
sjrl: Let's also update the types here to be more explicit than `Any`. I believe we also need to support plain strings coming in like we did in Elastic Search https://github.com/deepset-ai/haystack-core-integrations/pull/2934

anakin87: Done in https://github.com/deepset-ai/haystack-core-integrations/pull/2967/commits/3f90d268ffeab5fde54b1dd28979e0c9b8d75ef8.

This made the code a bit more verbose but it's worth it.

sjrl (approved):
So I'd say this looks all good from my point of view. It would be great if @tstadel could give it a final look over.

integrations/opensearch/src/haystack_integrations/document_stores/opensearch/document_store.py
```diff
@@ -82,7 +82,7 @@ def __init__(
         mappings: dict[str, Any] | None = None,
         settings: dict[str, Any] | None = DEFAULT_SETTINGS,
         create_index: bool = True,
-        http_auth: Any = (
+        http_auth: tuple[Secret, Secret] | tuple[str, str] | list[str] | str | AWSAuth | None = (
```
tstadel, line 85: I guess that's fine to streamline, to be 100% accurrate better would be to also allow a callable/prototype with this signature (same as AWSAuth's __call__)
```
    def __call__(
        self,
        method: str,
        url: str,
        body: Any = None,
        headers: dict[str, str] | None = None,
    ) -> dict[str, str]:
```

anakin87: Makes sense, but we only support the explicitly handled `http_auth` types via `_resolve_http_auth()`. 

Adding a generic callable would be misleading unless we also pass it through at runtime (which we don't do in the current implementation), so I'd keep the type strict for now

Issue #2963, opened 2026-03-16
Update core integrations to test on python 3.14


sjrl:
As a follow up to https://github.com/deepset-ai/haystack/issues/10509 we should go through our integrations and update them to test on python 3.14 and if able/necessary update the code to support python 3.14

To avoid running too many tests we should replace the python 3.13 testing in our github workflows to python 3.14 and let's make sure to add python 3.14 listed support to our pyprojec.toml under the `classifiers` label

PR #2979, opened 2026-03-18
feat: Add Cognee integration

hande-k:
Closes https://github.com/deepset-ai/haystack-private/issues/240

## Summary
- Adds Cognee integration with 4 components: `CogneeWriter`, `CogneeCognifier`, `CogneeRetriever`, and `CogneeMemoryStore`
- `CogneeWriter` ingests Haystack Documents into Cognee memory via `cognee.add()` + optional `cognee.cognify()`
- `CogneeRetriever` searches Cognee's memory and returns Haystack Documents
- `CogneeCognifier` wraps `cognee.cognify()` as a standalone pipeline step
- `CogneeMemoryStore` implements the `MemoryStore` protocol from haystack-experimental for use with Haystack's experimental Agent

## Test plan
- [x] Unit tests pass via `hatch run test:unit`
- [x] Linting passes via `hatch run fmt-check`
- [x] Type checking passes via `hatch run test:types`
- [x] Demo scripts tested manually (`demo_pipeline.py`, `demo_memory_agent.py`)


integrations/cognee/src/haystack_integrations/components/connectors/cognee/memory_store.py
```diff
@@ -0,0 +1,171 @@
...
+        """
+        self.search_type = search_type
+        self.top_k = top_k
+        self.dataset_name = dataset_name
+
+    def add_memories(
+        self,
+        *,
+        messages: list[ChatMessage],
+        user_id: str | None = None,
```
davidsbatista, line 141:
```python
        user_id: str | None = None,
        agent_id: str | None = None,
        run_id: str | None = None,
        **kwargs: Any,
```

are not being used at all - can be removed

integrations/cognee/src/haystack_integrations/components/connectors/cognee/memory_store.py
```diff
@@ -0,0 +1,171 @@
...
+
+        run_sync(cognee.cognify())
+        logger.info("Added and cognified {count} messages as memories", count=len(messages))
+
+    def search_memories(
+        self,
+        *,
+        query: str | None = None,
+        top_k: int = 5,
+        user_id: str | None = None,
```
davidsbatista, line 193:
```python
        user_id: str | None = None,
        agent_id: str | None = None,
        run_id: str | None = None,
        **kwargs: Any,
```

same as above

integrations/cognee/src/haystack_integrations/components/connectors/cognee/memory_store.py
```diff
@@ -0,0 +1,171 @@
...
+
+        for item in raw_results[:effective_top_k]:
+            text = _extract_memory_text(item)
+            if text:
+                memories.append(ChatMessage.from_system(text))
+
+        logger.info("Found {count} memories for query '{query}'", count=len(memories), query=query[:80])
+        return memories
+
+    def delete_all_memories(
```
davidsbatista: this can be converted to a `@staticmethod`

integrations/cognee/src/haystack_integrations/components/connectors/cognee/memory_store.py
```diff
@@ -0,0 +1,171 @@
...
+        self,
+        *,
+        user_id: str | None = None,
```
davidsbatista:
```python
        user_id: str | None = None,
        agent_id: str | None = None,
        run_id: str | None = None,
        **kwargs: Any,
```

also not being used, can be removed

integrations/cognee/src/haystack_integrations/components/connectors/cognee/memory_store.py
```diff
@@ -0,0 +1,171 @@
...
+        :param agent_id: Optional agent identifier (reserved for future use).
+        :param run_id: Optional run identifier (reserved for future use).
+        """
+        for msg in messages:
+            text = msg.text
+            if not text:
+                continue
+            run_sync(cognee.add(text, dataset_name=self.dataset_name))
+
+        run_sync(cognee.cognify())
```
davidsbatista: This will always run even when no new messages are added. We can add a safeguard

davidsbatista: And I think we should cognify only the dataset we are adding messages to 

```python
run_sync(cognee.cognify(datasets=[self.dataset_name]))
```

integrations/cognee/src/haystack_integrations/components/connectors/cognee/writer.py
```diff
@@ -0,0 +1,79 @@
...
+            run_sync(cognee.add(doc.content, dataset_name=self.dataset_name))
+            written += 1
+
+        if self.auto_cognify and written > 0:
+            logger.info(
+                "Cognifying {count} documents in dataset '{dataset}'",
+                count=written,
+                dataset=self.dataset_name,
+            )
+            run_sync(cognee.cognify())
```
davidsbatista: We should also only cognify the dataset we are working on, e.g:

```python
run_sync(cognee.cognify(datasets=[self.dataset_name]))
```

integrations/cognee/src/haystack_integrations/components/connectors/cognee/memory_store.py
```diff
@@ -0,0 +1,171 @@
...
+            top_k=self.top_k,
+            dataset_name=self.dataset_name,
+        )
+
+    @classmethod
+    def from_dict(cls, data: dict[str, Any]) -> "CogneeMemoryStore":
+        return default_from_dict(cls, data)
+
+
+def _extract_memory_text(item: Any) -> str:
```
davidsbatista: This is duplicated in `retriever.py` - you can added to utils.py and call it both here and in the retriever.py

integrations/cognee/pyproject.toml
```diff
@@ -0,0 +1,160 @@
...
+  "Programming Language :: Python :: 3.11",
+  "Programming Language :: Python :: 3.12",
+  "Programming Language :: Python :: 3.13",
+  "Programming Language :: Python :: Implementation :: CPython",
+  "Programming Language :: Python :: Implementation :: PyPy",
+]
+dependencies = [
+  "haystack-ai>=2.24.0",
+  "cognee==0.5.4",
+  "mistralai>=1.0,<2.0",
```
davidsbatista: I think this can be removed - is it necessary?

integrations/cognee/src/haystack_integrations/components/connectors/cognee/writer.py
```diff
@@ -0,0 +1,79 @@
...
+    def run(self, documents: list[Document]) -> dict[str, Any]:
+        """
+        Add documents to Cognee and optionally cognify them.
+
+        :param documents: List of Haystack Documents to add.
+        :returns: Dictionary with key `documents_written` indicating how many
+            documents were successfully added.
+        """
+        written = 0
+        for doc in documents:
```
davidsbatista: cognee.add supports [list[str]](https://docs.cognee.ai/python-api/add)

This will increase performance, a single async call instead of n

```python
texts = [doc.content for doc in documents if doc.content]
run_sync(cognee.add(texts, dataset_name=self.dataset_name))
```

integrations/cognee/src/haystack_integrations/components/connectors/cognee/memory_store.py
```diff
@@ -0,0 +1,171 @@
...
+    ```python
+    from haystack_integrations.components.connectors.cognee import CogneeMemoryStore
+
+    store = CogneeMemoryStore(search_type="GRAPH_COMPLETION", top_k=5)
+    store.add_memories(messages=[ChatMessage.from_user("Remember: the project deadline is Friday.")])
+    results = store.search_memories(query="When is the project deadline?")
+    ```
+    """
+
+    def __init__(self, search_type: str = "GRAPH_COMPLETION", top_k: int = 5, dataset_name: str = "haystack_memory"):
```
davidsbatista: There are several search_types supported https://docs.cognee.ai/python-api/search

instead of a string, this function should accepted something like this for validation


```python
from typing import Literal

  SearchType = Literal[
      "SUMMARIES",
      "CHUNKS",
      "RAG_COMPLETION",
      "TRIPLET_COMPLETION",
      "GRAPH_COMPLETION",
      "GRAPH_SUMMARY_COMPLETION",
      "CYPHER",
      "NATURAL_LANGUAGE",
      "GRAPH_COMPLETION_COT",
      "GRAPH_COMPLETION_CONTEXT_EXTENSION",
      "FEELING_LUCKY",
      "TEMPORAL",
      "CODING_RULES",
      "CHUNKS_LEXICAL",
  ]
```

this can go into a utils.py or types.py file and then use it as

```python
search_type: CogneeSearchType = "GRAPH_COMPLETION"
```

The same can also be applied on the retriever.py

davidsbatista (commented):
@hande-k thank you for this contribution!

I left some initial comments/suggestions for improvements.

integrations/cognee/CHANGELOG.md
```diff
@@ -0,0 +1,3 @@
+# Changelog
+
+<!-- All notable changes to this project will be documented in this file. -->
```
sjrl: This file is auto-generated so should be removed

integrations/cognee/.gitignore
sjrl, line 1: This file should be removed we have a `.gitignore` file on the top-level of this repo. We don't add individual ones to each integration

integrations/cognee/README.md
sjrl, line 1: Please stick to [this format](https://github.com/deepset-ai/haystack-core-integrations/blob/main/integrations/chroma/README.md) and remove everything else.

We'll put these useful examples in documentation page + integration page

integrations/cognee/src/haystack_integrations/components/connectors/cognee/py.typed
sjrl, line 1: This file should live one folder up, so on the same level as the `cognee` folder. `./src/haystack_integrations/components/connectors/py.typed`

integrations/cognee/src/haystack_integrations/components/connectors/cognee/memory_store.py
sjrl: @davidsbatista if we update this to a staticmethod I'm not sure it would be valid with the MemoryStore Protocol anymore

integrations/cognee/src/haystack_integrations/components/retrievers/cognee/memory_retriever.py
sjrl, line 1: I think to follow way we normally structure Haystack this should live in the folder path `integrations/cognee/src/haystack_integrations/components/retrievers/cognee/memory_retriever.py`

So basically this should live in the `components/retrievers` category and not in the `components/connectors` category. What do you think @davidsbatista ?

integrations/cognee/src/haystack_integrations/components/writers/cognee/memory_writer.py
sjrl, line 1: Similar comment here as this one https://github.com/deepset-ai/haystack-core-integrations/pull/2979/changes#r2959708476

I think this should live in `integrations/cognee/src/haystack_integrations/components/writers/cognee/memory_writer.py` WDYT @davidsbatista ?

integrations/cognee/src/haystack_integrations/memory_stores/cognee/memory_store.py
sjrl, line 1: Similar comment here as this one https://github.com/deepset-ai/haystack-core-integrations/pull/2979/changes#r2959708476

This should live in `integrations/cognee/src/haystack_integrations/memory_stores/cognee/memory_store.py`

integrations/cognee/src/haystack_integrations/components/connectors/cognee/cognifier.py
```diff
@@ -0,0 +1,51 @@
...
+
+
+@component
+class CogneeCognifier:
+    """
+    Processes previously added data in Cognee's knowledge engine.
+
+    Wraps `cognee.cognify()` as a standalone pipeline step, useful when you
+    want to separate the add and cognify phases (e.g., batch-add first, then
+    cognify once).
```
sjrl: Could you explain in more detail what `cognee.cognify()` does? It's not clear from the docstrings nor where in a pipeline a user might want to add this. 

Would it be possible to show in one of your example files how this expected to be used?

integrations/cognee/src/haystack_integrations/components/connectors/cognee/cognifier.py
```diff
@@ -0,0 +1,51 @@
...
+    from haystack_integrations.components.connectors.cognee import CogneeCognifier
+
+    cognifier = CogneeCognifier()
+    result = cognifier.run()
+    assert result["cognified"] is True
+    ```
+    """
+
+    @component.output_types(cognified=bool)
+    def run(self) -> dict[str, Any]:
```
sjrl: Related to this comment https://github.com/deepset-ai/haystack-core-integrations/pull/2979/changes#r2959732723 by setting up the run method to not have any input params it's not possible to trigger this component in a Pipeline. 

It would automatically run first when added to a pipeline. If it needs to run only after previous components have run then we should add a input parameter that we can connect to.

integrations/cognee/src/haystack_integrations/components/connectors/cognee/memory_store.py
davidsbatista: I overlooked that.

@hande-k please disregard any of my comments that break the MemoryStore protocol.

integrations/cognee/pyproject.toml
davidsbatista: I mean `mistralai`

integrations/cognee/src/haystack_integrations/components/retrievers/cognee/memory_retriever.py
davidsbatista: although this is a very specific retriever I think still it should be  moved to components/retriever

integrations/cognee/src/haystack_integrations/components/writers/cognee/memory_writer.py
davidsbatista: same thing as above, agree that it should be moved to components/writers/

Although in the future, if we start adding more memorystores we could have a `components.memory_stores.retrievers/writers`

sjrl: I might follow what we did for adding `image` components which would be `components/retrievers/memory/writer.py` similar to `components/converters/image/...`

.github/workflows/cognee.yml
```diff
@@ -0,0 +1,79 @@
...
+
+jobs:
+  run:
+    name: Python ${{ matrix.python-version }} on ${{ startsWith(matrix.os, 'macos-') && 'macOS' || startsWith(matrix.os, 'windows-') && 'Windows' || 'Linux' }}
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ubuntu-latest, windows-latest, macos-latest]
+        python-version: ["3.10", "3.13"]
```
sjrl: We are in the progress of updating our core integrations to test on python 3.14. More info here https://github.com/deepset-ai/haystack-core-integrations/issues/2963

But basically could you replace 3.13 with 3.14?

.github/workflows/cognee.yml
sjrl, line 1: Also we recently updated our github CI files to pin the full SHAs of the specified actions to help boost security. Was done in this PR https://github.com/deepset-ai/haystack-core-integrations/pull/3025

It would be great if you could update this file with the pinned SHAs as well.

integrations/cognee/src/haystack_integrations/components/connectors/cognee/_utils.py
```diff
@@ -0,0 +1,65 @@
...
+    except RuntimeError:
+        return asyncio.run(coro)
+
+    # Already in an async context — run in a thread to avoid nested loop issues
+    with ThreadPoolExecutor(max_workers=1) as pool:
+        future = pool.submit(asyncio.run, coro)
+        return future.result()
+
+
+def extract_text(item: Any) -> str:
```
sjrl: Would it be possible to make the type for `item` more strict? Looking at the code below it looks like it could be one of three things `str`, `dict` and some Cognee specific object.

integrations/cognee/src/haystack_integrations/components/retrievers/cognee/memory_retriever.py
```diff
@@ -0,0 +1,106 @@
...
+            SUMMARIES, INSIGHTS, etc.
+        :param top_k: Maximum number of results to return.
+        :param dataset_name: Optional dataset name to restrict search scope.
+        """
+        self.search_type = search_type
+        self.top_k = top_k
+        self.dataset_name = dataset_name
+
+    @component.output_types(documents=list[Document])
+    def run(self, query: str, top_k: int | None = None) -> dict[str, Any]:
```
sjrl:
```suggestion
    def run(self, query: str, top_k: int | None = None) -> dict[str, list[Document]]:
```

integrations/cognee/src/haystack_integrations/memory_stores/cognee/memory_store.py
```diff
@@ -0,0 +1,123 @@
...
+    ```
+    """
+
+    def __init__(
+        self, search_type: CogneeSearchType = "GRAPH_COMPLETION", top_k: int = 5, dataset_name: str = "haystack_memory"
+    ):
+        """
+        :param search_type: Cognee search type for memory retrieval.
+        :param top_k: Default number of results for memory search.
+        :param dataset_name: Cognee dataset name for storing memories.
```
sjrl: Can you help me understand Cognee a bit better. It looks like from your code that `dataset_name` is required which makes sense. 

My question is what level of scope is `dataset_name` meant to have? For example, is it normal to have a new `dataset_name` per user or is the intention for dataset to be scoped to multiple users (e.g. like at an org level)?

integrations/cognee/tests/test_memory_store.py
sjrl, line 1: It would be great to also add some integration tests that require access to a Cognee account. We then mark those test with a header like
```python
@pytest.mark.skipif(
    not os.environ.get("COGNEE_API_KEY", None),
    reason="Export an env var called COGNEE_API_KEY containing the Cognee API key to run this test.",
)
@pytest.mark.integration
```

integrations/cognee/tests/__init__.py
sjrl, line 1: Could we also add tests for `CogneeCognifier` and for the `connectors/cognee/_utils.py` file?

integrations/cognee/pyproject.toml
```diff
@@ -0,0 +1,159 @@
...
+  "Programming Language :: Python :: 3.10",
+  "Programming Language :: Python :: 3.11",
+  "Programming Language :: Python :: 3.12",
+  "Programming Language :: Python :: 3.13",
+  "Programming Language :: Python :: Implementation :: CPython",
+  "Programming Language :: Python :: Implementation :: PyPy",
+]
+dependencies = [
+  "haystack-ai>=2.24.0",
+  "cognee==0.5.4",
```
sjrl: Usually we try to only lower bound our dependencies and occasionally add upper bounds e.g. `<1.0.0`. Could we do that here as well or is there a reason to pin exactly to this version?

integrations/cognee/pyproject.toml
```diff
@@ -0,0 +1,159 @@
...
+]
+
+[project.optional-dependencies]
+memory = ["haystack-experimental"]
```
sjrl: I think we can drop this the only thing that is useful from experimental is the protocol and I don't think that's being used anywhere currently.

integrations/cognee/pyproject.toml
```diff
@@ -0,0 +1,159 @@
...
+[tool.ruff]
+line-length = 120
+
+[tool.ruff.lint]
+select = [
+  "A",
+  "ARG",
+  "B",
+  "C",
+  "DTZ",
```
sjrl, line 97: We recently updated our ruff tool to enforce more docstrings in this PR https://github.com/deepset-ai/haystack-core-integrations/pull/3010 so let's make the same change here

```suggestion
  "C",
  "D102",   # Missing docstring in public method
  "D103",   # Missing docstring in public function
  "D205",   # 1 blank line required between summary line and description
  "D209",   # Closing triple quotes go to new line
  "D213",   # summary lines must be positioned on the second physical line of the docstring
  "D417",   # Missing argument descriptions in the docstring
  "D419",   # Docstring is empty
  "DTZ",
```

integrations/cognee/pyproject.toml
```diff
@@ -0,0 +1,159 @@
...
+[tool.ruff.lint.isort]
+known-first-party = ["haystack_integrations"]
+
+[tool.ruff.lint.flake8-tidy-imports]
+ban-relative-imports = "parents"
+
+[tool.ruff.lint.per-file-ignores]
+# Tests can use magic values, assertions, and relative imports
+"tests/**/*" = ["PLR2004", "S101", "TID252"]
+"examples/**/*" = ["T201"]
```
sjrl: Related to https://github.com/deepset-ai/haystack-core-integrations/pull/2979/changes#r2980234079
```suggestion
"tests/**/*" = ["D", "PLR2004", "S101", "TID252"]
"examples/**/*" = ["D", "T201"]
```

integrations/cognee/pyproject.toml
```diff
@@ -0,0 +1,159 @@
...
+
+[tool.hatch.envs.test]
+dependencies = [
+    "pytest",
+    "pytest-asyncio",
+    "pytest-cov",
+    "pytest-rerunfailures",
+    "mypy",
+    "pip",
+    "haystack-experimental",
```
sjrl: I don't think we are using haystack-experimental in tests right?

integrations/cognee/src/haystack_integrations/memory_stores/cognee/memory_store.py
sjrl: To provide more context on why I'm asking this question is that our existing Mem0 memory store, most memories are scoped to users which is why our experimental protocol typically requires a `user_id` in all of its methods since the idea is that we are always scoping the request to a specific user. For example our protocol for `add_memories` looks like
```python
def add_memories(self, *, messages: list[ChatMessage], user_id: str | None = None, **kwargs: Any) -> None:
```
with the idea being that `user_id` is set at run time to scope the request to a specific user. 

This way we don't need to create a new MemoryStore for every user which is useful when we usually want to easily reuse a Haystack Agent or Pipeline for many users. 

So my hope is that Cognee can also fit into this pattern.

integrations/cognee/src/haystack_integrations/components/connectors/cognee/_utils.py
hande-k: Cognee's search API returns Any internally. Results can be str (LLM completions), dict (structured outputs), or cognee model objects (DataPoint subclasses) depending on the search type. Since these internal types aren't part of cognee's API, I've kept Any but expanded the docstring to document the three expected categories. Would that work?

sjrl: Yeah expanding the docstring with the three supported types and the note that they are internal types would be great.

integrations/cognee/src/haystack_integrations/memory_stores/cognee/memory_store.py
hande-k: Cognee supports user-level scoping. Its `add()`, `search()`, and `cognify()` APIs all accept a `user` parameter for ACL-based access control. I've added `user_id` support to `CogneeMemoryStore` to match your existing pattern: a single store instance can serve multiple users via `user_id` at runtime. When `user_id` is provided, it's resolved to a cognee `User` and forwarded to all API calls. 
`dataset_name` groups data logically (like a collection), while `user_id` controls who can access it. Search is scoped to the store's dataset_name for the given user, and shared datasets (where another user granted read permission) are automatically resolved. I've updated the examples/demo_memory_agent.py for a demo of isolation + sharing. 
Does the current logic make sense for the integration?

integrations/cognee/src/haystack_integrations/memory_stores/cognee/memory_store.py
```diff
@@ -0,0 +1,197 @@
...
+        """
+        user = run_sync(_get_cognee_user(user_id)) if user_id else None
+
+        added = 0
+        for msg in messages:
+            text = msg.text
+            if not text:
+                continue
+            run_sync(cognee.add(text, dataset_name=self.dataset_name, user=user))
+            added += 1
```
sjrl: What happens if the `user` is `None`?

integrations/cognee/src/haystack_integrations/memory_stores/cognee/memory_store.py
sjrl: Yup that logic makes a lot of sense! Thanks for the changes. 

I'd only say it would be good to expose the user resolution to the Retriever and Writer components as well since those are how pipeline builders will typically interact with the Cognee Memory Store.

integrations/cognee/src/haystack_integrations/memory_stores/cognee/memory_store.py
```diff
@@ -0,0 +1,197 @@
...
+        Search Cognee's knowledge engine for relevant memories.
+
+        :param query: The search query.
+        :param top_k: Maximum number of memories to return.
+        :param user_id: Optional cognee user UUID to scope the search to a specific user.
+            Search is restricted to the store's ``dataset_name``. If the user owns the
+            dataset it is resolved by name; otherwise the store checks whether the user
+            has been granted read access (e.g. via shared permissions) and searches by
+            dataset UUID.
+            When ``None``, cognee's default user is used.
```
sjrl: Let's make sure to only only use single backticks to wrap code sections otherwise our doc build complains. 
```suggestion
            Search is restricted to the store's `dataset_name`. If the user owns the
            dataset it is resolved by name; otherwise the store checks whether the user
            has been granted read access (e.g. via shared permissions) and searches by
            dataset UUID.
            When `None`, cognee's default user is used.
```

integrations/cognee/tests/test_memory_store.py
hande-k: Added test_integration.py gated on LLM_API_KEY. This integration uses Cognee SDK locally as a library, so the only external dependency is the LLM provider (in default config it is OpenAI API key). 
We can extend the integration in the future to use our Cloud as well, which would require using a cognee account

integrations/cognee/src/haystack_integrations/components/retrievers/cognee/memory_retriever.py
```diff
@@ -0,0 +1,110 @@
...
+    retriever = CogneeRetriever(search_type="GRAPH_COMPLETION", top_k=5)
+    results = retriever.run(query="What is Cognee?")
+    for doc in results["documents"]:
+        print(doc.content)
+    ```
+    """
+
+    def __init__(
+        self, search_type: CogneeSearchType = "GRAPH_COMPLETION", top_k: int = 10, dataset_name: str | None = None
+    ):
```
sjrl: Following the pattern for our DocumentStores and their respective retrievers I'd expect the init method to take in a `CogneeMemoryStore` as init param. Check out our OpenSearchBM25Retriever as an example https://github.com/deepset-ai/haystack-core-integrations/blob/c40817bf229f804b5099bdd636e1169c8fe62fcb/integrations/opensearch/src/haystack_integrations/components/retrievers/opensearch/bm25_retriever.py#L20-L39 

The idea is for the retriever to call the method directly from the `Store`. E.g. this is how the bm25 retrieval is run in the linked component `docs = doc_store._bm25_retrieval(**bm25_args)`. 

So it would be great if we could follow that pattern here as well.

integrations/cognee/src/haystack_integrations/components/retrievers/cognee/memory_retriever.py
```diff
@@ -0,0 +1,110 @@
...
+            SUMMARIES, INSIGHTS, etc.
+        :param top_k: Maximum number of results to return.
+        :param dataset_name: Optional dataset name to restrict search scope.
+        """
+        self.search_type = search_type
+        self.top_k = top_k
+        self.dataset_name = dataset_name
+
+    @component.output_types(documents=list[Document])
+    def run(self, query: str, top_k: int | None = None) -> dict[str, list[Document]]:
```
sjrl: Related to this comment https://github.com/deepset-ai/haystack-core-integrations/pull/2979/changes#r3027766776 the run method should more or less accept all of the same arguments as the `CogneeMemoryStore.search_memories` function

integrations/cognee/tests/test_memory_store.py
sjrl: Oh okay good to know. When using the Cognee SDK locally like this where do the memories get stored? Is the cognee library running a local db in the background or storing them in memory?

integrations/cognee/src/haystack_integrations/components/connectors/cognee/_utils.py
hande-k: Done — expanded the docstring to document the three supported shapes (`str`, `dict`, cognee model objects like `DataPoint` subclasses) and explained why `item` stays typed as `Any` (these are cognee-internal types, not part of its public API).

integrations/cognee/src/haystack_integrations/memory_stores/cognee/memory_store.py
hande-k: Done! `CogneeRetriever` now accepts `user_id` in `run()`, and `CogneeWriter` also accepts `user_id` in `run()` — both pass it through to the store. This way a single pipeline instance can serve many users at runtime.

integrations/cognee/tests/test_memory_store.py
hande-k: By default cognee stores everything locally. it uses SQLite for relational data, Kuzu (an embedded graph DB) for the knowledge graph, and LanceDB for vector embeddings. All stored under `~/.cognee/` (configurable). No external services needed for local usage, so the integration tests in `test_integration.py` just need an LLM API key (for the entity extraction / graph construction step) but no separate cognee account or server.

integrations/cognee/tests/__init__.py
hande-k: These were already added in the previous commit — `test_cognifier.py` (10 tests) and `test_utils.py` (18 tests covering `extract_text` and `run_sync`).

integrations/cognee/src/haystack_integrations/memory_stores/cognee/memory_store.py
hande-k: When `user_id` is `None` (the default), `_get_cognee_user` is not called and `user=None` is passed to `cognee.add()` / `cognee.search()`. Cognee treats `None` as its built-in default user, so the call succeeds normally — it just uses the default permissions scope. Updated the docstring to clarify this.

integrations/cognee/src/haystack_integrations/memory_stores/cognee/memory_store.py
hande-k: Applied — switched all double backticks to single backticks in the docstrings.

integrations/cognee/src/haystack_integrations/components/retrievers/cognee/memory_retriever.py
hande-k: Refactored — `CogneeRetriever.__init__` now takes a `CogneeMemoryStore` instance (following the `OpenSearchBM25Retriever` pattern). The retriever delegates search to `store.search_memories()` and handles `to_dict`/`from_dict` serialization of the nested store. Tests, examples, and integration tests all updated.

integrations/cognee/src/haystack_integrations/components/retrievers/cognee/memory_retriever.py
hande-k: Done — `run()` now accepts `query`, `top_k`, and `user_id`, matching the `CogneeMemoryStore.search_memories` signature. The retriever forwards all of them to the store.

hande-k:
Hi @sjrl sorry for the delay! it's been an intense few weeks but I'm back on this. I've addressed all the open review comments above.                                                                                                                              

The one remaining issue is the Python 3.14 on Windows CI failure. It's caused by kuzu (a dependency via cognee, the embedded graph DB i menioned above). Linux and macOS pass fine.                                                                         

Would it be okay to exclude the Windows + 3.14 combination from the CI matrix for now? We'd still have 3.14 coverage on Linux and macOS, and Windows coverage via 3.10. Let me know what you think.

sjrl:
> Hi @sjrl sorry for the delay! it's been an intense few weeks but I'm back on this. I've addressed all the open review comments above.
> 
> The one remaining issue is the Python 3.14 on Windows CI failure. It's caused by kuzu (a dependency via cognee, the embedded graph DB i menioned above). Linux and macOS pass fine.
> 
> Would it be okay to exclude the Windows + 3.14 combination from the CI matrix for now? We'd still have 3.14 coverage on Linux and macOS, and Windows coverage via 3.10. Let me know what you think.

No problem! Thanks for getting back to this. I should have some time to review this in the next few days. 

But to answer your question yes it would be fine to remove the windows + 3.14 combo for now, let's just leave a comment there explaining the problem.

integrations/cognee/src/haystack_integrations/components/connectors/cognee/_utils.py
```diff
@@ -0,0 +1,126 @@
...
+
+
+async def _get_cognee_user(user_id: str) -> Any:
+    """
+    Resolve a user_id string to a cognee User object.
+
+    Converts the given UUID string to a cognee User via ``cognee.modules.users.methods.get_user``.
+
+    :param user_id: UUID string identifying the cognee user.
+    :returns: A cognee ``User`` object.
```
sjrl:
```suggestion
    :returns: A cognee `User` object.
```

integrations/cognee/src/haystack_integrations/components/connectors/cognee/_utils.py
```diff
@@ -0,0 +1,126 @@
...
+                return item[key]
+
+    return str(item)
+
+
+async def _get_cognee_user(user_id: str) -> Any:
+    """
+    Resolve a user_id string to a cognee User object.
+
+    Converts the given UUID string to a cognee User via ``cognee.modules.users.methods.get_user``.
```
sjrl:
```suggestion
    Converts the given UUID string to a cognee User via `cognee.modules.users.methods.get_user`.
```

.github/workflows/cognee.yml
```diff
@@ -0,0 +1,79 @@
...
+name: Test / cognee
+
+on:
+  schedule:
+    - cron: "0 0 * * *"
+  pull_request:
+    paths:
+      - "integrations/cognee/**"
+      - "!integrations/cognee/*.md"
+      - ".github/workflows/cognee.yml"
```
sjrl, line 12: The push tag looks to be missing
```suggestion
  pull_request:
    paths:
      - "integrations/cognee/**"
      - "!integrations/cognee/*.md"
      - ".github/workflows/cognee.yml"
  push:
    branches:
      - main
    paths:
      - "integrations/cognee/**"
      - "!integrations/cognee/*.md"
      - ".github/workflows/cognee.yml"
```

.github/workflows/cognee.yml
```diff
@@ -0,0 +1,79 @@
...
+  run:
+    working-directory: integrations/cognee
+
+concurrency:
+  group: cognee-${{ github.head_ref }}
+  cancel-in-progress: true
+
+env:
+  PYTHONUNBUFFERED: "1"
+  FORCE_COLOR: "1"
```
sjrl, line 31: Let's go ahead and add the LLM api key here that we will need e.g. `LLM_API_KEY: ${{ secrets.OPENAI_API_KEY }}`. 

Which provider's api key would be appropriate to use?

.github/workflows/cognee.yml
```diff
@@ -0,0 +1,79 @@
...
+          python-version: ${{ matrix.python-version }}
+
+      - name: Install Hatch
+        run: pip install --upgrade hatch
+      - name: Lint
+        if: matrix.python-version == '3.10' && runner.os == 'Linux'
+        run: hatch run fmt-check && hatch run test:types
+
+      - name: Run tests
+        run: hatch run test:cov-retry
```
sjrl: We like to separate unit and integration test runs
```suggestion
      - name: Run unit tests
        run: hatch run test:unit-cov-retry
```

.github/workflows/cognee.yml
```diff
@@ -0,0 +1,79 @@
...
+
```
sjrl, line 69: Let's add the integration tests as well
```suggestion
      - name: Run integration tests
        run: hatch run test:integration-cov-append-retry
```

integrations/cognee/pyproject.toml
```diff
@@ -0,0 +1,160 @@
...
+    "pytest-rerunfailures",
+    "mypy",
+    "pip",
+]
+
+[tool.hatch.envs.test.scripts]
+unit = 'pytest -m "not integration" {args:tests}'
+integration = 'pytest -m "integration" {args:tests}'
+all = 'pytest {args:tests}'
+cov-retry = 'pytest --cov=haystack_integrations --reruns 3 --reruns-delay 30 -x {args:tests}'
```
sjrl: Apologies we've made some changes to the default pyproject.toml since this PR was opened.
```suggestion
unit-cov-retry = 'pytest --cov=haystack_integrations --reruns 3 --reruns-delay 30 -x -m "not integration" {args:tests}'
integration-cov-append-retry = 'pytest --cov=haystack_integrations --cov-append --reruns 3 --reruns-delay 30 -x -m "integration" {args:tests}'
```

integrations/cognee/examples/demo_memory_agent.py
sjrl, line 1: Could we rename this to `demo_memory_store.py` since we aren't actually using an Agent component in this demo?

integrations/cognee/src/haystack_integrations/components/writers/cognee/memory_writer.py
```diff
@@ -0,0 +1,92 @@
...
+
+        :param dataset_name: Name of the Cognee dataset to add documents to.
+        :param auto_cognify: If True, automatically runs `cognee.cognify()` after adding
+            documents to process them into the knowledge engine.
+        """
+        self.dataset_name = dataset_name
+        self.auto_cognify = auto_cognify
+
+    @component.output_types(documents_written=int)
+    def run(self, documents: list[Document], user_id: str | None = None) -> dict[str, Any]:
```
sjrl:
```suggestion
    def run(self, documents: list[Document], user_id: str | None = None) -> dict[str, int]:
```

integrations/cognee/src/haystack_integrations/memory_stores/cognee/memory_store.py
```diff
@@ -0,0 +1,198 @@
...
+        search_type_enum = SearchType[self.search_type]
+        effective_top_k = top_k or self.top_k
+
+        try:
+            raw_results = run_sync(
+                cognee.search(
+                    query_text=query,
+                    query_type=search_type_enum,
+                    user=user,
+                    datasets=[self.dataset_name],
```
sjrl: It seems we should pass the `top_k` directly to `cognee.search` which accepts a `top_k` param

sjrl:
Hey @hande-k I wanted to provide you with an example script of how I imagined the cognee integration to work with Haystack. The main use case for us is to enable memories for our Agent component so I think the Retriever and Writer components need to be reworked to return and accept list of ChatMessages respectively. 

Here is a demo script I tried out that assumes these changes to the components have already been made. Feel free to include this in the examples folder if you'd like. 

```python
#!/usr/bin/env python
"""
Demo: Memory-Augmented Agent with CogneeRetriever and CogneeWriter

Shows the core memory-agent loop where Cognee enriches every conversation
turn with memories from past sessions alongside the live chat history:

1. Pre-seeded memories: facts from previous sessions are loaded into Cognee
   before the conversation starts, simulating persistent long-term memory.
2. Before each turn: CogneeRetriever fetches relevant memories and the
   OutputAdapter prepends them to the live chat history + current user message.
3. The Agent processes the full context (memories + history + user message).
4. After each turn: CogneeWriter stores the agent's output messages so future
   sessions can recall what happened in this one.

Pipeline structure (per turn):

    query ──► CogneeRetriever ──► memories (list[ChatMessage])  ──┐
                                                                   ├──► OutputAdapter 
    history + user_message (pipeline inputs) ─────────────────────┘
                                                                         │
                                                                      Agent ──► messages, last_message
                                                                                      │
                                                                               CogneeWriter ──► messages_written

NOTE: This demo assumes CogneeRetriever and CogneeWriter have been updated to
work with ChatMessages rather than Documents. See the review comments for the
required changes.

Prerequisites:
    pip install -e "integrations/cognee"

Set your LLM and vector DB keys (Cognee uses them internally):
    export OPENAI_API_KEY="sk-..."
"""

import logging
import os

# Must be set before importing cognee — its setup_logging() reads LOG_LEVEL at
# import time and installs structlog handlers that bypass standard logging config.
os.environ.setdefault("LOG_LEVEL", "WARNING")

import asyncio

import cognee

from haystack import Pipeline

logging.basicConfig(level=logging.WARNING)
from haystack.components.agents import Agent
from haystack.components.converters import OutputAdapter
from haystack.components.generators.chat import OpenAIChatGenerator

from haystack_integrations.components.retrievers.cognee import CogneeRetriever
from haystack_integrations.components.writers.cognee import CogneeWriter
from haystack_integrations.memory_stores.cognee import CogneeMemoryStore

DATASET_NAME = "agent_memory_demo"

# Facts that would have been stored in a previous session.
SEEDED_MEMORIES = [
    "My name is Alice. I'm a senior data scientist at Acme Corp specialising in NLP and knowledge graphs.",
    "My current project is building an internal documentation search system powered by Haystack and Cognee.",
    "My team: Bob is the ML engineer and Carol handles infrastructure.",
    "I prefer concise answers with Python code examples over long prose explanations.",
]

SYSTEM_PROMPT = (
    "You are a helpful assistant with access to a persistent memory of past conversations. "
    "Any system messages at the start of the conversation contain relevant memories retrieved "
    "from previous interactions — use them to personalise your responses and maintain continuity "
    "across turns. "
    "Be concise. Prefer short answers and Python code examples over long prose unless the user "
    "asks for more detail."
)


def build_memory_agent_pipeline(store: CogneeMemoryStore) -> Pipeline:
    """
    Wire the memory-augmented agent pipeline.

    The OutputAdapter merges three inputs into one flat ChatMessage list for
    the Agent:
      - memories:      retrieved from Cognee by CogneeRetriever
      - history:       accumulated chat messages from previous turns
      - user_messages: the current user message

    unsafe=True is required so that the Jinja NativeEnvironment returns the
    concatenated list as a real Python object rather than its string representation.
    """
    pipeline = Pipeline()

    pipeline.add_component("retriever", CogneeRetriever(memory_store=store))
    pipeline.add_component(
        "injector",
        OutputAdapter(
            template="{{ memories + history + user_messages }}",
            output_type=list[ChatMessage],
            unsafe=True,
        ),
    )
    pipeline.add_component(
        "agent",
        Agent(
            chat_generator=OpenAIChatGenerator(model="gpt-4o-mini"),
            system_prompt=SYSTEM_PROMPT,
        ),
    )
    pipeline.add_component("writer", CogneeWriter(dataset_name=store.dataset_name))

    pipeline.connect("retriever.messages", "injector.memories")
    pipeline.connect("injector.output", "agent.messages")
    pipeline.connect("agent.messages", "writer.messages")

    return pipeline


def run_turn(
    pipeline: Pipeline,
    user_text: str,
    history: list[ChatMessage],
) -> str:
    """
    Run one conversation turn, update history in-place, and return the agent reply.

    History is maintained outside the pipeline so it contains only the clean
    user/assistant exchange — not the memory context messages, which are
    injected fresh on every turn.
    """
    result = pipeline.run(
        {
            "retriever": {"query": user_text},
            "injector": {
                "history": history,
                "user_messages": [ChatMessage.from_user(user_text)],
            },
        }
    )
    last_message = result["agent"]["last_message"]
    reply = last_message.text or "(no reply)"

    # Append this turn to history so the next turn has full conversational context.
    history.append(ChatMessage.from_user(user_text))
    history.append(last_message)

    return reply


async def main():
    print("=== Cognee Memory Agent Pipeline Demo ===\n")

    print("Pruning previous data for a clean start...")
    await cognee.prune.prune_data()
    await cognee.prune.prune_system(metadata=True)
    print("Done.\n")

    # -------------------------------------------------------------------------
    # Pre-seed long-term memories from a simulated previous session.
    # In a real application these would already be in Cognee from earlier runs.
    # -------------------------------------------------------------------------
    print("Seeding memories from a previous session...")
    store = CogneeMemoryStore(search_type="GRAPH_COMPLETION", dataset_name=DATASET_NAME)
    store.add_memories(messages=[ChatMessage.from_user(fact) for fact in SEEDED_MEMORIES])
    print(f"  {len(SEEDED_MEMORIES)} facts stored.\n")

    pipeline = build_memory_agent_pipeline(store)
    history: list[ChatMessage] = []

    # -------------------------------------------------------------------------
    # Multi-turn conversation. The Agent receives retrieved memories + live
    # history + the current message on every turn, so it can both recall past
    # facts and follow the thread of the current exchange.
    # -------------------------------------------------------------------------
    turns = [
        # Turn 1 — recall a specific memory (no history yet)
        "Hi! Can you remind me what project I'm currently working on?",
        # Turn 2 — follow-up that requires turn 1 to make sense
        "What's the tech stack we're using for it?",
        # Turn 3 — recall a different memory fact
        "Who else is on my team, and what are their roles?",
        # Turn 4 — requires both a recalled preference *and* conversation context
        "Based on what you know about me, give me a quick tip for structuring a new Haystack pipeline component.",
    ]

    for user_text in turns:
        print(f"User:  {user_text}")
        reply = run_turn(pipeline, user_text, history)
        print(f"Agent: {reply}\n")

    print("=== Done ===")


if __name__ == "__main__":
    asyncio.run(main())
```

sjrl:
Hey @hande-k soft ping.

integrations/cognee/pyproject.toml
```diff
@@ -0,0 +1,163 @@
...
+keywords = []
+authors = [{ name = "deepset GmbH", email = "info@deepset.ai" }]
+classifiers = [
+  "License :: OSI Approved :: Apache Software License",
+  "Development Status :: 3 - Alpha",
+  "Programming Language :: Python",
+  "Programming Language :: Python :: 3.10",
+  "Programming Language :: Python :: 3.11",
+  "Programming Language :: Python :: 3.12",
+  "Programming Language :: Python :: 3.13",
```
sjrl, line 21: It looks like we run tests on Python 3.14 as well so let's at this tag
```suggestion
  "Programming Language :: Python :: 3.13",
  "Programming Language :: Python :: 3.14",
```

integrations/cognee/src/haystack_integrations/memory_stores/cognee/memory_store.py
```diff
@@ -0,0 +1,270 @@
...
+            user = await _resolve_user(user_id)
+            await cognee.forget(dataset=self.dataset_name, user=user)
+
+        _run_sync(_delete())
+        logger.info("Deleted '{ds}'", ds=self.dataset_name)
+
+    def delete_memory(self, memory_id: str) -> None:
+        """Single-memory deletion is not exposed by cognee's public API yet."""
+        msg = "CogneeMemoryStore does not support deleting individual memories."
+        raise NotImplementedError(msg)
```
sjrl: Let's go ahead and remove this. We are in the process of rethinking and relaxing our protocol from experimental and we've decided not to require this method.

integrations/cognee/src/haystack_integrations/memory_stores/cognee/memory_store.py
```diff
@@ -0,0 +1,270 @@
...
+            await cognee.improve(
+                dataset=self.dataset_name,
+                session_ids=[target_session] if target_session else None,
+                user=user,
+            )
+
+        _run_sync(_improve())
+        logger.info("Improved '{ds}' (session={s})", ds=self.dataset_name, s=target_session)
+
+    def delete_all_memories(self, *, user_id: str | None = None, **_: Any) -> None:
```
sjrl:
```suggestion
    def delete_all_memories(self, *, user_id: str | None = None) -> None:
```
Same here we are relaxing our protocol to not require acceptance of kwargs so we can drop this.

integrations/cognee/src/haystack_integrations/memory_stores/cognee/memory_store.py
```diff
@@ -0,0 +1,270 @@
...
+            s=target_session,
+        )
+
+    def search_memories(
+        self,
+        *,
+        query: str | None = None,
+        top_k: int | None = None,
+        user_id: str | None = None,
+        **_: Any,
```
sjrl:
```suggestion
```
Also here we can drop the unneeded kwargs

integrations/cognee/src/haystack_integrations/memory_stores/cognee/memory_store.py
```diff
@@ -0,0 +1,270 @@
...
+        self.session_id = session_id
+        self.self_improvement = self_improvement
+
+    def add_memories(
+        self,
+        *,
+        messages: list[ChatMessage],
+        user_id: str | None = None,
+        session_id: str | None = None,
+        **_: Any,
```
sjrl:
```suggestion
```
Same here

integrations/cognee/tests/test_integration.py
```diff
@@ -0,0 +1,71 @@
...
+        writer = CogneeWriter(memory_store=store)
+        retriever = CogneeRetriever(memory_store=store, top_k=3)
+
+        writer.run(messages=[ChatMessage.from_user("Ada Lovelace wrote the first computer program.")])
+        out = retriever.run(query="Who wrote the first computer program?")
+
+        assert out["messages"], "retriever returned no messages"
+        assert any("lovelace" in m.text.lower() or "program" in m.text.lower() for m in out["messages"])
+
+    def test_improve_and_forget(self, dataset_name: str):
```
sjrl: This test fails for me when running the integration tests locally. Are you able to get it to pass?

integrations/cognee/src/haystack_integrations/memory_stores/cognee/memory_store.py
```diff
@@ -0,0 +1,270 @@
...
+        asyncio.get_running_loop()
+    except RuntimeError:
+        return asyncio.run(coro)
+
+    global _background_loop  # noqa: PLW0603
+    with _loop_lock:
+        if _background_loop is None or _background_loop.is_closed():
+            _background_loop = asyncio.new_event_loop()
+            threading.Thread(target=_background_loop.run_forever, daemon=True).start()
+    return asyncio.run_coroutine_threadsafe(coro, _background_loop).result()
```
sjrl: We should probably add a timeout otherwise this can wait indefinitely e.g. 
```suggestion
    return asyncio.run_coroutine_threadsafe(coro, _background_loop).result(timeout=300)
```

integrations/cognee/tests/__init__.py
sjrl, line 1: Could we also add a conftest.py file with content something like 
```python
# SPDX-FileCopyrightText: 2022-present deepset GmbH <info@deepset.ai>
#
# SPDX-License-Identifier: Apache-2.0

import logging
import os

# Set before cognee is imported so setup_logging() picks up WARNING level.
os.environ.setdefault("LOG_LEVEL", "WARNING")

_NOISY_LOGGERS = ("aiosqlite", "sqlalchemy", "sqlalchemy.engine", "alembic")


def _silence_noisy_loggers() -> None:
    for name in _NOISY_LOGGERS:
        logging.getLogger(name).setLevel(logging.WARNING)

_silence_noisy_loggers()
```
Or if you have a better way of getting the loggers to print less info that would be great. It makes it hard to debug the integration tests if they fail.

integrations/cognee/src/haystack_integrations/memory_stores/cognee/memory_store.py
hande-k: in place after rework, let me know if this works!

integrations/cognee/tests/test_integration.py
hande-k: The failure was cognee.recall raising DatasetNotFoundError on a forgotten dataset, added test_search_after_forget_raises which proves forget actually deletes and guards against catching the exception in search_memories.

integrations/cognee/src/haystack_integrations/memory_stores/cognee/memory_store.py
hande-k: Done 300s. Just wanted to flag: single-message remember() runs comfortably under it, but bulk ingestion can push past. Happy to expose it as an init param on `CogneeMemoryStore` now if you'd prefer

hande-k:
Thanks @sjrl for the patience on this one! cognee 1.0 shipping mid-review meant rebuilding the integration on the new apis. It is simpler and the agent-memory shape from your demo is the default path now. Appreciate the thorough review! let me know what else needs to be done

integrations/cognee/src/haystack_integrations/memory_stores/cognee/memory_store.py
sjrl: > Happy to expose it as an init param on CogneeMemoryStore now if you'd prefer

Yeah that would be great!

sjrl:
Hey @hande-k just two last comments 
- My reply here https://github.com/deepset-ai/haystack-core-integrations/pull/2979#discussion_r3274569594 
- Add if I run the example script locally I get an SQL error. Does the demo script work for you?

Otherwise it looks ready to go!

integrations/cognee/src/haystack_integrations/memory_stores/cognee/memory_store.py
hande-k: done!

hande-k:
> Hey @hande-k just two last comments
> 
> * My reply here [feat: Add Cognee integration #2979 (comment)](https://github.com/deepset-ai/haystack-core-integrations/pull/2979#discussion_r3274569594)
> * Add if I run the example script locally I get an SQL error. Does the demo script work for you?
> 
> Otherwise it looks ready to go!

hey @sjrl , the first one is done! For the SQL error, I couldn't reproduce on my end (cognee 1.0.9, Python 3.12, macOS). It could be due to a stale ~/.cognee/ from a pre-1.0 install, `rm -rf ~/.cognee/ ` would clear that but happy to look at the error if if you could share the trace with the cognee version

sjrl:
@hande-k I was able to figure it out. I had langfuse configured via my .env file and it looks like trying to run the demo with the langfuse observability turned on then it seems to result in a SQL error when "langfuse<3.0.0" is installed. I removed the dep and the .env file and the demo worked!

sjrl (approved):
Thanks for the contribution! I'll let you know once we have this released.

Issue #699, opened 2024-04-29
Elasticsearch: support dense, sparse, hybrid with inference in Elasticsearch


maxjakob:
## Summary and motivation

Elasticsearch offers multiple retrieval features including
- approximate dense vector retrieval with embedding inference in Python or in Elasticsearch
- exact dense vector retrieval with embedding inference in Python
- sparse vector retrieval with embedding inference in Elasticsearch
- hybrid retrieval (dense+BM25) with embedding inference in Elasticsearch

Other libraries such as LangChain already have all these options integrated. It would be great to also have them available in Haystack. Elastic is currently working on a Python package that will make the integration of these features easier. Here we want to discuss how to best make them available.

### Questions
- Does Haystack want to enable inference in Elasticsearch? The current design assumes that mapping from input string to embedding vector is done in Python before calling a retriever. With inference in Elasticsearch, this would change. For example, users could configure a dense vector model in Elasticsearch and then use input strings in Haystack.
- The options mentioned above require different ways of indexing the data. How to best incorporate this requirement? The current document store abstraction kind of assumes that there is only one way of indexing.

## Detailed design

**Concrete proposal:**

1. `ElasticsearchDocumentStore` takes an argument `retrieval_strategy` similarly to how it is down in [LangChain](https://github.com/langchain-ai/langchain-elastic/blob/97c0016220a44bd5e370e11c939559d97c927615/libs/elasticsearch/langchain_elasticsearch/vectorstores.py#L599). Calls to `write_documents` make use of the retrieval strategy to know how to index the data.
2. We add a number of different retrievers (`ElasticsearchDenseVectorRetriever`, `ElasticsearchSparseVectorRetriever`, `ElasticsearchHybridRetriever`, ...) that get initialized with an `ElasticsearchDocumentStore`. The retrieval strategy has to match the expectation of the individual retrievers. We check that the expectation is met upon initialization. For retrieving documents, the retrievers call a search method on the document store as this is the established pattern.

## Checklist

If the request is accepted, ensure the following checklist is complete before closing this issue.
```[tasklist]
### Tasks
- [ ] The code is documented with docstrings and was merged in the `main` branch
- [ ] Docs are published at https://docs.haystack.deepset.ai/
- [ ] There is a Github workflow running the tests for the integration nightly and at every PR
- [ ] A label named like `integration:<your integration name>` has been added to this repo
[...]

PR #2989, opened 2026-03-19
feat: add sparse vector storage to ElasticsearchDocumentStore

### LINKED ISSUE #2939: Sparse vector storage in `ElasticsearchDocumentStore`
- Add optional init parameter `sparse_vector_field: str | None = None`.
- When set: add a `sparse_vector` mapping for that field and, in `write_documents()`, write each document’s `sparse_embedding` to that field.
- ES: [sparse_vector field type](https://www.elastic.co/guide/en/elasticsearch/reference/current/sparse-vector.html).

GunaPalanivel:
### Related Issues
- fixes #2939
- partially addresses #699 

### Proposed Changes

Add sparse vector storage to `ElasticsearchDocumentStore`. Previously, `sparse_embedding` on `Document` was ignored during `write_documents`.

Key changes:

- Add optional `sparse_vector_field` to `ElasticsearchDocumentStore.__init__`
- Update `_default_mappings` to include `sparse_vector` when configured
- Transform `SparseEmbedding` (`indices`, `values`) → `{index: value}` in `write_documents` and `write_documents_async`
- Use `zip(..., strict=True)` to validate index/value alignment
- Add type hints to `_default_mappings` for mypy
- Extend `to_dict` / `from_dict` to include the new field
- Improve warnings when sparse embeddings are provided without a configured field

### How I tested

- Unit:
  - Updated `test_document_store.py`, `test_document_store_async.py` (init + serialization)
- Integration:
  - Added sparse write tests (sync + async) against Elasticsearch 8.13
  - Verified raw indexed structure
- Quality:
  - `hatch run fmt-check`
  - `hatch run test:types`
- Regression:
  - All existing tests pass (175)

### Notes for reviewers

- Follows existing Elasticsearch integration patterns
- Conversion happens during bulk indexing
- `strict=True` avoids silent misalignment bugs in sparse vectors
- Type hints added to satisfy mypy

### Checklist

- I've used a [conventional commit type](https://www.conventionalcommits.org/en/v1.0.0/) in the PR title

### DIFF
--- a/integrations/elasticsearch/tests/test_document_store.py
+++ b/integrations/elasticsearch/tests/test_document_store.py
@@ -43,6 +43,36 @@ def test_init_with_custom_mapping_injects_sparse_vector():
     assert store._custom_mapping["properties"]["my_sparse_vec"] == {"type": "sparse_vector"}
 
 
+def test_handle_sparse_embedding_no_op_when_absent():
+    store = ElasticsearchDocumentStore(hosts="testhost")
+    doc_dict = {"id": "doc-1", "content": "hello"}
+    store._handle_sparse_embedding(doc_dict, "doc-1")
+    assert doc_dict == {"id": "doc-1", "content": "hello"}
+
+
+def test_handle_sparse_embedding_converts_to_es_format():
+    store = ElasticsearchDocumentStore(hosts="testhost", sparse_vector_field="my_sparse")
+    doc_dict = {
+        "id": "doc-1",
+        "sparse_embedding": {"indices": [0, 5], "values": [0.3, 0.7]},
+    }
+    store._handle_sparse_embedding(doc_dict, "doc-1")
+    assert "sparse_embedding" not in doc_dict
+    assert doc_dict["my_sparse"] == {"0": 0.3, "5": 0.7}
+
+
+def test_handle_sparse_embedding_warns_when_no_field_configured(caplog):
+    store = ElasticsearchDocumentStore(hosts="testhost")
+    doc_dict = {
+        "id": "doc-1",
+        "content": "hello",
+        "sparse_embedding": {"indices": [0, 1], "values": [0.5, 0.5]},
+    }
+    store._handle_sparse_embedding(doc_dict, "doc-1")
+    assert "but `sparse_vector_field` is not configured" in caplog.text
+    assert "sparse_embedding" not in doc_dict
+
+
 @patch("haystack_integrations.document_stores.elasticsearch.document_store.Elasticsearch")
 def test_headers_are_supported(_mock_es_client):
     _ = ElasticsearchDocumentStore(


integrations/elasticsearch/src/haystack_integrations/document_stores/elasticsearch/document_store.py
davidsbatista, line 136: We need to add a check here in case `sparse_vector_field` is part of the `SPECIAL_FIELDS`

```python
if sparse_vector_field in SPECIAL_FIELDS:
      raise ValueError(f"sparse_vector_field '{sparse_vector_field}' conflicts with a reserved field name.")
```

davidsbatista, line 115: the `sparse_vector_field` param needs to be added to the docstring

davidsbatista, line 437: Here, we need to convert from the ES sparse embedding format to Haystack's SparseEmbedding

ES stores sparse vectors as {str(index): float} (e.g. {"0": 0.5}), while Haystack's SparseEmbedding.to_dict() / from_dict() uses {"indices": [...], "values": [...]}. 

We need to convert between them, example:

```python
if self._sparse_vector_field and self._sparse_vector_field in data:
      es_sparse = data.pop(self._sparse_vector_field)
      sorted_items = sorted(es_sparse.items(), key=lambda x: int(x[0]))
      data["sparse_embedding"] = {
          "indices": [int(k) for k, _ in sorted_items],
          "values": [v for _, v in sorted_items],
      }
```

integrations/elasticsearch/src/haystack_integrations/document_stores/elasticsearch/document_store.py
```diff
@@ -458,12 +463,18 @@ def write_documents(
             if "sparse_embedding" in doc_dict:
```
davidsbatista: This is duplicated code for the the sync and async version that can be reduced with a helper function, for example:

```python
def _handle_sparse_embedding(self, doc_dict: dict[str, Any], doc_id: str) -> None:
      if "sparse_embedding" not in doc_dict:
          return
      sparse_embedding = doc_dict.pop("sparse_embedding")
      if not sparse_embedding:
          return
      if self._sparse_vector_field:
          doc_dict[self._sparse_vector_field] = {
              str(idx): val
              for idx, val in zip(sparse_embedding["indices"], sparse_embedding["values"], strict=True)
          }
      else:
          logger.warning(
              "Document {doc_id} has the `sparse_embedding` field set, "
              "but `sparse_vector_field` is not configured for this ElasticsearchDocumentStore. "
              "The `sparse_embedding` field will be ignored.",
              doc_id=doc_id,
          )
```

Then both write methods can use it:

```python
  for doc in documents:
      doc_dict = doc.to_dict()
      self._handle_sparse_embedding(doc_dict, doc.id)
      ...
```

integrations/elasticsearch/src/haystack_integrations/document_stores/elasticsearch/document_store.py
```diff
@@ -135,7 +137,7 @@ def __init__(
             raise ValueError(msg)
```
davidsbatista, line 144: if both `custom_mapping`and `sparse_vector_field` are provided the sparse_vector_field is  ignored when 

The sparse field is never added to the custom mapping . The index gets created without the sparse_vector field type.

A possible fix is when a custom mapping is provided, automatically inject it into the custom mapping:

```python
  if self._custom_mapping and self._sparse_vector_field:
      self._custom_mapping.setdefault("properties", {})
      self._custom_mapping["properties"][self._sparse_vector_field] = {"type": "sparse_vector"}
```

integrations/elasticsearch/tests/test_document_store_async.py
```diff
@@ -152,13 +152,30 @@ async def test_write_documents_async_with_sparse_embedding_warning(self, documen
...
         assert results[0].id == "1"
         assert not hasattr(results[0], "sparse_embedding") or results[0].sparse_embedding is None
 
+    @pytest.mark.asyncio
+    async def test_write_documents_async_with_sparse_vectors(self):
+        """Test write_documents with document containing sparse_embedding field"""
+        store = ElasticsearchDocumentStore(
+            hosts=["http://localhost:9200"], index="test_async_sparse", sparse_vector_field="sparse_vec"
+        )
+        store.client.options(ignore_status=[400, 404]).indices.delete(index="test_async_sparse")
```
davidsbatista: this should be using the async client:

`await store.async_client.indices.delete(...)`

integrations/elasticsearch/tests/test_document_store.py
```diff
@@ -145,6 +150,7 @@ def test_from_dict_with_api_keys_str():
             "api_key": "my_api_key",
             "api_key_id": "my_api_key_id",
             "embedding_similarity_function": "cosine",
+            "sparse_vector_field": None,
         },
     }
```
davidsbatista, line 205: Can you also add a test verifying that a serialized config that doesn't have `sparse_vector_field` works

integrations/elasticsearch/src/haystack_integrations/document_stores/elasticsearch/document_store.py
GunaPalanivel: Added. The check now raises `ValueError` if `sparse_vector_field` matches any name in `SPECIAL_FIELDS`

GunaPalanivel: Added `:param sparse_vector_field:` to the `init` docstring

GunaPalanivel: `_deserialize_document` now pops the sparse vector field, sorts by integer key, and reconstructs `{"indices": [...], "values": [...]}`. Added retrieval assertions in both sync and async integration tests.

integrations/elasticsearch/src/haystack_integrations/document_stores/elasticsearch/document_store.py
GunaPalanivel: Extracted into `_handle_sparse_embedding(self, doc_dict, doc_id)`. Both `write_documents` and `write_documents_async` now call it.

integrations/elasticsearch/src/haystack_integrations/document_stores/elasticsearch/document_store.py
GunaPalanivel: When both are provided, the sparse vector type is now auto-injected into `self._custom_mapping["properties"]`. Added a unit test `test_init_with_custom_mapping_injects_sparse_vector` to verify.

integrations/elasticsearch/tests/test_document_store_async.py
GunaPalanivel: Changed to `await store.async_client.options(...).indices.delete(...)` for setup and `await store.async_client.indices.delete(...)` for teardown.

integrations/elasticsearch/tests/test_document_store.py
GunaPalanivel: Added `test_from_dict_without_sparse_vector_field` which deserializes a config dict that has no  `sparse_vector_field` key and asserts the field defaults to `None`

Issue #2864, opened 2026-02-20
Anthropic - remove redacted thinking handling


anakin87:
Stems from #2683

Redacted thinking was a Claude 3.7 behavior, removed in Claude 4 models.
> Occasionally Claude’s internal reasoning will be flagged by our safety systems. When this occurs, we encrypt some or all of the thinking block and return it to you as a redacted_thinking block. redacted_thinking blocks are decrypted when passed back to the API, allowing Claude to continue its response without losing context.

Sources:
- https://platform.claude.com/docs/en/build-with-claude/extended-thinking
- https://docs.aws.amazon.com/bedrock/latest/userguide/claude-messages-thinking-encryption.html (here the removal is explicitly specified)

Handling redacted thinking makes our code more complex. Now that 3.7 models are retired, we should consider simplifying the code.

PR #2998, opened 2026-03-20
chore: drop redacted thinking support for AmazonBedrockChatGenerator

tstadel:
### Related Issues

- partially fixes https://github.com/deepset-ai/haystack-core-integrations/issues/2864

### Proposed Changes:
- simplify reasoning parsing
- update models for integration tests (some were marked as legacy and couldn't be accessed anymore)

### How did you test it?
- added/adjusted tests

### Checklist

### DIFF
--- a/integrations/amazon_bedrock/tests/test_chat_generator.py
+++ b/integrations/amazon_bedrock/tests/test_chat_generator.py
@@ -744,12 +744,16 @@ class TestAmazonBedrockChatGeneratorInference:
             model=model_name,
             generation_kwargs={
                 "maxTokens": 8192,
-                **({
-                    "thinking": {
-                        "type": "enabled",
-                        "budget_tokens": 1024,
+                **(
+                    {
+                        "thinking": {
+                            "type": "enabled",
+                            "budget_tokens": 1024,
+                        }
                     }
-                } if "claude" in model_name else {})
+                    if "claude" in model_name
+                    else {}
+                ),
             },
         )
         results = component.run(messages=initial_messages)
@@ -767,12 +771,16 @@ class TestAmazonBedrockChatGeneratorInference:
             tools=tools,
             generation_kwargs={
                 "maxTokens": 8192,
-                **({
-                    "thinking": {
-                        "type": "enabled",
-                        "budget_tokens": 1024,
+                **(
+                    {
+                        "thinking": {
+                            "type": "enabled",
+                            "budget_tokens": 1024,
+                        }
                     }
-                } if "claude" in model_name else {})
+                    if "claude" in model_name
+                    else {}
+                ),
             },
         )
         results = component.run(messages=initial_messages)


integrations/amazon_bedrock/src/haystack_integrations/components/generators/amazon_bedrock/chat/utils.py
```diff
@@ -583,15 +583,20 @@ def _convert_event_to_streaming_chunk(
...
-            reasoning_text = reasoning_content.get("text", "")
+                reasoning_text = "[REDACTED]"
+            elif "signature" in reasoning_content:
+                reasoning_text = ""
             streaming_chunk = StreamingChunk(
                 content="",
                 index=block_idx,
                 reasoning=ReasoningContent(
                     reasoning_text=reasoning_text,
-                    extra={"reasoning_contents": [{"index": block_idx, "reasoning_content": reasoning_content}]},
```
sjrl, line 594: Could you explain why this isn't needed anymore? This info was used when converting a ChatMessage back into the bedrock format. In `_format_reasoning_content` which doesn't look updated here.

integrations/amazon_bedrock/src/haystack_integrations/components/generators/amazon_bedrock/chat/utils.py
```diff
@@ -630,86 +635,6 @@ def _convert_event_to_streaming_chunk(
     return streaming_chunk
 
 
-def _process_reasoning_contents(chunks: list[StreamingChunk]) -> ReasoningContent | None:
-    """
-    Process reasoning contents from a list of StreamingChunk objects into the Bedrock expected format.
-
-    :param chunks: List of StreamingChunk objects potentially containing reasoning contents.
-
-    :returns: List of Bedrock formatted reasoning content dictionaries
```
sjrl, line 639: I see you've removed this but previously this was needed since bedrock could return multiple reasoning content blocks. 

It seems now we are assuming bedrock can only ever return one content block? Is that correct?

integrations/amazon_bedrock/src/haystack_integrations/components/generators/amazon_bedrock/chat/utils.py
tstadel: That's a good one. ~~Looks like _format_reasoning_content is not covered by tests.~~ Will add that.

tstadel: For the next comment, we are adding only exactly one reasoning_content here (per chunk).

integrations/amazon_bedrock/src/haystack_integrations/components/generators/amazon_bedrock/chat/utils.py
tstadel: As we saw above we only added exactly one reasoning_content per chunk to chunk.reasoning.extra["resoning_content"]. As far as I can see the code there's no other way there could be multiple ones. Am I missing something?

integrations/amazon_bedrock/src/haystack_integrations/components/generators/amazon_bedrock/chat/utils.py
```diff
@@ -291,22 +291,23 @@ def _repair_tool_result_messages(bedrock_formatted_messages: list[dict[str, Any]
...
-        formatted_contents.append(formatted_content)
-    return formatted_contents
+    formatted_content = {
+        "reasoningContent": {
+            "reasoningText": {
+                "text": reasoning_content.reasoning_text,
+                **({"signature": reasoning_content.extra["signature"]} if reasoning_content.extra.get("signature") else {}),
+            }
+        }
+    }
```
tstadel, line 313: format according to https://docs.aws.amazon.com/bedrock/latest/userguide/conversation-inference-call.html ~~(before we dind't have signature correclty positioned)
I'll add a integration test to verify~~

tstadel:
@sjrl Ok, I've been missing quite a lot. Thanks for notifying. There's no fundamental functional improvement to the previous version left (only the trace is also applied for async streaming). 
The PR is now merely a complexity reduction by getting rid of the redacted handling that died with Claude 3.7.

integrations/amazon_bedrock/src/haystack_integrations/components/generators/amazon_bedrock/chat/utils.py
tstadel: Previously it was also correct, just wrapped into the "reasoning_contents" list. Note that signature was implicitly handled by the whole "reasoning_text/reasoningText" object.

integrations/amazon_bedrock/tests/test_chat_generator_utils.py
```diff
@@ -445,19 +387,24 @@ def test_format_textual_assistant_message(self):
...
-                    ]
+                    "signature": "reasoning_signature",
                 },
             ),
         )
         formatted_message = _format_textual_assistant_message(assistant_message)
         assert formatted_message == {
             "role": "assistant",
             "content": [
-                {"reasoningContent": {"reasoningText": "This is the reasoning behind the message."}},
```
tstadel, line 460: This one was actually misleading as in reality it'd be 
`{"reasoningContent": {"reasoningText": {"text": "This is the reasoning behind the message."}}},`

integrations/amazon_bedrock/src/haystack_integrations/components/generators/amazon_bedrock/chat/utils.py
```diff
@@ -492,23 +491,25 @@ def _parse_completion_response(response_body: dict[str, Any], model: str) -> lis
...
+            reasoning_extra = {}
             reasoning_text = ""
-            for content in reasoning_contents:
-                if "reasoning_text" in content["reasoning_content"]:
-                    reasoning_text += content["reasoning_content"]["reasoning_text"]["text"]
-                elif "redacted_content" in content["reasoning_content"]:
-                    reasoning_text += "[REDACTED]"
+            if reasoning_content:
+                if "redacted_content" in reasoning_content:
+                    reasoning_text = "[REDACTED]"
```
sjrl: Since we are removing redacted support since Sonnet 3.7 was retired do we need this if statement anymore?

integrations/amazon_bedrock/src/haystack_integrations/components/generators/amazon_bedrock/chat/utils.py
sjrl: Right the reason it was wrapped in a list and provided a block_idx was to reassamble the correct reasoning content blocks at this stage. Since it was possible the reasoning content was actually split up into multiple content blocks. E.g. I believe it was technically possible to get something like `reasoning_text` --> `redacted_content` --> `reasoning_text`. 

So I don't think it's possible or a worry now but before I think we were being defensive since their API technically allowed it.

integrations/amazon_bedrock/src/haystack_integrations/components/generators/amazon_bedrock/chat/utils.py
```diff
@@ -583,15 +584,20 @@ def _convert_event_to_streaming_chunk(
...
         elif "reasoningContent" in delta:
             reasoning_content = delta["reasoningContent"]
-            if "redactedContent" in reasoning_content:
+            if "text" in reasoning_content:
+                reasoning_text = reasoning_content.pop("text", "")
+            elif "redactedContent" in reasoning_content:
+                # redactedContent was only used for Claude 3.7 which is marked as legacy
                 reasoning_content["redacted_content"] = reasoning_content.pop("redactedContent")
-            reasoning_text = reasoning_content.get("text", "")
+                reasoning_text = "[REDACTED]"
```
sjrl: Same question here can we go ahead and fully remove this since Sonnet 3.7 was retired in Feb of 2026? https://platform.claude.com/docs/en/about-claude/model-deprecations

integrations/amazon_bedrock/src/haystack_integrations/components/generators/amazon_bedrock/chat/utils.py
tstadel: Not really, I left it as a safety-net in case there's still some bedrock model emitting `redactedContent`. We could also log a warning instead.

integrations/amazon_bedrock/src/haystack_integrations/components/generators/amazon_bedrock/chat/utils.py
tstadel: Jep, let me know if we should still have some warning, etc. Otherwise I'll remove it.

sjrl: If you could update the PR name to reflect that this is being dropped that should be enough. The PR name get's pulled into our automatic release notes. Since the model is deprecated anyway I think we would only need a minor release version bump.

integrations/amazon_bedrock/src/haystack_integrations/components/generators/amazon_bedrock/chat/utils.py
sjrl: Is there any chance that an existing bedrock model would still do this?

integrations/amazon_bedrock/src/haystack_integrations/components/generators/amazon_bedrock/chat/utils.py
tstadel: done

integrations/amazon_bedrock/src/haystack_integrations/components/generators/amazon_bedrock/chat/utils.py
tstadel: I don't think so. I also double checked bedrock guardrails, but they also don't use it.

sjrl: okay let's go ahead and drop

tstadel: done

integrations/amazon_bedrock/tests/test_chat_generator.py
```diff
@@ -13,42 +13,47 @@
 
 CLASS_TYPE = "haystack_integrations.components.generators.amazon_bedrock.chat.chat_generator.AmazonBedrockChatGenerator"
 MODELS_TO_TEST = [
-    "anthropic.claude-3-5-sonnet-20240620-v1:0",
-    "cohere.command-r-plus-v1:0",
```
sjrl, line 17: Could you mention why all cohere models were dropped?

integrations/amazon_bedrock/tests/test_chat_generator.py
```diff
@@ -13,42 +13,47 @@
...
 
 MODELS_TO_TEST_WITH_THINKING = [
-    "us.anthropic.claude-3-7-sonnet-20250219-v1:0",
-    "us.anthropic.claude-sonnet-4-20250514-v1:0",
+    "global.anthropic.claude-sonnet-4-6",
+]
+
+MODELS_TO_TEST_WITH_THINKING_TOOLS_STREAMING = [
+    "global.anthropic.claude-sonnet-4-6",
 ]
```
sjrl, line 59: These are the same, could we just reuse `MODELS_TO_TEST_WITH_THINKING` instead of adding `MODELS_TO_TEST_WITH_THINKING_TOOLS_STREAMING`

integrations/amazon_bedrock/tests/test_chat_generator.py
tstadel: They are legacy: 
image
Interestingly this is nowhere documented, neither as active, EoL, or Legacy under https://docs.aws.amazon.com/bedrock/latest/userguide/model-lifecycle.html

integrations/amazon_bedrock/tests/test_chat_generator.py
tstadel: Ok, added deepseek r1 to thinking, so these two are now actually different as deepseek r1 does not support tool calls.

integrations/amazon_bedrock/tests/test_chat_generator.py
sjrl: thanks for digging! And odd, I would have thought they would bring in the new ones ...

PR #3014, opened 2026-03-22
feat: Add support for Azure `TokenCredentials` to `AzureAISearchDocumentStore`

### LINKED ISSUE #1549: feat: support azure.identity.DefaultAzureCredential in AzureAISearchDocumentStore
Referring to https://github.com/deepset-ai/haystack/issues/8952 and https://github.com/deepset-ai/haystack/pull/9012, the same logic for àzure.identity.DefaultAzureCredential` layered authentication applies to `haystack_integrations.document_stores.azure_ai_search.AzureAISearchDocumentStore`

Same solution applied in https://github.com/deepset-ai/haystack/pull/9012

See https://github.com/deepset-ai/haystack/issues/8952

maxdswain:
### Related Issues

- fixes #1549

### Proposed Changes:

 
 
Add new `azure_token_provider` parameter to `AzureAISearchDocumentStore` allowing users to authenticate using `azure.identity.ClientSecretCredential`, `azure.identity.ManagedIdentityCredential` as well as their own class that matches the `TokenCredentials` protocol. Looking at what was implemented in https://github.com/deepset-ai/haystack/pull/9012, I saw that the token provider took priority, over the API key so I made this the case in this change as well.

### How did you test it?

I didn't add any tests as I don't changes are significant enough to require any.

### Notes for the reviewer

I wasn't sure on how to serialise the `azure_token_provider` parameter as it's important to accept both `azure.identity` objects as well as custom user-defined objects. This mean that I didn't want to have to enforce `to_dict`/`from_dict` in a protocol as this would mean `azure.identity` objects wouldn't be able to be used directly and would require more work from the user to implement.

I also looked into trying to write a automatic serialiser/deserialiser but as the protocol does not specify types/arguments for the `__init__` method, it would be difficult as objects could be defined by the user that were not easily serialisable or secrets could accidentally be serialised which is a security risk.

All things considered I decided to instead add a warning when a user tries to serialise `AzureAISearchDocumentStore` when `azure_token_provider` has been defined. This seemed like the best solution to me, but happy to get any feedback :)

### Checklist


integrations/azure_ai_search/src/haystack_integrations/document_stores/azure_ai_search/document_store.py
```diff
@@ -121,6 +121,7 @@ def __init__(
         metadata_fields: dict[str, SearchField | type] | None = None,
         vector_search_configuration: VectorSearch | None = None,
         include_search_metadata: bool = False,
+        azure_token_provider: TokenCredential | None = None,
```
bogdankostic: I'd rename this to `azure_token_credential`.

integrations/azure_ai_search/src/haystack_integrations/document_stores/azure_ai_search/document_store.py
```diff
@@ -155,6 +156,8 @@ def __init__(
             in the returned documents. When set to True, the `meta` field of the returned
             documents will contain the @search.score, @search.reranker_score, @search.highlights,
             @search.captions, and other fields returned by Azure AI Search.
+        :param azure_token_provider: A class with a `get_token` method that returns an Azure token,
+            will be invoked on every request.
```
bogdankostic:
```suggestion
        :param azure_token_credential: An Azure `TokenCredential` instance used to authenticate requests.
    When provided, this takes priority over `api_key`.
```

integrations/azure_ai_search/src/haystack_integrations/document_stores/azure_ai_search/document_store.py
```diff
@@ -315,6 +324,12 @@ def to_dict(self) -> dict[str, Any]:
         :returns:
             Dictionary with serialized data.
         """
+        if self._azure_token_provider:
+            logger.warning(
+                "`azure_token_provider` was used in initialization of AzureAISearchDocumentStore, "
+                "but does not support serializing it."
```
bogdankostic:
```suggestion
                "AzureAISearchDocumentStore was initialized with `azure_token_provider`, "
                "which cannot be serialized. It will be excluded from the serialized output "
                "and must be provided again when deserializing."
```

PR #3033, opened 2026-03-23
chore: increase lower pins for 3.14 support in some integrations + test with 3.14

anakin87:
### Related Issues

- part of https://github.com/deepset-ai/haystack-core-integrations/issues/2963

### Proposed Changes:

These integrations are compatible with Python 3.14, but the lower-bound dependency pins are not correct.

For example, an integration may declare "dependency>1.0.0" but `dependency==1.0.0` is not compatible with 3.14. Although standard package resolution installs the latest package versions, pins are not completely correct and can lead to resolution conflicts in projects with many dependencies.

Given that this is a residual use case, I'm doing these updates in a single PR and don't plan to release this change immediately, but wait for the next release.

- increase lower-bound pins for Python 3.14 support
- test integrations with Python 3.14 (instead of 3.13)
- add the 3.14 compatibility info to pyproject (not needed to make integrations installable, but good to have)

### How did you test it?
CI

### Checklist

### DIFF
--- a/integrations/azure_ai_search/pyproject.toml
+++ b/integrations/azure_ai_search/pyproject.toml
@@ -24,7 +24,7 @@ classifiers = [
   "Programming Language :: Python :: Implementation :: PyPy",
 ]
 dependencies = [
-  "haystack-ai>=2.26.0",
+  "haystack-ai>=2.26.1",
   "azure-search-documents>=11.5",
   "azure-identity"
 ]
--- a/integrations/azure_ai_search/tests/test_document_store.py
+++ b/integrations/azure_ai_search/tests/test_document_store.py
@@ -19,12 +19,17 @@ from azure.search.documents.indexes.models import (
 from haystack.dataclasses.document import Document
 from haystack.errors import FilterError
 from haystack.testing.document_store import (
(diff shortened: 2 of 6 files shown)


.github/workflows/hanlp.yml
```diff
@@ -31,13 +31,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-latest, windows-latest, macos-latest]
-        python-version: ["3.10", "3.13"]
-
-        # sentencepiece cannot be installed on Windows with Python 3.13
-        # https://github.com/google/sentencepiece/issues/1111
-        exclude:
-          - os: windows-latest
-            python-version: "3.13"
```
anakin87, line 40: this problem has been fixed on sentencepiece

.github/workflows/optimum.yml
```diff
@@ -50,9 +50,7 @@ jobs:
         run: pip install hatch "virtualenv<21.0.0"
 
       - name: Lint
-        # we check types with python 3.13 because with 3.9, the installation of some type stubs fails
-        # due to incompatibilities
```
anakin87, line 54: no longer true

PR #3058, opened 2026-03-26
fix: Nvidia - fix structured output syntax

### LINKED ISSUE #2838: Nvidia guided JSON test fails
**Describe the bug**
https://github.com/deepset-ai/haystack-core-integrations/actions/runs/21953077883/job/63409084941?pr=2835

**To Reproduce**
```python
def test_live_run_with_guided_json_schema(self):
      json_schema = {
          "type": "object",
          "properties": {"title": {"type": "string"}, "rating": {"type": "number"}},
          "required": ["title", "rating"],
      }
      chat_messages = [
          ChatMessage.from_user(
              """
          Return the title and the rating based on the following movie review according
          to the provided json schema.
          Review: Inception is a really well made film. I rate it four stars out of five."""
          )
      ]

      component = NvidiaChatGenerator(
          model="meta/llama-3.1-70b-instruct",
          generation_kwargs={"extra_body": {"guided_json": json_schema}, "temperature": 0.0},
      )

      results = component.run(chat_messages)
      assert len(results["replies"]) == 1
      message = results["replies"][0].text
      output = json.loads(message)
      assert output["title"] == "Inception"
      assert "rating" in output
```

```
FAILED tests/test_nvidia_chat_generator.py::TestNvidiaChatGenerator::test_live_run_with_guided_json_schema - json.decoder.JSONDecodeError: Expecting value: line 1 column 1 (char 0)
```

**What I tried**
- Switching to new [NIM syntax](https://docs.nvidia.com/nim/large-language-models/1.15.0/structured-generation.html#basic-example-movie-review): `extra_body={"guided_json": json_schema},` (no `nvext` key)
- Changing the model

This does not work and my impression is that the JSON schema is not correctly passed.

I can make the test pass including the schema in the prompt and telling the model to just return a JSON and nothing else, but at this point, the test would be meaningless.

Akash504-ai:
### Related Issues

* fixes #2838

---

### Proposed Changes

The `guided_json` schema passed via `generation_kwargs["extra_body"]` was incorrectly nested when sent to the NVIDIA API.

This resulted in payloads like:

```json
{
  "extra_body": {
    "extra_body": {
      "guided_json": {...}
    }
  }
}
```

Because of this, the model ignored the schema and returned non-JSON output, causing `JSONDecodeError` in tests.

#### Fix

* Flatten nested `extra_body` before making API calls
* Apply fix in both `run` and `run_async`
* Use explicit parent class calls instead of `super()` to avoid issues with the `@component` decorator

This ensures `guided_json` is correctly forwarded to the NVIDIA API.

---

### How did you test it?

* Ran all NVIDIA integration tests:

```bash
pytest integrations/nvidia/tests -vv
```

* Result:

  * 180 tests passed
  * 17 skipped (expected live tests)

---

### Notes for the reviewer

* The fix is scoped only to the NVIDIA integration
* No changes were made to core Haystack classes
* This improves correctness of structured generation without altering existing APIs

---

### Checklist

* [x] I have read the contributors guidelines and code of conduct
* [x] I added unit tests and updated the docstrings (no new tests needed, existing tests now pass)
* [x] I've used a conventional commit type (`fix:`)


integrations/nvidia/src/haystack_integrations/components/generators/nvidia/chat/chat_generator.py
```diff
@@ -127,6 +127,82 @@ def __init__(
...
+
+        return await OpenAIChatGenerator.run_async(
+            self,
+            messages=messages,
+            streaming_callback=streaming_callback,
+            generation_kwargs=generation_kwargs,
+            tools=tools,
+            tools_strict=tools_strict,
+        )
+
```
anakin87:
```suggestion
```

Sorry let me explain better. All this code can be removed.

What I suggest is

- changing the docstring below to reflect the new syntax
https://github.com/deepset-ai/haystack-core-integrations/blob/91688459761b98170577bf2e614b3f5e2c90278c/integrations/nvidia/src/haystack_integrations/components/generators/nvidia/chat/chat_generator.py#L90

- adapting the related test

PR #3088, opened 2026-04-02
feat: add structured output support via `outputConfig` to `AmazonBedrockChatGenerator`

### LINKED ISSUE #3067: Add structured output support for `AmazonBedrockChatGenerator`
We should add structured output support to `AmazonBedrockChatGenerator`. 

Here are the AWS docs explaining the feature https://docs.aws.amazon.com/bedrock/latest/userguide/structured-output.html. It will require adding support for `outputConfig.textFormat` in our response to the AWS Converse API. We currently support other configs like inferenceConfig, toolConfig, guardrailConfig, etc.

This came up in our call with @ju-gu and @bogdankostic where this feature is needed and currently being accomplished with a custom parsing component to handle the cases when the output of the LLM is not JSON compliant.

Akash504-ai:
### Related Issues

* Fixes #3067

---

### Proposed Changes

This PR adds **structured output support** to `AmazonBedrockChatGenerator` by introducing handling for the `outputConfig` parameter in requests to the Amazon Bedrock Converse API.

#### What was the issue?

Previously, any structured output configuration (e.g. `outputConfig.textFormat`) passed via `generation_kwargs` would be incorrectly forwarded under `additionalModelRequestFields`. However, according to AWS Bedrock API requirements, `outputConfig` must be provided as a **top-level parameter** in the request.

#### What was changed?

* Extracted `outputConfig` from `generation_kwargs`
* Ensured it is **removed from `additionalModelRequestFields`**
* Added it correctly to the request payload as:

  ```python
  params["outputConfig"] = output_config
  ```

#### Result

Users can now pass structured output configurations like:

```python
generation_kwargs={
    "outputConfig": {
        "textFormat": {
            "type": "json",
            "schema": {...}
        }
    }
}
```

and have them properly interpreted by the Bedrock Converse API.

---

### How did you test it?

* Verified that `outputConfig` is:

  * correctly extracted from `generation_kwargs`
  * not included in `additionalModelRequestFields`
  * correctly added to top-level request params
* Manually validated request construction logic against AWS documentation
* Ensured no regressions in existing parameter handling (`inferenceConfig`, `toolConfig`, etc.)

> Note: Full test suite execution was limited due to local environment conflicts (NumPy / Torch), but the change is isolated and does not affect unrelated components.

---

### Notes for the reviewer

* The change is minimal and scoped only to request preparation logic (`_prepare_request_params`)
* No breaking changes introduced
* Fully aligned with AWS Bedrock structured output documentation
* Works consistently for both sync and async execution paths

---

### Checklist

* [x] I have read the contributors guidelines and code of conduct
* [ ] I added unit tests and updated the docstrings *(can be added if required)*
* [x] I've used a conventional commit type (`feat:`) for the PR title

### DIFF
--- a/integrations/amazon_bedrock/src/haystack_integrations/components/generators/amazon_bedrock/chat/chat_generator.py
+++ b/integrations/amazon_bedrock/src/haystack_integrations/components/generators/amazon_bedrock/chat/chat_generator.py
@@ -1,3 +1,4 @@
+import json
 from typing import Any
 
 import aioboto3
@@ -35,6 +36,7 @@ from haystack_integrations.components.generators.amazon_bedrock.chat.utils impor
     _parse_completion_response,
     _parse_streaming_response,
     _parse_streaming_response_async,
+    _parse_structured_output,
     _validate_and_format_cache_point,
     _validate_guardrail_config,
 )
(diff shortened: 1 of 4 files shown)


davidsbatista:
run `hatch run fmt` - that should fix it

Akash504-ai:
I followed your suggestion and ran formatting specifically on the affected test files. Locally Black and Ruff both report that all files are already correctly formatted, and no changes are produced. I also aligned the Black version with CI, but the result is still the same.
It seems like there might be a mismatch between the CI formatting environment and the local setup.Could you please confirm if there is a specific command or environment I should replicate exactly? Happy to adjust accordingly.

integrations/amazon_bedrock/src/haystack_integrations/components/generators/amazon_bedrock/chat/chat_generator.py
davidsbatista, line 440: we can add `outputConfig` here

davidsbatista, line 601: we can add `outputConfig` here

davidsbatista, line 668: we can add `outputConfig` here

PR #3130, opened 2026-04-10
feat: (docling) Drop temp files for ByteStream sources

sjrl:
### Related Issues

- fixes #issue-number

### Proposed Changes:

 
 

In our docling converter `ByteStream` sources previously required writing to a temporary file. During a large indexing job this could create I/O issues which could cause a performance bottleneck. Docling also has a `DocumentStream` API which accepts a `BytesIO` directly, so the temp file write and `os.unlink` are no longer needed.

Also we try to do our best effort in guessing the filename passed to `DocumentStream` by checking `file_path`, `filename`, and `name` meta keys in that order, and falls back to guessing an extension from `mime_type` when no extension is present. This is to help docling reliably detect the file format even for `ByteStream` objects.

The conversion logic is extracted into a `_bytestream_to_document_stream` helper.

### How did you test it?

Added unit tests.

### Checklist

### DIFF
--- a/integrations/docling/src/haystack_integrations/components/converters/docling/converter.py
+++ b/integrations/docling/src/haystack_integrations/components/converters/docling/converter.py
@@ -9,6 +9,7 @@ from io import BytesIO
 from pathlib import Path
 from typing import Any
 
+from docling_core.types.io import DocumentStream
 from haystack import Document, component
 from haystack.components.converters.utils import normalize_metadata
 from haystack.dataclasses import ByteStream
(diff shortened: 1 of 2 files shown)


sjrl:
@anakin87 since you recently worked on this I thought you might be best to review

integrations/docling/src/haystack_integrations/components/converters/docling/converter.py
```diff
@@ -18,6 +19,27 @@
...
+    raw_name = meta.get("file_path") or meta.get("file_name") or meta.get("name")
+    if raw_name:
+        name = Path(raw_name).name
+        if not Path(name).suffix and source.mime_type:
+            ext = mimetypes.guess_extension(source.mime_type)
+            if ext:
+                name = f"{name}{ext}"
+    else:
+        ext = mimetypes.guess_extension(source.mime_type) if source.mime_type else ""
+        name = f"document{ext or ''}"
```
anakin87: I was originally a bit skeptical, but after investigating, I have found that Docling actually tries to use the extension to detect file type.

Would you be open to a simplification like this?

```python
      if raw_name:
          name = Path(raw_name).name                                                                                               
      else:                  
          name = "document"
                                                                                                                                   
      if not Path(name).suffix and source.mime_type:
          ext = mimetypes.guess_extension(source.mime_type)                                                                        
          if ext:            
              name = f"{name}{ext}"
```

anakin87 (approved):
Good idea!

I might have found a simplification opportunity

integrations/docling/src/haystack_integrations/components/converters/docling/converter.py
sjrl: done in https://github.com/deepset-ai/haystack-core-integrations/pull/3130/commits/1f5a3eabb7f142a5184a47a1710c89bf89ae05e8

PR #3131, opened 2026-04-10
fix: Ollama - better reasoning streaming support

anakin87:
### Related Issues

When experimenting with Gemma 4, I realized that on Ollama we have the same issue @sjrl noticed on the draft implementation of vLLM: `chunk.start` not correctly set when we pass from `reasoning` from `content`.

When used with `print_streaming_chunk`, this does not print `[ASSISTANT]`.

### Proposed Changes:
- better set `chunk.start` 

### How did you test it?
New test, tested locally in several scenarios (no reasoning, reasoning, tools)

### Checklist


sjrl (approved):
looks good and thanks!

PR #2482, opened 2025-10-30
Fastembed Colbert Reranker

xoaryaa:
### Related Issues

- fixes https://github.com/deepset-ai/haystack/issues/8245

### Proposed Changes:

**Feature:** Add `FastembedColbertReranker` — a ColBERT late-interaction (MaxSim) reranker inside the FastEmbed integration.

- New component: `haystack_integrations.components.rankers.fastembed.FastembedColbertReranker`
- Uses FastEmbed’s `LateInteractionTextEmbedding` to encode query and documents (token-level).
- Implements ColBERT MaxSim scoring: `sum_q max_d (sim(q_i, d_j))` with configurable `similarity={"cosine","dot"}` and optional L2 normalization.
- CPU-friendly ONNX backend via FastEmbed; no new core Haystack deps.
- Batching support (`batch_size`) and optional token limits (`max_query_tokens`, `max_doc_tokens`).
- Stable sorting with deterministic tie-break; parameter validation and robust warm-up tolerant to minor FastEmbed kwargs drift.
- Small README note + CHANGELOG entry for the FastEmbed integration.
- Example script: `examples/fastembed_colbert_reranker.py`.

**Why:** Cross-encoder rankers don’t cover late-interaction/bi-encoder models like ColBERT v2. This adds a first-class ColBERT reranker over ~100–500 retrieved candidates.

### How did you test it?

- **Unit tests**
  - Pure math checks for MaxSim and L2 normalization.
  - Parameter validation (invalid `similarity`, `batch_size`, `top_k`).
  - Deterministic ranking & `top_k` via monkeypatched encoders (no model download required).

- **Manual verification**
  1. `pip install -e integrations/fastembed`
  2. Run `python examples/fastembed_colbert_reranker.py`  
     - First run downloads the ColBERT model via FastEmbed (cached afterward).
     - Verified that documents mentioning “late interaction/ColBERT” rank at the top.

### Notes for the reviewer

- Lives entirely under **`integrations/fastembed/src/...`** (no changes to core Haystack).
- Warm-up handles minor API differences in FastEmbed by retrying without unknown kwargs.
- Intended usage: **rerank ~100–500 candidates** from a first-stage retriever (BM25 or dense).
- Example import path:
  ```python
  from haystack_integrations.components.rankers.fastembed import FastembedColbertReranker

PR #3135, opened 2026-04-13
feat: add FastembedLateInteractionRanker for ColBERT late-interaction reranking

keosung:
Add a new ranker component that uses FastEmbed's LateInteractionTextEmbedding to perform ColBERT MaxSim scoring for document reranking.

Related: deepset-ai/haystack#8245

### Related Issues

- Related to deepset-ai/haystack#8245

### Proposed Changes:

- Added `FastembedColbertRanker` component that uses FastEmbed's `LateInteractionTextEmbedding` for ColBERT-based document reranking
- Follows the existing `FastembedRanker` pattern (same parameters, serialization, test structure)
- Scoring uses MaxSim: for each query token, find the max similarity across all document tokens, then sum

### How did you test it?

- 12 unit tests covering init, serialization, input validation, metadata embedding, and MaxSim scoring with mocks
- 1 integration test with `colbert-ir/colbertv2.0` model

### Notes for the reviewer

- Two previous attempts exist (deepset-ai/haystack#10458, #2482). Incorporated feedback from those reviews: implementation lives in fastembed integration, follows existing component pattern, no extra `__init__.py` files, no separate utility files.

### Checklist

### DIFF
--- a/integrations/fastembed/pydoc/config_docusaurus.yml
+++ b/integrations/fastembed/pydoc/config_docusaurus.yml
@@ -5,6 +5,7 @@ loaders:
       - haystack_integrations.components.embedders.fastembed.fastembed_sparse_document_embedder
       - haystack_integrations.components.embedders.fastembed.fastembed_sparse_text_embedder
       - haystack_integrations.components.rankers.fastembed.ranker
+      - haystack_integrations.components.rankers.fastembed.late_interaction_ranker
     search_path: [../src]
 processors:
   - type: filter
--- a/integrations/fastembed/pyproject.toml
+++ b/integrations/fastembed/pyproject.toml
@@ -22,7 +22,7 @@ classifiers = [
   "Programming Language :: Python :: Implementation :: CPython",
   "Programming Language :: Python :: Implementation :: PyPy",
 ]
-dependencies = ["haystack-ai>=2.22.0", "fastembed>=0.4.2"]
+dependencies = ["haystack-ai>=2.24.1", "fastembed>=0.4.2"]
 
 [project.urls]
 Source = "https://github.com/deepset-ai/haystack-core-integrations"
--- a/integrations/fastembed/src/haystack_integrations/components/rankers/fastembed/__init__.py
+++ b/integrations/fastembed/src/haystack_integrations/components/rankers/fastembed/__init__.py
@@ -2,6 +2,7 @@
 #
 # SPDX-License-Identifier: Apache-2.0
 
+from .late_interaction_ranker import FastembedLateInteractionRanker
 from .ranker import FastembedRanker
 
-__all__ = ["FastembedRanker"]
+__all__ = ["FastembedLateInteractionRanker", "FastembedRanker"]
--- a/integrations/fastembed/src/haystack_integrations/components/rankers/fastembed/late_interaction_ranker.py
+++ b/integrations/fastembed/src/haystack_integrations/components/rankers/fastembed/late_interaction_ranker.py
@@ -1,0 +1,230 @@
+# SPDX-FileCopyrightText: 2024-present deepset GmbH <info@deepset.ai>
+#
+# SPDX-License-Identifier: Apache-2.0
+
+from dataclasses import replace
+from typing import Any
+
+import numpy as np
+from haystack import Document, component, default_from_dict, default_to_dict, logging
+from haystack.utils.misc import _deduplicate_documents
+
+from fastembed import LateInteractionTextEmbedding
+
+logger = logging.getLogger(__name__)
+
+
+@component
+class FastembedLateInteractionRanker:
+    """
+    Ranks Documents based on their similarity to the query using ColBERT models via Fastembed.
+
+    Uses late interaction (MaxSim) scoring to compute token-level similarity between
+    query and document embeddings, then ranks documents accordingly.
+
+    See https://qdrant.github.io/fastembed/examples/Supported_Models/ for supported models.
+
+    Usage example:
+    ```python
+    from haystack import Document
+    from haystack_integrations.components.rankers.fastembed import FastembedLateInteractionRanker
+
+    ranker = FastembedLateInteractionRanker(model_name="colbert-ir/colbertv2.0", top_k=2)
+
+    docs = [Document(content="Paris"), Document(content="Berlin")]
+    query = "What is the capital of germany?"
+    output = ranker.run(query=query, documents=docs)
+    print(output["documents"][0].content)
+
+    # Berlin
+    ```
+    """
+
+    def __init__(
+        self,
+        model_name: str = "colbert-ir/colbertv2.0",
+        top_k: int = 10,
+        cache_dir: str | None = None,
+        threads: int | None = None,
+        batch_size: int = 64,
+        parallel: int | None = None,
+        local_files_only: bool = False,
+        meta_fields_to_embed: list[str] | None = None,
+        meta_data_separator: str = "\n",
+        score_threshold: float | None = None,
+    ) -> None:
+        """
+        Creates an instance of the 'FastembedLateInteractionRanker'.
+
+        :param model_name: Fastembed ColBERT model name. Check the list of supported models in the
+            [Fastembed documentation](https://qdrant.github.io/fastembed/examples/Supported_Models/).
+        :param top_k: The maximum number of documents to return.
+        :param cache_dir: The path to the cache directory.
+                Can be set using the `FASTEMBED_CACHE_PATH` env variable.
+                Defaults to `fastembed_cache` in the system's temp directory.
+        :param threads: The number of threads single onnxruntime session can use. Defaults to None.
+        :param batch_size: Number of strings to encode at once.
+        :param parallel:
+                If > 1, data-parallel encoding will be used, recommended for offline encoding of large datasets.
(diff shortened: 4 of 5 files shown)


dina-deifallah:
## PR Review — `FastembedColbertRanker` (#3135)

**Reviewed with:** manual code review + functional testing against a RAG pipeline using `colbert-ir/colbertv2.0`

---

### Testing results

All 4 tests passed after one fix (see pipeline wiring note below):

| Test | Result | Notes |
|------|--------|-------|
| Instantiation & serialization | ✅ | `to_dict`/`from_dict` round-trip works correctly |
| Ranking comparison | ✅ | ColBERT changed order on 2/3 ambiguous queries — reranking adds value |
| Edge cases | ✅ | Single doc, `top_k > len(docs)`, empty query, metadata preservation all handled |
| Pipeline integration | ✅ | Full retriever→ranker→LLM pipeline ran in ~4s, correct answer |

---

### Issues

**1. Missing document deduplication** *(medium)*

Other rankers (`SentenceTransformersSimilarityRanker`, `FastembedRanker`) call `_deduplicate_documents()` before scoring to avoid returning duplicate documents. This ranker skips that step. Suggest adding:

```python
from haystack.utils.misc import _deduplicate_documents

# in run(), before _prepare_fastembed_input_docs:
documents = _deduplicate_documents(documents)
```

---

**2. Missing `score_threshold` parameter** *(medium)*

`FastembedRanker` supports a `score_threshold` parameter to filter out low-confidence results. For consistency, this ranker should too. Note that ColBERT scores are unnormalized sums (observed range ~3–25 in testing), so the docs should clarify what a reasonable threshold looks like.

---

**3. Missing `_get_telemetry_data`** *(low)*

`FastembedRanker` implements this for usage analytics. Should be added for consistency:

```python
def _get_telemetry_data(self) -> dict[str, Any]:
    return {"model": self.model_name}
```

---

**4. `next(iter(...))` is redundant** *(low)*

```python
# Current:
query_embedding = next(iter(self._model.query_embed(query)))
# Simpler — query_embed already returns a generator:
query_embedding = next(self._model.query_embed(query))
```

---

**5. Pipeline ergonomics — `query` input must be passed via `run()`** *(worth documenting)*

The ranker's `query` input cannot be wired from an upstream component (e.g. you can't do `pipeline.connect("some_component.text", "ranker.query")`). It must be supplied directly in `pipeline.run()`:

```python
pipeline.run({
    "text_embedder": {"text": question},
    "ranker":        {"query": question},   # must be passed here explicitly
    "prompt_builder":{"question": question},
})
```

This is valid Haystack behaviour, but it's non-obvious for users coming from `SentenceTransformersSimilarityRanker`. Worth adding a note to the docstring or usage example.

---

### Test file issues

**6. Self-connection bug in the example script**

The `.py` test script contains:
```python
rag.connect("ranker.query", "ranker.query")   # BUG — connects component to itself
```
This raises `PipelineConnectError`. The `query` input should be passed via `run()` instead (as in point 5 above). The notebook version already has the correct pattern.

**7. Missing `warm_up` idempotency test** *(low)*

There's no test verifying that calling `warm_up()` twice only initializes the model once. Given the model is ~500 MB, this matters:

```python
def test_warm_up_called_once(self):
    ranker = FastembedColbertRanker()
    with patch("...LateInteractionTextEmbedding") as mock:
        ranker.warm_up()
        ranker.warm_up()
        mock.assert_called_once()
```

---

### What's good

- MaxSim scoring implementation is correct: `matmul → max over doc tokens → sum over query tokens`
- Correctly uses `query_embed` for the query and `embed` for documents (ColBERT requires different encoding paths)
- Clean parameter design, good docstrings, proper serialization
- Mock-based `test_run_with_mock` test is excellent — it validates the actual scoring math with controlled embeddings
- 12 unit tests + 1 integration test is solid coverage
[test_colbert_ranker.py](https://github.com/user-attachments/files/26672260/test_colbert_ranker.py)

keosung:
Thank you for the thorough review and for testing against a real RAG pipeline!

I've addressed all the feedback:
  - Simplified next(iter(...)) to next(...)
  - Added _deduplicate_documents() before scoring
  - Added score_threshold parameter
  - Added _get_telemetry_data()
  - Added a test_warm_up_called_once test to verify the model is only initialized once
  - Added a note to the class docstring clarifying that query must be passed explicitly via pipeline.run()

dina-deifallah:
Thank you! I am now updating the documentation.

keosung:
I ran into two issues while applying the feedback:
1. _deduplicate_documents causes an ImportError in CI. No other integration in this repo uses it either. Should I remove it, or is there a preferred way to handle this?
3. Changing next(iter(...)) to next(...) causes a type check error - query_embed() is typed as Iterable in fastembed, so next() requires iter() to wrap it first. I've kept next(iter(...)) for now.

integrations/fastembed/src/haystack_integrations/components/rankers/fastembed/colbert_ranker.py
```diff
@@ -0,0 +1,238 @@
...
+from haystack import Document, component, default_from_dict, default_to_dict, logging
+from haystack.utils.misc import _deduplicate_documents
+
+from fastembed import LateInteractionTextEmbedding
+
+logger = logging.getLogger(__name__)
+
+
+@component
+class FastembedColbertRanker:
```
anakin87: To be more future-proof, we thought of naming this `FastembedLateInteractionRanker`
@julian-risch right?

integrations/fastembed/src/haystack_integrations/components/rankers/fastembed/colbert_ranker.py
```diff
@@ -0,0 +1,238 @@
...
+    See https://qdrant.github.io/fastembed/examples/Supported_Models/ for supported models.
+
+    Note: The `query` input cannot be connected to an upstream component in a pipeline.
+    Pass it directly via `pipeline.run()`:
+    ```python
+    pipeline.run({
+        "retriever": {"query": question},
+        "ranker": {"query": question},  # must be passed explicitly
+    })
+    ```
```
anakin87:
```suggestion
    Note: The `query` input cannot be connected to an upstream component in a pipeline.
    Pass it directly via `pipeline.run()`:
    ```python
    pipeline.run({
        "retriever": {"query": question},
        "ranker": {"query": question},  # must be passed explicitly
    })
    ```
```

Sorry, I know this was suggested but I don't think it's correct 🙂

integrations/fastembed/src/haystack_integrations/components/rankers/fastembed/colbert_ranker.py
```diff
@@ -0,0 +1,238 @@
...
+
+        if self._model is None:
+            self.warm_up()
+
+        documents = _deduplicate_documents(documents)
+        fastembed_input_docs = self._prepare_fastembed_input_docs(documents)
+
+        query_embedding = next(iter(self._model.query_embed(query)))  # type: ignore[union-attr]
+        document_embeddings = list(
+            self._model.embed(  # type: ignore[union-attr]
```
anakin87:
```suggestion
        assert self._model is not None    # noqa: S101

        query_embedding = next(iter(self._model.query_embed(query)))
        document_embeddings = list(
            self._model.embed(
```

this should help with mypy errors (we use this pattern)

anakin87 (changes requested):
A good work! I left some comments...

anakin87:
> I ran into two issues while applying the feedback:
> 
> 1. _deduplicate_documents causes an ImportError in CI. No other integration in this repo uses it either. Should I remove it, or is there a preferred way to handle this?
> 2. Changing next(iter(...)) to next(...) causes a type check error - query_embed() is typed as Iterable in fastembed, so next() requires iter() to wrap it first. I've kept next(iter(...)) for now.

1. This utility method was added in Haystack 2.24. So I recommend changing https://github.com/deepset-ai/haystack-core-integrations/blob/101e23f2a2af5540589f432fd6be6d502d9880b9/integrations/fastembed/pyproject.toml#L25
to use `haystack-ai>=2.24.1`. (The failing test just checks if our lower-bound pins are precise)

2. OK

keosung:
Thanks for the feedback! I've addressed all the comments.
Should I also rename the file from colbert_ranker.py to late_interaction_ranker.py for consistency?

integrations/fastembed/src/haystack_integrations/components/rankers/fastembed/colbert_ranker.py
julian-risch: Yes, seems better to me too

anakin87:
> Thanks for the feedback! I've addressed all the comments. Should I also rename the file from colbert_ranker.py to late_interaction_ranker.py for consistency?

Yes, please

keosung:
Done, renamed accordingly.

anakin87:
Thank you.

Last thing to fix: we should add the new component to https://github.com/deepset-ai/haystack-core-integrations/blob/main/integrations/fastembed/pydoc/config_docusaurus.yml

This ensures that the API reference for documentation is generated from docstrings

keosung:
Done! Added late_interaction_ranker to config_docusaurus.yml.

anakin87 (approved):
Very good work!

Thank you

PR #3139, opened 2026-04-13
refactor: improve docling-serve integration

cbrumm:
## Summary

**Draft created at PyCon DE Sprints — feedback and review welcome!**

Relates to #2960

- Adds a new `DoclingServeConverter` component that converts documents via a running [docling-serve](https://github.com/docling-project/docling-serve) REST API instance
- Separate from the existing `docling-haystack` integration to avoid heavy local dependencies (PyTorch, etc.)
- Supports file paths, `ByteStream` objects, and URLs as sources
- Implements both `run()` (sync via `httpx.Client`) and `run_async()` (via `httpx.AsyncClient`)
- Integration tests run against a real `docling-serve-cpu` Docker container on Linux CI runners

**Note that `docling-server` docker images are quite large, I think around 4 GB! See [pypi docling-serve](https://pypi.org/project/docling-serve/0.16.1/#:~:text=docling%2Dserve%2Dcu128-,quay.io%2Fdocling%2Dproject%2Fdocling%2Dserve%2D,Get%20help%20and).**

### Design decisions
- **Keyword-only init** with `base_url`, `api_key` (`Secret`), `timeout`, `convert_options` (single dict for all docling-serve parameters)
- **Dual endpoint support**: URL strings route to `/v1/convert/source` (v1 `sources` format with `kind` discriminator), local files and ByteStreams to `/v1/convert/file`
- **`httpx`** for both sync and async HTTP in a single dependency
- **Graceful error handling**: HTTP errors are logged as warnings; processing continues for remaining sources
- **Integration tests**: CI starts `ghcr.io/docling-project/docling-serve-cpu` via `docker run` on Linux runners, with health-check polling. macOS/Windows skip integration tests.

## Test plan
- [x] `hatch run fmt` — all checks passed
- [x] `hatch run test:types` — no issues found
- [x] `hatch run test:unit` — 30/30 passed
- [x] Integration test: file conversion via `/v1/convert/file`
- [x] Integration test: URL conversion via `/v1/convert/source` (validates v1 `sources` format)
- [x] Review metadata attached to returned Documents
- [x] Review convert_options passthrough behavior

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #2960, opened 2026-03-13
Add new docling-serve integration


maxdswain:
## Summary and motivation

Docling is an extremely popular document information extraction library and in production uses cases, [docling-serve](https://github.com/docling-project/docling-serve) can be used to host Docling in a more performant scalable fashion.

## Detailed design

Follow the design of the already existing [docling-haystack](https://github.com/docling-project/docling-haystack) introducing a new `DoclingServeConverter` component accepting a list of `str`/`pathlib.Path` and producing a `dict[str, list[Document]]` but instead using the Docling serve endpoints following Docling serve's [documentation](https://github.com/docling-project/docling-serve/blob/main/docs/usage.md). An `run_async` method would also be good to implement as the requests to Docling serve with be blocking.

I considered extending the existing Docling integration, but it has large dependencies (e.g., pytorch) which are unneeded when sending requests to Docling serve.

## Checklist

If the request is accepted, ensure the following checklist is complete before closing this issue.

### Tasks
- [ ] The code is documented with docstrings and was merged in the `main` branch
- [ ] Docs are published at https://docs.haystack.deepset.ai/
- [ ] There is a Github workflow running the tests for the integration nightly and at every PR
- [ ] A new label named like `integration:<your integration name>` has been added to the list of labels for this [repository](https://github.com/deepset-ai/haystack-core-integrations/labels)
- [ ] The [labeler.yml](https://github.com/deepset-ai/haystack-core-integrations/blob/main/.github/labeler.yml) file has been updated
- [ ] The package has been released on PyPI
- [ ] An integration tile with a usage example has been added to https://github.com/deepset-ai/haystack-integrations
- [ ] The integration has been listed in the [Inventory section](https://github.com/deepset-ai/haystack-core-integrations#inventory) of this repo README
- [ ] The feature was announced through social media

julian-risch:
@maxdswain We do have some draft code here in a duplicate issue if you are still interested in working on this and have time. If not, not problem at all. Just let us know. https://github.com/deepset-ai/haystack-core-integrations/issues/3128#issue-4237257957 and copied below.

julian-risch:
One of our Sols @OGuggenbuehl has a custom docling-serve version of the docling converter that we should build on for the docling-serve integration. 

@sjrl provided a **draft version** below that could help us get started. We should make sure to update the draft version with our best practices such as:
- enforce keyword only arguments in the init
- make sure to use our Secret class for the api_key
[...]

PR #3150, opened 2026-04-13
fix: skip invalid docling serve sources

turkeshp:
### Related Issues

- Related to #2960
- Follow-up to #3139

### Proposed Changes:

- handle missing or invalid local sources in `DoclingServeConverter.run()`
- handle the same case in `run_async()`
- log a warning and continue processing remaining sources
- add sync and async test coverage for this behavior

### How did test it?

- `python -m py_compile src\haystack_integrations\components\converters\docling_serve\converter.py tests\test_converter.py`
- `python -m pytest tests\test_converter.py -q -k missing_file_logged_and_skipped`

### Notes for the reviewer

This PR is intentionally small and builds on top of `#3139`.
It does not change the integration design; it only makes invalid local file inputs non-fatal in sync and async conversion.

### Checklist

 I have read the contributors guidelines and the code of conduct
 I have updated the related issue with new insights and changes
 I added unit tests and updated the docstrings
 I've used one of the conventional commit types for my PR title

This is a follow-up on top of #3139, not a competing implementation.

The only intended change here is making invalid local sources non-fatal in `run()` and `run_async()`, with focused test coverage for that behavior.

Happy to work this PR and have the fix cherry-picked into #3139 if that is the preferred workflow.


sjrl:
Hey @turkeshp could you recheck if your changes are needed? We recently updated the docling serve integration.

julian-risch:
I've ported the commit that contains the fix onto the current `main` in #3449 and credited you @turkeshp as co-author. The fix is still valuable. The current `run()`/`run_async()` don't catch local-file read failures, so a missing/unreadable file raises `FileNotFoundError` and aborts the whole batch.

PR #3177, opened 2026-04-17
fix(google-genai): include cached_content_token_count in streaming responses

### LINKED ISSUE #3168: `cached_content_token_count` is missing from streaming responses in Google GenAI chat generator
**Describe the bug**
The `cached_content_token_count` field from the Google GenAI API's `GenerateContentResponseUsageMetadata` is not being extracted in streaming responses. Additionally, the streaming code path lacks the explicit check for this field entirely.

**Steps to Reproduce:**

1. Use `GoogleGenAIChatGenerator` with streaming enabled
2. Make a request that uses Google Cloud context caching (or any request where `cachedContentTokenCount` is non-zero in the API response)
3. Inspect the `meta.usage` of the returned `ChatMessage`

Expected behavior:

`meta.usage` should contain `cached_content_token_count` when the API returns it, e.g.:

```
{
    "prompt_tokens": 19882,
    "completion_tokens": 185,
    "total_tokens": 20451,
    "thoughts_token_count": 384,
    "cached_content_token_count": 16214,  # ← should be present
}
```

Actual behavior:

`cached_content_token_count` is completely missing from `meta.usage` in streaming mode.

**Root Cause:**

In `_convert_google_chunk_to_streaming_chunk`, only `prompt_tokens`, `completion_tokens`, `total_tokens`, and `thoughts_token_count` are extracted. There is no check for `cached_content_token_count`, unlike in `_convert_google_genai_response_to_chatmessage` (non-streaming path) where an explicit check exists.

Additionally, `cached_content_token_count` is also missing from `_usage_attr_names` in `_convert_usage_metadata_to_serializable`, meaning it can only be added via the explicit check — which streaming code doesn't have at all.

The field only appears on the final streaming chunk (index N) where `cache_tokens_details` is also populated. All earlier chunks have `cached_content_token_count=None`.

**Proposed Fix:**

1. Add `cached_content_token_count` to `_usage_attr_names` in _convert_usage_metadata_to_serializable` so it's always extracted when present
2. Add an explicit check for `cached_content_token_count` in `_convert_google_chunk_to_streaming_chunk`, similar to the existing `thoughts_token_count` check
3. Consider adding it to the non-streaming path's `_convert_google_genai_response_to_chatmessage` as well for consistency

Additional context: The same fix should likely be applied to `_aggregate_streaming_chunks_with_reasoning` if it merges usage metadata, to ensure the final aggregated message includes the cached token count from the final chunk.

 - OS: `Linux Ubuntuzarp 6.17.0-20-generic #20~24.04.1-Ubuntu SMP PREEMPT_DYNAMIC Thu Mar 19 01:28:37 UTC 2 x86_64 x86_64 x86_64 GNU/Linux`
 - Haystack version: 2.24.1
 - Integration version: 3.11.0+

Aftabbs:
## Summary

Fixes #3168

`cached_content_token_count` from the Google GenAI API usage metadata was
being populated in **non-streaming** responses but was completely absent
from **streaming** responses.

**Root cause (two missing pieces):**

1. `_convert_google_chunk_to_streaming_chunk` only checked for
   `thoughts_token_count` but had no equivalent check for
   `cached_content_token_count`. The usage dict built for each chunk
   therefore never included the cached token count.

2. `_aggregate_streaming_chunks_with_reasoning` forwarded
   `thoughts_token_count` from the final chunk into the aggregated
   message, but had no corresponding logic for
   `cached_content_token_count`.

**Fix:**

- Add an explicit `cached_content_token_count` check in
  `_convert_google_chunk_to_streaming_chunk`, matching the pattern
  already used by the non-streaming path.
- Propagate `cached_content_token_count` through
  `_aggregate_streaming_chunks_with_reasoning`, mirroring how
  `thoughts_token_count` is handled.
- Add two unit tests covering the streaming chunk conversion and the
  aggregator.

## How to test

```python
from haystack.dataclasses import StreamingChunk, ComponentInfo
from haystack_integrations.components.generators.google_genai.chat.utils import (
    _convert_google_chunk_to_streaming_chunk,
    _aggregate_streaming_chunks_with_reasoning,
)
from unittest.mock import Mock

# Streaming chunk with cached tokens
mock_usage = Mock()
mock_usage.prompt_token_count = 1000
mock_usage.candidates_token_count = 10
mock_usage.total_token_count = 1010
mock_usage.thoughts_token_count = None
mock_usage.cached_content_token_count = 800
# ... (build mock chunk) ...
assert chunk.meta["usage"]["cached_content_token_count"] == 800

# Aggregated message
result = _aggregate_streaming_chunks_with_reasoning([chunk1, final_chunk])
assert result.meta["usage"]["cached_content_token_count"] == 800
```

### DIFF
--- a/integrations/google_genai/CHANGELOG.md
+++ b/integrations/google_genai/CHANGELOG.md
@@ -1,11 +1,5 @@
 # Changelog
 
-## [Unreleased]
-
-### 🐛 Bug Fixes
-
-- Fix `cached_content_token_count` missing from streaming responses in `GoogleGenAIChatGenerator` (#3168)
-
 ## [integrations/google_genai-v4.0.1] - 2026-04-13
 
 ### 🐛 Bug Fixes
--- a/integrations/google_genai/CHANGELOG.md
+++ b/integrations/google_genai/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## [Unreleased]
+
+### 🐛 Bug Fixes
+
+- Fix `cached_content_token_count` missing from streaming responses in `GoogleGenAIChatGenerator` (#3168)
+
 ## [integrations/google_genai-v4.0.1] - 2026-04-13
 
 ### 🐛 Bug Fixes
--- a/integrations/google_genai/src/haystack_integrations/components/generators/google_genai/chat/utils.py
+++ b/integrations/google_genai/src/haystack_integrations/components/generators/google_genai/chat/utils.py
@@ -734,7 +734,7 @@ def _aggregate_streaming_chunks_with_reasoning(chunks: list[StreamingChunk]) -> 
                 # We'll keep the last set of signatures as they represent the complete state
                 thought_signatures = signature_deltas
 
-        # Extract thinking token usage (from the last chunk that has it)
+        # Extract token usage metadata (from the last chunk that has it)
         if chunk.meta and "usage" in chunk.meta:
             chunk_usage = chunk.meta["usage"]
             if "thoughts_token_count" in chunk_usage:
--- a/integrations/google_genai/src/haystack_integrations/components/generators/google_genai/chat/utils.py
+++ b/integrations/google_genai/src/haystack_integrations/components/generators/google_genai/chat/utils.py
@@ -625,6 +625,14 @@ def _convert_google_chunk_to_streaming_chunk(
     if usage_metadata and hasattr(usage_metadata, "thoughts_token_count") and usage_metadata.thoughts_token_count:
         usage["thoughts_token_count"] = usage_metadata.thoughts_token_count
 
+    # Add cached content token count if available (context caching)
+    if (
+        usage_metadata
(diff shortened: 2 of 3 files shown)


integrations/google_genai/CHANGELOG.md
bogdankostic, line 1: Please revert the changes made here, the Changelog is generated automatically whenever we release a new version.

integrations/google_genai/src/haystack_integrations/components/generators/google_genai/chat/utils.py
bogdankostic, line 737: Let's update this comment, we don't only extract thinking token usage here.

bogdankostic (changes requested):
Thanks @Aftabbs, looks good overall, just left two minor comments.

bogdankostic (approved):
Looking good to me, thanks @Aftabbs!

PR #3178, opened 2026-04-17
fix: docling validation error

ArzelaAscoIi:
## Summary

Fix a `ValidationError` raised by `DoclingConverter.run` when processing documents that contain section headers with a heading level greater than 100.

Docling's `DocumentConverter` builds its internal document representation using `model_construct()`, which bypasses Pydantic validation. This allows `SectionHeaderItem.level` to hold values above the upstream `LevelNumber` maximum of 100. The error surfaces later in `DoclingConverter.run` when the document is passed to chunking or export operations that trigger Pydantic re-validation.

**Changes:**
- Added `_clamp_section_header_levels(dl_doc)` helper that iterates `dl_doc.texts` and replaces any `SectionHeaderItem` with `level > 100` with a valid copy capped at 100 (using `model_copy`, not direct mutation). The cap of 100 matches the `LevelNumber` constraint in `docling-core`.
- The helper is called immediately after each document is converted, before any export path (DOC_CHUNKS, MARKDOWN, JSON), so all three paths benefit.
- Updated existing tests whose mocks passed plain strings as `document` objects — these are now proper `SimpleNamespace(texts=[])` objects so the new iteration over `dl_doc.texts` doesn't fail.
- Added `TestClampSectionHeaderLevels` unit tests covering: level capped, level unchanged at/below 100, non-`SectionHeaderItem` entries left untouched, mixed-item lists.
- Added `test_run_doc_chunks_with_high_section_header_level` end-to-end unit test verifying the converter completes without error and the level is capped in the output.

## Test plan

- [ ] `hatch run test:unit` passes (33 tests)
- [ ] `hatch run test:types` passes
- [ ] `hatch run fmt` produces no changes

### DIFF
--- a/integrations/docling/src/haystack_integrations/components/converters/docling/converter.py
+++ b/integrations/docling/src/haystack_integrations/components/converters/docling/converter.py
@@ -9,6 +9,7 @@ from io import BytesIO
 from pathlib import Path
 from typing import Any
 
+from docling_core.types.doc.document import SectionHeaderItem
 from docling_core.types.io import DocumentStream
 from haystack import Document, component
 from haystack.components.converters.utils import normalize_metadata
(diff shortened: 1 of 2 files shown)


sjrl:
Hey @davidsbatista I can take the review of this one

sjrl:
Hey @ArzelaAscoIi thanks for opening the PR! A few questions so I better understand why this change is needed. 

- What kinds of documents did you run into that ran into this validation error?

- And would it be worth you opening an issue for this in docling's repo?

sjrl:
Hey @ArzelaAscoIi I'm going to close for now since when running the doc in question locally with this integration the doc successfully was converted. 

I commented on linear that I think the issue is with the custom component implementation.

PR #3195, opened 2026-04-21
feat: add e2b-haystack integration for E2B cloud sandbox tools

tholor:
partially addresses #3227  

## Summary

Migrates the E2B sandbox toolset prototype from [deepset-ai/haystack-experimental#448](https://github.com/deepset-ai/haystack-experimental/pull/448) into a proper `e2b-haystack` integration package.

- Adds `E2BSandbox` for managing the lifecycle of an E2B cloud sandbox (lazy `warm_up`, `close`, full serialisation round-trip)
- Adds four Haystack `Tool` subclasses that share a single sandbox instance: `RunBashCommandTool`, `ReadFileTool`, `WriteFileTool`, `ListDirectoryTool`
- Adds `E2BToolset` (a `Toolset` subclass) that bundles all four tools as a convenience — just pass `E2BToolset()` to any Haystack `Agent`
- Follows the `haystack_integrations/tools/` module path convention (same as `mcp-haystack` and `github-haystack`)

## Test plan

- [x] 38 unit tests pass (`hatch run test:unit`) — all sandbox calls are mocked, no API key needed
- [x] Lint clean (`hatch run fmt-check`)
- [ ] Review examples in `integrations/e2b/examples/` for correctness
- [x] Manual smoke test with a real E2B API key (integration test, requires `E2B_API_KEY`)
- [x] Maintainer to add `E2B_API_KEY` secret to GitHub repo for CI integration tests


.github/labeler.yml
```diff
@@ -64,10 +64,17 @@ integration:docling:
...
+integration:e2b:
+  - changed-files:
+      - any-glob-to-any-file: "integrations/e2b/**/*"
+      - any-glob-to-any-file: ".github/workflows/e2b.yml"
+=======
 integration:dspy:
   - changed-files:
       - any-glob-to-any-file: "integrations/dspy/**/*"
       - any-glob-to-any-file: ".github/workflows/dspy.yml"
+>>>>>>> main
```
bogdankostic: Let's fix this unresolved merge conflict and keep this list in alphabetical order.

```suggestion
integration:dspy:
  - changed-files:
      - any-glob-to-any-file: "integrations/dspy/**/*"
      - any-glob-to-any-file: ".github/workflows/dspy.yml"

integration:e2b:
  - changed-files:
      - any-glob-to-any-file: "integrations/e2b/**/*"
      - any-glob-to-any-file: ".github/workflows/e2b.yml"
```

integrations/e2b/examples/e2b_agent_example.py
```diff
@@ -0,0 +1,97 @@
...
+
+Demonstrates that all four tools (run_bash_command, read_file, write_file,
+list_directory) share the same sandbox instance, so the agent can write a
+file in one step and read it back / execute it in the next.
+
+Requirements:
+    pip install e2b-haystack openai
+
+Environment variables:
+    E2B_API_KEY   - your E2B API key
```
bogdankostic, line 16: Let's align the descriptions of the API keys here.
```suggestion
    E2B_API_KEY    - your E2B API key
```

integrations/e2b/examples/e2b_pipeline_example.py
```diff
@@ -0,0 +1,101 @@
...
+
+    assert type(rest_ts).__name__ == "E2BToolset", "Toolset type mismatch"
+    assert [t.name for t in rest_ts] == [t.name for t in orig_ts], "Tool names mismatch"
+    assert rest_ts.sandbox.sandbox_template == orig_ts.sandbox.sandbox_template
+    assert rest_ts.sandbox.timeout == orig_ts.sandbox.timeout
+
+    sandbox_ids = {id(t._e2b_sandbox) for t in rest_ts}
+    assert len(sandbox_ids) == 1, "Tools should share a single sandbox after round-trip"
+
+    print("All assertions passed: YAML round-trip preserves pipeline structure.\n")
```
bogdankostic, line 82: In my opinion, we can remove this from the example, as this is rather testing that serialization works as expected than showcasing how to use this integration.

integrations/e2b/pydoc/config_docusaurus.yml
```diff
@@ -0,0 +1,14 @@
+loaders:
+  - modules:
```
bogdankostic, line 2: I think we should also include the remaining modules here to be displayed in our API reference docs.

integrations/e2b/src/haystack_integrations/tools/e2b/e2b_sandbox.py
```diff
@@ -0,0 +1,185 @@
...
+
+class E2BSandbox:
+    """
+    Manages the lifecycle of an E2B cloud sandbox.
+
+    Instantiate this class and pass it to one or more E2B tool classes
+    (``RunBashCommandTool``, ``ReadFileTool``, ``WriteFileTool``,
+    ``ListDirectoryTool``) to share a single sandbox environment across all
+    tools.  All tools that receive the same ``E2BSandbox`` instance operate
+    inside the same live sandbox process.
```
bogdankostic, line 28: We use single backticks in doc strings.
```suggestion
    Instantiate this class and pass it to one or more E2B tool classes
    (`RunBashCommandTool`, `ReadFileTool`, `WriteFileTool`,
    `ListDirectoryTool`) to share a single sandbox environment across all
    tools.  All tools that receive the same `E2BSandbox` instance operate
    inside the same live sandbox process.
```

integrations/e2b/src/haystack_integrations/tools/e2b/e2b_sandbox.py
```diff
@@ -0,0 +1,185 @@
...
+    ```python
+    sandbox.warm_up()
+    # ... use tools ...
+    sandbox.close()
+    ```
+    """
+
+    def __init__(
+        self,
+        api_key: Secret | None = None,
```
bogdankostic: We usually directly use the secret as a default value.
```suggestion
        api_key: Secret = Secret.from_env_var("E2B_API_KEY", strict=True)
```

integrations/e2b/src/haystack_integrations/tools/e2b/e2b_sandbox.py
```diff
@@ -0,0 +1,185 @@
...
+        sandbox_template: str = "base",
+        timeout: int = 120,
+        environment_vars: dict[str, str] | None = None,
+    ) -> None:
+        """
+        Create an E2BSandbox instance.
+
+        :param api_key: E2B API key. Defaults to ``Secret.from_env_var("E2B_API_KEY")``.
+        :param sandbox_template: E2B sandbox template name. Defaults to ``"base"``.
+        :param timeout: Sandbox inactivity timeout in seconds. Defaults to ``120``.
```
bogdankostic: In my opinion we can remove the default values here, as users can infer them from the signature.

integrations/e2b/src/haystack_integrations/tools/e2b/e2b_sandbox.py
```diff
@@ -0,0 +1,185 @@
...
+    # ------------------------------------------------------------------
+
+    def _require_sandbox(self) -> "Sandbox":
+        """Return the active sandbox or raise a helpful error."""
+        if self._sandbox is None:
+            msg = (
+                "E2B sandbox is not running. Call warm_up() before using the tools, "
+                "or add the sandbox to a Haystack pipeline/agent which calls warm_up() automatically."
+            )
+            raise RuntimeError(msg)
```
bogdankostic: I think instead of raising the error here that the sandbox is not running, we could just call warm_up.

integrations/e2b/README.md
bogdankostic, line 1: We typically keep these READMEs minimal and just add a sentence that an API key is required to run the integration tests (see for example the [Google GenAI README](https://github.com/deepset-ai/haystack-core-integrations/blob/main/integrations/google_genai/README.md)).

Usage instructions are added to https://github.com/deepset-ai/haystack-integrations once the integration is released.

integrations/e2b/src/haystack_integrations/tools/e2b/bash_tool.py
```diff
@@ -0,0 +1,94 @@
...
+
+    def to_dict(self) -> dict[str, Any]:
+        """Serialize this tool to a dictionary."""
+        return {
+            "type": generate_qualified_class_name(type(self)),
+            "data": {"sandbox": self._e2b_sandbox.to_dict()},
+        }
+
+    @classmethod
+    def from_dict(cls, data: dict[str, Any]) -> "RunBashCommandTool":
```
bogdankostic, line 91: If a user builds an `Agent` with the four tools passed individually (sharing one `E2BSandbox`) rather than via `E2BToolset`, and then serializes/deserializes the pipeline, each tool's from_dict constructs its own independent `E2BSandbox`, breaking the shared state of the tools.

I tried this out by adapting `build_pipeline` in `e2b_pipeline_example`:
```python
def build_pipeline() -> Pipeline:
    sandbox = E2BSandbox(sandbox_template="base", timeout=120)
    tools = [
        WriteFileTool(sandbox=sandbox),
        RunBashCommandTool(sandbox=sandbox),
    ]
    agent = Agent(
        chat_generator=OpenAIChatGenerator(model="gpt-4o-mini"),
        tools=tools,
        system_prompt=(
            "You are a helpful coding assistant with access to a live Linux sandbox. "
            "Use the available tools freely to write files and run commands."
        ),
        max_agent_steps=10,
    )
    pipeline = Pipeline()
    pipeline.add_component("agent", agent)
    return pipeline
```

Executing the script with this version of `build_pipeline`, we get the following error:
```
e2b_pipeline_example.py", line 88, in verify_roundtrip
    assert len(sandbox_ids) == 1, "Tools should share a single sandbox after round-trip"
           ^^^^^^^^^^^^^^^^^^^^^
AssertionError: Tools should share a single sandbox after round-trip
```

bogdankostic (changes requested):
Looking forward to having this integration soon in Haystack! 

I've left a few minor comments, mostly focused on polishing the docstrings. The only major blocker we should address IMO is the deserialization logic for individual tools to ensure the sandbox environments remain consistent within one pipeline.

integrations/e2b/src/haystack_integrations/tools/e2b/bash_tool.py
tholor: @bogdankostic good point - do we have a similar pattern somewhere else in haystack that we can follow here for serializing a shared object?

bogdankostic: I don't think there's a straightforwar way to preserve a shared `E2BSandbox` across individually-serialized tools. The round-trip would need some form of cross-tool deduplication.

A quick solution would be to log a warning in each tool's `to_dict` pointing users at `E2BToolset` for the serialize/deserialize path. 

Something like:

```python
def to_dict(self) -> dict[str, Any]:
    logger.warning(
        "Serializing %s standalone will not preserve a shared E2BSandbox "
        "across tools after deserialization. If you need to serialize an "
        "agentic pipeline (e.g. to YAML) with multiple E2B tools sharing "
        "one sandbox, use E2BToolset instead.",
        type(self).__name__,
    )
    return {
        "type": generate_qualified_class_name(type(self)),
        "data": {"sandbox": self._e2b_sandbox.to_dict()},
    }
```
WDYT?

integrations/e2b/src/haystack_integrations/tools/e2b/py.typed
sjrl, line 1: This file should be removed. Only need a `py.typed` file on the same level as the `e2b` directory

integrations/e2b/e2b.md
sjrl, line 1: I think this file is more appropriate to have in the https://github.com/deepset-ai/haystack-integrations repo

integrations/e2b/pyproject.toml
```diff
@@ -0,0 +1,167 @@
...
+
+[tool.hatch.envs.test]
+dependencies = [
+    "pytest",
+    "pytest-asyncio",
+    "pytest-cov",
+    "pytest-rerunfailures",
+    "mypy",
+    "pip",
+    "e2b>=2.0.0",
```
sjrl: We can drop this dep from the tests env since it's already in the main project deps

integrations/e2b/e2b.md
bogdankostic: I think this is the API reference, so doesn't fit the haystack-integrations repo either but will be automatically created once this PR is merged, so we should just remove this file from here.

integrations/e2b/src/haystack_integrations/tools/e2b/bash_tool.py
tholor: Hm ok ... I guess the above warning but we the simplest approach. 
I explored one alternative to make the round-trip work (see latest [commit](https://github.com/deepset-ai/haystack-core-integrations/pull/3195/changes/987889aabd99068cfaebcb81aab5c6891f389758)):  

- **Identity-based dedup inside `E2BSandbox.from_dict`** — give each `E2BSandbox` a UUID at construction, serialize it, and have `from_dict` consult a class-level `weakref.WeakValueDictionary` to return the already-restored instance for the same UUID. Transparent to all callers (tools, toolset, ad-hoc usage).

Pro: Enables production / platform use cases, where you regularly have to serialize / deserialize pipelines. Showing an warning there is not really useful and we should then rather disable the standalone tools for such cases and always push users directly to "E2BToolset" 

Con: Makes the code more complex.

What do you think? @sjrl @bogdankostic

bogdankostic: I just tried it out, works like a charm 🙌🏼

integrations/e2b/src/haystack_integrations/tools/e2b/e2b_sandbox.py
```diff
@@ -0,0 +1,238 @@
...
+        :param sandbox_template: E2B sandbox template name.
+        :param timeout: Sandbox inactivity timeout in seconds.
+        :param environment_vars: Optional environment variables to inject into the sandbox.
+        :param instance_id: Stable identifier preserved across `to_dict`/`from_dict`. When
+            omitted, a fresh UUID is generated. Tools that share the same `E2BSandbox`
+            instance inherit this id, which is what lets them re-share the instance after
+            a serialization round-trip. Distinct from the cloud-side sandbox id assigned
+            by E2B at warm-up.
+        """
+        self.api_key = api_key or Secret.from_env_var("E2B_API_KEY")
```
bogdankostic: The default is already `Secret.from_env_var("E2B_API_KEY")`.
```suggestion
        self.api_key = api_key
```

bogdankostic (approved):
Nice, looking good now, just found one minor improvement related to changing the default value of `api_key` that I will directly change.

PR #3197, opened 2026-04-21
fix: remove double-wrapping of tools_cachepoint_config in _format_tools

### LINKED ISSUE #3181: AmazonBedrockChatGenerator Bug: `tools_cachepoint_config` causes `ParamValidationError`
**Describe the bug**

When passing `tools_cachepoint_config={"type": "default"}` to `AmazonBedrockChatGenerator`,
the resulting Bedrock API call contains a double-wrapped `cachePoint` block and fails with a `ParamValidationError`.

The issue is that `__init__` pre-formats the value via `_validate_and_format_cache_point`,
storing `{"cachePoint": {"type": "default"}}` on `self.tools_cachepoint_config`.
`_format_tools` then wraps it a second time, emitting `{"cachePoint": {"cachePoint": {"type": "default"}}}`,
which Bedrock rejects.

```
ParamValidationError: Parameter validation failed:
Missing required parameter in toolConfig.tools[N].cachePoint: "type"
Unknown parameter in toolConfig.tools[N].cachePoint: "cachePoint", must be one of: type, ttl
```

**Root cause**

`_format_tools` in `utils.py` treats `tools_cachepoint_config` as a raw user value and wraps it:

```python
# utils.py — current (broken)
if tools_cachepoint_config:
    tool_specs.append({"cachePoint": tools_cachepoint_config})
#                     ^^^^^^^^^^^ wraps again — tools_cachepoint_config is already {"cachePoint": {...}}
```

But by the time `_format_tools` is called, `tools_cachepoint_config` has already been formatted by
`_validate_and_format_cache_point` in `__init__` and is `{"cachePoint": {"type": "default"}}`.

The fix is a single-line change in `_format_tools`:

```python
# utils.py — fixed
if tools_cachepoint_config:
    tool_specs.append(tools_cachepoint_config)  # already {"cachePoint": {...}}, append directly
```

**To Reproduce**

```python
from haystack_integrations.components.generators.amazon_bedrock.chat.chat_generator import (
    AmazonBedrockChatGenerator,
)

def think(thought: str) -> str:
    """Think step."""
    return ""

    model="anthropic.claude-3-5-sonnet-20241022-v2:0",
    tools=[Tool.from_function(think)],
    tools_cachepoint_config={"type": "default"},  # documented API
)

# Fails immediately with ParamValidationError — no actual Bedrock call is made
generator.run(messages=[ChatMessage.from_user("Hello")])
```

The double-wrapping can be observed without any API call:

```python
from haystack_integrations.components.generators.amazon_bedrock.chat.utils import (
    _validate_and_format_cache_point,
    _format_tools,
)

stored = _validate_and_format_cache_point({"type": "default"})
print(stored)
# → {"cachePoint": {"type": "default"}}   (stored as self.tools_cachepoint_config)

class FakeTool:
    name, description, parameters = "think", "think", {}

result = _format_tools([FakeTool()], tools_cachepoint_config=stored)
print(result["tools"][-1])
# → {"cachePoint": {"cachePoint": {"type": "default"}}}   ← sent to Bedrock, rejected
# expected: {"cachePoint": {"type": "default"}}
```

- OS: macOS 15 (Darwin)
- Haystack version: 2.25.2
- Integration version: amazon-bedrock-haystack 6.8.0

SyedShahmeerAli12:
## Summary

Fixes #3181

- `AmazonBedrockChatGenerator.__init__` calls `_validate_and_format_cache_point` which wraps `{"type": "default"}` → `{"cachePoint": {"type": "default"}}` and stores it as `self.tools_cachepoint_config`
- `_format_tools` was then wrapping it a **second time**, producing `{"cachePoint": {"cachePoint": {"type": "default"}}}` which Bedrock rejects with `ParamValidationError`
- Fix: append `tools_cachepoint_config` directly in `_format_tools` since it is already in Bedrock format by the time it arrives
- Updated the existing `test_format_tools` test to pass the pre-formatted value (matching real call flow)

## Test plan

- [x] Existing unit tests pass (291 passed)
- [x] `test_format_tools` updated to reflect actual call flow (pre-wrapped value passed to `_format_tools`)

### DIFF
--- a/integrations/amazon_bedrock/src/haystack_integrations/components/generators/amazon_bedrock/chat/utils.py
+++ b/integrations/amazon_bedrock/src/haystack_integrations/components/generators/amazon_bedrock/chat/utils.py
@@ -93,7 +93,7 @@ def _format_tools(
         )
 
     if tools_cachepoint_config:
-        tool_specs.append({"cachePoint": tools_cachepoint_config})
+        tool_specs.append(tools_cachepoint_config)
 
     return {"tools": tool_specs}
 
--- a/integrations/amazon_bedrock/tests/test_chat_generator_utils.py
+++ b/integrations/amazon_bedrock/tests/test_chat_generator_utils.py
@@ -67,7 +67,7 @@ def tools():
 
 class TestAmazonBedrockChatGeneratorUtils:
     def test_format_tools(self, tools):
-        formatted_tool = _format_tools(tools, tools_cachepoint_config={"type": "default"})
+        formatted_tool = _format_tools(tools, tools_cachepoint_config={"cachePoint": {"type": "default"}})
         assert formatted_tool == {
             "tools": [
                 {


anakin87:
Sorry @SyedShahmeerAli12, your fix was right but I merged https://github.com/deepset-ai/haystack-core-integrations/pull/3199 because it also contained an additional test.

Thank you again!

PR #3198, opened 2026-04-21
feat!: change default fuzziness from AUTO to 0 in OpenSearch retrievers

### LINKED ISSUE #3187: Turn fuzziness off by default in Open Search integration
## Summary and motivation

We have had cases that the turning fuzziness on by default in the OpenSearch integration has caused large performance issues for clients when they have a lot of documents in their index.

More details on the performance issues can be found here: https://linear.app/deepset/issue/BUI-327/turn-off-fuzziness-in-opensearchhybridretriever-in-pipeline-templates

To summarize we have had cases where search times with the OpenSearchHybridRetriever went from 30 seconds to 6 seconds when turning fuzziness off.

So the proposal is to switch the default value of fuzziness from `AUTO` to `0` so it's off by default in the `OpenSearchHybridRetriever` and the `OpenSearchBm25Retriever`. This would be a breaking change so we should make a new major version release.

SyedShahmeerAli12:
## Summary

Closes #3187

- `fuzziness="AUTO"` was causing severe performance regressions on large indexes — search times going from **30s → 6s** when turned off
- Changed the default from `"AUTO"` to `0` (exact matching) in:
  - `OpenSearchBM25Retriever`
  - `OpenSearchHybridRetriever`
  - `OpenSearchDocumentStore._bm25_retrieval` / `_bm25_retrieval_async`
- Users who need fuzzy matching can still pass `fuzziness="AUTO"` explicitly — no functionality removed, just the default changes

## Test plan

- [x] All 143 unit tests pass
- [x] Updated `to_dict` serialization tests to reflect new default
- [x] Updated `run()` call assertion tests to reflect new default
- [x] `from_dict` tests that explicitly pass `"AUTO"` are unchanged (backward compat with old serialized configs)


sjrl:
Hey @SyedShahmeerAli12 you also touched amazon bedrock files which are unrelated to your PR, please remove them.

SyedShahmeerAli12:
@sjrl removed the Amazon Bedrock files, the PR now only contains the OpenSearch fuzziness  changes.

integrations/opensearch/src/haystack_integrations/components/retrievers/opensearch/bm25_retriever.py
sjrl: Update docstring to indicate what the new default is

sjrl (approved):
Thanks!

PR #3209, opened 2026-04-23
feat: Add better language support presidio

sjrl:
### Related Issues

- fixes #issue-number

### Proposed Changes:

 
 

Update the presidio integration to properly pass `language` to the `supported_languages` param of `AnalyzerEngine` and add a new param called `models` which is required if using languages other than English to tell Presidio what model to use to power the NlpEngine.

For example, to run the document cleaner in german this now works and the german model is automatically downloaded at warm up time. 
```python
from haystack import Document
from haystack_integrations.components.preprocessors.presidio import PresidioDocumentCleaner

cleaner = PresidioDocumentCleaner(
    language="de",
    models=[{"lang_code": "de", "model_name": "de_core_news_lg"}],
)
cleaner.warm_up()
docs = [Document(content="Mein Name ist Hans Müller und meine E-Mail ist hans@example.com")]
result = cleaner.run(documents=docs)
```

### How did you test it?

Added unit and integration tests.

### Checklist

### DIFF
--- a/integrations/presidio/src/haystack_integrations/components/common/presidio/utils.py
+++ b/integrations/presidio/src/haystack_integrations/components/common/presidio/utils.py
@@ -1,0 +1,33 @@
+# SPDX-FileCopyrightText: 2022-present deepset GmbH <info@deepset.ai>
+#
+# SPDX-License-Identifier: Apache-2.0
+
+# Maps ISO 639-1 language codes to the largest available spaCy model for that language.
+# Used to automatically configure the NLP engine when only `language` is specified.
+# See https://spacy.io/models for the full list of available models.
+SPACY_DEFAULT_MODELS: dict[str, str] = {
+    "ca": "ca_core_news_lg",
+    "zh": "zh_core_web_lg",
+    "hr": "hr_core_news_lg",
+    "da": "da_core_news_lg",
+    "nl": "nl_core_news_lg",
+    "en": "en_core_web_lg",
+    "fi": "fi_core_news_lg",
+    "fr": "fr_core_news_lg",
+    "de": "de_core_news_lg",
+    "el": "el_core_news_lg",
+    "it": "it_core_news_lg",
+    "ja": "ja_core_news_lg",
+    "ko": "ko_core_news_lg",
+    "lt": "lt_core_news_lg",
+    "mk": "mk_core_news_lg",
+    "nb": "nb_core_news_lg",
+    "pl": "pl_core_news_lg",
+    "pt": "pt_core_news_lg",
+    "ro": "ro_core_news_lg",
+    "ru": "ru_core_news_lg",
+    "sl": "sl_core_news_lg",
+    "es": "es_core_news_lg",
+    "sv": "sv_core_news_lg",
+    "uk": "uk_core_news_lg",
+}
--- a/integrations/presidio/src/haystack_integrations/components/extractors/presidio/presidio_entity_extractor.py
+++ b/integrations/presidio/src/haystack_integrations/components/extractors/presidio/presidio_entity_extractor.py
@@ -112,9 +112,7 @@ class PresidioEntityExtractor:
                 raise ValueError(msg)
             models = [{"lang_code": self.language, "model_name": self.SPACY_DEFAULT_MODELS[self.language]}]
 
-        nlp_engine = NlpEngineProvider(
-            nlp_configuration={"nlp_engine_name": "spacy", "models": models}
-        ).create_engine()
+        nlp_engine = NlpEngineProvider(nlp_configuration={"nlp_engine_name": "spacy", "models": models}).create_engine()
         supported_languages = [m["lang_code"] for m in models]
         self._analyzer = AnalyzerEngine(nlp_engine=nlp_engine, supported_languages=supported_languages)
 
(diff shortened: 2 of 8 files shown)


bogdankostic (commented):
Looks good in principle, I'm just wondering if we should remove the `language` parameter, given that if a user wants to use a non-english model, they have to set the `models` parameter. In that case `self.language` is not used to initialize the `AnalyzerEngine`.

sjrl:
> Looks good in principle, I'm just wondering if we should remove the `language` parameter, given that if a user wants to use a non-english model, they have to set the `models` parameter. In that case `self.language` is not used to initialize the `AnalyzerEngine`.

Maybe instead we could add a set of pre-defined mappings between language and model so users could just use the language param?

bogdankostic:
> > Looks good in principle, I'm just wondering if we should remove the `language` parameter, given that if a user wants to use a non-english model, they have to set the `models` parameter. In that case `self.language` is not used to initialize the `AnalyzerEngine`.
> 
> Maybe instead we could add a set of pre-defined mappings between language and model so users could just use the language param?

SOunds good, that way users wouldn't need to figure out which model to use for their language.

integrations/presidio/src/haystack_integrations/components/common/presidio/utils.py
```diff
@@ -0,0 +1,33 @@
+# SPDX-FileCopyrightText: 2022-present deepset GmbH <info@deepset.ai>
+#
+# SPDX-License-Identifier: Apache-2.0
+
+# Maps ISO 639-1 language codes to the largest available spaCy model for that language.
+# Used to automatically configure the NLP engine when only `language` is specified.
+# See https://spacy.io/models for the full list of available models.
+SPACY_DEFAULT_MODELS: dict[str, str] = {
```
sjrl, line 8: @bogdankostic I added a default mapping for models here

integrations/presidio/src/haystack_integrations/components/extractors/presidio/presidio_entity_extractor.py
```diff
@@ -40,18 +44,29 @@ class PresidioEntityExtractor:
     ```
     """
 
+    SPACY_DEFAULT_MODELS: ClassVar[dict[str, str]] = _SPACY_DEFAULT_MODELS
+    """Mapping from ISO 639-1 language code to the largest available spaCy model for that language.
+
+    Used to automatically select an NLP model when `models` is not specified.
+    See https://spacy.io/models for the full list of available spaCy models.
+    """
```
sjrl, line 52: I also added it as a class var so it should appear in api docstrings like we have for ChatGenerators (example [here](https://docs.haystack.deepset.ai/reference/generators-api#supported_models))

integrations/presidio/src/haystack_integrations/components/extractors/presidio/presidio_entity_extractor.py
```diff
@@ -40,18 +44,29 @@ class PresidioEntityExtractor:
     ```
     """
 
+    SPACY_DEFAULT_MODELS: ClassVar[dict[str, str]] = _SPACY_DEFAULT_MODELS
+    """Mapping from ISO 639-1 language code to the largest available spaCy model for that language.
+
+    Used to automatically select an NLP model when `models` is not specified.
+    See https://spacy.io/models for the full list of available spaCy models.
```
bogdankostic:
```suggestion
    See [spaCy documentation](https://spacy.io/models) for the full list of available spaCy models.
```

integrations/presidio/src/haystack_integrations/components/preprocessors/presidio/presidio_document_cleaner.py
```diff
@@ -36,18 +41,29 @@ class PresidioDocumentCleaner:
     ```
     """
 
+    SPACY_DEFAULT_MODELS: ClassVar[dict[str, str]] = _SPACY_DEFAULT_MODELS
+    """Mapping from ISO 639-1 language code to the largest available spaCy model for that language.
+
+    Used to automatically select an NLP model when `models` is not specified.
+    See https://spacy.io/models for the full list of available spaCy models.
```
bogdankostic:
```suggestion
    See [spaCy documentation](https://spacy.io/models) for the full list of available spaCy models.
```

integrations/presidio/src/haystack_integrations/components/preprocessors/presidio/presidio_text_cleaner.py
```diff
@@ -33,18 +38,29 @@ class PresidioTextCleaner:
     ```
     """
 
+    SPACY_DEFAULT_MODELS: ClassVar[dict[str, str]] = _SPACY_DEFAULT_MODELS
+    """Mapping from ISO 639-1 language code to the largest available spaCy model for that language.
+
+    Used to automatically select an NLP model when `models` is not specified.
+    See https://spacy.io/models for the full list of available spaCy models.
```
bogdankostic:
```suggestion
    See [spaCy documentation](https://spacy.io/models) for the full list of available spaCy models.
```

bogdankostic (approved):
Looking good to me now! Just added a minor suggestion to improve the doc string of the class var.

PR #3223, opened 2026-04-24
feat: implement chonkie integration with four chunkers

### LINKED ISSUE #2237: Add new Chonkie integration with RecursiveChunker, SemanticChunker and others
## Summary and motivation

[Chonkie](https://github.com/chonkie-inc/chonkie) is a chunking library. With [release 1.2](https://github.com/chonkie-inc/chonkie/releases/tag/v1.2.0), Chonkie comes with [8 chunkers](https://github.com/chonkie-inc/chonkie?tab=readme-ov-file#chunkers) that could all be made available in a Haystack integration. In particular, SemanticChunker and RecursiveChunker.

### Tasks
- [ ] A new label named like `integration:<your integration name>` has been added to the list of labels for this [repository](https://github.com/deepset-ai/haystack-core-integrations/labels)
- [x] The [labeler.yml](https://github.com/deepset-ai/haystack-core-integrations/blob/main/.github/labeler.yml) file has been updated
- [x] The integration has been listed in the [Inventory section](https://github.com/deepset-ai/haystack-core-integrations#inventory) of this repo README
- [ ] An integration tile with a usage example has been added to https://github.com/deepset-ai/haystack-integrations

yugborana:
### Related Issues

- fixes #2237 

### Proposed Changes:

 Implemented the Chonkie integration for Haystack by building four modular preprocessing components: ChonkieSemanticChunker, ChonkieRecursiveChunker, ChonkieTokenChunker, and ChonkieSentenceChunker.

Each component initializes its respective Chonkie engine with parameters (like thresholds, chunk_overlap, and custom delimiter rules). When executed in a pipeline (run()), the components delegate the input text to the underlying Chonkie algorithm. Finally, they package the resulting chunks back into new Haystack Document objects—preserving the original metadata while injecting new, precise annotations like start_index, end_index, and token_count.
### How did you test it?

Unit Tests: Built 16 tests covering initialization, to_dict/from_dict serialization, and run() logic for all 4 chunkers. Ran hatch run test:unit and achieved a 100% pass rate.
Integration Tests: N/A (Chonkie runs locally and doesn't rely on external databases or cloud APIs).
Manual Verification: Successfully executed live_test.py on real models (minishlab/potion-base-32M) to confirm dynamic semantic splits, token overlaps, and custom delimiter extraction without using mocks.
Instructions for Manual Tests:
cd integrations/chonkie
Run unit tests: hatch run test:unit
Run the live demonstration: hatch run python live_test.py
### Checklist

yes - I have read the [contributors guidelines](https://github.com/deepset-ai/haystack-core-integrations/blob/main/CONTRIBUTING.md) and the [code of conduct](https://github.com/deepset-ai/haystack-core-integrations/blob/main/CODE_OF_CONDUCT.md)
yes - I added unit tests and updated the docstrings
yes - I've used one of the [conventional commit types](https://www.conventionalcommits.org/en/v1.0.0/) for my PR title: `fix:`, `feat:`, `build:`, `chore:`, `ci:`, `docs:`, `style:`, `refactor:`, `perf:`, `test:`.


.github/workflows/chonkie.yml
```diff
@@ -0,0 +1,139 @@
...
+
+concurrency:
+  group: chonkie-${{ github.head_ref || github.sha }}
+  cancel-in-progress: true
+
+env:
+  PYTHONUNBUFFERED: "1"
+  FORCE_COLOR: "1"
+  TEST_MATRIX_OS: '["ubuntu-latest", "windows-latest", "macos-latest"]'
+  TEST_MATRIX_PYTHON: '["3.10", "3.13"]'
```
sjrl, line 33: In our latest template script `create_new_integration.py` this is automatically `TEST_MATRIX_PYTHON: '["3.10", "3.14"]'`. Does Chonkie not yet support python 3.14 or should this be updated?

integrations/chonkie/src/haystack_integrations/components/preprocessors/chonkie/recursive_splitter.py
```diff
@@ -0,0 +1,113 @@
...
+        print(result["documents"])
+    """
+
+    def __init__(
+        self,
+        tokenizer: str = "character",
+        chunk_size: int = 2048,
+        min_characters_per_chunk: int = 24,
+        rules: Any = None,
+    ) -> None:
```
sjrl, line 42: For new components we typically like to enforce keyword only params so  
```suggestion
    def __init__(
        self,
        *,
        tokenizer: str = "character",
        chunk_size: int = 2048,
        min_characters_per_chunk: int = 24,
        rules: Any = None,
    ) -> None:
```

integrations/chonkie/src/haystack_integrations/components/preprocessors/chonkie/recursive_splitter.py
```diff
@@ -0,0 +1,113 @@
...
+        """
+        Initializes the ChonkieRecursiveChunker.
+
+        :param tokenizer: The tokenizer to use for chunking. Defaults to "character".
```
sjrl, line 46: Let's link to chonkie docs that explains all available tokenizer values.

integrations/chonkie/src/haystack_integrations/components/preprocessors/chonkie/recursive_chunker.py
```diff
@@ -0,0 +1,113 @@
...
+        tokenizer: str = "character",
+        chunk_size: int = 2048,
+        min_characters_per_chunk: int = 24,
+        rules: Any = None,
+    ) -> None:
+        """
+        Initializes the ChonkieRecursiveChunker.
+
+        :param tokenizer: The tokenizer to use for chunking. Defaults to "character".
+        :param chunk_size: The maximum size of each chunk.
```
sjrl: Let's explain what `chunk_size` refers to. Is it number of characters? Does it depend on what the tokenizer is?

integrations/chonkie/src/haystack_integrations/components/preprocessors/chonkie/recursive_splitter.py
```diff
@@ -0,0 +1,113 @@
...
+        :param chunk_size: The maximum size of each chunk.
+        :param min_characters_per_chunk: The minimum number of characters per chunk.
+        :param rules: Custom rules for recursive chunking. If None, default rules are used.
```
sjrl, line 51: Let's link to chonkie docs so users can understand what rules can look like. Also can we type this more narrowly than `Any`? 

We want to make sure we support serialization/deserialization of our components to JSON and it looks like `rules` might not be easy to do that with if we don't know it's type.

integrations/chonkie/tests/test_integration.py
```diff
@@ -0,0 +1,70 @@
...
+from haystack_integrations.components.preprocessors.chonkie import (
+    ChonkieRecursiveChunker,
+    ChonkieSentenceChunker,
+    ChonkieTokenChunker,
+)
+
+
+@pytest.mark.integration
+class TestChonkieChunkersIntegration:
+    def test_recursive_chunker(self):
```
sjrl: Let's put the integration tests for each chunker in it's own existing file.

And then get rid of this `test_integration.py` file.

integrations/chonkie/tests/test_recursive_splitter.py
```diff
@@ -0,0 +1,68 @@
...
+
+
+class TestChonkieRecursiveChunker:
+    def test_init_default(self):
+        chunker = ChonkieRecursiveChunker()
+        assert chunker.chunk_size == 2048
+        assert chunker.tokenizer == "character"
+        assert chunker.min_characters_per_chunk == 24
+
+    def test_to_dict(self):
```
sjrl, line 18: Let's add another `to_dict` test for when you specify `rules` as well.

integrations/chonkie/tests/test_recursive_splitter.py
```diff
@@ -0,0 +1,68 @@
...
+            "type": "haystack_integrations.components.preprocessors.chonkie.recursive_chunker.ChonkieRecursiveChunker",
+            "init_parameters": {
+                "chunk_size": 1024,
+                "tokenizer": "word",
+                "min_characters_per_chunk": 10,
+                "rules": None,
+            },
+        }
+
+    def test_from_dict(self):
```
sjrl, line 60: Same comment as [here](https://github.com/deepset-ai/haystack-core-integrations/pull/3223/changes#r3135846249) let's add another from_dict test (or extend this one) to include when rules is not None.

integrations/chonkie/tests/test_recursive_chunker.py
```diff
@@ -0,0 +1,68 @@
...
+        chunker = ChonkieRecursiveChunker(chunk_size=10, min_characters_per_chunk=2)
+        doc = Document(content="Hello world! This is a test string for chunking.")
+        result = chunker.run(documents=[doc])
+
+        assert "documents" in result
+        chunks = result["documents"]
+        assert len(chunks) > 0
+        assert chunks[0].meta["source_id"] == doc.id
+        assert "start_index" in chunks[0].meta
+        assert "end_index" in chunks[0].meta
```
sjrl: Could we do a full meta assertion instead of one-by-one keys? I'd like to make sure we are testing all returned meta and that's easier to see with 

```suggestion
        assert chunks[0].meta == {"source_id": doc.id, ...}
```

Also let's test all chunks not just the first one.

integrations/chonkie/tests/test_recursive_chunker.py
```diff
@@ -0,0 +1,68 @@
...
+        assert chunker.min_characters_per_chunk == 10
+
+    def test_run(self):
+        chunker = ChonkieRecursiveChunker(chunk_size=10, min_characters_per_chunk=2)
+        doc = Document(content="Hello world! This is a test string for chunking.")
+        result = chunker.run(documents=[doc])
+
+        assert "documents" in result
+        chunks = result["documents"]
+        assert len(chunks) > 0
```
sjrl: Let's make this explicit so test for the exact number of chunks that should be returned.

integrations/chonkie/src/haystack_integrations/components/preprocessors/chonkie/recursive_chunker.py
```diff
@@ -0,0 +1,113 @@
...
+            "chunk_size": chunk_size,
+            "min_characters_per_chunk": min_characters_per_chunk,
+        }
+        if rules is not None:
+            kwargs["rules"] = rules
+
+        self._chunker = chonkie.RecursiveChunker(**kwargs)
+
+    @component.output_types(documents=list[Document])
+    def run(self, documents: list[Document]) -> dict[str, Any]:
```
sjrl:
```suggestion
    def run(self, documents: list[Document]) -> dict[str, list[Document]]:
```

integrations/chonkie/src/haystack_integrations/components/preprocessors/chonkie/recursive_chunker.py
```diff
@@ -0,0 +1,113 @@
...
+        :returns: A dictionary with the "documents" key containing the list of chunks.
+        """
+        if not isinstance(documents, list) or (documents and not isinstance(documents[0], Document)):
+            msg = "ChonkieRecursiveChunker expects a list of Document objects."
+            raise TypeError(msg)
+
+        chunked_documents = []
+        for doc in documents:
+            if not doc.content:
+                continue
```
sjrl: Let's follow the same pattern we use in our other splitters in Haystack which is using a code block like this. So raise an error if content is `None` and a new init param called `skip_empty_documents` to control the behavior of whether empty documents should be skipped or not (check [here](https://github.com/deepset-ai/haystack/blob/602c4976ba6f494d2fd664582a37a770716429b8/haystack/components/preprocessors/document_splitter.py#L93-L95) to see this param in our DocumentSplitter) 
```python
        for doc in documents:
            if doc.content is None:
                raise ValueError(
                    f"DocumentSplitter only works with text documents but content for document ID {doc.id} is None."
                )
            if doc.content == "" and self.skip_empty_documents:
                logger.warning("Document ID {doc_id} has an empty content. Skipping this document.", doc_id=doc.id)
                continue
```

integrations/chonkie/src/haystack_integrations/components/preprocessors/chonkie/recursive_chunker.py
```diff
@@ -0,0 +1,113 @@
...
+
+            chunks = self._chunker.chunk(doc.content)
+            for chunk in chunks:
+                meta = doc.meta.copy() if doc.meta else {}
+                meta["source_id"] = doc.id
+                meta["start_index"] = chunk.start_index
+                meta["end_index"] = chunk.end_index
+                meta["token_count"] = chunk.token_count
```
sjrl: We have a standard-ish set of meta data we like to add to our split documents. Here is an example
```python
        "source_id": ...,
        "split_id": 0,
        "page_number": 1,
        "split_idx_start": 0,
```
so can we rename `start_index` to `split_idx_start`, `end_index` to `split_idx_end` and make sure to add `split_id` and `page_number`.

`split_id` should simply start at 0 for the first produced chunk and then increment by 1 for each produced chunk after. 

For `page_number` this could require more work depending on whether Chonkie returns this info or not. If it doesn't I'd recommend checking out `_split_documents_by_markdown_headers` [here](https://github.com/deepset-ai/haystack/blob/602c4976ba6f494d2fd664582a37a770716429b8/haystack/components/preprocessors/markdown_header_splitter.py#L291) to see how we handle page numbers by counting and keeping track of the page break character `"\f"`

integrations/chonkie/tests/test_recursive_splitter.py
```diff
@@ -0,0 +1,68 @@
...
+            },
+        }
+        chunker = ChonkieRecursiveChunker.from_dict(data)
+        assert chunker.chunk_size == 1024
+        assert chunker.tokenizer == "word"
+        assert chunker.min_characters_per_chunk == 10
+
+    def test_run(self):
+        chunker = ChonkieRecursiveChunker(chunk_size=10, min_characters_per_chunk=2)
+        doc = Document(content="Hello world! This is a test string for chunking.")
```
sjrl, line 104: Could we add a bit more of a complicated example? Especially when we add `page_number` support as mentioned in this [comment](https://github.com/deepset-ai/haystack-core-integrations/pull/3223/changes#r3135908340) it would be good to see that `page_numbers` are correctly calculated when passing a string with page break characters.

sjrl:
Hey @yugborana thanks for opening the contribution! I left an in depth review for one of the chunkers if you could also apply the same comments to other chunkers where appropriate that would be very appreciated!

integrations/chonkie/chonkie.md
sjrl, line 1: I think this file should be dropped. It looks like a API docs file which will be auto generated and adding to our docs website once this PR is merged.

yugborana:
@sjrl thanks for the review. It was pretty fast. I will take forward from here...

integrations/chonkie/src/haystack_integrations/components/preprocessors/chonkie/recursive_splitter.py
yugborana: No explicit docs on this. They have just names "character"
"gpt2"
"cl100k_base"

sjrl:
Thanks @yugborana! 

Also I discussed with the team and we think to be more consistent with our existing component name we should rename the Chonkie components to be like `ChonkieRecursiveChunker` --> `ChonkieRecursiveDocumentSplitter`. We tend to use Splitter in our library rather than Chunker (fine to call it chunking in the docs) and we like to be explicit if the i/o type is a Document by putting it in the name of the component. If you could update the names of all components that would be great.

integrations/chonkie/src/haystack_integrations/components/preprocessors/chonkie/recursive_splitter.py
sjrl: Okay if those are the three only options lets much them in our docs here as well

.github/workflows/chonkie.yml
yugborana: it does not yet support python 3.14

yugborana:
@sjrl resolved all the issues.

integrations/chonkie/src/haystack_integrations/components/preprocessors/chonkie/recursive_splitter.py
sjrl, line 1: In addition the the component renames could we also rename the files to reflect the change as well? E.g. `recursive_splitter.py`

integrations/chonkie/src/haystack_integrations/components/preprocessors/chonkie/recursive_chunker.py
```diff
@@ -0,0 +1,148 @@
+# SPDX-FileCopyrightText: 2024-present deepset GmbH <info@deepset.ai>
+#
+# SPDX-License-Identifier: Apache-2.0
+
+import logging
+from dataclasses import asdict
+from typing import Any
+
+from haystack import Document, component, default_from_dict, default_to_dict
```
sjrl: We should use the logging import from Haystack so 
```suggestion
from dataclasses import asdict
from typing import Any

from haystack import Document, component, default_from_dict, default_to_dict, logging
```

Our logger formatting is slightly different. E.g. it should look like this 
```python
logger.warning(
    "Agent reached maximum agent steps of {max_agent_steps}, stopping.",
    max_agent_steps=self.max_agent_steps,
)
```
if you want to format the string. Note that we do not use f-strings.

integrations/chonkie/src/haystack_integrations/components/preprocessors/chonkie/recursive_chunker.py
```diff
@@ -0,0 +1,148 @@
...
+
+
+@component
+class ChonkieRecursiveDocumentSplitter:
+    """
+    A Document Splitter that uses Chonkie's RecursiveChunker to split documents.
+
+    Usage::
+
+        from haystack import Document
```
sjrl: We usually format our docstrings like 
```
    ### Usage example

    ```python
    ...
```

integrations/chonkie/src/haystack_integrations/components/preprocessors/chonkie/recursive_splitter.py
```diff
@@ -0,0 +1,148 @@
...
+        """
+        self.tokenizer = tokenizer
+        self.chunk_size = chunk_size
+        self.min_characters_per_chunk = min_characters_per_chunk
+        self.skip_empty_documents = skip_empty_documents
+        self.page_break_character = page_break_character
+
+        if isinstance(rules, dict):
+            levels = [RecursiveLevel(**level) for level in rules.get("levels", [])]
+            self.rules: RecursiveRules | None = RecursiveRules(levels=levels)
```
sjrl, line 64: I believe you added this to handle deserialization of rules when using `from_dict`. Normally we'd just put this into `from_dict` itself. E.g. this is how the from_dict method looks like for our `DocumentSplitter`

```python
    @classmethod
    def from_dict(cls, data: dict[str, Any]) -> "DocumentSplitter":
        """
        Deserializes the component from a dictionary.
        """
        init_params = data.get("init_parameters", {})

        splitting_function = init_params.get("splitting_function", None)
        if splitting_function:
            init_params["splitting_function"] = deserialize_callable(splitting_function)

```

but this should be fine to leave as-is, just to let you know.

integrations/chonkie/src/haystack_integrations/components/preprocessors/chonkie/recursive_chunker.py
```diff
@@ -0,0 +1,148 @@
...
+            raise TypeError(msg)
+
+        chunked_documents = []
+        for doc in documents:
+            if doc.content is None:
+                msg = f"ChonkieRecursiveDocumentSplitter works only with text documents but doc ID {doc.id} is None."
+                raise ValueError(msg)
+
+            if doc.content == "" and self.skip_empty_documents:
+                logger.warning("Document ID %s has an empty content. Skipping this document.", doc.id)
```
sjrl: Related to this [comment](https://github.com/deepset-ai/haystack-core-integrations/pull/3223/changes#r3137308621) this should be updated to 
```suggestion
                logger.warning("Document ID {doc_id} has an empty content. Skipping this document.", doc_id=doc.id)
```

integrations/chonkie/src/haystack_integrations/components/preprocessors/chonkie/recursive_chunker.py
```diff
@@ -0,0 +1,148 @@
...
+                    {
+                        "source_id": doc.id,
+                        "page_number": current_page,
+                        "split_id": split_id,
+                        "split_idx_start": chunk.start_index,
+                        "split_idx_end": chunk.end_index,
+                        "token_count": chunk.token_count,
+                    }
+                )
+                current_page += chunk.text.count(self.page_break_character)
```
sjrl: Just to double check does Chonkie ever support having overlapping chunks of text shared between chunks? E.g.
```python
text = "This is a sentence"
chunk1 = "This is a"
chunk2 = "a sentence"
```
where `"a"` is shared as an overlap?

If not then all good, but if so technically this page counting could be incorrect in the case where the page break character is shared between two chunks. 

I couldn't immediately see from Chonkie's docs if they do support overlap.

integrations/chonkie/tests/test_recursive_chunker.py
```diff
@@ -0,0 +1,213 @@
...
+        pages_seen = sorted({chunk.meta["page_number"] for chunk in chunks})
+
+        assert 10 in pages_seen
+        assert pages_seen[-1] >= 12
+        last_chunk = chunks[-1]
+        text_before_last_chunk = text[: last_chunk.meta["split_idx_start"]]
+        expected_last_page = 10 + text_before_last_chunk.count("\f")
+        assert last_chunk.meta["page_number"] == expected_last_page
+
+    def test_custom_rules_serialization(self):
```
sjrl: I don't think we need this test right? It looks to be covered already by `test_from_dict_with_rules` and `test_to_dict_with_rules`

integrations/chonkie/tests/test_recursive_chunker.py
```diff
@@ -0,0 +1,213 @@
...
+        result = chunker.run(documents=docs)
+        chunks = result["documents"]
+        assert len(chunks) == 5
+        for split_id, chunk in enumerate(chunks):
+            assert chunk.meta["source_id"] == docs[0].id
+            assert chunk.meta["split_id"] == split_id
+            assert "page_number" in chunk.meta
+            assert "split_idx_start" in chunk.meta
+            assert "split_idx_end" in chunk.meta
+            assert "token_count" in chunk.meta
```
sjrl: Let's assert their actual values not just that they appear

integrations/chonkie/src/haystack_integrations/components/preprocessors/chonkie/semantic_splitter.py
```diff
@@ -0,0 +1,179 @@
...
+            "include_delim": include_delim,
+            "skip_window": skip_window,
+            "filter_window": filter_window,
+            "filter_polyorder": filter_polyorder,
+            "filter_tolerance": filter_tolerance,
+        }
+        if delim is not None:
+            kwargs["delim"] = delim
+
+        self._chunker = chonkie.SemanticChunker(**kwargs)
```
sjrl, line 108: This comment is only relevant for chunkers that require loading ML models. 

We typically put all model loading into our `warm_up` functions. Check out the example [here](https://github.com/deepset-ai/haystack/blob/08a029af98dc098ba5855c8ab456b8a044c752e9/haystack/components/generators/chat/hugging_face_local.py#L282-L297) for our `HuggingFaceLocalChatGenerator`

We do this so initializing a component always remains fast so we can easily do things like pipeline validation and then we can easily control when we want to do the heavy stuff

sjrl: Then make sure in `run` to add automatic warm_up in the beginning. e.g.
```python
if not self._is_warmed_up:
    self.warm_up()
```

integrations/chonkie/src/haystack_integrations/components/preprocessors/chonkie/recursive_chunker.py
yugborana: there is chunk overlap parameter in chonkie's docs so it does support.

yugborana:
@sjrl now I think it looks fine

yugborana:
@sjrl

integrations/chonkie/src/haystack_integrations/components/preprocessors/chonkie/sentence_splitter.py
```diff
@@ -0,0 +1,163 @@
...
+            "chunk_size": chunk_size,
+            "chunk_overlap": chunk_overlap,
+            "min_sentences_per_chunk": min_sentences_per_chunk,
+            "min_characters_per_sentence": min_characters_per_sentence,
+            "approximate": approximate,
+        }
+        if delim is not None:
+            kwargs["delim"] = delim
+        if include_delim is not None:
+            kwargs["include_delim"] = include_delim
```
sjrl: `include_delim` is typed as a string in the init so I'd either update it to allow it to be optional in the init type or drop the if check.

sjrl:
@yugborana one more minor comment then ready to go!

yugborana:
@sjrl done

sjrl (approved):
Thanks for the contribution!

PR #3238, opened 2026-04-27
feat: Add anthropic foundry support

sjrl:
### Related Issues

- fixes https://github.com/deepset-ai/haystack-private/issues/254

### Proposed Changes:

 
 

Add `AnthropicFoundryChatGenerator` to support anthropic models hosted on Azure Foundry. Similar in spirit to the `AnthropicVertexChatGenerator`. Basically we need to support a different client initialization to all clients to connect to anthropic models hosted through Azure Foundry. 

### How did you test it?

- Added new unit tests
- I'm not sure if we should keep the integration tests b/c I don't think this is easy for us to test. @bglearning would it be possible if you can test if this component works as expected?

### Checklist

### DIFF
--- a/integrations/anthropic/src/haystack_integrations/components/generators/anthropic/chat/foundry_chat_generator.py
+++ b/integrations/anthropic/src/haystack_integrations/components/generators/anthropic/chat/foundry_chat_generator.py
@@ -229,9 +229,8 @@ class AnthropicFoundryChatGenerator(AnthropicChatGenerator):
         """
         if not self._is_warmed_up:
             self.warm_up()
-        return super(AnthropicFoundryChatGenerator, self).run(
-            messages=messages, streaming_callback=streaming_callback,
-            generation_kwargs=generation_kwargs, tools=tools
+        return super().run(
+            messages=messages, streaming_callback=streaming_callback, generation_kwargs=generation_kwargs, tools=tools
         )
 
     @component.output_types(replies=list[ChatMessage])
@@ -256,9 +255,8 @@ class AnthropicFoundryChatGenerator(AnthropicChatGenerator):
         """
         if not self._is_warmed_up:
             self.warm_up()
-        return await super(AnthropicFoundryChatGenerator, self).run_async(
-            messages=messages, streaming_callback=streaming_callback,
-            generation_kwargs=generation_kwargs, tools=tools
+        return await super().run_async(
+            messages=messages, streaming_callback=streaming_callback, generation_kwargs=generation_kwargs, tools=tools
         )
 
     def to_dict(self) -> dict[str, Any]:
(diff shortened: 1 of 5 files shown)


integrations/anthropic/src/haystack_integrations/components/generators/anthropic/chat/foundry_chat_generator.py
```diff
@@ -0,0 +1,306 @@
...
+            Timeout for Anthropic client calls. If not set, it defaults to the default set by the Anthropic client.
+        :param max_retries:
+            Maximum number of retries to attempt for failed requests. If not set, it defaults to the default set by
+            the Anthropic client.
+        :param azure_ad_token_provider: A function that returns an Azure AD token for authentication.
+            Can be used instead of `api_key` for enhanced security.
+            See [Azure Identity documentation](https://learn.microsoft.com/en-us/azure/developer/python/sdk/authentication/overview)
+            for more details.
+        """
+        if api_key is None and azure_ad_token_provider is None:
```
bogdankostic, line 149: `api_key` is never `None`, except if explicitly set to `None` by the user. I'm wondering if we should change the default value of `api_key` to `None` or set `strict=True`.

bogdankostic (commented):
Looking good in general from my POV, just one minor comment regarding the default value for `api_key`.
Also, it would be great if we could run the integration tests at least once - not sure how feasible this is though.

integrations/anthropic/src/haystack_integrations/components/generators/anthropic/chat/foundry_chat_generator.py
sjrl: I don't think we should set it to `None` since it's nice to have the component work right away if the appropriate env var is set. 

I'll see if `strict=True` makes sense though since it's used when calling `api_key.resolve_value()` which happens at warm_up time and throwing an error then if the param is missing could make sense.

bglearning:
Ran through fine with this example:

```python
# To run this example, you will need to set a `ANTHROPIC_FOUNDRY_API_KEY` environment variable.

from haystack import Pipeline
from haystack.components.generators.utils import print_streaming_chunk

from haystack_integrations.components.generators.anthropic import AnthropicFoundryChatGenerator

messages = [
    ChatMessage.from_system("You are an expert on Football"),
    ChatMessage.from_user(
        "Who won the FIFA World Cup 2014?"
    ),
]
pipeline = Pipeline()
pipeline.add_component(
    "llm",
    AnthropicFoundryChatGenerator(
        model="...",
        endpoint="...",
        streaming_callback=print_streaming_chunk,
    )
)

print(pipeline.run(
    data={
        'llm': {'messages': messages}
    }
))
```

integrations/anthropic/src/haystack_integrations/components/generators/anthropic/chat/foundry_chat_generator.py
sjrl: Set strict to True in https://github.com/deepset-ai/haystack-core-integrations/pull/3238/commits/5cd844ff215e795e98191acaf111d99861c7bdc0

bogdankostic (approved):

bglearning:
Nice to have it in, thank you! :D

PR #3258, opened 2026-04-30
build: add supply chain hardening via uv exclude-newer and pip uploaded-prior-to

julian-risch:
### Related Issues

- Similar to Haystack PR https://github.com/deepset-ai/haystack/pull/11170

As recent motivation for this PR, there has been a compromised PyPI release of the mistralai package. Only version 2.4.6 was affected. It was uploaded on May 12, 2026 at 00:05 UTC and has been removed on May 12, 2026 at 03:05 UTC [according to Mistral AI](https://docs.mistral.ai/resources/security-advisories). Details [here](https://github.com/mistralai/client-python/security/advisories/GHSA-wx9m-wx4f-4cmg).

Given the short time window, uv exclude-newer and pip uploaded-prior-to seem to be a reasonable supply-chain security improvement.

### Proposed Changes:

- Add uv.toml with exclude-newer = "24 hours" so all uv pip installs skip packages published within the last day
- Add Dependabot cooldown (default-days: 1) for github-actions and pip
- Upgrade pip before each pip install step and add --uploaded-prior-to=P1D (pip 26.1 relative duration) to all direct pip install commands in CI
- Add UV_NO_CONFIG=1 to the llama_stack "Test Llama Stack Server" step to exempt it from exclude-newer: the PyTorch wheel index omits upload dates for sympy, causing uv to use the cutoff itself as the fallback timestamp — placing those packages exactly at the exclusive boundary and excluding them regardless of the date set

### How did you test it?
Ran `uv pip install mistralai` locally, which seems to also pick up the root uv.toml. 

### Notes for the reviewer
I am unsure if the changes to root uv.toml are enough or if the setting should be in pyproject.toml of all individual integrations. My understanding is that the uv.toml is enough because this will protect our CI.

I would merge despite failing tests once approved.

### Checklist


.github/workflows/CI_coverage_comment.yml
```diff
@@ -16,6 +16,7 @@ on:
       - "Test / chroma"
       - "Test / cohere"
       - "Test / cometapi"
+      - "Test / db2"
```
anakin87: I don't think we have such an integration

anakin87 (changes requested):
Small request change.

I'll also experiment quickly with uv.toml

anakin87:
- uv.toml works correctly: I tested locally with `weave` and different time settings
- Llama Stack failure (https://github.com/deepset-ai/haystack-core-integrations/actions/runs/25907956855/job/76145648661) seems related to the new uv config. Can you take a look?

.github/workflows/CI_coverage_comment.yml
julian-risch: True, thanks for spotting. This came from local experiments with our scaffolding script that got added to this PR by mistake. Removed now.

julian-risch:
> * uv.toml works correctly: I tested locally with `weave` and different time settings
> * Llama Stack failure (https://github.com/deepset-ai/haystack-core-integrations/actions/runs/25907956855/job/76145648661) seems related to the new uv config. Can you take a look?

Thanks. I found the issue with Llama Stack and it should be fixed now. I'll re-request a review once these tests pass.

anakin87 (approved):
👍

PR #3301, opened 2026-05-13
Fix: S3Downloader concurrency races in cache writing and eviction

### LINKED ISSUE #3294: `S3Downloader` cache races cause corrupt downloads under concurrent pipeline runs
**Describe the bug**

When multiple pipeline runs execute in parallel, `S3Downloader` corrupts files in its on-disk cache, causing downstream components like `DocumentToImageContent` to fail with errors such as `PDFium: Data format error`, `Failed to load page`, or `KeyError` on a missing page.

The component shares one cache directory (`file_root_path`) across all concurrent `run()` calls, and two races compound - could be more:

1. `_download_file` writes straight to the final cache path, so a concurrent reader can observe a partial PDF mid-write
2. `_cleanup_cache` only protects files referenced by the current run, so it can delete files another in-flight run is about to read

**To Reproduce**

1. Wire `S3Downloader → DocumentToImageContent` in a pipeline.
2. Call pipeline.run() from multiple threads simultaneously, requesting documents whose total file count exceeds `max_cache_size` default 100. This forces cache eviction while runs are in flight.
3. Observe intermittent pdfium failures on `DocumentToImageContent`. Failure rate scales with parallelism.

ritikraj2425:
- fixes #3294 

### Proposed Changes:

This PR fixes two critical race conditions in the `S3Downloader` component that occur when multiple pipeline runs execute in parallel with a shared cache directory:

1.  **Partial Write Corruption**: `_download_file` previously wrote directly to the final cache path. Concurrent readers could observe a partially written file, leading to parsing errors (e.g., `PDFium: Data format error`). 
    *   **Solution**: Implemented atomic writes by downloading to a unique temporary file (`.tmp.<uuid>`) and using `pathlib.Path.replace()` only once the download is complete.
2.  **Premature Cache Eviction**: `_cleanup_cache` would identify files from other active runs as "misses" and delete them if the cache was full, leading to `KeyError` or missing files for downstream components.
    *   **Solution**: Updated the eviction logic to calculate overflow correctly but strictly protect any file accessed within the last **60 seconds**. It also now safely ignores and cleans up stale `.tmp.` files older than 5 minutes.

### How did you test it?

- **Unit Tests**: Ran the full suite using `hatch run test:unit`.
- **Regression Testing**: Updated `test_cleanup_cache_evicts_old_files` in `tests/test_s3_downloader.py` to verify that stale files are still evicted while recently accessed files (the new "fresh" file) are correctly protected.
- **Quality Checks**: Verified all checks pass with `hatch run fmt` and `hatch run test:types`.

### Notes for the reviewer

- The 60-second time-based heuristic for cache protection was chosen as a lightweight alternative to complex file locking, effectively preventing races in concurrent pipeline executions without adding new dependencies. This approach was refined using AI assistance to balance safety and performance.

### Checklist

### DIFF
--- a/integrations/amazon_bedrock/src/haystack_integrations/components/downloaders/s3/s3_downloader.py
+++ b/integrations/amazon_bedrock/src/haystack_integrations/components/downloaders/s3/s3_downloader.py
@@ -3,6 +3,8 @@
 # SPDX-License-Identifier: Apache-2.0
 
 import os
+import time
+import uuid
 from collections.abc import Callable
 from concurrent.futures import ThreadPoolExecutor
 from pathlib import Path
@@ -207,7 +209,14 @@ class S3Downloader:
         else:
             s3_key = self.s3_key_generation_function(document) if self.s3_key_generation_function else file_name
             # we know that _storage is not None after warm_up() is called, but mypy does not know that
-            self._storage.download(key=s3_key, local_file_path=file_path)  # type: ignore[union-attr]
+            temp_file_path = file_path.with_name(f"{file_path.name}.tmp.{uuid.uuid4().hex}")
+            try:
+                self._storage.download(key=s3_key, local_file_path=temp_file_path)  # type: ignore[union-attr]
+                temp_file_path.replace(file_path)
+            except Exception:
+                if temp_file_path.exists():
+                    temp_file_path.unlink()
+                raise
 
         document.meta["file_path"] = str(file_path)
         return document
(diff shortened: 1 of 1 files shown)


sjrl:
Hey @ritikraj2425 this is a sensitive fix that we would like to tackle internally. So thanks for the contribution but I'll close this PR

ritikraj2425:
No worries! Thanks for letting me know, @sjrl. Happy to help where I can!

PR #3321, opened 2026-05-15
fix(ollama): preserve tool call IDs to fix repeated same-tool calls

### LINKED ISSUE #3303: OllamaChatGenerator drops Ollama tool call IDs, breaking repeated same-tool calls
## Describe the bug
  `OllamaChatGenerator` does not preserve tool call IDs from Ollama responses. This makes repeated
  calls to the same tool in a single assistant message unreliable, and can also break agent/HITL
  disambiguation because `tool_call_id` is missing.

### Ollama POV
  Upstream Ollama issues about this are still open:
  - `#6708` https://github.com/ollama/ollama/issues/6708
  - `#7510` https://github.com/ollama/ollama/issues/7510

  However, current Ollama source appears to already assign `call_...` IDs in native `/api/chat`:
  - ID helper: https://github.com/ollama/ollama/blob/main/server/routes.go#L2169-L2175
  - non-streaming assignment: https://github.com/ollama/ollama/blob/main/server/routes.go#L2547
  - streaming assignment: https://github.com/ollama/ollama/blob/main/server/routes.go#L2594
  - tests expecting IDs:
    - https://github.com/ollama/ollama/blob/main/server/routes_generate_test.go#L581-L585
    - https://github.com/ollama/ollama/blob/main/server/routes_generate_test.go#L703-L704

### Haystack POV
  Haystack’s Ollama integration still drops those IDs:
  - non-streaming conversion:
  https://github.com/deepset-ai/haystack-core-integrations/blob/main/integrations/ollama/src/haystack_integrations/components/generators/ollama/chat/chat_generator.py#L170-L173
  - chunk conversion does not propagate ID:
  https://github.com/deepset-ai/haystack-core-integrations/blob/main/integrations/ollama/src/haystack_integrations/components/generators/ollama/chat/chat_generator.py#L195-L201
  - streaming reconstruction drops IDs:
    - https://github.com/deepset-ai/haystack-core-integrations/blob/main/integrations/ollama/src/haystack_integrations/components/generators/ollama/chat/chat_generator.py#L440
    - https://github.com/deepset-ai/haystack-core-integrations/blob/main/integrations/ollama/src/haystack_integrations/components/generators/ollama/chat/chat_generator.py#L522

  Relevant downstream impact in Haystack:
  - known HITL limitation when IDs are missing:
  https://github.com/deepset-ai/haystack/blob/main/haystack/human_in_the_loop/strategies.py#L513-L521
  - existing duplicate-name/no-id failure test:
  https://github.com/deepset-ai/haystack/blob/main/test/human_in_the_loop/test_strategies.py#L319-L330

  ## To Reproduce
  1. Use `OllamaChatGenerator` in an agent with a tool that can be called more than once in one
  assistant turn.
  2. Prompt the model so it emits two calls to the same tool with different arguments.
  3. Inspect the resulting Haystack `ChatMessage.tool_calls`.
  4. Observe that `ToolCall.id` is not preserved.
  5. If HITL is enabled, duplicate same-name tool calls cannot be reliably disambiguated.

## Describe your environment (please complete the following information):
   - OS: Linux (GitHub Codespaces)
   - Haystack version: `v2.30.0-rc0-16-g34b221b19`
   - Integration version: `integrations/ollama-v6.3.0-170-gd83a037a`

SyedShahmeerAli12:
### Related Issues

- fixes #3303

### Proposed Changes:

`OllamaChatGenerator` was silently dropping tool call IDs, which broke any scenario where the same tool is called more than once in a single assistant turn. In the streaming accumulator, both calls mapped to the same key (the tool name since ID was `None`), so the second call overwrote the first and it was lost.

Fix: generate a synthetic ID (`call_N`) for each tool call in both the streaming and non-streaming paths, and propagate it to the final `ToolCall` objects. `ollama_tc.get("id")` is checked first so the code will use server-assigned IDs automatically if a future Ollama SDK version exposes them.

Affected locations:
- `_convert_ollama_response_to_chatmessage`  non-streaming path
- `_build_chunk` streaming chunk builder
- `_handle_streaming_response` + `_handle_streaming_response_async`  streaming reconstruction

### How did you test it?

- Reproduced the bug: two streaming chunks calling `weather` with different cities  only the second call survived before the fix
- After the fix both calls are returned with distinct IDs
- Added `test_convert_ollama_response_to_chatmessage_with_repeated_tool`
- Added `test_handle_streaming_response_repeated_tool_calls`
- Updated existing tests that previously asserted `id=None` on `ToolCallDelta` and `ToolCall` objects
- All 41 unit tests pass, ruff clean

### Notes for the reviewer

The Ollama Python SDK (0.6.1) drops the `id` field from the server JSON response during Pydantic parsing `Message.ToolCall` has no `id` field defined. Synthetic IDs (`call_0`, `call_1`, …) are the only option today. The `get("id")` fallback is a forward-compatible hook for when the SDK adds the field.

### Checklist

### DIFF
--- a/integrations/ollama/src/haystack_integrations/components/generators/ollama/chat/chat_generator.py
+++ b/integrations/ollama/src/haystack_integrations/components/generators/ollama/chat/chat_generator.py
@@ -165,10 +165,10 @@ def _convert_ollama_response_to_chatmessage(ollama_response: ChatResponse) -> Ch
     tool_calls: list[ToolCall] = []
 
     if ollama_tool_calls := ollama_message.get("tool_calls"):
-        for idx, ollama_tc in enumerate(ollama_tool_calls):
+        for ollama_tc in ollama_tool_calls:
             tool_calls.append(
                 ToolCall(
-                    id=ollama_tc.get("id") or f"call_{idx}",
+                    id=ollama_tc.get("id"),
                     tool_name=ollama_tc["function"]["name"],
                     arguments=ollama_tc["function"]["arguments"],
                 )
@@ -209,7 +209,7 @@ def _build_chunk(
             tool_calls_list.append(
                 ToolCallDelta(
                     index=tool_call_index,
-                    id=tool_call.get("id") or f"call_{tool_call_index}",
+                    id=tool_call.get("id"),
                     tool_name=tool_call["function"]["name"],
                     arguments=json.dumps(tool_call["function"]["arguments"])
                     if tool_call["function"]["arguments"]
@@ -372,10 +372,11 @@ class OllamaChatGenerator:
         component_info = ComponentInfo.from_component(self)
         chunks: list[StreamingChunk] = []
 
-        # Accumulators
-        arg_by_id: dict[str, str] = {}
-        name_by_id: dict[str, str] = {}
-        id_order: list[str] = []
+        # Accumulators keyed by tool_call.index (always unique per call, even for repeated tool names)
+        arg_by_index: dict[str, str] = {}
+        name_by_index: dict[str, str] = {}
+        id_by_index: dict[str, str | None] = {}
+        index_order: list[str] = []
         tool_call_index: int = 0
 
         # track reasoning and content blocks to correctly set start=True on the first chunk of each block
@@ -402,15 +403,14 @@ class OllamaChatGenerator:
 
             if chunk.tool_calls:
                 for tool_call in chunk.tool_calls:
-                    # id is always set by _build_chunk (either from the server or synthetic "call_N").
-                    # Fall back to tool_name only as a last resort for callers that bypass _build_chunk.
-                    tool_call_id = tool_call.id or tool_call.tool_name or ""
+                    key = str(tool_call.index)
                     args = tool_call.arguments or ""
 
-                    if tool_call_id not in id_order:
-                        id_order.append(tool_call_id)
-                        name_by_id[tool_call_id] = tool_call.tool_name or ""
-                    arg_by_id[tool_call_id] = args
+                    if key not in index_order:
+                        index_order.append(key)
+                        name_by_index[key] = tool_call.tool_name or ""
+                        id_by_index[key] = tool_call.id
+                    arg_by_index[key] = args
 
             if callback:
                 callback(chunk)
@@ -423,10 +423,10 @@ class OllamaChatGenerator:
             reasoning += c.reasoning.reasoning_text if c.reasoning else ""
 
         tool_calls = []
-        for tool_call_id in id_order:
-            arguments: str = arg_by_id.get(tool_call_id, "")
+        for key in index_order:
+            arguments: str = arg_by_index.get(key, "")
             tool_calls.append(
-                ToolCall(id=tool_call_id, tool_name=name_by_id[tool_call_id], arguments=json.loads(arguments))
+                ToolCall(id=id_by_index[key], tool_name=name_by_index[key], arguments=json.loads(arguments))
             )
 
         # We can't use _convert_streaming_chunks_to_chat_message because
@@ -453,10 +453,11 @@ class OllamaChatGenerator:
         component_info = ComponentInfo.from_component(self)
         chunks: list[StreamingChunk] = []
 
-        # Accumulators
-        arg_by_id: dict[str, str] = {}
-        name_by_id: dict[str, str] = {}
-        id_order: list[str] = []
+        # Accumulators keyed by tool_call.index (always unique per call, even for repeated tool names)
+        arg_by_index: dict[str, str] = {}
+        name_by_index: dict[str, str] = {}
+        id_by_index: dict[str, str | None] = {}
+        index_order: list[str] = []
         tool_call_index: int = 0
 
         # track reasoning and content blocks to correctly set start=True on the first chunk of each block
@@ -484,13 +485,14 @@ class OllamaChatGenerator:
 
             if chunk.tool_calls:
                 for tool_call in chunk.tool_calls:
-                    tool_call_id = tool_call.id or tool_call.tool_name or ""
+                    key = str(tool_call.index)
                     args = tool_call.arguments or ""
 
-                    if tool_call_id not in id_order:
-                        id_order.append(tool_call_id)
-                        name_by_id[tool_call_id] = tool_call.tool_name or ""
-                    arg_by_id[tool_call_id] = args
+                    if key not in index_order:
+                        index_order.append(key)
+                        name_by_index[key] = tool_call.tool_name or ""
+                        id_by_index[key] = tool_call.id
+                    arg_by_index[key] = args
 
             if callback is not None:
                 await callback(chunk)
@@ -505,10 +507,10 @@ class OllamaChatGenerator:
             reasoning += c.reasoning.reasoning_text if c.reasoning else ""
 
         tool_calls = []
-        for tool_call_id in id_order:
-            arguments: str = arg_by_id.get(tool_call_id, "")
+        for key in index_order:
+            arguments: str = arg_by_index.get(key, "")
             tool_calls.append(
-                ToolCall(id=tool_call_id, tool_name=name_by_id[tool_call_id], arguments=json.loads(arguments))
+                ToolCall(id=id_by_index[key], tool_name=name_by_index[key], arguments=json.loads(arguments))
             )
 
         # We can't use _convert_streaming_chunks_to_chat_message because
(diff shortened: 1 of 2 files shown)


SyedShahmeerAli12:
Quick reproduction showing the bug and the fix

**Before**  two streaming calls to the same tool, one silently lost:
```
tool_calls: 1
  tool='weather'  args={'city': 'London'}   ← Paris overwritten
```

**After**  both calls preserved with distinct IDs:
```
tool_calls: 2
  id='call_1'  tool='weather'  args={'city': 'Paris'}
  id='call_2'  tool='weather'  args={'city': 'London'}
```

The root cause was that `_build_chunk` created `ToolCallDelta` with no `id`, so the streaming accumulator fell back to `tool_name` as the deduplication key  collapsing two calls to the same tool into one.

sjrl:
Please remove all changes unrelated to the PR. So all changes in falkdordb should be removed.

integrations/ollama/src/haystack_integrations/components/generators/ollama/chat/chat_generator.py
```diff
@@ -165,9 +165,10 @@ def _convert_ollama_response_to_chatmessage(ollama_response: ChatResponse) -> Ch
     tool_calls: list[ToolCall] = []
 
     if ollama_tool_calls := ollama_message.get("tool_calls"):
-        for ollama_tc in ollama_tool_calls:
+        for idx, ollama_tc in enumerate(ollama_tool_calls):
             tool_calls.append(
                 ToolCall(
+                    id=ollama_tc.get("id") or f"call_{idx}",
```
sjrl: We should only use the ones from ollama, don't generate synthetic IDs.

integrations/ollama/src/haystack_integrations/components/generators/ollama/chat/chat_generator.py
```diff
@@ -208,6 +209,7 @@ def _build_chunk(
             tool_calls_list.append(
                 ToolCallDelta(
                     index=tool_call_index,
+                    id=tool_call.get("id") or f"call_{tool_call_index}",
```
sjrl: Same here just use the ones from ollama

sjrl:
@SyedShahmeerAli12 could you add information about when Ollama started returning tool call IDs (e.g. which version of ollama is this supported?). Also were you able to reproduce the bug with a live ollama server and not just through the unit tests?

integrations/ollama/src/haystack_integrations/components/generators/ollama/chat/chat_generator.py
SyedShahmeerAli12: ok let check this in a while

integrations/ollama/tests/test_chat_generator.py
sjrl, line 1: Let's update the integration tests that use tools like `test_live_run_with_tools`, `test_live_run_with_thinking_and_tools` to check that tool call IDs are now present.

SyedShahmeerAli12:
**Ollama version:** Tool call IDs were added in **v0.12.10** (released Nov 5, 2025)  the release notes explicitly say _"Ollama will now return tool call IDs from the `/api/chat` API"_.

**Live reproduction:** Yes, tested with Ollama v0.24.0. The raw HTTP response confirms IDs are present:
```json
[
  {"id": "call_lyywui55", "function": {"name": "get_weather", "arguments": {"city": "Paris"}}},
  {"id": "call_0scw2dos", "function": {"name": "get_weather", "arguments": {"city": "London"}}}
]
```

However, the **Ollama Python SDK (0.6.1) silently drops the `id` field**  `Message.ToolCall` does not define it, so Pydantic strips it during parsing. `model_dump()` never includes it. This means even against a live server, `tool_call.id` is always `None` through the current SDK.

I'll address the inline comments about synthetic IDs and update the PR accordingly.

sjrl:
@SyedShahmeerAli12 do you know if there is an open issue for the SDK to add the tool call ID support?

SyedShahmeerAli12:
No open issue found in ollama/ollama-python for this. The SDK's `Message.ToolCall` Pydantic model only defines `function: Function` with no `id` field, so server-returned IDs are silently dropped during parsing.

Also addressed the inline comments  removed synthetic ID generation. The streaming accumulator now keys by `tool_call.index` (the positional counter assigned per chunk, always unique even for repeated tool names), so the duplicate-tool-call bug is fixed without fabricating IDs. Server-provided IDs are passed through as-is; when Ollama doesn't return one, `id` is `None`.

sjrl:
> No open issue found in ollama/ollama-python for this. The SDK's `Message.ToolCall` Pydantic model only defines `function: Function` with no `id` field, so server-returned IDs are silently dropped during parsing.
> 
> Also addressed the inline comments removed synthetic ID generation. The streaming accumulator now keys by `tool_call.index` (the positional counter assigned per chunk, always unique even for repeated tool names), so the duplicate-tool-call bug is fixed without fabricating IDs. Server-provided IDs are passed through as-is; when Ollama doesn't return one, `id` is `None`.

Hey @SyedShahmeerAli12 if this is the case that the ollama python SDK ignores the tool call ID then your changes here won't have any effect right? The tool call ID will always be `None` or am I missing something?

integrations/ollama/src/haystack_integrations/components/generators/ollama/chat/chat_generator.py
```diff
@@ -370,10 +372,11 @@ def _handle_streaming_response(
         component_info = ComponentInfo.from_component(self)
         chunks: list[StreamingChunk] = []
 
-        # Accumulators
-        arg_by_id: dict[str, str] = {}
-        name_by_id: dict[str, str] = {}
-        id_order: list[str] = []
+        # Accumulators keyed by tool_call.index (always unique per call, even for repeated tool names)
```
sjrl, line 375: I see so instead accumulate by tool call index. Could you update the integration tests to check that the tool call indices are unique by checking their values?

SyedShahmeerAli12:
The PR fixes two things:

1. **The actual bug**  the streaming accumulator was keyed by `tool_name`, so two calls to the same tool (e.g. two `get_weather`) collapsed into one. Now it keys by `tool_call.index` which is always unique. That fix has nothing to do with IDs.

2. **The ID cleanup**  the old code invented fake IDs (`call_0`, `call_1`). Since the SDK drops real ones anyway, those were just noise. Now `id=None` honestly when the server doesn't provide one  and when the SDK eventually adds ID support, the pass-through will just work.

So yes, IDs are always `None` right now  but the bug this PR fixes was never about IDs.

integrations/ollama/src/haystack_integrations/components/generators/ollama/chat/chat_generator.py
sjrl: Actually even better please add an integration test that causes two tool calls to occur in one go so we can make sure that the disambiguation is working. Add a new test under `TestOllamaChatGeneratorLiveInference` that triggers the LLM to call the same tool twice in one turn and then send it back to the LLM to see if it responds correctly. E.g look at this test https://github.com/deepset-ai/haystack/blob/ec1777a49c05ed241abb4936c7f33d31920c2c32/test/components/generators/chat/test_openai.py#L1047

integrations/ollama/tests/test_chat_generator.py
SyedShahmeerAli12: Since the SDK always returns `id=None`, we can't assert IDs are present — that would just be asserting `None`. Added `test_live_run_with_repeated_tool_calls` instead, which directly tests the fix by triggering the same tool twice in one turn and asserting both calls are preserved.

integrations/ollama/src/haystack_integrations/components/generators/ollama/chat/chat_generator.py
SyedShahmeerAli12: Added `test_live_run_with_repeated_tool_calls`  it asks for weather in Paris and London in one turn, asserts exactly 2 `ToolCall` objects come back with distinct cities, then sends the tool results back to the LLM. That proves the index-based disambiguation is working.

SyedShahmeerAli12: Done  added `test_live_run_with_repeated_tool_calls` under `TestOllamaChatGeneratorLiveInference`. It triggers the same `weather` tool twice in one turn (Paris and London), asserts both calls are returned, invokes them via `ToolInvoker`, and sends results back to check the LLM responds correctly. Runs in both streaming and non-streaming via parametrize.

sjrl:
@SyedShahmeerAli12 looking good and thanks for the additional tests!

Last comment would be to add async versions of the unit test `test_handle_streaming_response_repeated_tool_calls` and then it should be good to go.

SyedShahmeerAli12:
Added `test_handle_streaming_response_async_repeated_tool_calls`  mirrors the sync test, feeds the same two-city chunks into `_handle_streaming_response_async` via an async generator, and asserts both tool calls are preserved with correct names and arguments.

sjrl (approved):
Thanks!

PR #3324, opened 2026-05-18
fix: Fix langfuse tracer when input messages is `None`

sjrl:
### Related Issues

- fixes https://github.com/deepset-ai/haystack-private/issues/358

### Proposed Changes:

 
 

The LLM component accepts `messages=None` as valid input which can break the Langfuse Tracer since it always assumes that if `messages` is present that it's always a `list[ChatMessage]`. This PR updates to handle the `None` case. 

We also add the same guard for `replies` returning as `None`. This is not a known case but done for safety. 

### How did you test it?

Added new unit tests.

### Checklist

### DIFF
--- a/integrations/langfuse/src/haystack_integrations/tracing/langfuse/tracer.py
+++ b/integrations/langfuse/src/haystack_integrations/tracing/langfuse/tracer.py
@@ -90,7 +90,7 @@ class LangfuseSpan(Span):
             return
         if key.endswith(".input"):
             if "messages" in value:
-                messages = [m.to_openai_dict_format(require_tool_call_ids=False) for m in value["messages"]]
+                messages = [m.to_openai_dict_format(require_tool_call_ids=False) for m in (value.get("messages") or [])]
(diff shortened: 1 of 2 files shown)


integrations/langfuse/src/haystack_integrations/tracing/langfuse/tracer.py
```diff
@@ -90,7 +90,7 @@ def set_content_tag(self, key: str, value: Any) -> None:
             return
         if key.endswith(".input"):
             if "messages" in value:
-                messages = [m.to_openai_dict_format(require_tool_call_ids=False) for m in value["messages"]]
+                messages = [m.to_openai_dict_format(require_tool_call_ids=False) for m in (value.get("messages") or [])]
```
sjrl, line 93: We do `value.get("messages") or []` to catch the case that `messages` key is present in `value` but it returns `None`. Doing `value.get("messages", [])` would not catch it since the back up `[]` only occurs if `messages` is missing.

anakin87 (approved):
👍 

(I'm not a big fan of the `LLM` component 😃)

PR #3325, opened 2026-05-18
feat(llama_hub): refactor connector to support LlamaIndex v0.14+ dire…

### LINKED ISSUE #575: Build LlamaHub Integration
LlamaHub has added a function to convert outputs to haystack format, which luckily, seems compatible with 2.0 although it was built for 1.x.

The 'verified' data loaders have a standardized way to be loaded and used. Let's create an integration that consists of a custom component where the run function takes only a few things, such as: the name of the loader for `download_loader('name of loader')`, and the input expected for `load_data()`

image

anshuS1310:
---

---

## 🛠️ Key Changes

---


anakin87:
Based on https://github.com/run-llama/llama_index/issues/21703#issuecomment-4477048537, it seems that a unified LlamaHub interface no longer exists. Also, the project seems quite unmaintained. At this point, and with better alternatives, having an integration does not seem to provide significant value to Haystack users.

For this reason, I'll close this PR and the original issue.

Thank you!

anshuS1310:
😭😭😭😭😭

PR #3336, opened 2026-05-20
fix: support aliases in all OpenSearchDocumentStore methods

tstadel:
### Related Issues

OpenSearch's indices.get_mapping() and indices.get() APIs always key their response by the real (concrete) index name, not by the alias name used in the request. Every call site that did mapping[self._index] would raise KeyError when self._index was an alias.

### Proposed Changes:
document_store.py — 6 call sites fixed using actual_index = next(iter(response)):

- _ensure_index_exists / _ensure_index_exists_async — reading mappings on startup
- delete_all_documents / delete_all_documents_async — reading index settings before recreation
- count_unique_metadata_by_filter / count_unique_metadata_by_filter_async — reading field mappings
- get_metadata_fields_info / get_metadata_fields_info_async — reading field mappings

Alias guard for destructive recreation: delete_all_documents(recreate_index=True) now raises DocumentStoreError early when self._index is an alias. Deleting and recreating through an alias would drop the backing index and create a new concrete index named after the alias, permanently breaking the alias relationship. The recreate_index=False (delete-by-query) path is unaffected and works normally with aliases.

### How did you test it?

Unit tests (mocked):

- test_delete_all_documents_recreate_raises_for_alias / _works_for_concrete_index (sync + async variants)
- test_ensure_index_exists_{direct_index,with_alias,creates_index,no_create} (sync + async variants, added in prior commits)

Integration tests:

- test_document_store_with_alias / test_document_store_async_with_alias — basic read/write via alias
- test_delete_all_documents_with_alias / _async_ — asserts DocumentStoreError with recreate_index=True
- test_delete_all_documents_no_recreate_with_alias / _async_ — verifies delete-by-query clears all docs and index remains writable
- test_count_unique_metadata_by_filter_with_alias / _async_ — mapping lookup works through alias
- test_get_metadata_fields_info_with_alias / _async_ — mapping lookup works through alias

### Checklist

### DIFF
--- a/integrations/opensearch/src/haystack_integrations/document_stores/opensearch/document_store.py
+++ b/integrations/opensearch/src/haystack_integrations/document_stores/opensearch/document_store.py
@@ -774,6 +774,8 @@ class OpenSearchDocumentStore:
 
         :param recreate_index: If True, the index will be deleted and recreated with the original mappings and
             settings. If False, all documents will be deleted using the `delete_by_query` API.
+            ``recreate_index=True`` is not supported when the configured index name is an alias; a
+            :class:`haystack.document_stores.errors.DocumentStoreError` is raised in that case.
         :param refresh: If True, OpenSearch refreshes all shards involved in the delete by query after the request
             completes. If False, no refresh is performed. For more details, see the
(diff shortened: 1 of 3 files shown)


davidsbatista (approved):
Looks good!

PR #3369, opened 2026-05-31
Fix: use AsyncPipeline with OpensearchHybridRetriever super_component

Nitro963:
### Proposed Changes:

Use AsyncPipeline so that OpensearchHybridRetriever component work within sync and async pipelines

### How did you test it?
Created Unit tests for usage within async pipelines

### DIFF
--- a/integrations/elasticsearch/src/haystack_integrations/components/retrievers/elasticsearch/elasticsearch_hybrid_retriever.py
+++ b/integrations/elasticsearch/src/haystack_integrations/components/retrievers/elasticsearch/elasticsearch_hybrid_retriever.py
@@ -4,7 +4,7 @@
 
 from typing import TYPE_CHECKING, Any
 
-from haystack import Document, Pipeline, default_from_dict, default_to_dict, logging, super_component
+from haystack import Document, AsyncPipeline, default_from_dict, default_to_dict, logging, super_component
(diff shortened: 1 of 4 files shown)


sjrl:
Hey @Nitro963 thanks for the contribution! Could you separate these changes into two separate PRs, so one for OpenSearch and one for ElasticSearch? It helps keep our release notes for the different packages well separated.

Nitro963:
@sjrl DONE, kindly check #3374 pull request

sjrl:
@Nitro963 thanks! Please also drop the changes to the elastic search integration in this PR

sjrl:
Hey @Nitro963 I just wanted to give you an update that we need to resolve an issue I describe in this PR https://github.com/deepset-ai/haystack/pull/11457 before we can merge your changes. We are still arriving at the best solution so this might take a little while.

sjrl:
Hey @Nitro963 thanks again for opening a PR! We've decided that we don't want make this change for now. However, in Haystack v3 we have merged `Pipeline` and `AsyncPipeline` into one Pipeline object so the async capability of this component should automatically take effect when Haystack v3 is released. The PR where we merged the two can be found here https://github.com/deepset-ai/haystack/pull/11587

Haystack v3 isn't out yet but will be in the near future.

PR #3371, opened 2026-06-01
feat: add async job mode to Docling Serve converter

### LINKED ISSUE #3345: docling_serve: support the async-job endpoint for long-running conversions

`DoclingServeConverter.run()` and `run_async()` both POST to docling-serve's synchronous endpoints (`/v1/convert/file`, `/v1/convert/source`) and block until the server returns the conversion result. This pattern hits two real limits for non-trivial documents:

1. **Server-side cap.** docling-serve enforces a 120s ceiling on synchronous requests via the `max_sync_wait` setting ([`docling-serve/docling_serve/settings.py#L146`](https://github.com/docling-project/docling-serve/blob/main/docling_serve/settings.py#L146)). Can be raised via `DOCLING_SERVE_MAX_SYNC_WAIT`, however its still sync.

2. **Network-path timeouts.** A long-held HTTP request must survive every hop (reverse proxies, load balancers, NAT keepalive windows, corporate firewalls) where defaults vary.

For PDFs with VLM picture description, OCR-heavy scans, or large multi-page documents, both limits get hit routinely in production.

Add opt-in support for docling-serve's async-job endpoint triplet:

- `POST /v1/convert/file/async` → returns `task_id` immediately
- `GET /v1/status/poll/{task_id}?wait=N` → server-side long-poll, returns on status transition
- `GET /v1/result/{task_id}` → fetch result after terminal status

Concretely, add a constructor parameter:

```python
DoclingServeConverter(
    base_url="...",
    mode: Literal["sync", "async"] = "sync",  # opt-in, default unchanged
    poll_interval: float = 2.0,
    job_timeout: float = 600.0,
)
```

When `mode="async"`, `run()`/`run_async()` submit, long-poll `/v1/status/poll/{task_id}?wait=...` until `task_status` is `success` or `failure`, then GET `/v1/result/{task_id}`. Errors are surfaced explicitly: HTTP errors, task failures (`task_status == "failure"` with `error_message`), per-document failures (`status in {"failure", "skipped"}`), and job timeout.

The async-job pattern is what docling's own reference Python client uses ([`docling/service_client/client.py`](https://github.com/docling-project/docling/blob/main/docling/service_client/client.py)) for exactly this case. So the proposal mirrors upstream convention.

- **Bump `DOCLING_SERVE_MAX_SYNC_WAIT` server-side.** Solves the docling-serve cap but leaves every intermediate proxy timeout in place. Brittle and ops-config-dependent; doesn't scale beyond a single known network path.
- **Subclass `DoclingServeConverter` downstream.** Currently doable (the `_post_file*` helpers are subclass-friendly), but every consumer reinvents the same ~100 lines of polling logic. Better to land it once upstream.

**Additional context**

- Backwards compatible: existing users see no change with default `mode="sync"`.

2830500285:
## Summary
- Add an async job mode for DoclingServeConverter that submits conversions to Docling Serve's async endpoints and polls until completion.
- Support async jobs for both URL sources and uploaded file/ByteStream sources.
- Preserve the existing synchronous mode as the default and cover serialization, failure handling, and sync/async execution paths with tests.

Closes #3345

## Validation
- SETUPTOOLS_SCM_PRETEND_VERSION=0.0.0 uvx hatch run fmt-check src/haystack_integrations/components/converters/docling_serve/converter.py src/haystack_integrations/components/converters/docling_serve/__init__.py tests/test_converter.py
- SETUPTOOLS_SCM_PRETEND_VERSION=0.0.0 uvx hatch run test:unit tests/test_converter.py
- SETUPTOOLS_SCM_PRETEND_VERSION=0.0.0 uvx hatch run test:types


integrations/docling_serve/src/haystack_integrations/components/converters/docling_serve/converter.py
```diff
@@ -277,6 +517,12 @@ def run(
                         source=source,
                         error=e,
                     )
+                except DoclingServeConversionError as e:
```
davidsbatista, line 540: This execption `DoclingServeConversionError` raised on timeout ends up being caught by the same except block in run() / run_async() that also handles ordinary conversion failures:

If a job exceeds job_timeout (default 600s), the caller silently gets an empty documents list after waiting 10 minutes — indistinguishable from "file had no content". A timeout feels like a different signal from a conversion failure.

One option would be a dedicated subclass:

```python
class DoclingServeTimeoutError(DoclingServeConversionError):
    """Raised when a job exceeds job_timeout."""
```

That way callers can catch it separately, and you could re-raise it in run() or at least use a more specific log message.

Same for the run_async()

integrations/docling_serve/src/haystack_integrations/components/converters/docling_serve/converter.py
```diff
@@ -108,12 +133,29 @@ def __init__(
             API key for authenticating with a secured DoclingServe instance. Reads from the
             `DOCLING_SERVE_API_KEY` environment variable by default. Set to `None` to disable
             authentication.
+        :param mode:
+            Conversion mode. `sync` uses DoclingServe's synchronous endpoints. `async` submits
+            conversion jobs to DoclingServe's async endpoints and polls until completion.
+        :param poll_interval:
+            Maximum server-side long-poll wait in seconds when `mode="async"`.
```
davidsbatista:
```suggestion
            Controls both the server-side long-poll wait (?wait= parameter) and the maximum local sleep between polls. A higher value reduces round-trips; a lower value increases polling frequency.
```

integrations/docling_serve/src/haystack_integrations/components/converters/docling_serve/converter.py
```diff
@@ -108,12 +133,29 @@ def __init__(
...
+        :param job_timeout:
+            Maximum time in seconds to wait for each async conversion job.
         """
+        if poll_interval <= 0:
```
davidsbatista: those are only used in async, so we can do:

```python
if self.mode == ConversionMode.ASYNC:
      if poll_interval <= 0:
          msg = "poll_interval must be greater than 0."
          raise ValueError(msg)
      if job_timeout <= 0:
          msg = "job_timeout must be greater than 0."
          raise ValueError(msg)
```

integrations/docling_serve/src/haystack_integrations/components/converters/docling_serve/converter.py
```diff
@@ -184,6 +264,24 @@ def _post_file(self, client: httpx.Client, source: str | Path | ByteStream) -> d
         response.raise_for_status()
         return response.json()
```
davidsbatista: This is a pre-existing issue, not introduced by this PR, but worth fixing.

The sync endpoints (_post_file, _post_url) return the raw response JSON without checking the status field in the body. This means if DoclingServe returns a 200 OK with {"status": "failure", "errors": [...]}, the error details are silently discarded — the code falls through to _extract_content, gets None, and logs "No content returned for source" with no indication of why.

The async job path handles this correctly via _raise_for_failed_conversion in _fetch_job_result. 

Let's apply the same check to _post_file and _post_url to make the behaviour consistent.

integrations/docling_serve/tests/test_converter.py
```diff
@@ -404,6 +508,32 @@ async def test_run_async_multiple_sources(self):
 
         assert len(result["documents"]) == 2
 
+    @pytest.mark.asyncio
+    async def test_run_async_async_mode_uses_job_endpoints(self):
```
davidsbatista, line 622: There's a test for `run_async()` with ASYNC mode and a URL source, but no equivalent for a file source. 

Could you add a test that verifies run_async() with mode=ASYNC and a file path hits /v1/convert/file/async and returns the converted document?

integrations/docling_serve/tests/test_converter.py
```diff
@@ -296,6 +341,65 @@ def test_run_convert_options_merged_into_payload(self):
...
+
+        assert result["documents"][0].content == "async content"
+
+        post_url = mock_post.call_args[0][0]
+        data = mock_post.call_args[1]["data"]
+        assert post_url.endswith("/v1/convert/file/async")
+        assert data["to_formats"] == "md"
+        assert data["target_type"] == "inbody"
+
+    def test_run_async_mode_skips_failed_job(self, caplog):
```
davidsbatista, line 422: Could you add a test for the timeout path? 

When `job_timeout` is exceeded, `DoclingServeConversionError` is raised and caught — the source should be skipped with a warning containing "Timed out".

davidsbatista: Also a test for the "skipped" status? 

A job can complete with task_status: "success" but the conversion result can still have status: "skipped"

davidsbatista:
Hi @2830500285! 

Is the PR already ready for a second review? What were the changes did you introduced you want me to look into?

2830500285:
Hi @davidsbatista, yes, this is ready for a second review.

I addressed the review feedback in 601b473:
- added DoclingServeTimeoutError and a specific timeout warning path in run() / run_async()
- clarified poll_interval docs
- validate poll_interval and job_timeout only when mode="async"
- check DoclingServe response body status for the sync/direct endpoints as well
- added tests for async file job conversions, timeout handling, skipped conversion results, and direct endpoint failure details

CI is green and the coverage report improved to 96.7% overall / 98.59% PR coverage.

davidsbatista (approved):
Looks good!

PR #3374, opened 2026-06-01
Fix: Use Async pipeline with ElasticsearchHybridRetriever

Nitro963:
### Proposed Changes:

 Use AsyncPipeline so that ElasticsearchHybridRetriever component work within sync and async pipelines

### How did you test it?

Created Unit tests for usage within async pipelines


sjrl:
Hey @Nitro963 thanks again for opening a PR! We've decided that we don't want make this change for now. However, in Haystack v3 we have merged `Pipeline` and `AsyncPipeline` into one Pipeline object so the async capability of this component should automatically take effect when Haystack v3 is released. The PR where we merged the two can be found here https://github.com/deepset-ai/haystack/pull/11587

Haystack v3 isn't out yet but will be in the near future.

PR #3380, opened 2026-06-01
feat(supabase): add async support to SupabaseGroongaDocumentStore and SupabaseGroongaBM25Retriever

### LINKED ISSUE #3379: Add async support to `SupabaseGroongaDocumentStore` and `SupabaseGroongaBM25Retriever`
`SupabaseGroongaDocumentStore` and SupabaseGroongaBM25Retriever currently only expose synchronous methods. 

`supabase-py` ships a full async client (AsyncClient / acreate_client) with an async def execute() query builder, so adding async counterparts is straightforward

`supabase-py` exposes two parallel clients:

```python
from supabase import create_client 
from supabase import acreate_client
from supabase import AsyncClient
```

The AsyncClient.table() returns an AsyncRequestBuilder where all filter methods (.eq(), .neq(), .in_(), .or_(), etc.) are identical to the sync client, but .execute() is an async def and must be
awaited.

---
Work required

### 1. SupabaseGroongaDocumentStore

- Add _async_client: AsyncClient | None = None to __init__
- Add async def warm_up_async() — calls acreate_client(url, key) and await self._setup_table_async()
- Add async counterparts for all public methods:
  - async def count_documents_async() -> int
  - async def filter_documents_async(filters) -> list[Document]
  - async def write_documents_async(documents, policy) -> int
  - async def delete_documents_async(document_ids) -> None
  - async def delete_all_documents_async() -> None
  - async def delete_by_filter_async(filters) -> int
  - async def update_by_filter_async(filters, meta) -> int
- Add async def _groonga_retrieval_async(query, top_k, filters) used by the retriever

The internal _apply_filters / _apply_condition logic is reusable as-is since it builds the query object without executing it — only the final .execute() call changes.

### 2. SupabaseGroongaBM25Retriever

The run_async() method currently just delegates to run():

async def run_async(self, query, filters=None, top_k=None):
    return self.run(query=query, filters=filters, top_k=top_k)

This should be updated to call _groonga_retrieval_async() instead, and the class docstring note saying async is not supported should be removed.

### 3. Tests

- Add async unit tests to test_groonga_document_store.py and test_groonga_retriever.py
- Add async integration tests to test_groonga_integration.py using the mixins from haystack.testing.document_store_async if available

## 4. Documentation

- Remove the note from the SupabaseGroongaBM25Retriever class docstring that says async is not supported
- Update the SupabaseGroongaDocumentStore class docstring to document the async workflow

---

Aftabbs:
## Summary

- Adds `warm_up_async()` and a full set of `*_async()` methods to `SupabaseGroongaDocumentStore` using `supabase-py`'s `AsyncClient` and `acreate_client`
- Fixes `SupabaseGroongaBM25Retriever.run_async()`, which was a stub that delegated to the synchronous `run()` instead of using real async I/O
- Adds 14 new unit tests for the async path across both classes

Closes #3379

## Root Cause

`supabase-py` ships two parallel clients: `create_client` (sync) and `acreate_client` (async). The original implementation only wired up the sync client. `run_async()` on the retriever simply called `self.run()`, blocking the event loop on every invocation — making `AsyncPipeline` usage effectively synchronous.

## Changes

| File | What changed |
|------|-------------|
| `groonga_document_store.py` | Added `_async_client` field; `warm_up_async()`, `_setup_table_async()`, and async counterparts for all public methods (`count_documents_async`, `filter_documents_async`, `write_documents_async`, `delete_documents_async`, `delete_all_documents_async`, `delete_by_filter_async`, `update_by_filter_async`, `_groonga_retrieval_async`) |
| `groonga_bm25_retriever.py` | Fixed `run_async()` to call `_groonga_retrieval_async()`; removed stale docstring note saying async is unsupported |
| `test_groonga_document_store.py` | Added `TestDocumentStoreAsync` class with 12 async unit tests using `AsyncMock` for the async client |
| `test_groonga_retriever.py` | Updated `test_run_async` to verify `_groonga_retrieval_async` is actually awaited (not the sync path) |

## Testing

- [x] Existing sync tests pass (no sync code was changed)
- [x] New async tests: `TestDocumentStoreAsync` — 12 tests covering warm_up, count, filter, write (OVERWRITE/SKIP/FAIL policies), delete, delete_all, retrieval, and uninitialized-client guard
- [x] Retriever test: `test_run_async_uses_async_retrieval` — asserts `_groonga_retrieval_async` is awaited
- [x] `ruff check` passes with no errors

## Impact

Users running `SupabaseGroongaBM25Retriever` inside an `AsyncPipeline` now get true non-blocking I/O instead of a blocking call that defeats the purpose of async execution.


davidsbatista:
@Aftabbs thanks for your contribution - can you please make the CI green so that I can start the review

davidsbatista:
Regarding the tests, since we are testing against a local Docker instance no need to mock up; please convert them to integration tests

integrations/supabase/src/haystack_integrations/components/retrievers/supabase/groonga_bm25_retriever.py
```diff
@@ -21,8 +21,8 @@ class SupabaseGroongaBM25Retriever:
     This retriever works without embeddings — it searches documents using plain text queries.
     It can be used alongside SupabasePgvectorEmbeddingRetriever in hybrid search pipelines.
 
-    Note: async operations are not supported as the supabase-py sync client does not expose
-    awaitable query methods. Use the sync run() method instead.
+    Supports both synchronous and asynchronous retrieval. For async usage, initialize the
```
davidsbatista: Let's remove all this text, including the existing one.

davidsbatista:
Hey @Aftabbs, thanks for the fix to `run_async()`

The new `warm_up_async()` method is inconsistent with how every other integration in this repo handles async initialization.

Looking at others, none of them expose a `warm_up_async()`. Instead, they use lazy initialization: each async method calls a private _initialize_async_client() at the top before doing any work, creating the client on first use. For example, from the Qdrant store:

```python
  async def _initialize_async_client(self) -> None:
      if self._async_client is None:
          self._async_client = qdrant_client.AsyncQdrantClient(...)
          await self._set_up_collection_async()

  async def count_documents_async(self) -> int:
      await self._initialize_async_client()
```

davidsbatista (approved):
I have handled the remaining issues. Thanks for your contribution

PR #3407, opened 2026-06-05
feat: move HuggingFaceTEIRanker from Haystack

anakin87:
### Related Issues

- part of https://github.com/deepset-ai/haystack-private/issues/365 (initially forgotten)

### Proposed Changes:
- move `HuggingFaceTEIRanker` into the HF API integration: it is similar to components in this integration (they call an external service) but I kept the TEI name since it require Text Embeddings Inference to work (locally or on HF Inference Endpoints)

### How did you test it?
CI
In the future, we might want to spin up a TEI container and add an integration test

### Checklist


integrations/huggingface_api/src/haystack_integrations/components/rankers/huggingface_api/ranker.py
```diff
@@ -0,0 +1,297 @@
...
+    def from_dict(cls, data: dict[str, Any]) -> "HuggingFaceTEIRanker":
+        """
+        Deserializes the component from a dictionary.
+
+        :param data:
+            Dictionary to deserialize from.
+        :returns:
+            Deserialized component.
+        """
+        return default_from_dict(cls, data)
```
julian-risch, line 124: We could probably leave out this method entirely as it's only doing the default. Let's focus in this PR only on the migration though. No change request for this PR.

julian-risch (approved):

PR #3410, opened 2026-06-06
fix: async streaming chunk indices in GoogleGenAIChatGenerator start at 0, not 1

abhay23-AI:
### Related Issues

No existing issue; I came across this while reading through the GoogleGenAIChatGenerator streaming code.

### Proposed Changes:

`_handle_streaming_response_async` increments its chunk counter before passing it to `_convert_google_chunk_to_streaming_chunk`, so the first async chunk is handed `index=1`. The sync path uses `enumerate(...)` and starts at `0`, so the two diverge.

That index isn't cosmetic. In `chat/utils.py` the chunk's `start` flag is computed as `start = index == 0 or len(tool_calls) > 0`, and `start` is what Haystack's default `print_streaming_chunk` uses to decide whether to emit the `[ASSISTANT]` header and the leading newline. With the off-by-one, a text-only first chunk coming from `run_async` gets `index=1`, so `start` is `False`, and the assistant header is dropped and newline handling shifts compared to the identical sync stream.

The fix moves the increment to the end of the loop body so async chunk indices line up with the sync path (`0, 1, 2, ...`). Aggregation is unaffected since it joins chunks in list order, so the change is scoped to the index/`start` parity.

### How did you test it?

Added a unit test in `TestGoogleGenAIChatGeneratorRun` that streams two chunks through `run_async` and asserts the callback receives indices `[0, 1]` and that the first chunk has `start=True`. It fails before the change (indices come through as `[1, 2]`, first chunk `start=False`) and passes after.

Ran the full `google_genai` unit suite (115 passed), `hatch run fmt-check`, and `hatch run test:types` (no issues). The fix is pure internal control flow so a mocked async stream covers it; the existing `@pytest.mark.integration` live async-streaming test still works against `gemini-2.5-flash` for anyone who wants to confirm end to end.

### Notes for the reviewer

The one behavioral line is the `i += 1` move in `_handle_streaming_response_async`; everything else is the regression test. Worth a quick look at `chat/utils.py:669` (`start = index == 0 or ...`) to confirm the `start`-flag reasoning.

### Checklist

- [ ] I have updated the related issue with new insights and changes (no related issue)


abhay23-AI:
Confirmed end to end against the real API (`gemini-2.5-flash`), streaming through `run_async` with a callback that records each chunk's `index` and `start`:

Before the change:
```
chunks received: 2
  (1, False, 'The ocean is a vast ')
  (2, False, ' planet. It teems wi')
indices start at: 1
```

After the change:
```
chunks received: 2
  (0, True, 'The ocean is a vast ')
  (1, False, ' of life, from micro')
indices start at: 0
first chunk start flag: True
```

So the first async chunk now carries `index=0` / `start=True` and matches the sync path, which is what `print_streaming_chunk` relies on for the `[ASSISTANT]` header.

abhay23-AI:
Hi @anakin87, this one's ready when you get a chance. It's a small sync/async parity fix in the GoogleGenAIChatGenerator streaming path: the first async chunk was getting `index=1` while the sync path starts at `0`, which flips the `start` flag that `print_streaming_chunk` relies on for the `[ASSISTANT]` header. CI is green, and I added a live before/after against `gemini-2.5-flash` in the comment just above. Thanks for taking a look.

abhay23-AI:
Ran the google_genai integration suite locally against `gemini-2.5-flash` (`hatch run test:integration`), since the fork PR can't access secrets in CI.

26 of 27 passed, including every async streaming integration test, which are the ones this change actually exercises.

The single failure is `test_concurrent_async_calls`, and it's a free tier quota limit rather than a code issue: once the rest of the suite uses up my key's per minute free tier quota, the API returns `429 RESOURCE_EXHAUSTED` (`generate_content_free_tier_requests`). That test just fires three concurrent `run_async` calls and has nothing to do with chunk indexing, so it's unrelated to this PR and passes on an account with normal quota. Happy to re-run anything specific if useful.

anakin87 (approved):
Thank you!

PR #3419, opened 2026-06-08
feat(oauth): Add OAuth integration

bogdankostic:
### Related Issues

  - Part of https://github.com/deepset-ai/haystack-private/issues/346

  ### Proposed Changes:

  Adds a new **`oauth`** integration package (`oauth-haystack`) that resolves a user-scoped OAuth 2.0 access token at pipeline runtime, in a provider-neutral way.

  - **`OAuthResolver`** (`@component`): a thin graph node that emits `access_token` on an output socket (to be connected to any component that needs a bearer token, e.g. a SharePoint/GDrive retriever). It delegates to a pluggable **`TokenSource`** strategy and stays agnostic to the provider.
  - **Dynamic input sockets**: the resolver declares a **mandatory `subject_token` run input only when the configured source needs it** (sources expose `requires_subject_token`). Config-only sources declare no input and run as source nodes.
  - **Two protocols**: `TokenSource` (config-only, `resolve()`) and `SubjectTokenSource` (per-request, `resolve(subject_token)`).
  - **Three built-in sources** (All RFC-based. No provider-specific code, providers are pure configuration):
    - `RefreshTokenSource` — RFC 6749 refresh-token grant; single-identity; in-process cache with expiry buffer + `on_rotate` hook for providers that rotate refresh tokens.
    - `TokenExchangeSource` — RFC 8693 token exchange / Microsoft on-behalf-of; multi-user **without persistent storage** (the per-request `subject_token` is the identity). 
    - `StaticTokenSource` — long-lived tokens (e.g. PATs).

  ### How did you test it?

  - **Unit tests**: init/config validation; per-grant request shaping; caching, expiry-buffer refresh, refresh-token rotation + 
  `on_rotate`, per-subject LRU eviction; error mapping; sync **and** async paths; `to_dict`/`from_dict` round-trips; and  the dynamic mandatory-socket 
  - **Gated integration test** (`test_run_against_real_idp`, skipped unless `OAUTH_TOKEN_URL` / `OAUTH_CLIENT_ID` / `OAUTH_REFRESH_TOKEN` are set).
- **Manual live verification against two real providers**, exercising `OAuthResolver` end to end:
    - **Microsoft Entra ID** — `RefreshTokenSource` (refresh grant → Graph `/me/drive/root/children` `200`; also confirmed refresh-token rotation via `on_rotate` and the in-process cache) and `TokenExchangeSource` (on-behalf-of → Graph `/me` returns the signed-in user).
    - **Google** — `RefreshTokenSource` and `StaticTokenSource` (token resolved → Google Drive `GET /drive/v3/files` returns `200` with the user's files; second `run()` served from cache).

### Checklist


integrations/oauth/pydoc/config_docusaurus.yml
```diff
@@ -0,0 +1,14 @@
+loaders:
+  - modules:
+      - haystack_integrations.components.connectors.oauth.resolver
+      - haystack_integrations.components.connectors.oauth.sources
```
sjrl: I'm not sure what we normally do, but would it make sense to expose the errors classes as well?

integrations/oauth/src/haystack_integrations/components/connectors/oauth/resolver.py
```diff
@@ -0,0 +1,113 @@
...
+from .errors import OAuthConfigError
+from .sources import SubjectTokenSource, TokenSource, deserialize_token_source
+
+
+@component
+class OAuthResolver:
+    """
+    Resolves an OAuth access token at pipeline runtime and emits it on the `access_token` output socket.
+
+    The resolver is a thin graph node over a pluggable token source that decides *where* the token comes from:
```
sjrl: I think we want to say something like component here instead of node
```suggestion
    The resolver component is a thin wrapper over a pluggable token source that decides *where* the token comes from:
```

integrations/oauth/src/haystack_integrations/components/connectors/oauth/resolver.py
```diff
@@ -0,0 +1,113 @@
...
+            client_id="aaa-bbb-ccc",
+            refresh_token=Secret.from_env_var("MS_REFRESH_TOKEN"),
+            scopes=["https://graph.microsoft.com/Files.Read.All", "offline_access"],
+        ),
+    )
+    access_token = resolver.run()["access_token"]
+    ```
+    """
+
+    def __init__(self, token_source: "TokenSource | SubjectTokenSource") -> None:
```
sjrl: I don't think we need to defer the types here 
```suggestion
    def __init__(self, token_source: TokenSource | SubjectTokenSource) -> None:
```

integrations/oauth/src/haystack_integrations/components/connectors/oauth/sources.py
```diff
@@ -0,0 +1,524 @@
...
+    """Validate an OAuth token endpoint response and extract the access token, its TTL, and the raw payload."""
+    if response.is_error:
+        msg = f"Token request failed: {response.status_code} {response.text}"
+        raise TokenRefreshError(msg, status_code=response.status_code)
+    payload = response.json()
+    access_token = payload.get("access_token")
+    if not access_token:
+        msg = "Token response did not contain an access_token."
+        raise TokenRefreshError(msg)
+    expires_in = float(payload.get("expires_in", DEFAULT_ACCESS_TOKEN_TTL_SECONDS))
```
sjrl: minor but we could add a try: except block around `expires_in = float(payload.get("expires_in", DEFAULT_ACCESS_TOKEN_TTL_SECONDS))` which defaults to `DEFAULT_ACCESS_TOKEN_TTL_SECONDS` if the float conversion fails.

integrations/oauth/src/haystack_integrations/utils/oauth/sources.py
```diff
@@ -0,0 +1,524 @@
...
+    Provider differences are expressed as configuration: `grant_type`, `subject_token_param` (for example
+    `assertion` for Microsoft), `scopes`, and `extra_token_params` (for example
+    `{"requested_token_use": "on_behalf_of"}`).
+    """
+
+    requires_subject_token = True
+
+    def __init__(
+        self,
+        token_url: str,
```
sjrl, line 244: Let's enforce keyword-only for all or starting after `client_id`. This is what we've been adopting for component's in general and especially with long init signatures.

integrations/oauth/src/haystack_integrations/components/connectors/oauth/sources.py
```diff
@@ -0,0 +1,524 @@
...
+            to `subject_token` (RFC 8693). Some providers expect a different name, such as `assertion`.
+        :param subject_token_type: The RFC 8693 identifier for the type of the supplied subject token, sent as the
+            `subject_token_type` form parameter (omitted when not set). Required by RFC 8693 token exchange
+            (e.g. `urn:ietf:params:oauth:token-type:access_token`); not used by Microsoft's on-behalf-of flow.
+        :param requested_token_type: The RFC 8693 identifier for the token to return, sent as the
+            `requested_token_type` form parameter (omitted when not set). Optional.
+        :param scopes: The scopes to request, joined with `scope_delimiter`.
+        :param scope_delimiter: The delimiter used to join scopes. Defaults to a space.
+        :param extra_token_params: Additional form parameters included verbatim in every request (for example
+            `{"requested_token_use": "on_behalf_of"}`).
```
sjrl: Maybe worth mentioning anything in the docs that anything in `extra_token_params` will overwrite other config options like `requested_token_type`, etc. ?

integrations/oauth/src/haystack_integrations/utils/oauth/sources.py
```diff
@@ -0,0 +1,524 @@
...
+        return request_data
+
+    def resolve(self, subject_token: str) -> str:
+        """Exchange the per-request `subject_token` for an access token (cached per subject token)."""
+        if not subject_token:
+            msg = "TokenExchangeSource requires a non-empty subject_token."
+            raise OAuthConfigError(msg)
+        cache_key = self._cache_key(subject_token)
+        with self._lock:
+            cached = self._cached_if_valid(cache_key)
```
sjrl, line 376: Maybe we could add a dev comment explaining why we want a lock here.

integrations/oauth/src/haystack_integrations/components/connectors/oauth/sources.py
```diff
@@ -0,0 +1,524 @@
...
+            if client_secret:
+                request_data["client_secret"] = client_secret
+        if self.scopes:
+            request_data["scope"] = self.scope_delimiter.join(self.scopes)
+        if self.extra_token_params:
+            request_data.update(self.extra_token_params)
+        return request_data
+
+    def resolve(self, subject_token: str) -> str:
+        """Exchange the per-request `subject_token` for an access token (cached per subject token)."""
```
sjrl: Let's add `:param subject_token:` and `:returns :` docstrings

integrations/oauth/src/haystack_integrations/components/connectors/oauth/sources.py
```diff
@@ -0,0 +1,524 @@
...
+        """Return a valid access token."""
+        ...
+
+    async def resolve_async(self) -> str:
+        """Asynchronous counterpart of `resolve`."""
+        ...
+
+    def to_dict(self) -> dict[str, Any]:
+        """Serialize the source to a dictionary."""
+        ...
```
sjrl: Should we also enforce/include a `from_dict` method in the protocol?

integrations/oauth/src/haystack_integrations/components/connectors/oauth/sources.py
```diff
@@ -0,0 +1,524 @@
...
+class SubjectTokenSource(Protocol):
+    """
+    A token source that resolves an access token by exchanging a per-request subject token.
+
+    The `subject_token` is a controller-injected per-request credential (for example an incoming user assertion),
+    not chosen by an end user. Implemented by `TokenExchangeSource`. Such sources set the class attribute
+    `requires_subject_token = True`, which makes `OAuthResolver` declare a mandatory `subject_token` run input.
+    """
+
+    requires_subject_token: bool
```
sjrl: I might not know enough about protocols, but is it possible to set `requires_subject_token: bool = True` here since this must be true to be a valid `SubjectTokenSource` instance? At least going off of the docstrings.

integrations/oauth/src/haystack_integrations/components/connectors/oauth/sources.py
```diff
@@ -0,0 +1,524 @@
...
+    """
+
+    requires_subject_token = False
+
+    def __init__(
+        self,
+        token_url: str,
+        client_id: str,
+        refresh_token: Secret,
+        client_secret: Secret | None = None,
```
sjrl: Should we add default Secrets here to give users a known env var they could set like we do in other components?

integrations/oauth/src/haystack_integrations/components/connectors/oauth/sources.py
```diff
@@ -0,0 +1,524 @@
...
+        new_refresh_token = payload.get("refresh_token")
+        if new_refresh_token:
+            self._current_refresh_token = new_refresh_token
+            if self.on_rotate is not None:
+                self.on_rotate(new_refresh_token)
+        logger.debug("Refreshed access token via {token_url}.", token_url=self.token_url)
+        return access_token
+
+    def resolve(self) -> str:
+        """Return a cached access token, or run the refresh-token grant to obtain a fresh one."""
```
sjrl: Lets add a `:returns :` docstring here

integrations/oauth/src/haystack_integrations/utils/oauth/sources.py
```diff
@@ -0,0 +1,524 @@
...
+        if new_refresh_token:
+            self._current_refresh_token = new_refresh_token
+            if self.on_rotate is not None:
+                self.on_rotate(new_refresh_token)
+        logger.debug("Refreshed access token via {token_url}.", token_url=self.token_url)
+        return access_token
+
+    def resolve(self) -> str:
+        """Return a cached access token, or run the refresh-token grant to obtain a fresh one."""
+        with self._sync_lock:
```
sjrl, line 169: Let's also add a dev comment here about why the lock is needed

integrations/oauth/src/haystack_integrations/utils/oauth/sources.py
sjrl, line 1: Hmm this deviates a bit from our structure. These sources aren't components but are still user facing so my impression is that they shouldn't live within the `components` folder. 

One idea could be to do something put them into a new folder like `src/haystack_integrations/utils/oauth/`
for the files
```
__init__.py          # exports sources, protocols, errors, deserialize_token_source
sources.py
errors.py
py.typed
```
and perhaps we could move the protocols into their own file separate from the implementations which is what we normally do in haystack core. 

WDYT?

integrations/oauth/README.md
```diff
@@ -0,0 +1,6 @@
+# oauth-haystack
+
+[![PyPI - Version](https://img.shields.io/pypi/v/oauth-haystack.svg)](https://pypi.org/project/oauth-haystack)
+[![PyPI - Python Version](https://img.shields.io/pypi/pyversions/oauth-haystack.svg)](https://pypi.org/project/oauth-haystack)
+
```
sjrl, line 5: I believe we also typically mention some set up instructions on how to be able to run the integration tests. E.g. for https://github.com/deepset-ai/haystack-core-integrations/tree/main/integrations/langfuse

integrations/oauth/src/haystack_integrations/components/connectors/oauth/sources.py
```diff
@@ -0,0 +1,524 @@
...
+                "client_id": self.client_id,
+                "refresh_token": self.refresh_token.to_dict(),
+                "client_secret": self.client_secret.to_dict() if self.client_secret else None,
+                "scopes": self.scopes,
+                "scope_delimiter": self.scope_delimiter,
+                "expiry_buffer_seconds": self.expiry_buffer_seconds,
+                "timeout": self.timeout,
+                "on_rotate": serialize_callable(self.on_rotate) if self.on_rotate else None,
+            },
+        }
```
sjrl: It seems like this follows our component serialization pattern. Could we reuse `default_to_dict` here then we could also remove the `_type_name` function.

integrations/oauth/src/haystack_integrations/components/connectors/oauth/sources.py
```diff
@@ -0,0 +1,524 @@
...
+    def from_dict(cls, data: dict[str, Any]) -> "RefreshTokenSource":
+        """Deserialize the source from a dictionary."""
+        init_params = data["init_parameters"]
+        secret_keys = ["refresh_token"]
+        if init_params.get("client_secret") is not None:
+            secret_keys.append("client_secret")
+        deserialize_secrets_inplace(init_params, keys=secret_keys)
+        if init_params.get("on_rotate"):
+            init_params["on_rotate"] = deserialize_callable(init_params["on_rotate"])
+        return cls(**init_params)
```
sjrl: Also we should use the `default_from_dict` here right?

integrations/oauth/src/haystack_integrations/utils/oauth/sources.py
```diff
@@ -0,0 +1,524 @@
...
+        secret_keys = ["refresh_token"]
+        if init_params.get("client_secret") is not None:
+            secret_keys.append("client_secret")
+        deserialize_secrets_inplace(init_params, keys=secret_keys)
+        if init_params.get("on_rotate"):
+            init_params["on_rotate"] = deserialize_callable(init_params["on_rotate"])
+        return cls(**init_params)
+
+
+class TokenExchangeSource:
```
sjrl: I think I could go either way on this but we normally have one file per implementation. Should we make something like a `sources` folder that would have a separate file for the protocols and each of the implementations?

integrations/oauth/src/haystack_integrations/utils/oauth/sources.py
```diff
@@ -0,0 +1,524 @@
...
+                headers={"Accept": "application/json"},
+                timeout=self.timeout,
+            )
+            return self._handle_response(response)
+
+    async def resolve_async(self) -> str:
+        """Asynchronous counterpart of `resolve`. Use a single instance in either sync or async mode, not both."""
+        if self._async_lock is None:
+            self._async_lock = asyncio.Lock()
+        async with self._async_lock:
```
sjrl, line 187: Any reason to create the _async_lock once at run time instead of like the sync one which we do in the init method?

integrations/oauth/src/haystack_integrations/components/connectors/oauth/sources.py
```diff
@@ -0,0 +1,524 @@
...
+# RFC 8693 OAuth 2.0 Token Exchange grant
+DEFAULT_TOKEN_EXCHANGE_GRANT = "urn:ietf:params:oauth:grant-type:token-exchange"
+DEFAULT_EXPIRY_BUFFER_SECONDS = 300
+DEFAULT_ACCESS_TOKEN_TTL_SECONDS = 3600
+DEFAULT_CACHE_MAX_SIZE = 1000
+DEFAULT_TIMEOUT_SECONDS = 30.0
+
+
+@runtime_checkable
+class TokenSource(Protocol):
```
sjrl: Hey I noticed that `TokenSource` and `SubjectTokenSource` are currently identical to `isinstance` (`@runtime_checkable` only checks member presence, not signatures) so the two protocols are indistinguishable at runtime.

They only differ in docstrings so I don't think they really are serving the purpose of being protocols. So I have two potential suggestions:

**Idea 1:** Merge into one protocol.
  Single `TokenSource` with `requires_subject_token: bool` and a unified `resolve(self, subject_token: str | None = None)`. The `bool` drives whether the resolver declares the `subject_token` socket; the resolver drops both casts.

**Idea 2:** Make the split real via `Literal` discriminant. Keep two protocols, but type the flag as `requires_subject_token: Literal[False]` vs `Literal[True]`, turning them into a tagged union mypy can narrow, so `resolve()` vs `resolve(subject_token)` are statically distinct and no casts are needed.

I think I'm leaning to idea 1 but I'm open for discussion :)

integrations/oauth/src/haystack_integrations/utils/oauth/sources.py
sjrl: Same goes for the other user-facing classes in this package.

integrations/oauth/src/haystack_integrations/components/connectors/oauth/sources.py
sjrl: Sorry this should only apply to `refresh_token` since `client_secret` already has a default.

integrations/oauth/src/haystack_integrations/components/connectors/oauth/sources.py
```diff
@@ -0,0 +1,524 @@
...
+        """
+        self.token = token
+
+    def resolve(self) -> str:
+        """Return the configured token."""
+        value = self.token.resolve_value()
+        if not value:
+            msg = "StaticTokenSource has no token value."
+            raise TokenRefreshError(msg)
+        return value
```
sjrl: Minor: depending on the secret type (e.g. `EnvVarSecret` this line `value = self.token.resolve_value()` can raise a `ValueError` when the resolution results in `None`. So for a consistent error message it would be good to catch errors from `self.token.resolve_value()` and re-raise as a `TokenRefreshError`

integrations/oauth/src/haystack_integrations/components/connectors/oauth/resolver.py
```diff
@@ -0,0 +1,113 @@
...
+    @classmethod
+    def from_dict(cls, data: dict[str, Any]) -> "OAuthResolver":
+        """
+        Deserialize this component from a dictionary.
+
+        :param data: The dictionary representation of this component.
+        :returns: The deserialized component instance.
+        """
+        init_params = data["init_parameters"]
+        init_params["token_source"] = deserialize_token_source(init_params["token_source"])
```
sjrl: I think we could use the `component_from_dict` method here instead of a custom `deserialize_token_source`

integrations/oauth/src/haystack_integrations/components/connectors/oauth/sources.py
bogdankostic: Good point. I think I'd prefer Option 2. I don't like that merging into a single protocol forces a unified `resolve(self, subject_token: str | None = None) signature` resulting in the config-only sources `RefreshTokenSource` and  `StaticTokenSource` having to accept a subject_token they never use, just to satisfy the protocol.

integrations/oauth/pydoc/config_docusaurus.yml
bogdankostic: Done in eaae730e8

integrations/oauth/src/haystack_integrations/components/connectors/oauth/resolver.py
bogdankostic: Done in de8369735

integrations/oauth/src/haystack_integrations/components/connectors/oauth/resolver.py
bogdankostic: Done in 78dbd4494

integrations/oauth/src/haystack_integrations/components/connectors/oauth/sources.py
bogdankostic: Done in 7eeec0387

integrations/oauth/src/haystack_integrations/utils/oauth/sources.py
bogdankostic: Done in 78dbd4494

integrations/oauth/src/haystack_integrations/components/connectors/oauth/sources.py
bogdankostic: Done in de8369735

integrations/oauth/src/haystack_integrations/utils/oauth/sources.py
bogdankostic: Done in 0e66b5113

integrations/oauth/src/haystack_integrations/components/connectors/oauth/sources.py
bogdankostic: Done in de8369735

integrations/oauth/src/haystack_integrations/components/connectors/oauth/sources.py
bogdankostic: Done in 78dbd4494

integrations/oauth/src/haystack_integrations/components/connectors/oauth/sources.py
bogdankostic: Done in de8369735

integrations/oauth/src/haystack_integrations/utils/oauth/sources.py
bogdankostic: Done in 0e66b5113

integrations/oauth/src/haystack_integrations/utils/oauth/sources.py
bogdankostic: Done in 964c48b4a

integrations/oauth/README.md
bogdankostic: Done in eaae730e8

integrations/oauth/src/haystack_integrations/components/connectors/oauth/sources.py
bogdankostic: Done in 3346945e2

integrations/oauth/src/haystack_integrations/components/connectors/oauth/sources.py
bogdankostic: Done in 3346945e2

integrations/oauth/src/haystack_integrations/utils/oauth/sources.py
bogdankostic: I split the protocols into their own file but kept the three sources grouped in `sources.py` as they're rather small classes, but I'm not very opinionated on this so happy to split them into individual files if we prefer that.

integrations/oauth/src/haystack_integrations/utils/oauth/sources.py
bogdankostic: Added a comment in 0e66b5113. Basically, an `asyncio.Lock` must bind to the running event loop, but `__init__` is synchronous and may run without one, so it's created lazily on first use.

integrations/oauth/src/haystack_integrations/components/connectors/oauth/sources.py
bogdankostic: Done in 7eeec0387

integrations/oauth/src/haystack_integrations/utils/oauth/sources.py
sjrl: All good happy to keep them in the same file

integrations/oauth/src/haystack_integrations/utils/oauth/sources.py
sjrl: thanks for the explanation!

integrations/oauth/src/haystack_integrations/utils/oauth/sources.py
```diff
@@ -0,0 +1,476 @@
...
+    """
+    Resolves access tokens by running the RFC 6749 refresh-token grant against an OAuth token endpoint.
+
+    Given a stored refresh token plus client credentials, it exchanges them for an access token and caches it in
+    process until shortly before expiry. If the identity provider rotates the refresh token on exchange, the new value
+    is kept for the lifetime of the process and surfaced through the optional `on_rotate` callback so it can be
+    persisted.
+
+    This source is **single-identity**: one refresh token per instance. For per-user, multi-tenant resolution that
+    needs no persistent storage, use `TokenExchangeSource` instead. It takes no per-request input.
```
sjrl: I'm struggling to understand a bit if this token source would work in a server backend like FastAPI with replicas of the pipeline running. 

Since the token is stored on the instance would a rotation update all replicas using this token source?

bogdankostic: No, a rotation on one replica won't update the others. Each `RefreshTokenSource` keeps its token cache in process memory, and the per-instance lock only serializes refreshes within a single process.
 
Whether that matters depends on the identity provider:
  - Non-rotating refresh tokens (common case): fine across replicas. Each replica just exchanges the shared refresh token for its own access token.
  - Rotating / single-use refresh tokens: replicas sharing one seed token will step on each other. To run this multi-replica, a developer would need to take the refresh token out of per-instance memory by persisting rotations via the `on_rotate` callback to a shared store and have a single owner do the refresh with the replicas consuming the resulting access token. 

For multi-replica, multi-user backends the intended source is `TokenExchangeSource`. The per-request `subject_token` is the identity and its cache is just in-memory, so a lost cache only triggers a re-exchange.

integrations/oauth/src/haystack_integrations/components/connectors/oauth/sources.py
bogdankostic: I implemented Option 2 in 6f67cf3 and e31d274.

integrations/oauth/src/haystack_integrations/utils/oauth/sources.py
sjrl: Thanks for the explanation! And I think better understand why its still useful to have this even though we also have a `StaticTokenSource`. 

Could we update the docs to mention that for "multi-replica" cases `TokenExchangeSource` should be used. 

And make it more clear when a user should choose this over the `StaticTokenSource`?

integrations/oauth/src/haystack_integrations/components/connectors/oauth/resolver.py
```diff
@@ -0,0 +1,113 @@
...
+
+from typing import Any
+
+from haystack import component, default_from_dict, default_to_dict
+
+from haystack_integrations.utils.oauth import OAuthConfigError, SubjectTokenSource, TokenSource
+
+
+@component
+class OAuthResolver:
```
sjrl: I was wondering if this should be renamed to `OAuthTokenResolver` to make it more clear what's being resolved. WDYT?

integrations/oauth/src/haystack_integrations/components/connectors/oauth/resolver.py
```diff
@@ -0,0 +1,113 @@
...
+        Resolve an access token and emit it.
+
+        :param kwargs: Carries `subject_token` when the configured source requires it (declared as a mandatory
+            input in that case, injected by the application/controller per request). For config-only sources no
+            input is declared and `kwargs` is empty.
+        :returns: A dictionary with a single `access_token` key containing a bearer token string.
+        """
+        source = self.token_source
+        if source.requires_subject_token:
+            token = source.resolve(kwargs.get("subject_token", ""))
```
sjrl: Do you think we should check for the presence/value of subject_token here (e.g. raise if it's an empty string) or are we relying on the source.resolve to raise a relevant error?

integrations/oauth/src/haystack_integrations/utils/oauth/sources.py
```diff
@@ -0,0 +1,476 @@
...
+        logger.warning(
+            "Token response had a non-numeric 'expires_in' value ({value}); falling back to {default}s.",
+            value=raw_expires_in,
+            default=DEFAULT_ACCESS_TOKEN_TTL_SECONDS,
+        )
+        expires_in = float(DEFAULT_ACCESS_TOKEN_TTL_SECONDS)
+    return access_token, expires_in, payload
+
+
+class RefreshTokenSource:
```
sjrl: Since this is Oauth specific should we rename to something like `RefreshOAuthTokenSource`? More wordy but more consistent with how we name our components when they are integration specific.

integrations/oauth/src/haystack_integrations/utils/oauth/sources.py
```diff
@@ -0,0 +1,476 @@
...
+    ) -> None:
+        """
+        Initialize the source.
+
+        :param token_url: The OAuth 2.0 token endpoint.
+        :param client_id: The OAuth client identifier.
+        :param refresh_token: The refresh token to exchange. Defaults to the value of the `OAUTH_REFRESH_TOKEN`
+            environment variable.
+        :param client_secret: The client secret for confidential clients. Omit it for public clients.
+        :param scopes: The scopes to request, joined with `scope_delimiter`.
```
sjrl: Are these provider specific or is there an OAuth docstring link we could provide here for users to understand what values they could set?

integrations/oauth/src/haystack_integrations/utils/oauth/sources.py
```diff
@@ -0,0 +1,476 @@
...
+    def from_dict(cls, data: dict[str, Any]) -> "RefreshTokenSource":
+        """Deserialize the source from a dictionary."""
+        init_params = data["init_parameters"]
+        deserialize_secrets_inplace(init_params, keys=["refresh_token", "client_secret"])
+        if init_params.get("on_rotate"):
+            init_params["on_rotate"] = deserialize_callable(init_params["on_rotate"])
+        return default_from_dict(cls, data)
+
+
+class TokenExchangeSource:
```
sjrl: Same question here about naming to `OAuthTokenExchangeSource`

sjrl:
@bogdankostic looking good! A few minor comments mostly around naming and then should be good to go.

integrations/oauth/src/haystack_integrations/utils/oauth/sources.py
bogdankostic: These are provider specific. I made this explicit in the docstring.

integrations/oauth/src/haystack_integrations/utils/oauth/sources.py
bogdankostic: I tried to make the doc string a bit clearer regarding this.

sjrl (approved):
Looks good! Excited to have this released!

PR #3423, opened 2026-06-10
feat: support str in addition to Secret for aws_region_name

tstadel:
### Related Issues

- fixes #issue-number

### Proposed Changes:
When not using secrets for region, it will be as easy as setting:
```yaml
...
init_parameters:
  aws_region_name: MY_REGION
```

instead of

```yaml
...
init_parameters:
  boto3_config:
    region_name: MY_REGION
```

Using flattened settings is especially easier when dealing with UI-driven configs.

Changes applied across all six Bedrock components (chat generator, generator, text/document/document-image embedders, ranker, S3 downloader)

### Checklist

### DIFF
--- a/integrations/amazon_bedrock/tests/test_chat_generator.py
+++ b/integrations/amazon_bedrock/tests/test_chat_generator.py
@@ -307,7 +307,6 @@ class TestAmazonBedrockChatGenerator:
         assert generator.model == "global.anthropic.claude-sonnet-4-6"
         assert generator.aws_region_name.resolve_value() == "my-fake-region"
 
-
     def test_default_constructor(self, mock_boto3_session, mock_aioboto3_session, set_env_variables):
         """
         Test that the default constructor sets the correct values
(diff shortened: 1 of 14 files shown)


integrations/amazon_bedrock/tests/test_chat_generator.py
```diff
@@ -291,6 +291,22 @@ def test_from_dict(self, mock_boto3_session: Any, boto3_config: dict[str, Any] |
         assert generator.streaming_callback == print_streaming_chunk
         assert generator.boto3_config == boto3_config
 
+    def test_from_dict_aws_region_name(self, mock_boto3_session: Any):
+        """
+        Test that the from_dict method returns the correct object
```
tstadel:
```suggestion
        Test that aws_region_name as str value is correctly parsed
```

julian-risch (changes requested):
Could you please add a roundtrip test with the `aws_region_name` as str? from_dict, to_dict, from_dict. I believe the `to_dict` step will fail because we do not serialize token-based Secrets. `aws_region_name` would need to be a str again to get properly serialized. At that point allowing `aws_region_name` to be of type str or Secret might work better.
Another aspect is consistency with other AmazonBedrock components but I would be okay with deferring that to a later PR.

julian-risch (approved):

PR #3424, opened 2026-06-10
feat(mongodb_atlas): support custom schema mapping via meta_project_mapping

### LINKED ISSUE #690: Extending MongoDBAtlasDocumentStore to support custom schema
The current implementation of MongoDBAtlasDocumentStore only supports specific MongoDB document schema. Content is expected to be stored in the `content` field, and metadata must be within a `meta` subdocument. This schema requirement is enforced by the `$project` stage in the aggregation pipeline executed by `_embedding_retrieval` function:

```python
           {
                "$vectorSearch": {
                    "index": self.vector_search_index,
                    "path": "embedding",
                    "queryVector": query_embedding,
                    "numCandidates": 100,
                    "limit": top_k,
                    "filter": filters,
                }
            },
            {
                "$project": {
                    "_id": 0,
                    "content": 1,
                    "dataframe": 1,
                    "blob": 1,
                    "meta": 1,
                    "embedding": 1,
                    "score": {"$meta": "vectorSearchScore"},
                }
            }
```

This tightly couples the Haystack Document representation with the database schema, which can be inconvenient. I have a vector store in MongoDB with an existing schema defined when I was using langchaig. Specifically, I have the document's content stored in a `text` field, and I have some metadata stored in different fields of a MongoDB document (like `source` storing the original document location reference). I would prefer to avoid migrating to a new schema dictated by MongoDBAtlasDocumentStore.

I propose adding the ability to override the `$project` stage of the aggregation pipeline partially, optionally, while retaining the existing behavior as a default. For example, initializing the MongoDBAtlasDocumentStore could look like this:
```python
MongoDBAtlasDocumentStore(
    database_name="db",
    collection_name="embedded_docs",
    vector_search_index='index',
    content_field_key='text', # maps "text" field in MongoDB to Document's content
    meta_project_mapping={
       {"source": "$source"}  # allows to flexibly build meta from a MongoDB doc fields
    }
```

`self.content_field_key` and `self.meta_project_mapping` would be then used in the `$project` aggregation pipeline stage. What do you think?

I extended MongoDBAtlasDocumentStore in my project and made the described change. While this approach works, I was wondering if it would be beneficial to include it in the library.

**Additional context**
I can submit a PR :)

PranavSShetty:
### Related Issues

- fixes #690

### Proposed Changes:

This PR adds support for custom MongoDB schema mappings in MongoDBAtlasDocumentStore through a new optional meta_project_mapping configuration parameter.

When provided, keys in the Haystack Document meta dictionary can be mapped to specific root or nested MongoDB document fields.

Key additions:

1. Nested field helper utilities
   - _get_nested_value
   - _set_nested_value
   - _pop_nested_value

2. Filter translation support
   - Added _translate_filters to map metadata filter fields to corresponding MongoDB paths.

3. Document conversion enhancements
   - Updated MongoDB ↔ Haystack document conversion logic.

4. Metadata and aggregation support
   - Updated metadata analysis pipelines and retrieval logic.

5. Serialization support
   - Added serialization and deserialization support for meta_project_mapping.
   
- Added unit tests for nested path helper functions
- Added tests for document conversion
- Added tests for filter translation
- Added tests for metadata aggregation and retrieval
- Added serialization/deserialization tests

Test Results:

80 passed, 133 skipped

Backward compatibility has been preserved.

Existing behaviour remains unchanged when meta_project_mapping is not provided.


integrations/mongodb_atlas/src/haystack_integrations/document_stores/mongodb_atlas/document_store.py
```diff
@@ -1224,6 +1363,14 @@ def _mongo_doc_to_haystack_doc(self, mongo_doc: dict[str, Any]) -> Document:
             mongo_doc["content"] = mongo_doc.pop(self.content_field, None)
         if self.embedding_field != "embedding":
             mongo_doc["embedding"] = mongo_doc.pop(self.embedding_field, None)
+
+        if self.meta_project_mapping:
+            meta = mongo_doc.setdefault("meta", {})
+            for meta_key, mongo_field in self.meta_project_mapping.items():
+                val = self._pop_nested_value(mongo_doc, mongo_field)
+                if val is not None:
+                    meta[meta_key] = val
```
bogdankostic: This raises `ValueError` when the stored doc has any root-level field that isn't a Haystack Document field and isn't fully consumed by the mapping, as long as at least one mapped field populated meta. `Document.from_dict` rejects receiving both a non-empty meta and "flattened" top-level keys:

```python
store = MongoDBAtlasDocumentStore(..., meta_project_mapping={"source": "source"})
mongo_doc = {"id": "x1", "content": "hello", "source": "url", "category": "news", "meta": {}}
store._mongo_doc_to_haystack_doc(mongo_doc)
# ValueError: You can pass either the 'meta' parameter or flattened metadata keys ...
```

integrations/mongodb_atlas/src/haystack_integrations/document_stores/mongodb_atlas/document_store.py
```diff
@@ -1212,6 +1277,80 @@ async def _fulltext_retrieval_async(
...
+                    val = d.pop(key)
+                    return val, len(d) == 0
+                return None, False
+
+            if key in d and isinstance(d[key], dict):
+                val, should_delete_parent = rec_pop(d[key], path_parts[1:])
+                if should_delete_parent:
+                    d.pop(key)
+                return val, len(d) == 0
+            return None, False
```
bogdankostic: These should be static methods as they don't make use of `self`.

integrations/mongodb_atlas/src/haystack_integrations/document_stores/mongodb_atlas/document_store.py
```diff
@@ -105,6 +108,7 @@ def __init__(
         self.full_text_search_index = full_text_search_index
         self.embedding_field = embedding_field
         self.content_field = content_field
+        self.meta_project_mapping = meta_project_mapping
```
bogdankostic: The mapping values get a leading `$` stripped in ~6 places, while the aggregation builders conditionally prepend one. Prepending `$` where an aggregation expression references a field value is correct and necessary. But accepting values that already start with `$` — and therefore stripping it everywhere else — looks like an avoidable convenience that's leaking complexity across the file.

Could we instead define mapping values as bare field paths (no `$`), document that in this param, and normalize once here (e.g. strip a leading `$` when storing `self.meta_project_mapping`)? Then every downstream site can assume a clean field name, the aggregation builders are the only place that adds `$`, and the five `startswith("$")` strips go away.

bogdankostic (changes requested):
Thanks @PranavSShetty! I added a few in-line comments that should be addressed.

integrations/mongodb_atlas/src/haystack_integrations/document_stores/mongodb_atlas/document_store.py
PranavSShetty: Thanks for the review and the detailed feedback. I've gone through the comments and I agree there are a few issues that need to be addressed, particularly the "Document.from_dict" case and the mapping normalization suggestions.

I'm currently unavailable for the next week, but I'll work through the feedback and submit an updated revision as soon as I'm back.

Thanks again for taking the time to review this.

bogdankostic:
Hi @PranavSShetty, are you planning to continue working on this PR?

PranavSShetty:
> Hi @PranavSShetty, are you planning to continue working on this PR?

Yes
Sorry just got stuck with my personal work
And I have completed 90% of it
Only the last one which you had asked is remaining
And I will get it done by this weekend
Sorry for the delay

integrations/mongodb_atlas/src/haystack_integrations/document_stores/mongodb_atlas/document_store.py
bogdankostic, line 562: `_normalize_filters` is called without `_translate_filters`

integrations/mongodb_atlas/src/haystack_integrations/document_stores/mongodb_atlas/document_store.py
```diff
@@ -1262,6 +1399,23 @@ def _mongo_doc_to_haystack_doc(self, mongo_doc: dict[str, Any]) -> Document:
...
+
+            # Document.from_dict() raises ValueError when a non-empty 'meta' dict coexists with
+            # unrecognised top-level keys (it treats them as flattened metadata).  When the
+            # mapping has populated meta, drop any leftover root-level fields that are not
+            # recognised Haystack Document fields so that reconstruction always succeeds.
+            if meta:
+                for key in list(mongo_doc.keys()):
+                    if key not in self._DOCUMENT_FIELDS:
+                        mongo_doc.pop(key)
+
```
bogdankostic, line 1439: The `ValueError` is gone, thanks. But the fix leaves three related problems in this block:

**a) The strip is conditional on `meta` being non-empty**, so two documents from the same collection get different metadata semantics:

```python
store = MongoDBAtlasDocumentStore(..., meta_project_mapping={"source": "source_field"})
store._mongo_doc_to_haystack_doc({"id": "1", "content": "c", "source_field": "url", "category": "news"}).meta
# {'source': 'url'}      <- category dropped
store._mongo_doc_to_haystack_doc({"id": "2", "content": "c", "category": "news"}).meta
# {'category': 'news'}   <- category flattened into meta
```

**b) `meta` is captured before `_pop_nested_value` can detach it.** `_pop_nested_value` prunes emptied parents, so a mapping targeting a path under `meta` writes into an orphaned dict:

```python
store = MongoDBAtlasDocumentStore(..., meta_project_mapping={"author": "meta.author"})
store._mongo_doc_to_haystack_doc({"id": "1", "content": "c", "meta": {"author": "jane"}}).meta
# {}   <- value silently lost
store._mongo_doc_to_haystack_doc({"id": "2", "content": "c", "meta": {"author": "jane", "x": 1}}).meta
# {'x': 1, 'author': 'jane'}   <- works only because the sibling keeps meta alive
```

**c) A mapped value of `None` is lost rather than preserved** — `_pop_nested_value` can't distinguish absent from stored `null`, and `if val is not None` skips it after the key has already been popped out of the doc. `Document(meta={"source": None})` -> write -> read gives `meta == {}`.

Collecting into a local dict and assigning back at the end fixes all three:

```python
if self.meta_project_mapping:
    mapped_meta = {}
    for meta_key, mongo_field in self.meta_project_mapping.items():
        val = self._pop_nested_value(mongo_doc, mongo_field, default=_MISSING)
        if val is not _MISSING:
            mapped_meta[meta_key] = val

    # With an explicit mapping, unmapped root-level fields are not metadata.
    for key in list(mongo_doc):
        if key not in self._DOCUMENT_FIELDS:
            mongo_doc.pop(key)

    meta = mongo_doc.get("meta")
    if not isinstance(meta, dict):  # same guard as _haystack_doc_to_mongo_doc
        meta = {}
    meta.update(mapped_meta)
    mongo_doc["meta"] = meta
```

integrations/mongodb_atlas/tests/test_document_store.py
```diff
@@ -232,6 +272,375 @@ def test_document_conversion_methods_with_custom_field_names(self):
...
+        doc: dict = {}
+        MongoDBAtlasDocumentStore._set_nested_value(doc, "x.y", "val")
+        assert doc == {"x": {"y": "val"}}
+
+    def test_pop_nested_value_is_static(self):
+        doc = {"a": {"b": 1, "c": 2}}
+        val = MongoDBAtlasDocumentStore._pop_nested_value(doc, "a.b")
+        assert val == 1
+        assert doc == {"a": {"c": 2}}
+
```
bogdankostic: I don't think these tests are needed.

bogdankostic (changes requested):
Thanks for addressing the feedback @PranavSShetty! I added a few more inline comment that need to be fixed before we can merge.

PranavSShetty:
> Thanks for addressing the feedback @PranavSShetty! I added a few more inline comment that need to be fixed before we can merge.

Will look into it as soon as possible and will try to resolve those.

PranavSShetty:
@bogdankostic
Is the fix proper or are there any more changes required??

bogdankostic:
[This comment](https://github.com/deepset-ai/haystack-core-integrations/pull/3424#discussion_r3766632475) wasn't addressed yet, I pushed the necessary changes in bef9cf9.

bogdankostic (approved):
Looks good now, thanks @PranavSShetty!

PranavSShetty:
@bogdankostic thank u very much and please let me know if anything else is required from my end.
I would be happy to work

PR #3426, opened 2026-06-10
feat: add Oracle embedders and hybrid search support

fileames:
### Proposed Changes:

Adds missing Oracle integration capabilities from the downstream implementation into the upstream Haystack Oracle integration:

- Adds OracleTextEmbedder and OracleDocumentEmbedder using OracleConnectionConfig.
- Adds OracleHybridRetriever for DBMS hybrid vector search.
- Adds hybrid vector index support through OracleVectorizerPreference, create_hybrid_vector_index, and
  create_hybrid_vector_index_async.
- Extends vector index configuration with opt-in IVF support while keeping HNSW as the default.
- Adds direct async paths where Oracle async APIs are available, with fallback behavior preserved.
- Adds explicit close() and close_async() lifecycle APIs for sync/async pools.
- Extends filter support with contains, not contains, and hybrid filter conversion.
- Improves DBMS_SEARCH keyword index cleanup by using DBMS_SEARCH.DROP_INDEX.

### How did you test it?

- Unit tests
- Live Oracle feature integration tests

### Checklist


julian-risch:
@fileames Thank you for opening this PR! We need you to agree to the CLA please, before we can merge this PR. https://github.com/deepset-ai/haystack-core-integrations/pull/3426#issuecomment-4673198958

integrations/oracle/src/haystack_integrations/document_stores/oracle/filters.py
```diff
@@ -135,6 +150,102 @@ def _field_to_sql(field: str, value: Any) -> str:
...
+
+
+def _hybrid_filter_path(field: str) -> str:
+    if not field.startswith("meta."):
+        msg = "Oracle hybrid retrieval supports only filters under the 'meta.' field."
+        raise FilterError(msg)
+    if not re.match(_JSON_FIELD_NAME, field):
+        msg = f"Invalid metadata field name: {field!r}"
+        raise FilterError(msg)
+    return field
```
fede-kamel: **Blocking bug: hybrid retrieval with any Haystack filter raises ORA-00904.** The path is passed through verbatim (`meta.lang`), but the table column is `metadata`, so `DBMS_HYBRID_VECTOR.SEARCH` resolves it as a column reference and fails. Reproduced live on 26ai:

```
ORA-20000: Oracle Text error:
DRG-50857: oracle error in drvdml.fusion_search
ORA-00904: "THEBASE"."META"."LANG": invalid identifier
```

I verified the fix on the same database: with `path = "metadata.lang"` the identical query succeeds and filters correctly (and no `FILTER BY` clause is needed on the index — in fact `FILTER BY metadata` is rejected with ORA-29850 since JSON columns aren't allowed there). So this function should map `meta.<key>` -> `metadata.<key>`, mirroring what `FilterTranslator._field_to_json_path` already does for the regular filter path:

```python
return "metadata." + field[len("meta.") :]
```

Could you also add a hybrid-retrieval-with-filters case to `test_oracle_features_integration.py`? It's the one new code path the live tests don't currently touch.

integrations/oracle/src/haystack_integrations/document_stores/oracle/document_store.py
```diff
@@ -899,19 +1588,45 @@ async def _embedding_retrieval_async(
...
+                try:
+                    await _maybe_await(cur.execute(sql, params))
+                except oracledb.DatabaseError as e:
+                    logger.debug("Async embedding retrieval failed. SQL: %s\nParams: %s", sql, params)
+                    msg = (
+                        f"Async embedding retrieval failed. Error: {e!r}. "
+                        "You can find the SQL query and the parameters in the debug logs."
+                    )
+                    raise DocumentStoreError(msg) from e
+                rows = await _maybe_await(cur.fetchall())
```
fede-kamel: **Bug (confirmed live): the direct async path returns broken Documents when `fetch_lobs` is at its default.** `TestOracleDocumentStoreAsync::test_async_write_and_retrieve` fails against a real 26ai database:

```
ValueError: The `content` field must be a string or None.
RuntimeWarning: coroutine 'AsyncLOB.read' was never awaited
```

The async connection returns the CLOB `text` column as `AsyncLOB`, and `_row_to_document` calls `text.read()` synchronously — for an `AsyncLOB` that returns an un-awaited coroutine, which ends up as `Document.content`. `_hybrid_retrieval_async` has the same problem (its row fetch also goes through `_row_to_document`).

It didn't show up in this PR's CI because the integration job dies earlier at fixture teardown, and it's masked whenever an embedder runs first in the process because of the global `oracledb.defaults.fetch_lobs = False` in the embedders (see comment there). Suggested fix: await LOB reads on the async paths (an async-aware variant of `_row_to_document`), or set `fetch_lobs=False` on the pool/connection params instead of relying on row objects.

integrations/oracle/src/haystack_integrations/components/embedders/oracle/text_embedder.py
```diff
@@ -0,0 +1,270 @@
...
+
+    def _serialize_proxy(self) -> Any:
+        return _serialize_secret(self.proxy)
+
+    def _proxy_value(self) -> str | None:
+        proxy = _resolve_secret(self.proxy)
+        return str(proxy) if proxy else None
+
+    def _embed_documents(self, texts: list[str]) -> list[list[float]]:
+        oracledb.defaults.fetch_lobs = False
```
fede-kamel: `oracledb.defaults.fetch_lobs = False` mutates a **process-wide global** (here and in the async variant below) — it changes LOB fetch behavior for every other oracledb consumer in the process, including `OracleDocumentStore` itself. Concretely, it currently *hides* the `AsyncLOB` bug in `_embedding_retrieval_async` whenever an embedder happens to run first, which makes failures order-dependent. python-oracledb supports `fetch_lobs` as a per-connection / per-pool parameter — passing it in `_connect_kwargs()` would keep the behavior scoped to this component.

(Also flagged by mypy — note that `haystack_integrations.components.embedders.oracle` is not currently included in the `test:types` script in `pyproject.toml`, so none of this package is type-checked in CI.)

fede-kamel (commented):
Thanks for upstreaming this — the feature set is really useful and the overall structure is solid (validated identifiers, bound parameters, opt-in IVF, backwards-compatible `to_dict`). I tested the branch against a live Oracle Autonomous Database 26ai (23.26.2.2.0) with an in-database ONNX embedding model, since the CI container can't exercise most of the new functionality (see below). Results:

- `fmt-check` and the configured `test:types` pass; 50 mocked unit tests pass
- Full integration suite against the live 26ai DB: **115 passed, 2 failed, 3 skipped** — both failures are reproducible bugs (inline comments)
- Two extra live tests for hybrid retrieval **with filters** (not covered by the PR's tests) found a blocking bug (inline comment in `filters.py`)

**Why CI is red (and what it means):** the CI container is `gvenzl/oracle-free:23-slim`, and the slim flavor [removes Oracle Text entirely](https://github.com/gvenzl/oci-oracle-free/blob/main/ImageDetails.md), which `DBMS_SEARCH` is built on. The CI logs show every fixture teardown failing with `PLS-00201: identifier 'DBMS_SEARCH.DROP_INDEX' must be declared`. This was invisible before this PR because `_ensure_keyword_index()` swallows all `DatabaseError`s on create (so the keyword index has silently never existed in CI), while the new `_drop_keyword_index()` only forgives ORA-942/ORA-1418/DRG-10502 — the "package does not exist" case (ORA-06550/PLS-00201) falls through and raises. Suggestion: either treat ORA-06550/PLS-00201 as benign in `_is_missing_object_error`, or (better) switch CI to the non-slim `gvenzl/oracle-free:23` image, which would give the keyword and hybrid paths real CI coverage for the first time.

**Type-checking gap (not visible in this diff):** `tool.hatch.envs.test.scripts.types` in `pyproject.toml` only checks `document_stores.oracle` and `components.retrievers.oracle`, so the new embedders package is never type-checked. Running `mypy -p haystack_integrations.components.embedders.oracle` yields 4 errors: the `oracledb.defaults.fetch_lobs` assignment, and two Liskov violations because `OracleDocumentEmbedder.run(documents: list[Document])` overrides `OracleTextEmbedder.run(text: str)`. Please add the package to the types script and consider making the two embedders independent components (per Haystack convention) instead of subclassing.

**Pre-existing latent bug surfaced by live testing (not introduced here, but related):** on a database where `DBMS_SEARCH` actually works, `test_write_documents_duplicate_overwrite` fails with `ORA-06531: Reference to uninitialized collection` raised from the `DBMS_SEARCH`-generated maintenance trigger (`DR$...$TRIG`) during the overwrite. So `DuplicatePolicy.OVERWRITE` is currently broken on any database with a functioning keyword index — worth a separate issue, and it will surface in CI as soon as the image is switched to non-slim.

Smaller points: `pydoc/config_docusaurus.yml` wasn't updated with the new modules (see `oracle.md` comment); the `VECDB_*` env fallbacks in `tests/conftest.py` look downstream-specific, and conftest doesn't read `ORACLE_WALLET_LOCATION`/`ORACLE_WALLET_PASSWORD` while `test_oracle_features_integration.py` does — adding them would let the non-mocked tests run against wallet-based databases (I had to patch this locally to test against ADB).

integrations/oracle/src/haystack_integrations/document_stores/oracle/document_store.py
```diff
@@ -498,17 +1008,39 @@ async def count_documents_async(self) -> int:
         """
         return await asyncio.to_thread(self.count_documents)
 
+    def _keyword_index_name(self) -> str:
+        index_name = f"{self.table_name}_search_idx"
+        return index_name[:MAX_INDEX_NAME_LEN]
+
+    def _drop_keyword_index(self, cur: Any) -> None:
```
fede-kamel, line 1069: This is the line making CI red: `gvenzl/oracle-free:23-slim` strips Oracle Text, so `DBMS_SEARCH` doesn't exist there, and `DBMS_SEARCH.DROP_INDEX` fails with `ORA-06550` / `PLS-00201: identifier 'DBMS_SEARCH.DROP_INDEX' must be declared` — which `_is_missing_object_error` doesn't recognize, so every test fixture teardown raises `DocumentStoreError`.

Note the asymmetry with `_ensure_keyword_index()`, which swallows *all* `DatabaseError`s on create — so on databases without `DBMS_SEARCH`, creation silently no-ops but deletion hard-fails. Options:

1. Treat ORA-06550/PLS-00201 as benign here (keeps the slim CI image), and/or
2. Switch CI to non-slim `gvenzl/oracle-free:23` — which would also give the keyword/hybrid features actual CI coverage.

I verified on a full 26ai install that create + drop both work fine, so this is purely about environments lacking Oracle Text.

integrations/oracle/src/haystack_integrations/document_stores/oracle/document_store.py
```diff
@@ -858,6 +1377,176 @@ async def get_metadata_field_unique_values_async(
...
+            "values": ["rowid", "score", "vector_score", "text_score"],
+            "format": "JSON",
+        }
+        return search_params
+
+    @staticmethod
+    def _merge_hybrid_scores(
+        search_rows: list[dict[str, Any]], documents: list[Document], *, return_scores: bool
+    ) -> None:
+        for row, document in zip(search_rows, documents, strict=False):
```
fede-kamel: Two issues with the score merge:

1. **Misalignment risk:** the per-rowid re-fetch loop in `_hybrid_retrieval` can return zero rows for a rowid that was deleted between the search and the fetch; `zip(..., strict=False)` then silently shifts every subsequent score onto the wrong document. Pairing the score with the row at fetch time (e.g. building `(row, score)` tuples inside the loop) removes the failure mode.
2. **N+1 queries:** one `SELECT ... WHERE ROWID = :rid` per hit; a single `WHERE ROWID IN (...)` (or fetching alongside the search) would cut `top_k` round-trips to one.

integrations/oracle/src/haystack_integrations/document_stores/oracle/document_store.py
```diff
@@ -268,33 +676,135 @@ def create_keyword_index(self) -> None:
...
+
+        Either provide an existing ``vectorizer_preference`` or a text embedder from which a new
+        preference can be created. The returned preference can be dropped by the caller if it was
+        created only for this index.
+        """
+        if vectorizer_preference is None:
+            if text_embedder is None:
+                msg = "text_embedder is required when vectorizer_preference is not provided."
+                raise ValueError(msg)
+            vectorizer_preference = OracleVectorizerPreference.create(self, text_embedder)
```
fede-kamel, line 780: If the index DDL below fails after this auto-created preference is committed, the preference is orphaned (the caller never gets the handle back to drop it). Consider wrapping the DDL in `try/except` and dropping the preference on failure when it was created internally. Same applies to `create_hybrid_vector_index_async`.

integrations/oracle/src/haystack_integrations/components/embedders/oracle/document_embedder.py
```diff
@@ -0,0 +1,88 @@
...
+from haystack import component, default_to_dict
+from haystack.dataclasses import Document
+
+from haystack_integrations.document_stores.oracle import OracleConnectionConfig
+
+from .text_embedder import OracleTextEmbedder
+
+
+@component
+class OracleDocumentEmbedder(OracleTextEmbedder):
```
fede-kamel: Subclassing `OracleTextEmbedder` here violates LSP (`run(text: str)` -> `run(documents: list[Document])`), which mypy flags as soon as the embedders package is added to the types script:

```
error: Argument 1 of "run" is incompatible with supertype ... [override]
```

Haystack's other integrations implement Text/Document embedders as two independent components sharing a helper — that would resolve the override errors and avoid `isinstance(doc_embedder, OracleTextEmbedder)` being true in pipelines.

Minor: `document.embedding = embedding` triggers Haystack's mutation warning in the test run; current guidance is `dataclasses.replace(...)` (same for `document.score` in `_merge_hybrid_scores`).

integrations/oracle/oracle.md
```diff
@@ -0,0 +1,786 @@
+---
```
fede-kamel: This looks like a committed build artifact (`hatch run docs` output) — no other integration checks in its rendered API markdown, and this copy appears to have been generated from the downstream sources (it contains `__init__` docstrings that don't exist in this branch). Meanwhile `pydoc/config_docusaurus.yml` still lists only `embedding_retriever` and `document_store`, so the published docs won't include the new embedders or the hybrid retriever.

Suggest: delete `oracle.md` and instead add the new modules (`components.embedders.oracle.*`, `components.retrievers.oracle.hybrid_retriever`, ideally also the pre-existing `keyword_retriever`) to the pydoc config. That's also why the PR got auto-labeled `type:documentation`.

integrations/oracle/tests/conftest.py
```diff
@@ -11,9 +12,9 @@
 
 from haystack_integrations.document_stores.oracle import OracleConnectionConfig, OracleDocumentStore
 
-_USER = "haystack"
-_PASSWORD = "haystack"
-_DSN = "localhost:1521/freepdb1"
+_USER = os.getenv("ORACLE_USER") or os.getenv("VECDB_USER") or "haystack"
```
fede-kamel: The `VECDB_USER`/`VECDB_PASS`/`VECDB_HOST` fallbacks look specific to the downstream environment — upstream, `ORACLE_*` alone seems cleaner. Conversely, it would be genuinely useful for conftest to also read `ORACLE_WALLET_LOCATION`/`ORACLE_WALLET_PASSWORD` (as `test_oracle_features_integration.py` already does) so the non-mocked tests can run against wallet-based databases like Autonomous DB — I had to patch this locally to run the suite against a live 26ai instance.

fede-kamel:
## Review — validated against a live Oracle 26ai database

Thanks @fileames! I went through the embedders + hybrid-search work and ran it against a live Oracle 26ai database with Oracle Text / `DBMS_SEARCH`, `DBMS_VECTOR_CHAIN`, `DBMS_HYBRID_VECTOR` and an in-database ONNX embedding model. Unit tests, `ruff` and `mypy` are all clean. Two issues reproduce against a real DB but are **invisible in CI**, because the `gvenzl/oracle-free` image ships without Oracle Text — so the keyword index (and its trigger) is never created in CI.

### 🔴 1. Hybrid retrieval + any metadata filter → `ORA-20000` / `ORA-00904`
`to_hybrid_filter()` turns a Haystack filter such as `{"field": "meta.lang", "operator": "==", "value": "en"}` into `filter_by={"path": "meta.lang", ...}`, but `DBMS_HYBRID_VECTOR.SEARCH`'s `filter_by.path` resolves to a **column on the base table**, not a JSON path into the `metadata` column:

```
ORA-00904: "THEBASE"."META"."LANG": invalid identifier   -- path "meta.lang"
ORA-00904: "THEBASE"."LANG": invalid identifier          -- even with the meta. prefix stripped
```

So hybrid retrieval raises for *any* metadata filter. `test_hybrid_retriever_live` never passes `filters`, so this wasn't caught (the unit test only asserts the JSON shape against a mock).

**Fix:** apply Haystack metadata filters as a SQL predicate while fetching the ranked hits (post-filter); native `filter_by` over columns declared with `FILTER BY` at index-creation time is still available via `params`. This also removes a positional score-misalignment that occurred whenever a ranked hit was dropped.

### 🔴 2. `write_documents(policy=OVERWRITE)` → `ORA-06531`
The OVERWRITE `MERGE` (combining `WHEN MATCHED UPDATE` with `WHEN NOT MATCHED INSERT`) fails inside the `DBMS_SEARCH` keyword-index trigger — even on an empty table where every row is an insert:

```
ORA-06531: Reference to uninitialized collection
ORA-06512: at "...DR$...$TRIG"
```

`NONE` (plain INSERT) and `SKIP` (`MERGE ... WHEN NOT MATCHED`) both work — only the combined update+insert MERGE trips the trigger.

**Fix:** delete-then-insert for OVERWRITE (rows de-duplicated by id, last wins).

### Fix branch + verification
I pushed both fixes plus live integration tests that cover them:

**`fede-kamel:fix/oracle-hybrid-filter-and-overwrite`**
https://github.com/fede-kamel/haystack-core-integrations/tree/fix/oracle-hybrid-filter-and-overwrite

After the fix, against the live DB: the new `test_hybrid_retriever_with_metadata_filter_live` and `test_write_documents_overwrite_policy_live` pass, the existing hybrid test still passes, and the full base/features suite is green (unit + ruff + mypy also clean). To apply it onto this branch:

```bash
git fetch https://github.com/fede-kamel/haystack-core-integrations fix/oracle-hybrid-filter-and-overwrite
git cherry-pick 08cd15f1
```

### Minor (non-blocking)
- **`oracle.md`** (786 lines) is a generated pydoc artifact — no other integration commits one, and it's already drifting from the code (module path `...components.document_stores...`, docstrings that don't match the classes). Suggest deleting it and instead adding the new modules — `text_embedder`, `document_embedder`, `hybrid_retriever`, `keyword_retriever` — to `pydoc/config_docusaurus.yml`, which today lists only `embedding_retriever` + `document_store`, so the new components are missing from the generated docs.
- `OracleDocumentEmbedder` mutates `document.embedding` in place → Haystack emits a warning; consider `dataclasses.replace`.
- An empty `in` / `not in` list produces invalid SQL `... IN ()`.
- `_has_async_pool()` actually checks whether the *library* supports async pools (always true on oracledb 2.x), so the async methods always open a second pool alongside the sync one; `__del__` only closes the sync pool.

_Separate / pre-existing (not introduced here): `_keyword_retrieval` with filters builds `WHERE t. (...)` via `where.replace("WHERE", "WHERE t.")`._

fede-kamel:
@fileames I packaged the two fixes from my review above as a small delta PR **against your branch** so it's easy to pull in (I don't have push access to `oracle-upstream-compat-features` directly):

👉 **fileames/haystack-core-integrations#2** — `fix(oracle): hybrid metadata filters + OVERWRITE upsert`

It branches directly off `oracle-upstream-compat-features`, so it contains only the delta on top of your work: the hybrid-filter post-filtering fix, the OVERWRITE delete-then-insert fix, the two live integration tests, and wallet support in conftest. All verified against a live Oracle 26ai DB (details in my comment above).

**To get it into this PR — whichever is easiest for you:**

1. **Merge it** — merging #2 into `oracle-upstream-compat-features` updates this PR (#3426) automatically, since they share the same head branch.
2. **Cherry-pick the commit:**
   ```bash
   git checkout oracle-upstream-compat-features
   git fetch https://github.com/fede-kamel/haystack-core-integrations fix/oracle-hybrid-filter-and-overwrite
   git cherry-pick 08cd15f1
   git push
   ```
3. **Pull the branch directly:**
   ```bash
   git checkout oracle-upstream-compat-features
   git pull https://github.com/fede-kamel/haystack-core-integrations fix/oracle-hybrid-filter-and-overwrite
   git push
   ```

Happy to adjust the approach (e.g. the post-filter top_k semantics) if you'd prefer something different.

davidsbatista:
@fileames you need to sign CLA otherwise even if it's approved we can are not able to merge the PR.

fede-kamel:
@fileames you're right — I had branched off an earlier commit ("Fix lint issues") and missed your "Address feedback" changes. I re-checked the latest HEAD and validated both items I'd flagged against a live Oracle Autonomous DB with Oracle Text/`DBMS_SEARCH` and `DBMS_HYBRID_VECTOR` enabled (the CI `gvenzl` image lacks these, so neither path is exercised in CI). Both are fixed:

**1. `OVERWRITE` → `ORA-06531` in the keyword-index trigger — fixed.**
Switching from the combined `MERGE` to the PL/SQL `UPDATE`-then-`INSERT` block avoids the `DBMS_SEARCH` trigger fault. Differential check on the same DB: restoring the old `MERGE` still raises `ORA-06531: Reference to uninitialized collection` / `error during execution of trigger 'DR$...$TRIG'`, while your current code passes — verified for update-existing, insert-new, and repeated-id-within-a-batch.

**2. Hybrid retrieval + metadata filters → `ORA-00904` — fixed.**
Correcting the `filter_by` path prefix from `meta.` to `metadata.` (the real JSON column) resolves it. Verified live with `==`, `>=` (range), `in`, `not in`, and a nested `AND` — all return correctly filtered results. The per-row fetch + skip-on-missing-rowid also keeps scores aligned when a hit drops out.

Full integration suite against that DB: **119 passed, 3 skipped**. The one failure (`test_update_by_filter`) was `ORA-04036 PGA_AGGREGATE_LIMIT` — a memory cap on the small test instance, unrelated to this PR.

So nothing further needed from your side on these two. My separate fix branch is redundant and I'm dropping it. Thanks for the quick turnaround!

fileames:
@davidsbatista Thanks for all the help. I will sign the CLA very soon, in the meantime, is there anything else blocking in this branch? How can I resolve [Core / Add label on docstrings edit / label (pull_request_target)](https://github.com/deepset-ai/haystack-core-integrations/actions/runs/27951865075/job/82710445545?pr=3426) error?

davidsbatista:
> @davidsbatista Thanks for all the help. I will sign the CLA very soon, in the meantime, is there anything else blocking in this branch? How can I resolve [Core / Add label on docstrings edit / label (pull_request_target)](https://github.com/deepset-ai/haystack-core-integrations/actions/runs/27951865075/job/82710445545?pr=3426) error?

fix is here: https://github.com/deepset-ai/haystack-core-integrations/pull/3484

davidsbatista:
it's fixed now

davidsbatista:
@fede-kamel thank you very much for the detailed review!

@fileames, thanks again for working on this. What's the status of the review comments, where they addressed? I can take it over if you don't have time, but again, for this to be merged, the CLA needs to be signed.

fileames:
Hi @davidsbatista the review comments are addressed. For the CLA, is it possible to sign it as Oracle instead of an individual?

davidsbatista:
> Hi @davidsbatista the review comments are addressed. For the CLA, is it possible to sign it as Oracle instead of an individual?

It always needs to be a GitHub account to sign it. You need to open a PR with this code from the Oracle GitHub account.

davidsbatista:
@fileames, I'm working to get this PR signed by Oracle.

Can you point me to the company representative for the specific Oracle Legal Entity who can sign an agreement for Oracle?

julian-risch:
@fede-kamel @fileames Any updates on the contributors license agreement? If it's a show stopper, an alternative is that you publish the integration yourselves, using our repository template https://github.com/deepset-ai/custom-component. We'd promote the integration in the same way.

PR #3427, opened 2026-06-11
ci: add hf token to CI env

sjrl:
### Related Issues

- should hopeful fix the 429 errors we are getting in our tests download HF models. e.g. https://github.com/deepset-ai/haystack-core-integrations/actions/runs/27316341057/job/80697640152

### Proposed Changes:

 
 

Add our HF_TOKEN to the env so it's picked up by huggingface hub to give us higher api rate limits.

### How did you test it?

Haven't tested, we can observe to see if the 429 requests go away. 

### Checklist

### DIFF
--- a/.github/workflows/CI_license_compliance.yml
+++ b/.github/workflows/CI_license_compliance.yml
@@ -12,13 +12,14 @@ on:
 
 env:
   PYTHON_VERSION: "3.10"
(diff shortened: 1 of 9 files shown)


anakin87 (approved):
Let's also exclude llama.cpp from license compliance (it seems that the new version does not correctly ship license but it's MIT).

I think that this will help but won't fix the issue out of the box due to `del_hf_env_vars` fixture. I am already working to improve it.

PR #3429, opened 2026-06-11
feat(microsoft_sharepoint): Add Microsoft SharePoint retriever integration

bogdankostic:
### Related Issues

  - part of https://github.com/deepset-ai/haystack-private/issues/346

  ### Proposed Changes:

  Adds a new integration, `microsoft-sharepoint-haystack`, providing the `MSSharePointRetriever` component.

  Given a query, the retriever calls the Microsoft Graph Search API (`POST /search/query`) and returns matching SharePoint and OneDrive content as Haystack `Document`s. Each hit becomes a `Document` whose `content` is the search snippet (hit-highlight markup stripped, entities unescaped) and whose `meta` carries `file_name`, `web_url`, `entity_type`,
  `created_date_time`, `last_modified_date_time`, `created_by`, `last_modified_by`, `mime_type`, and `file_extension` (keys with no value are omitted).
  
  ### How did you test it?
  
  - Unit tests covering: init validation, serialization round-trip, hit-to-`Document` mapping, request-body and auth-header construction, KQL/`fields` omission, pagination with offset/size, `top_k` handling, error handling (401/403/other 4xx, 429 retry-then-succeed, give-up after max retries), running inside a `Pipeline`, and the async path.
  - Manual live verification against a real Microsoft 365 tenant, including a two-user per-user-scoping check: a user with access to a restricted site receives its content, while a user without access does not.

  ### Notes for the reviewer

 - The `access_token` is provided at run time. It can be obtained from the `OAuthResolver`, which ships in a separate PR: https://github.com/deepset-ai/haystack-core-integrations/pull/3419
  - The component does not fetch the files themselves. It returns the summary provided by the Search API in the `Document`'s `content` field, with the link to the file in the `meta` (`web_url`). Fetching the actual file contents requires additional downstream components.

### Checklist


.github/workflows/microsoft_sharepoint.yml
```diff
@@ -0,0 +1,140 @@
...
+      - "integrations/microsoft_sharepoint/**"
+      - "!integrations/microsoft_sharepoint/*.md"
+      - ".github/workflows/microsoft_sharepoint.yml"
+
+defaults:
+  run:
+    working-directory: integrations/microsoft_sharepoint
+
+concurrency:
+  group: microsoft_sharepoint-${{ github.head_ref || github.sha }}
```
sjrl: I think this should be a dash and not an underscore based off the same pattern for `amazon-bedrock`
```suggestion
  group: microsoft-sharepoint-${{ github.head_ref || github.sha }}
```

not sure how much it matters seems like other integrations use dash or underscore

integrations/microsoft_sharepoint/src/haystack_integrations/components/retrievers/microsoft_sharepoint/retriever.py
```diff
@@ -0,0 +1,354 @@
...
+        self.entity_types = entity_types
+        self.top_k = top_k
+        self.fields = fields
+        self.query_template = query_template
+        self.graph_url = graph_url.rstrip("/")
+        self.timeout = timeout
+        self.max_retries = max_retries
+
+    @component.output_types(documents=list[Document])
+    def run(self, query: str, access_token: str, top_k: int | None = None) -> dict[str, list[Document]]:
```
sjrl: Minor suggestion: Would it also be worth supporting passing a `Secret` for `access_token` in addition here?

integrations/microsoft_sharepoint/src/haystack_integrations/components/retrievers/microsoft_sharepoint/retriever.py
```diff
@@ -0,0 +1,354 @@
...
+
+
+@component
+class MSSharePointRetriever:
+    """
+    Retrieves content from Microsoft SharePoint and OneDrive via the Microsoft Search (Graph) API.
+
+    Given a query, the retriever calls `POST /search/query` and maps each hit to a Haystack `Document`
+    whose `content` is the search snippet and whose `meta` carries the resource metadata (`file_name`,
+    `web_url`, `entity_type`, author, timestamps, `mime_type`, and `file_extension`). It does not
```
sjrl:
```suggestion
    `web_url`, `entity_type`, `author`, `timestamps`, `mime_type`, and `file_extension`). It does not
```

integrations/microsoft_sharepoint/src/haystack_integrations/components/retrievers/microsoft_sharepoint/retriever.py
bogdankostic: Added support for `Secret` in e9ad7bc

integrations/microsoft_sharepoint/src/haystack_integrations/components/retrievers/microsoft_sharepoint/retriever.py
```diff
@@ -0,0 +1,374 @@
...
+        if not response.is_error:
+            return response.json()
+
+        status = response.status_code
+        if status == _HTTP_UNAUTHORIZED:
+            msg = (
+                "Microsoft Graph rejected the access token (401 Unauthorized). The token may be expired, invalid, "
+                "or missing the required delegated scopes (for example Files.Read.All / Sites.Read.All)."
+            )
+        elif status == _HTTP_FORBIDDEN:
```
sjrl: Minor comment: I feel like these status code values are pretty common. Do we need a global variable for them?

integrations/microsoft_sharepoint/src/haystack_integrations/components/retrievers/microsoft_sharepoint/retriever.py
```diff
@@ -0,0 +1,374 @@
...
+    The retriever takes a per-user `access_token` as a run input, typically wired
+    from an upstream `OAuthResolver`. The token must carry delegated Microsoft Graph permissions
+    (for example `Files.Read.All` and, for site/list scoping, `Sites.Read.All`). The Search API supports
+    delegated permissions only.
+
+    ### Usage example
+    ```python
+    from haystack import Pipeline
+    from haystack.utils import Secret
+    from haystack_integrations.components.connectors.oauth import OAuthResolver, RefreshTokenSource
```
sjrl: Maybe we could update this example to not use another integration? Then in the docs page we make for this I think we can add a full example where we can explain that users also need to install the other haystack integration.

integrations/microsoft_sharepoint/src/haystack_integrations/components/retrievers/microsoft_sharepoint/retriever.py
```diff
@@ -0,0 +1,374 @@
...
+        :param entity_types: The Microsoft Search entity types to query. Defaults to `["driveItem", "listItem"]`,
+            which covers files, folders, SharePoint pages and news, and list items. Other valid values are
+            `"list"` and `"site"`.
+        :param top_k: The maximum number of documents to return. Maps to the Search API `size` and is paginated
+            when it exceeds a single page.
+        :param fields: Optional list of resource properties to request via the Search API `fields` selection
+            (only honored for `listItem` and `driveItem` entity types).
+        :param query_template: Optional KQL query template used to scope the search, for example
+            `'{searchTerms} path:"https://contoso.sharepoint.com/sites/Team"'`. The literal `{searchTerms}`
+            placeholder is replaced by the run-time query.
```
sjrl: For the entity_types, fields and query_template are there microsoft docs links we could provide?

integrations/microsoft_sharepoint/src/haystack_integrations/components/retrievers/microsoft_sharepoint/retriever.py
```diff
@@ -0,0 +1,374 @@
...
+        if not isinstance(access_token, Secret):
+            return access_token
+        resolved = access_token.resolve_value()
+        if not isinstance(resolved, str):
+            msg = "The access_token Secret did not resolve to a string value."
+            raise SharePointConfigError(msg)
+        return resolved
+
+    def _build_request_body(self, query: str, offset: int, size: int) -> dict[str, Any]:
+        """Build the Microsoft Search `POST /search/query` request body for one page."""
```
sjrl: I realize it's internal but if possible providing a microsoft docs link for how to see what are valid params to this endpoint would be great for future devs

integrations/microsoft_sharepoint/src/haystack_integrations/components/retrievers/microsoft_sharepoint/retriever.py
```diff
@@ -0,0 +1,374 @@
...
+    def _post(self, client: httpx.Client, access_token: str, body: dict[str, Any]) -> dict[str, Any]:
+        """Send a search request, retrying on throttling/transient errors, and return the parsed JSON."""
+        url = f"{self.graph_url}/search/query"
+        headers = {"Authorization": f"Bearer {access_token}", "Content-Type": "application/json"}
+        attempt = 0
+        while True:
+            response = client.post(url, headers=headers, json=body)
+            if response.status_code in _RETRYABLE_STATUS and attempt < self.max_retries:
+                attempt += 1
+                delay = self._retry_delay(response, attempt)
```
sjrl: Maybe worth using tenacity instead of rolling our own retry mechanism?

We may even be able to reuse `request_with_retry` from `haystack` which uses tenacity and httpx under the hood.

sjrl: We also have an async version `async_request_with_retry` available in haystack in `haystack/utils/request_utils.py`

integrations/microsoft_sharepoint/src/haystack_integrations/components/retrievers/microsoft_sharepoint/retriever.py
```diff
@@ -0,0 +1,374 @@
...
+        self.entity_types = entity_types
+        self.top_k = top_k
+        self.fields = fields
+        self.query_template = query_template
+        self.graph_url = graph_url.rstrip("/")
+        self.timeout = timeout
+        self.max_retries = max_retries
+
+    @component.output_types(documents=list[Document])
+    def run(self, query: str, access_token: str | Secret, top_k: int | None = None) -> dict[str, list[Document]]:
```
sjrl, line 176: Out of curiosity is there a concept of `filters` we could also add here? Just to make it work more like our doc store retrievers

sjrl:
@bogdankostic one thing I noticed is that ideally we can use this new sharepoint retriever inside of the `MultiRetriever` since that is our recommended retriever to use in platform and allows for not needing to manually wire many retrievers all up to a document joiner. 

To do this though we will need two sets of changes. The main one I want to propose for this integration (I think) is number 2 where we create a wrapper component to align with the `TextRetriever` protocol. WDYT?

**1. `MultiRetriever` — add a per-retriever run-inputs map** (haystack)

A new optional run param (`run` + `run_async`) that maps retriever name → extra kwargs, merged only into that retriever's call. Generic, explicit routing, no signature introspection, protocol-safe (the protocol allows extra optional params), and not serialized (it's a run input, not init).

```python
def run(self, query, filters=None, top_k=None, *, active_retrievers=None,
        retriever_inputs: dict[str, dict[str, Any]] | None = None):
    retriever_inputs = retriever_inputs or {}
    unknown = set(retriever_inputs) - self.retrievers.keys()
    if unknown:
        raise ValueError(f"Unknown retriever name(s) in retriever_inputs: {sorted(unknown)}")
    ...
    executor.submit(
        retriever.run,
        query=query, filters=resolved_filters, top_k=resolved_top_k,
        **retriever_inputs.get(name, {}),   # only this retriever gets the extras
    )
```

**2. Create a new `SharePointTextRetriever` adapter**

Wraps an `OAuthResolver` + `MSSharePointRetriever`, conforms to our `TextRetriever` protocol, and resolves the token per run. If the resolver's token source needs a per-request `subject_token`, the adapter declares it as a **mandatory** input too — mirroring the resolver's contract. 

```python
@component
class SharePointTextRetriever:
    def __init__(self, *, retriever: MSSharePointRetriever, resolver: OAuthResolver):
        self.retriever = retriever
        self.resolver = resolver
        # public token-source attribute, not the resolver's private flag
        self._requires_subject_token = bool(getattr(resolver.token_source, "requires_subject_token", False))
        if self._requires_subject_token:
            component.set_input_type(self, "subject_token", str)  # no default => mandatory socket

    @component.output_types(documents=list[Document])
    def run(self, query: str, filters=None, top_k=None, **kwargs):
        if filters is not None:
            logger.warning("SharePoint retrieval ignores `filters`; scope via `query_template` instead.")
        access_token = self.resolver.run(**self._resolver_kwargs(kwargs))["access_token"]
        return self.retriever.run(query=query, access_token=access_token, top_k=top_k)

    @component.output_types(documents=list[Document])
    async def run_async(self, query: str, filters=None, top_k=None, **kwargs):
        # signature must match run() — enforced by @component
        if filters is not None:
            logger.warning("SharePoint retrieval ignores `filters`; scope via `query_template` instead.")
        result = await self.resolver.run_async(**self._resolver_kwargs(kwargs))
        return await self.retriever.run_async(query=query, access_token=result["access_token"], top_k=top_k)

    def _resolver_kwargs(self, kwargs):
        if not self._requires_subject_token:
            return {}
        return {"subject_token": kwargs.get("subject_token", "")}

    # also add to_dict / from_dict
```

integrations/microsoft_sharepoint/src/haystack_integrations/components/retrievers/microsoft_sharepoint/retriever.py
```diff
@@ -0,0 +1,374 @@
...
+    )
+    pipeline.add_component("retriever", MSSharePointRetriever(top_k=5))
+    pipeline.connect("resolver.access_token", "retriever.access_token")
+
+    result = pipeline.run({"retriever": {"query": "quarterly roadmap"}})
+    documents = result["retriever"]["documents"]
+    ```
+    """
+
+    def __init__(
```
sjrl, line 119: Do you think it would make sense to also allow passing a Secret in at init time to allow users to configure a default access token?

Or do you think it is better to enforce that users must pass this at run time?

integrations/microsoft_sharepoint/src/haystack_integrations/components/retrievers/microsoft_sharepoint/retriever.py
bogdankostic: Just found out that there is `httpx.codes.UNAUTHORIZED` and `httpx.codes.FORBIDDEN`, so there's no need to have our own global variables for them. I'll update to make use of httpx.

integrations/microsoft_sharepoint/src/haystack_integrations/components/retrievers/microsoft_sharepoint/retriever.py
bogdankostic: I looked into `request_with_retry` but it won't quite work for in this case: it uses tenacity's plain `wait_exponential()` and ignores the `Retry-After` header. The Graph Search API returns `Retry-After` on 429 throttling and MS recommends honoring it, which is what the current `_retry_delay` does.

tenacity directly should work, though. I'll implement it that way and update the PR.

integrations/microsoft_sharepoint/src/haystack_integrations/components/retrievers/microsoft_sharepoint/retriever.py
bogdankostic: Microsoft provides the [Keyword query language](https://learn.microsoft.com/en-us/sharepoint/dev/general-development/keyword-query-language-kql-syntax-reference) that allows for filtering directly in the query. We could either add a `filters` parameter here that would get concatenated to the query or just keep the `query` parameter and add more details to the doc string on how to apply filters. What would you prefer?

sjrl: In the spirt of this comment https://github.com/deepset-ai/haystack-core-integrations/pull/3429#issuecomment-4731428204 it would be great to add direct `filters` support if it's not too much work. 

Otherwise we would end up with a `filters` param on `SharePointTextRetriever` that wouldn't do anything.

integrations/microsoft_sharepoint/src/haystack_integrations/components/retrievers/microsoft_sharepoint/retriever.py
bogdankostic: I'd lean toward enforcing it at run time because Microsoft Graph access tokens are short-lived.

sjrl: sounds good!

bogdankostic:
> @bogdankostic one thing I noticed is that ideally we can use this new sharepoint retriever inside of the `MultiRetriever` since that is our recommended retriever to use in platform and allows for not needing to manually wire many retrievers all up to a document joiner.
> 
> To do this though we will need two sets of changes. The main one I want to propose for this integration (I think) is number 2 where we create a wrapper component to align with the `TextRetriever` protocol. WDYT?
> 
> **1. `MultiRetriever` — add a per-retriever run-inputs map** (haystack)
> 
> A new optional run param (`run` + `run_async`) that maps retriever name → extra kwargs, merged only into that retriever's call. Generic, explicit routing, no signature introspection, protocol-safe (the protocol allows extra optional params), and not serialized (it's a run input, not init).
> 
> ```python
> def run(self, query, filters=None, top_k=None, *, active_retrievers=None,
>         retriever_inputs: dict[str, dict[str, Any]] | None = None):
>     retriever_inputs = retriever_inputs or {}
>     unknown = set(retriever_inputs) - self.retrievers.keys()
>     if unknown:
>         raise ValueError(f"Unknown retriever name(s) in retriever_inputs: {sorted(unknown)}")
>     ...
>     executor.submit(
>         retriever.run,
>         query=query, filters=resolved_filters, top_k=resolved_top_k,
>         **retriever_inputs.get(name, {}),   # only this retriever gets the extras
>     )
> ```
> 
> **2. Create a new `SharePointTextRetriever` adapter**
> 
> Wraps an `OAuthResolver` + `MSSharePointRetriever`, conforms to our `TextRetriever` protocol, and resolves the token per run. If the resolver's token source needs a per-request `subject_token`, the adapter declares it as a **mandatory** input too — mirroring the resolver's contract.
> 
> ```python
> @component
> class SharePointTextRetriever:
>     def __init__(self, *, retriever: MSSharePointRetriever, resolver: OAuthResolver):
>         self.retriever = retriever
>         self.resolver = resolver
>         # public token-source attribute, not the resolver's private flag
>         self._requires_subject_token = bool(getattr(resolver.token_source, "requires_subject_token", False))
>         if self._requires_subject_token:
>             component.set_input_type(self, "subject_token", str)  # no default => mandatory socket
> 
>     @component.output_types(documents=list[Document])
>     def run(self, query: str, filters=None, top_k=None, **kwargs):
>         if filters is not None:
>             logger.warning("SharePoint retrieval ignores `filters`; scope via `query_template` instead.")
>         access_token = self.resolver.run(**self._resolver_kwargs(kwargs))["access_token"]
>         return self.retriever.run(query=query, access_token=access_token, top_k=top_k)
> 
>     @component.output_types(documents=list[Document])
>     async def run_async(self, query: str, filters=None, top_k=None, **kwargs):
>         # signature must match run() — enforced by @component
>         if filters is not None:
>             logger.warning("SharePoint retrieval ignores `filters`; scope via `query_template` instead.")
>         result = await self.resolver.run_async(**self._resolver_kwargs(kwargs))
>         return await self.retriever.run_async(query=query, access_token=result["access_token"], top_k=top_k)
> 
>     def _resolver_kwargs(self, kwargs):
>         if not self._requires_subject_token:
>             return {}
>         return {"subject_token": kwargs.get("subject_token", "")}
> 
>     # also add to_dict / from_dict
> ```

Before considering this I want to flag that `MSSharePointRetriever` doesn't return the full content of the files. Each hit maps to a Document whose content is just the Search snippet, plus metadata. It doesn't download or convert the underlying file, you need a downstream fetcher/converter on `web_url` to get the actual content. That's quite a difference from our doc-store retrievers, which return full document content.

sjrl:
> 
> Before considering this I want to flag that `MSSharePointRetriever` doesn't return the full content of the files. Each hit maps to a Document whose content is just the Search snippet, plus metadata. It doesn't download or convert the underlying file, you need a downstream fetcher/converter on `web_url` to get the actual content. That's quite a difference from our doc-store retrievers, which return full document content.

Fair point! So we'd need to be able to wire it up to something like our `LinkContentFetcher`. Okay I think then it's fair to skip the creation of `SharePointTextRetriever` for now. 

Do you think it makes sense to make sure this component is compatible with the `LinkContentFetcher` then? Or would we need a different `MSSharepointFetcher` since I'm guessing that we will still need to handle credentials somehow.

bogdankostic:
> Do you think it makes sense to make sure this component is compatible with the `LinkContentFetcher` then? Or would we need a different `MSSharepointFetcher` since I'm guessing that we will still need to handle credentials somehow.

Good point, `LinkContentFetcher` will probably not work without an access token, so it makes sense to add a `MSSharepointFetcher` - let's do that in a separate PR.

integrations/microsoft_sharepoint/src/haystack_integrations/components/retrievers/microsoft_sharepoint/retriever.py
sjrl: No longer a concern fine to just document the additional details of how to do this via the query param

sjrl:
> Good point, `LinkContentFetcher` will probably not work without an access token, so it makes sense to add a `MSSharepointFetcher` - let's do that in a separate PR.

@bogdankostic could you open an issue for this?

sjrl (approved):
Thanks looks good! 🚀

PR #3430, opened 2026-06-11
chore: improve consistency of integrations folder structure

anakin87:
### Related Issues

I realized that some folders in integrations were not consistent with our conventions.

### Proposed Changes:
- `common` should live on the same level of `components`, not inside
- some `py.typed` files were in the wrong place
- remove some `__init__.py` files that conflicted with [namespace packages conventions](https://packaging.python.org/en/latest/guides/packaging-namespace-packages/#native-namespace-packages): I verified that this is not causing issues for regular users but could be problematic in some setups (`pip install --target <dir>`)

### How did you test it?
CI

### Notes for the reviewer
Since these are mostly internal changes, we don't need a release for now.

### Checklist


davidsbatista (approved):
looks good!

PR #3445, opened 2026-06-16
fix: serialize ChromaDocumentStore metadata in to_dict

ly-wang19:
### Related Issues

None — found while reviewing `to_dict`/`from_dict` parity across the document-store integrations.

### Proposed Changes:

`ChromaDocumentStore.to_dict()` serialized every constructor parameter **except `metadata`**, so the field was silently dropped on a `to_dict`/`from_dict` round-trip (e.g. when a pipeline is serialized to YAML with `Pipeline.dumps()` and reloaded with `Pipeline.loads()`).

`metadata` is a documented, material parameter: it is forwarded to `client.create_collection(metadata=...)` and carries `hnsw:space` plus the other HNSW index settings for the collection. Dropping it on reload means the store recreates the collection with **default** index settings instead of the ones the user configured. `distance_function` and all other init params were already serialized — `metadata` was the only omission.

The fix adds `metadata=self._metadata` to the `default_to_dict(...)` payload. `from_dict()` uses `default_from_dict`, which already forwards init parameters to `__init__`, so no change is needed there.

### How did you test it?

- Added a `to_dict`/`from_dict` round-trip regression test (`test_to_dict_and_from_dict_preserves_metadata`) asserting a non-default `metadata` (e.g. `{"hnsw:space": "cosine", "hnsw:search_ef": 200}`) survives serialization. It fails before the fix (metadata absent / `None` after reload) and passes after.
- Updated the existing exact-dict assertions (`test_document_store.py::test_to_dict` and the two retriever `test_to_dict` tests, which embed the store dict) to include the new `metadata` key.
- Full chroma unit suite: `46 passed`; `ruff check` clean.

### Notes for the reviewer

One-line behavior change in `to_dict` output (now includes `"metadata"`); the three affected exact-dict assertions are updated accordingly. No `from_dict` change required.

### Checklist

- [x] I've used one of the conventional commit types for my PR title: `fix:`


julian-risch:
@ly-wang19 Thank you for opening this issue! We need you to agree to the CLA please: https://github.com/deepset-ai/haystack-core-integrations/pull/3445#issuecomment-4714596018 Otherwise we cannot merge your PR.

julian-risch (approved):
Looks good to me! Thank you for your contribution @ly-wang19 ! I applied one change to make sure we avoid mutating ChromaDocumentStore._metadata during initialization.

PR #3455, opened 2026-06-16
fix: serialize OllamaChatGenerator think parameter

ly-wang19:
### Related Issues

None — found while reviewing `to_dict`/`from_dict` parity across the generators.

### Proposed Changes:

`OllamaChatGenerator.to_dict()` omitted `think`, so it was lost on a `to_dict`/`from_dict` round-trip. `think` is passed to the Ollama client on every run and enables a thinking model's reasoning; a generator configured with `think=True` (or `"low"`/`"medium"`/`"high"`) reverted to `False` after reload, disabling reasoning. Added it to `to_dict()`; `from_dict` already forwards it via `default_from_dict`.

### How did you test it?

Added a round-trip regression test with `think="high"` — fails before the change, passes after. Updated the existing `test_to_dict` assertion. Non-live chat-generator tests + `ruff` pass.

### Notes for the reviewer

One-line change in `to_dict`; no `from_dict` change needed.

### Checklist

### DIFF
--- a/integrations/ollama/src/haystack_integrations/components/generators/ollama/chat/chat_generator.py
+++ b/integrations/ollama/src/haystack_integrations/components/generators/ollama/chat/chat_generator.py
@@ -341,6 +341,7 @@ class OllamaChatGenerator:
             streaming_callback=callback_name,
             tools=serialize_tools_or_toolset(self.tools),
             response_format=self.response_format,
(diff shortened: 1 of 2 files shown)


bogdankostic (approved):
Thanks @ly-wang19!

PR #1547, opened 2025-03-18
feat: Pgvector - async support (+ refactoring)

anakin87:
### Related Issues

- fixes #1393

### Proposed Changes:
- introduce async support for pgvector
- non-breaking refactoring to reduce code duplication where possible

### How did you test it?
- CI (original sync test are mostly unchanged and pass)
- new tests for async

### Notes for the reviewer
Sorry for the big PR

### Checklist

- I have read the [contributors guidelines](https://github.com/deepset-ai/haystack-core-integrations/blob/main/CONTRIBUTING.md) and the [code of conduct](https://github.com/deepset-ai/haystack-core-integrations/blob/main/CODE_OF_CONDUCT.md)
- I have updated the related issue with new insights and changes
- I added unit tests and updated the docstrings
- I've used one of the [conventional commit types](https://www.conventionalcommits.org/en/v1.0.0/) for my PR title: `fix:`, `feat:`, `build:`, `chore:`, `ci:`, `docs:`, `style:`, `refactor:`, `perf:`, `test:`.

PR #1352, opened 2025-02-04
feat: add array_contains and array_overlaps.

micha-f:
### Related Issues

- proposal for #1351

### Proposed Changes:

Add `array_contains` (all items must be present in array) and `array_overlaps` (at least on item must be present in array) operators for PGVector metadata filtering.

### How did you test it?

Since `filterable_docs` did not contain arrays in the document metadata, I added tests with custom documents containing arrays of tags in their metadata.

### Checklist

- I have read the [contributors guidelines](https://github.com/deepset-ai/haystack-core-integrations/blob/main/CONTRIBUTING.md) and the [code of conduct](https://github.com/deepset-ai/haystack-core-integrations/blob/main/CODE_OF_CONDUCT.md)
- I have updated the related issue with new insights and changes
- I added unit tests and updated the docstrings
- I've used one of the [conventional commit types](https://www.conventionalcommits.org/en/v1.0.0/) for my PR title: `fix:`, `feat:`, `build:`, `chore:`, `ci:`, `docs:`, `style:`, `refactor:`, `perf:`, `test:`.

PR #3459, opened 2026-06-17
feat: add array_contains and array_overlaps operators to pgvector filters

### LINKED ISSUE #1351: Support array contains and array overlaps in pgvector metadata filtering
We have quite a few cases where we have ingested documents into with metadata containing arrays (e.g. tags, belonging to multiple customers, projects etc.).

Currently metadata filtering does not support filtering on array values.

We have added an `array_overlaps` and an `array_contains` operator for PGVector and would like to discuss whether that's something to contribute back. We have only implemented this for PGVector 

```python
# tag array must contain one of the tags
{"field": "meta.tags", "operator": "array_overlaps", "value": ["tag1", "tag3"]}

# tag array must contain all of the tags
 {"field": "meta.tags", "operator": "array_contains", "value": ["tag1", "tag2"]} #
```

Serializing the array to a comma-separated string and using the `LIKE` operator (added in https://github.com/deepset-ai/haystack-core-integrations/pull/1341). Feels brittle.

**Additional context**
The example in the documentation here might wrongly lead one to think that the `in` operator could be used for arrays:
https://docs.haystack.deepset.ai/docs/metadata-filtering#filters-usage

specifically the use of `companies` (plural) here:

`{"field": "meta.companies", "operator": "in", "value": ["BMW", "Mercedes"]}`

abhay23-AI:
### What

Adds two metadata filter operators to the pgvector document store for querying JSONB array fields, as requested in #1351:

- `array_contains`: the meta array must contain **all** of the given values
- `array_overlaps`: the meta array must share **at least one** element with the given list

```python
# docs whose meta.tags contains both tag1 AND tag2
{"field": "meta.tags", "operator": "array_contains", "value": ["tag1", "tag2"]}

# docs whose meta.tags contains tag1 OR tag3
{"field": "meta.tags", "operator": "array_overlaps", "value": ["tag1", "tag3"]}
```

### How

meta is stored as a JSONB column, so list metadata lands as a native JSON array. The existing scalar operators access meta with the `->>` (text) accessor, which can't drive the JSONB array operators, so I added a small `_treat_meta_field_as_jsonb` helper that uses the `->` (jsonb) accessor with no cast:

- `array_contains` generates `meta->'field' @> %s` with the value wrapped as `Jsonb(...)`. `jsonb @> jsonb` is true when the left array contains every element of the right one.
- `array_overlaps` generates `meta->'field' ?| %s`. `jsonb ?| text[]` is true when any of the given strings exist as top level array elements.

Field names are still embedded via `SQLLiteral`, so this stays injection safe. `array_overlaps` validates that the values are strings (the `?|` operator only matches string elements) and both operators raise `FilterError` on a non list value, matching the style of the existing `in` / `not in` operators.

### Context

This picks up the work from the earlier #1352, which got closed for inactivity. @julian-risch said back then to feel free to reopen, and since the pgvector integration was refactored in #1547 to build SQL with psycopg `Composable` objects, I redid it fresh against the current code rather than reviving the stale branch. The old undefined `Jsonb` import concern from that PR is moot now since `Jsonb` is already imported in filters.py.

### Testing

- Unit tests for SQL rendering, param wrapping, and the validation error paths (no DB needed).
- Integration tests covering contains all, contains single, overlaps, no match, and a combined AND filter, run against a real `pgvector/pgvector:pg17` container (the same image CI uses).

Full pgvector suite locally: 162 integration passed, all unit passed, `fmt-check` and `mypy` clean.

One open question for reviewers: I kept `array_overlaps` string only with a hard `FilterError` for non string values, since `?|` only matches strings and the issue's use case is tags. If you'd rather support numeric/object arrays too I can switch it to a `jsonb_array_elements` based approach, just let me know.

Resolves #1351


abhay23-AI:
hi @julian-risch, gentle nudge on this one when you get a chance. it adds the array_contains and array_overlaps pgvector operators you invited me to reopen from the old #1352, rebased and green (the py 3.10 and 3.14 checks pass). fwiw the auto labeler tagged it type:documentation but it's a feature, in case that helps triage. happy to adjust anything.

ErenAta16 (commented):
Checked the parameter wrapping and the injection-safety of the new helper against the existing conventions in this file rather than in isolation.

`_array_overlaps` returning `[value]` (the list wrapped in another list) looked odd at first glance, but it matches the exact same pattern `_in`/`_not_in` already use for `= ANY(%s)`, with the same psycopg3 list-adaptation rationale (single param that adapts to a Postgres array), so that's consistent rather than a mistake.

`_treat_meta_field_as_jsonb` uses `psycopg.sql.Literal` for the field name, same as the existing `_treat_meta_field`, and there's a dedicated test confirming an injection attempt in the field name gets escaped rather than interpolated. `array_contains`'s value goes through `Jsonb(value)` (a real bind parameter), not string formatting, so there's no path for the filter value itself to reach the query as raw SQL either.

The two operator semantics match real Postgres `jsonb` behavior: `@>` (containment) requires every element of the right-hand array to be present in the left, and `?|` (existence, any) requires at least one string element to match, both to reject non-string elements up front for `array_overlaps` (since `?|` only matches string elements, so anything else would silently return zero rows instead of erroring) is a good call, saves someone a confusing "why does this filter always return nothing" debugging session.

Tests cover both operators, both error paths (non-list value, non-string list elements for overlaps), and a combined AND filter exercising both together against a real document store. Looks correct and complete.

julian-risch (approved):
Looks good to me! @abhay23-AI I just did some minor changes before merging. Thank you for your contribution and congratulations on another merged Haystack PR!

PR #3463, opened 2026-06-18
feat!: migrate to official PaddleOCRClient, drop paddlex dependency

alialiaaa0712-cell:
### Proposed Changes

Migrates the `PaddleOCR-VL` integration from a hand-rolled `requests`-based HTTP client and `paddlex[serving]` schema validation to the official `paddleocr` SDK's `PaddleOCRClient`.

- Replaces `api_url` with `base_url` (optional, falls back to `PADDLEOCR_BASE_URL` env var, then SDK default)
- Renames auth env var from `AISTUDIO_ACCESS_TOKEN` to `PADDLEOCR_ACCESS_TOKEN`
- Adds `model` parameter (default `Model.PADDLE_OCR_VL_16`) for selecting the document parsing model
- Removes `paddlex[serving]` and `requests` dependencies; bumps `paddleocr>=3.7.0`
- All inference parameters now passed via `PaddleOCRVLOptions`; `additional_params` mapped to `extra_options`
- Fixes `to_dict`/`from_dict` round-trip for `file_type` (integers 0/1 now accepted alongside `"pdf"`/`"image"`)

### How did you test it?

Unit tests (56) covering: init validation, serialization round-trip with `model` and `base_url`, all source types (str path, Path, ByteStream), multi-source processing, partial failure handling, unknown extension skipping, empty text, model/options forwarding, meta (single dict and per-source list), temp file cleanup, base URL and token forwarding to the SDK client.

Manual live verification against the PaddleOCR AI Studio API with a real PDF and image file.

### Notes for the reviewer

`additional_params` is now passed as `extra_options` to `PaddleOCRVLOptions` rather than merged into the top-level request dict — existing users relying on this for top-level API params will need to update.

### Checklist

### DIFF
--- a/.github/workflows/paddleocr.yml
+++ b/.github/workflows/paddleocr.yml
@@ -29,8 +29,8 @@ concurrency:
 env:
   PYTHONUNBUFFERED: "1"
   FORCE_COLOR: "1"
-  AISTUDIO_ACCESS_TOKEN: ${{ secrets.AISTUDIO_ACCESS_TOKEN }}
-  PADDLEOCR_VL_API_URL: ${{ secrets.PADDLEOCR_VL_API_URL }}
+  PADDLEOCR_ACCESS_TOKEN: ${{ secrets.PADDLEOCR_ACCESS_TOKEN }}
+  PADDLEOCR_VL_BASE_URL: ${{ secrets.PADDLEOCR_VL_BASE_URL }}
   TEST_MATRIX_OS: '["ubuntu-latest", "windows-latest", "macos-latest"]'
   TEST_MATRIX_PYTHON: '["3.10", "3.12"]'
 
--- a/.github/workflows/paddleocr.yml
+++ b/.github/workflows/paddleocr.yml
@@ -30,7 +30,7 @@ env:
   PYTHONUNBUFFERED: "1"
   FORCE_COLOR: "1"
   PADDLEOCR_ACCESS_TOKEN: ${{ secrets.PADDLEOCR_ACCESS_TOKEN }}
-  PADDLEOCR_VL_BASE_URL: ${{ secrets.PADDLEOCR_VL_BASE_URL }}
+  PADDLEOCR_BASE_URL: ${{ secrets.PADDLEOCR_BASE_URL }}
   TEST_MATRIX_OS: '["ubuntu-latest", "windows-latest", "macos-latest"]'
   TEST_MATRIX_PYTHON: '["3.10", "3.12"]'
 
--- a/integrations/paddleocr/src/haystack_integrations/components/converters/paddleocr/paddleocr_vl_document_converter.py
+++ b/integrations/paddleocr/src/haystack_integrations/components/converters/paddleocr/paddleocr_vl_document_converter.py
@@ -112,11 +112,11 @@ class PaddleOCRVLDocumentConverter:
         self,
         *,
         base_url: str | None = None,
-        access_token: Secret = Secret.from_env_var("PADDLEOCR_ACCESS_TOKEN"),
+        access_token: Secret = Secret.from_env_var(["PADDLEOCR_ACCESS_TOKEN", "AISTUDIO_ACCESS_TOKEN"]),
         model: Model | str = Model.PADDLE_OCR_VL_16,
         file_type: FileTypeInput = None,
-        use_doc_orientation_classify: bool | None = None,
-        use_doc_unwarping: bool | None = None,
+        use_doc_orientation_classify: bool | None = False,
+        use_doc_unwarping: bool | None = False,
         use_layout_detection: bool | None = None,
         use_chart_recognition: bool | None = None,
         use_seal_recognition: bool | None = None,
@@ -305,6 +305,12 @@ class PaddleOCRVLDocumentConverter:
             Deserialized component.
         """
         deserialize_secrets_inplace(data["init_parameters"], keys=["access_token"])
+        init_params = data["init_parameters"]
+        if "model" in init_params and isinstance(init_params["model"], str):
+            try:
+                init_params["model"] = Model(init_params["model"])
+            except ValueError:
+                pass
         return default_from_dict(cls, data)
 
     def _build_options(self) -> PaddleOCRVLOptions:
@@ -396,7 +402,7 @@ class PaddleOCRVLDocumentConverter:
                 try:
                     bytestream = get_bytestream_from_source(source)
                 except Exception as e:
-                    logger.warning(f"Could not read {source}. Skipping it. Error: {e}")
+                    logger.warning("Could not read {source}. Skipping it. Error: {error}", source=source, error=e)
                     continue
 
                 if self.file_type is not None:
@@ -406,13 +412,13 @@ class PaddleOCRVLDocumentConverter:
                     file_type = _infer_file_type_from_source(source, mime_type)
 
                 if file_type is None:
-                    logger.warning(f"Could not determine file type for {source}. Skipping it.")
+                    logger.warning("Could not determine file type for {source}. Skipping it.", source=source)
                     continue
 
                 try:
                     text, raw_resp = self._parse(bytestream.data, file_type, client)
                 except Exception as e:
-                    logger.warning(f"Could not convert {source} to Document, skipping. Error: {e}")
+                    logger.warning("Could not convert {source} to Document, skipping. Error: {error}", source=source, error=e)
                     continue
 
                 if not text:
--- a/integrations/paddleocr/src/haystack_integrations/components/converters/paddleocr/paddleocr_vl_document_converter.py
+++ b/integrations/paddleocr/src/haystack_integrations/components/converters/paddleocr/paddleocr_vl_document_converter.py
@@ -427,7 +427,9 @@ class PaddleOCRVLDocumentConverter:
 
                 if not text:
                     logger.warning(
-                        f"{self.__class__.__name__} could not extract text from {source}. Returning an empty document."
+                        f"{cls} could not extract text from {source}. Returning an empty document.",
+                        cls=self.__class__.__name__,
+                        source=source,
                     )
 
                 merged_metadata = {**bytestream.meta, **metadata}
--- a/integrations/paddleocr/src/haystack_integrations/components/converters/paddleocr/paddleocr_vl_document_converter.py
+++ b/integrations/paddleocr/src/haystack_integrations/components/converters/paddleocr/paddleocr_vl_document_converter.py
@@ -1,11 +1,11 @@
 # SPDX-FileCopyrightText: 2025-present deepset GmbH <info@deepset.ai>
 #
 # SPDX-License-Identifier: Apache-2.0
-import base64
+import asyncio
+import tempfile
 from pathlib import Path
 from typing import Any, Literal
 
-import requests
 from haystack import Document, component, default_from_dict, default_to_dict, logging
 from haystack.components.converters.utils import (
     get_bytestream_from_source,
@@ -13,37 +13,23 @@ from haystack.components.converters.utils import (
 )
 from haystack.dataclasses import ByteStream
 from haystack.utils import Secret, deserialize_secrets_inplace
-from paddlex.inference.serving.schemas.paddleocr_vl import (  # type: ignore[import-untyped]
-    InferRequest as PaddleOCRVLInferRequest,
-)
-from paddlex.inference.serving.schemas.paddleocr_vl import (  # type: ignore[import-untyped]
-    InferResult as PaddleOCRVLInferResult,
-)
-from paddlex.inference.serving.schemas.shared.ocr import FileType  # type: ignore[import-untyped]
+from paddleocr import AsyncPaddleOCRClient, Model, PaddleOCRVLOptions  # type: ignore[import-untyped]
 
 logger = logging.getLogger(__name__)
 
 
 FileTypeInput = Literal["pdf", "image"] | None
 
-# Supported image file extensions
-_IMAGE_EXTENSIONS = {
-    ".jpg",
-    ".jpeg",
-    ".png",
-    ".bmp",
-    ".tiff",
-    ".tif",
-    ".webp",
-}
-# Supported PDF file extensions
+_IMAGE_EXTENSIONS = {".jpg", ".jpeg", ".png", ".bmp", ".tiff", ".tif", ".webp"}
 _PDF_EXTENSIONS = {".pdf"}
+_FILE_TYPE_TO_STR = {0: "pdf", 1: "image"}
+_EXTENSION_FOR_FILE_TYPE = {0: ".pdf", 1: ".jpg"}
 
 
 def _infer_file_type_from_source(
     source: str | Path | ByteStream,
     mime_type: str | None = None,
-) -> FileType | None:
+) -> int | None:
     """
     Infer file type from file extension or MIME type.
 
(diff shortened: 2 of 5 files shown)


sjrl:
Hey @alialiaaa0712-cell thanks for opening a PR! Once the CI is green feel free to mark it ready for review and someone from the team will take a look at it.

alialiaaa0712-cell:
> Hey @alialiaaa0712-cell thanks for opening a PR! Once the CI is green feel free to mark it ready for review and someone from the team will take a look at it.

Thanks!  CI passed 🙌

alialiaaa0712-cell:
Hi! One heads-up on this PR: the internal implementation has switched from `AsyncPaddleOCRClient` + `asyncio.run()` to the synchronous `PaddleOCRClient`.

Since `run()` is a synchronous method, using the sync client is a more natural fit — it avoids the overhead of spinning up an event loop and eliminates edge cases around `asyncio.run()` failing when called from within an already-running loop (e.g. Jupyter, FastAPI).

The public interface (`run()` method signature, inputs, outputs) is unchanged, so existing pipeline code requires no modification. The behavioral difference is that multiple sources are now processed sequentially within a single HTTP session rather than concurrently — if you were relying on parallel processing for large batches, throughput may be lower.

Let me know if you have concerns or prefer to keep the async approach.

integrations/paddleocr/src/haystack_integrations/components/converters/paddleocr/paddleocr_vl_document_converter.py
```diff
@@ -346,146 +307,60 @@ def from_dict(cls, data: dict[str, Any]) -> "PaddleOCRVLDocumentConverter":
...
-            request_data["relevelTitles"] = self.relevel_titles
-        if self.visualize is not None:
-            request_data["visualize"] = self.visualize
-        if self.additional_params is not None:
-            request_data.update(self.additional_params)
-
-        # Validate input parameters
+    def _parse(self, data: bytes, file_type: int, client: PaddleOCRClient) -> tuple[str, dict[str, Any]]:
+        extension = _EXTENSION_FOR_FILE_TYPE[file_type]
+        with tempfile.NamedTemporaryFile(suffix=extension, delete=False) as tmp:
```
sjrl: Just to confirm that it's fine we are always either writing files with extension of pdf or jpg even if the image type is actually something like png?

integrations/paddleocr/src/haystack_integrations/components/converters/paddleocr/paddleocr_vl_document_converter.py
```diff
@@ -142,18 +111,19 @@ class PaddleOCRVLDocumentConverter:
     def __init__(
         self,
         *,
-        api_url: str,
-        access_token: Secret = Secret.from_env_var("AISTUDIO_ACCESS_TOKEN"),
+        base_url: str | None = None,
+        access_token: Secret = Secret.from_env_var("PADDLEOCR_ACCESS_TOKEN"),
```
sjrl: Do we really need to also change the default secret name for the access token? If we keep it the same it will make migration slightly easier for users.

sjrl: If it's the default value name used by the new SDK then its probably fine to change.

integrations/paddleocr/src/haystack_integrations/components/converters/paddleocr/paddleocr_vl_document_converter.py
```diff
@@ -142,18 +111,19 @@ class PaddleOCRVLDocumentConverter:
...
+        model: Model | str = Model.PADDLE_OCR_VL_16,
         file_type: FileTypeInput = None,
-        use_doc_orientation_classify: bool | None = False,
-        use_doc_unwarping: bool | None = False,
+        use_doc_orientation_classify: bool | None = None,
+        use_doc_unwarping: bool | None = None,
```
sjrl: Does switching from False to None here change the default behavior?

sjrl:
@alialiaaa0712-cell thanks for opening the PR! I've updated your GH title to include `feat!:` which is our way of indicating this PR will have breaking changes. 

Also some other general feedback on your changes:
- The https://github.com/deepset-ai/haystack-core-integrations/blob/main/integrations/paddleocr/README.md is out of date since it references old api keys
- Please update these lines in the CI workflow https://github.com/deepset-ai/haystack-core-integrations/blob/3fc27f67723a66081b47fe19e4c3d8f3250eccea/.github/workflows/paddleocr.yml#L32-L33 to the new environment variables.

integrations/paddleocr/tests/test_paddleocr_vl_document_converter.py
```diff
@@ -20,1022 +21,528 @@
...
-        assert converter.prompt_label is None
-        assert converter.format_block_content is None
-        assert converter.repetition_penalty is None
-        assert converter.temperature is None
-        assert converter.top_p is None
-        assert converter.min_pixels is None
-        assert converter.max_pixels is None
-        assert converter.prettify_markdown is None
-        assert converter.show_formula_number is None
-        assert converter.visualize is None
```
sjrl, line 65: Could we add these back? I don't see why we need to delete them.

integrations/paddleocr/tests/test_paddleocr_vl_document_converter.py
```diff
@@ -20,1022 +21,528 @@
...
+    "haystack_integrations.components.converters.paddleocr.paddleocr_vl_document_converter.PaddleOCRVLDocumentConverter"
+)
+
+_PATCH_CLIENT = "haystack_integrations.components.converters.paddleocr.paddleocr_vl_document_converter.PaddleOCRClient"
 
-    def test_init_default(self, monkeypatch):
-        monkeypatch.setenv("AISTUDIO_ACCESS_TOKEN", "test-access-token")
-        converter = PaddleOCRVLDocumentConverter(api_url="http://test-api-url.com")
 
-        assert converter.access_token == Secret.from_env_var("AISTUDIO_ACCESS_TOKEN")
```
sjrl, line 45: Same comment as here https://github.com/deepset-ai/haystack-core-integrations/pull/3463/changes#r3457854159

integrations/paddleocr/tests/test_paddleocr_vl_document_converter.py
```diff
@@ -20,1022 +21,528 @@
...
+
+        assert converter.base_url is None
+        assert converter.model == Model.PADDLE_OCR_VL_16
         assert converter.file_type is None
-        assert converter.use_doc_orientation_classify is False
-        assert converter.use_doc_unwarping is False
+        assert converter.use_doc_orientation_classify is None
+        assert converter.use_doc_unwarping is None
         assert converter.use_layout_detection is None
-        assert converter.use_chart_recognition is None
```
sjrl, line 51: Same comment as here https://github.com/deepset-ai/haystack-core-integrations/pull/3463/changes#r3457854159

integrations/paddleocr/src/haystack_integrations/components/converters/paddleocr/paddleocr_vl_document_converter.py
```diff
@@ -499,65 +374,54 @@ def run(
...
+        kwargs: dict[str, Any] = {"client_platform": "haystack", "token": token}
+        if self.base_url is not None:
+            kwargs["base_url"] = self.base_url
+
+        with PaddleOCRClient(**kwargs) as client:
+            for source, metadata in zip(sources, meta_list, strict=True):
+                try:
+                    bytestream = get_bytestream_from_source(source)
+                except Exception as e:
+                    logger.warning(f"Could not read {source}. Skipping it. Error: {e}")
```
sjrl: We avoid using f-strings in our logging statements. We have a specially patched logger, please do 
```suggestion
                    logger.warning("Could not read {source}. Skipping it. Error: {error}", source=source, error=e)
```

integrations/paddleocr/src/haystack_integrations/components/converters/paddleocr/paddleocr_vl_document_converter.py
```diff
@@ -499,65 +374,54 @@ def run(
...
+                    continue
+
+                if self.file_type is not None:
+                    file_type: int | None = self.file_type
+                else:
+                    mime_type = bytestream.mime_type if bytestream.mime_type else None
+                    file_type = _infer_file_type_from_source(source, mime_type)
+
+                if file_type is None:
+                    logger.warning(f"Could not determine file type for {source}. Skipping it.")
```
sjrl:
```suggestion
                    logger.warning("Could not determine file type for {source}. Skipping it.", source=source)
```

integrations/paddleocr/src/haystack_integrations/components/converters/paddleocr/paddleocr_vl_document_converter.py
```diff
@@ -499,65 +374,54 @@ def run(
...
+                    continue
+
+                try:
+                    text, raw_resp = self._parse(bytestream.data, file_type, client)
+                except Exception as e:
+                    logger.warning(f"Could not convert {source} to Document, skipping. Error: {e}")
```
sjrl:
```suggestion
                    logger.warning("Could not convert {source} to Document, skipping. Error: {error}", source=source, error=e)
```

integrations/paddleocr/src/haystack_integrations/components/converters/paddleocr/paddleocr_vl_document_converter.py
```diff
@@ -499,65 +374,54 @@ def run(
...
+                    continue
+
+                if not text:
+                    logger.warning(
+                        f"{self.__class__.__name__} could not extract text from {source}. Returning an empty document."
+                    )
```
sjrl, line 442:
```suggestion
                    logger.warning(
                        f"{cls} could not extract text from {source}. Returning an empty document.",
                        cls=self.__class__.__name__,
                        source=source,
                    )
```

integrations/paddleocr/src/haystack_integrations/components/converters/paddleocr/paddleocr_vl_document_converter.py
```diff
@@ -346,146 +307,60 @@ def from_dict(cls, data: dict[str, Any]) -> "PaddleOCRVLDocumentConverter":
         deserialize_secrets_inplace(data["init_parameters"], keys=["access_token"])
         return default_from_dict(cls, data)
```
sjrl, line 314: Now that the enum `Model` has been added it would be better if we reconstructed the enum here. Otherwise a roundtrip of to_dict and from_dict ends up with the Model enum being converted from a enum to a string.

integrations/paddleocr/tests/test_paddleocr_vl_document_converter.py
```diff
@@ -20,1022 +21,528 @@
...
-
-        result = converter.run(sources=source_files)
+    def test_run_with_files(
+        self,
+        test_files_path: Path,
+        source_files: list[str],
+        expected_docs: int,
+    ) -> None:
+        converter = PaddleOCRVLDocumentConverter(
+            base_url=os.environ["PADDLEOCR_VL_BASE_URL"],
```
sjrl: Just to double check is this line needed? You mention in the docstrings that the default value for `base_url` is auto-detected from `PADDLEOCR_VL_BASE_URL`. If so lets drop it from here.

integrations/paddleocr/src/haystack_integrations/components/converters/paddleocr/paddleocr_vl_document_converter.py
alialiaaa0712-cell: The _EXTENSION_FOR_FILE_TYPE map uses .jpg for all image types. The PaddleOCR API determines file type via the fileType field (0 = PDF, 1 = image), not the file extension — the extension is only used to create the temp file so the SDK recognizes it as a regular file path. We verified this against the live API with PNG content written to a .jpg-suffixed temp file and it parses correctly. Let me know if you want it to be exactly match.

integrations/paddleocr/src/haystack_integrations/components/converters/paddleocr/paddleocr_vl_document_converter.py
alialiaaa0712-cell: Changed to Secret.from_env_var(["PADDLEOCR_ACCESS_TOKEN", "AISTUDIO_ACCESS_TOKEN"]) — the new name takes priority, but the old AISTUDIO_ACCESS_TOKEN is tried as a fallback, so existing deployments continue to work without changes.

integrations/paddleocr/src/haystack_integrations/components/converters/paddleocr/paddleocr_vl_document_converter.py
alialiaaa0712-cell: Keeping the defaults as False (matching the original implementation). The SDK's PaddleOCRVLOptions defaults these to None (not sent to the API), but the original code explicitly passed False. These are functionally equivalent for the API, but keeping False avoids a behavior change for existing users.

alialiaaa0712-cell:
Updated README.md and .github/workflows/paddleocr.yml to replace AISTUDIO_ACCESS_TOKEN / PADDLEOCR_VL_API_URL with PADDLEOCR_ACCESS_TOKEN / PADDLEOCR_VL_BASE_URL. Note: the GitHub repo secrets will also need to be updated accordingly.

> @alialiaaa0712-cell thanks for opening the PR! I've updated your GH title to include `feat!:` which is our way of indicating this PR will have breaking changes.
> 
> Also some other general feedback on your changes:
> 
> * The https://github.com/deepset-ai/haystack-core-integrations/blob/main/integrations/paddleocr/README.md is out of date since it references old api keys
> * Please update these lines in the CI workflow https://github.com/deepset-ai/haystack-core-integrations/blob/3fc27f67723a66081b47fe19e4c3d8f3250eccea/.github/workflows/paddleocr.yml#L32-L33
>    to the new environment variables.

integrations/paddleocr/tests/test_paddleocr_vl_document_converter.py
alialiaaa0712-cell: Done.

integrations/paddleocr/tests/test_paddleocr_vl_document_converter.py
alialiaaa0712-cell: Same as above.

integrations/paddleocr/tests/test_paddleocr_vl_document_converter.py
alialiaaa0712-cell: Done.

integrations/paddleocr/src/haystack_integrations/components/converters/paddleocr/paddleocr_vl_document_converter.py
alialiaaa0712-cell: Done.

integrations/paddleocr/src/haystack_integrations/components/converters/paddleocr/paddleocr_vl_document_converter.py
alialiaaa0712-cell: Done.

integrations/paddleocr/src/haystack_integrations/components/converters/paddleocr/paddleocr_vl_document_converter.py
alialiaaa0712-cell: Done.

integrations/paddleocr/tests/test_paddleocr_vl_document_converter.py
alialiaaa0712-cell: Done, removed. The SDK automatically picks up PADDLEOCR_BASE_URL from the environment, so passing it explicitly was redundant.

integrations/paddleocr/src/haystack_integrations/components/converters/paddleocr/paddleocr_vl_document_converter.py
alialiaaa0712-cell: Done

integrations/paddleocr/src/haystack_integrations/components/converters/paddleocr/paddleocr_vl_document_converter.py
```diff
@@ -499,65 +380,60 @@ def run(
...
+                    logger.warning(
+                        "Could not convert {source} to Document, skipping. Error: {error}",
+                        source=source,
+                        error=e,
+                    )
+                    continue
+
+                if not text:
+                    logger.warning(
+                        f"{cls} could not extract text from {source}. Returning an empty document.",
```
sjrl: sorry made a typo in my original comment. should let the linting pass
```suggestion
                        "{cls} could not extract text from {source}. Returning an empty document.",
```

alialiaaa0712-cell:
PTAL

integrations/paddleocr/src/haystack_integrations/components/converters/paddleocr/paddleocr_vl_document_converter.py
sjrl: No all good thanks for the explanation.

.github/workflows/paddleocr.yml
```diff
@@ -29,8 +29,8 @@ concurrency:
 env:
   PYTHONUNBUFFERED: "1"
   FORCE_COLOR: "1"
-  AISTUDIO_ACCESS_TOKEN: ${{ secrets.AISTUDIO_ACCESS_TOKEN }}
-  PADDLEOCR_VL_API_URL: ${{ secrets.PADDLEOCR_VL_API_URL }}
+  PADDLEOCR_ACCESS_TOKEN: ${{ secrets.PADDLEOCR_ACCESS_TOKEN }}
+  PADDLEOCR_VL_BASE_URL: ${{ secrets.PADDLEOCR_VL_BASE_URL }}
```
sjrl: Hey @alialiaaa0712-cell I noticed an inconsistency between our integration and the PaddleOCR SDK. 

It seems the SDK expects the env variable `PADDLEOCR_BASE_URL` to be set (so no `VL` in the name). I think we should also use the same one, so could we update all instances of `PADDLEOCR_VL_BASE_URL` to `PADDLEOCR_BASE_URL`?

alialiaaa0712-cell: Good catch! Fixed — updated all instances of PADDLEOCR_VL_BASE_URL to PADDLEOCR_BASE_URL in the CI workflow, README, and tests to align with the SDK convention.

sjrl (approved):
Thanks for the contribution!

PR #3477, opened 2026-06-19
fix(amazon-bedrock): forward dimensions/normalize for Titan V2 embedders

### LINKED ISSUE #3476: AmazonBedrockTextEmbedder/AmazonBedrockDocumentEmbedder silently ignore Titan V2 `dimensions` and `normalize`
First, thanks for maintaining the Amazon Bedrock integration — the Cohere/Titan embedder split is clean and easy to follow.

### Describe the bug

`AmazonBedrockTextEmbedder` and `AmazonBedrockDocumentEmbedder` accept arbitrary `**kwargs` and the docstring states they are "Additional parameters to pass for model inference". For Cohere models the embedders forward `input_type` and `truncate` from those kwargs into the request body. For Amazon Titan models, the request body is built as `{"inputText": text}` only — so `dimensions` and `normalize` are silently dropped.

Amazon Titan Text Embeddings **V2** (`amazon.titan-embed-text-v2:0`) supports `dimensions` (1024 default, 512, 256) and `normalize` (default `true`) in the request body. Because the integration never forwards them, a user who configures e.g. `dimensions=512`/`normalize=False` always gets the default 1024-dim, normalized vector with no error and no warning.

Source (`main`):
- `text_embedder.py`, Titan branch in `run`: `body = {"inputText": text}`
- `document_embedder.py`, `_embed_titan`: `body = {"inputText": text}`

### To reproduce

```python
from unittest.mock import patch
import io
from haystack_integrations.components.embedders.amazon_bedrock import AmazonBedrockTextEmbedder

embedder = AmazonBedrockTextEmbedder(model="amazon.titan-embed-text-v2:0", dimensions=512, normalize=False)
with patch.object(embedder._client, "invoke_model") as m:
    m.return_value = {"body": io.StringIO('{"embedding": [0.1]}')}
    embedder.run(text="hello")
    print(m.call_args.kwargs["body"])
# main:     {"inputText": "hello"}            <- dimensions/normalize dropped
# expected: {"inputText": "hello", "dimensions": 512, "normalize": false}
```

### Expected behavior

`dimensions`/`normalize` passed via kwargs are forwarded to the request body for Titan **V2**, matching how `input_type`/`truncate` are already forwarded for Cohere. Titan G1 (`amazon.titan-embed-text-v1`) does not support inference parameters, so they must not be sent for v1.

### AWS reference

AWS Bedrock docs, "Amazon Titan Embeddings — Text" request/response:
https://docs.aws.amazon.com/bedrock/latest/userguide/model-parameters-titan-embed-text.html

- **V2 Request:** "The inputText parameter is required. The normalize and dimensions parameters are optional. … dimensions — accepted values: 1024 (default), 512, 256. … normalize — Defaults to true."
- **G1 Request:** "The only available field is `inputText`."

### Versions

- haystack-core-integrations `main`
- integration: `amazon_bedrock`

I'm happy to open a PR fixing this following the existing Cohere kwargs pattern.

kimnamu:
Thanks for the clean Bedrock embedder design — this PR fills in a gap in the existing kwargs-forwarding convention rather than adding a new behavior.

### Related Issues

- Fixes #3476

### Proposed Changes

The `AmazonBedrockTextEmbedder` and `AmazonBedrockDocumentEmbedder` Titan branches built the request body as `{"inputText": text}` only, so `dimensions`/`normalize` passed via `**kwargs` were silently dropped. The Cohere branch already forwards `input_type`/`truncate` from `self.kwargs` into the body; this PR does the same for Titan's `dimensions`/`normalize`, **gated on Titan V2** (`"v2" in self.model`).

Amazon Titan G1 (`amazon.titan-embed-text-v1`) does not support inference parameters — only `inputText` — so the gate ensures v1 requests are unchanged. The docstring already promised "Additional parameters to pass for model inference", so this matches the documented intent; it has been extended to name `dimensions`/`normalize` for Titan V2.

AWS reference: https://docs.aws.amazon.com/bedrock/latest/userguide/model-parameters-titan-embed-text.html — V2 request supports optional `dimensions` (1024 default / 512 / 256) and `normalize` (default `true`); G1 request supports only `inputText`.

### Before / After

| Case | Before | After |
|------|--------|-------|
| Titan **v2** + `dimensions=512, normalize=False` | ❌ body `{"inputText": "..."}` (params dropped, always 1024-dim normalized) | ✅ body `{"inputText": "...", "dimensions": 512, "normalize": false}` |
| Titan **v2**, no extra params | `{"inputText": "..."}` | ✅ `{"inputText": "..."}` (unchanged) |
| Titan **v1** + `dimensions`/`normalize` | `{"inputText": "..."}` | ✅ `{"inputText": "..."}` (v1 has no inference params; not sent) |
| Cohere `input_type`/`truncate` | ✅ forwarded | ✅ forwarded (unchanged) |
| Public API / constructor args / output keys | — | ✅ Unchanged |
| Titan batching (one-by-one), Cohere batching | — | ✅ Unchanged |

### How did you test it?

Added mock-based unit tests (no AWS keys needed) to both `test_text_embedder.py` and `test_document_embedder.py`, asserting the exact request `body`: v2 forwards the params, v2 without params is unchanged, and v1 ignores them.

To prove the new tests catch the bug, I reverted only the source (kept the new tests) and re-ran them — the two "v2 forwards params" tests fail:

```
E       assert '{"inputText": "some text"}' == '{"inputText"...lize": false}'
E         - {"inputText": "some text", "dimensions": 512, "normalize": false}
E         + {"inputText": "some text"}
FAILED tests/test_text_embedder.py::...::test_titan_v2_invocation_with_dimensions_and_normalize
FAILED tests/test_document_embedder.py::...::test_embed_titan_v2_with_dimensions_and_normalize
================== 2 failed, 4 passed, 35 deselected ==================
```

Restoring the source, all pass, plus types and format:

```
hatch run test:unit  -> 35 passed, 6 deselected
hatch run test:types -> Success: no issues found in 20 source files
hatch run fmt        -> All checks passed! / 37 files left unchanged
```

(run from `integrations/amazon_bedrock`)

### Notes for the reviewer

- Minimal diff: the Titan branch reuses the existing Cohere `self.kwargs.get(...)` idiom; no new params on `__init__`, no public API change.
- A more complete follow-up could also expose `embeddingTypes` (binary/float) for V2, but this PR is scoped to the reported `dimensions`/`normalize` drop.

### Checklist

- [x] I've used a conventional commit type for my PR title: `fix:`

---

*This contribution was prepared with the help of an AI agent (Claude Code); a human reviewed the change, the AWS-documented rationale, and the test results before submitting.*

### DIFF
--- a/integrations/amazon_bedrock/src/haystack_integrations/components/embedders/amazon_bedrock/document_embedder.py
+++ b/integrations/amazon_bedrock/src/haystack_integrations/components/embedders/amazon_bedrock/document_embedder.py
@@ -101,7 +101,7 @@ class AmazonBedrockDocumentEmbedder:
             Can be used to tune [retry behavior](https://docs.aws.amazon.com/boto3/latest/guide/retries.html)
             and other low-level settings like timeouts and connection management.
         :param kwargs: Additional parameters to pass for model inference. For example, `input_type` and `truncate` for
-            Cohere models.
+            Cohere models, or `dimensions` and `normalize` for Amazon Titan Text Embeddings V2.
         :raises ValueError: If the model is not supported.
         :raises AmazonBedrockConfigurationError: If the AWS environment is not configured correctly.
         """
@@ -209,9 +209,17 @@ class AmazonBedrockDocumentEmbedder:
 
         texts_to_embed = self._prepare_texts_to_embed(documents=documents)
 
+        titan_body: dict[str, Any] = {}
+        if "v2" in self.model:
+            # `dimensions` and `normalize` are only supported by Amazon Titan Text Embeddings V2
+            if (dimensions := self.kwargs.get("dimensions")) is not None:
+                titan_body["dimensions"] = dimensions
+            if (normalize := self.kwargs.get("normalize")) is not None:
+                titan_body["normalize"] = normalize
+
         all_embeddings = []
         for text in tqdm(texts_to_embed, disable=not self.progress_bar, desc="Creating embeddings"):
-            body = {"inputText": text}
+            body = {"inputText": text, **titan_body}
             try:
                 response = self._client.invoke_model(
                     body=json.dumps(body), modelId=self.model, accept="*/*", contentType="application/json"
--- a/integrations/amazon_bedrock/src/haystack_integrations/components/embedders/amazon_bedrock/document_embedder.py
+++ b/integrations/amazon_bedrock/src/haystack_integrations/components/embedders/amazon_bedrock/document_embedder.py
@@ -210,7 +210,7 @@ class AmazonBedrockDocumentEmbedder:
         texts_to_embed = self._prepare_texts_to_embed(documents=documents)
 
         titan_body: dict[str, Any] = {}
-        if "v2" in self.model:
+        if self.model.startswith("amazon.titan-embed-text-v2"):
             # `dimensions` and `normalize` are only supported by Amazon Titan Text Embeddings V2
             if (dimensions := self.kwargs.get("dimensions")) is not None:
                 titan_body["dimensions"] = dimensions
--- a/integrations/amazon_bedrock/src/haystack_integrations/components/embedders/amazon_bedrock/text_embedder.py
+++ b/integrations/amazon_bedrock/src/haystack_integrations/components/embedders/amazon_bedrock/text_embedder.py
@@ -82,7 +82,7 @@ class AmazonBedrockTextEmbedder:
             Can be used to tune [retry behavior](https://docs.aws.amazon.com/boto3/latest/guide/retries.html)
             and other low-level settings like timeouts and connection management.
         :param kwargs: Additional parameters to pass for model inference. For example, `input_type` and `truncate` for
-            Cohere models.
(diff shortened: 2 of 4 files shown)


integrations/amazon_bedrock/src/haystack_integrations/components/embedders/amazon_bedrock/document_embedder.py
```diff
@@ -209,9 +209,17 @@ def _embed_titan(self, documents: list[Document]) -> list[Document]:
 
         texts_to_embed = self._prepare_texts_to_embed(documents=documents)
 
+        titan_body: dict[str, Any] = {}
+        if "v2" in self.model:
```
bogdankostic: Not sure if checking only for `v2` in the model identifier is too broad, I think `v2` is likely to be also present in other model identifiers apart from Amazon Titan Text Embeddings V2.

bogdankostic (commented):
Thanks @kimnamu! Looks already good, just left a minor comment.

bogdankostic (approved):
Looks good, thanks @kimnamu!

PR #3483, opened 2026-06-22
fix: NvidiaRanker missing /ranking path for local NIM deployments

### LINKED ISSUE #1457: Nvidia Ranker URL issue

SyedShahmeerAli12:
## Summary

Closes #1457 - `NvidiaRanker` sends ranking requests to the wrong endpoint for self-hosted/local NIM deployments.

**The bug:**
`NimBackend.rank()` used `self.api_url` directly, while `embed()`, `generate()`, and `models()` all append their own endpoint suffix (`/embeddings`, `/chat/completions`, `/models`). Only `rank()` was missing its suffix.

This went unnoticed in the existing unit test because hosted ranking models get their `api_url` fully overridden by `validate_hosted_model()` to a model-specific endpoint that already includes the complete path (e.g. `.../nv-rerankqa-mistral-4b-v3/reranking`). For self-hosted or local NIM ranking containers - where no model-specific override applies - `self.api_url` stays as the bare base URL (e.g. `http://localhost:8000/v1`), so requests were sent to the wrong endpoint instead of `{api_url}/ranking`.

**The fix:**
Track whether the hosted override already supplied a full ranking path (`_has_custom_ranking_endpoint`). Only append `/ranking` when it did not, so both hosted models (unchanged behavior) and local/self-hosted NIM ranking deployments (the actual bug) hit the correct endpoint.

**Tests:**
- Added `test_rank_local_nim_appends_ranking_path` exercising the non-hosted path that the original bug affected
- Existing `test_rank` (hosted model with custom endpoint override) still passes unchanged
- All 204 unit tests pass
- `mypy` and `ruff` clean

### DIFF
--- a/integrations/nvidia/src/haystack_integrations/utils/nvidia/nim_backend.py
+++ b/integrations/nvidia/src/haystack_integrations/utils/nvidia/nim_backend.py
@@ -42,6 +42,11 @@ class NimBackend:
         self.session.headers.update(headers)
 
         self.api_url = api_url
+        # Hosted ranking models are assigned a per-model endpoint that already includes the
+        # full path (e.g. ".../nv-rerankqa-mistral-4b-v3/reranking"), unlike other hosted
+        # endpoints which only override the host. `rank()` uses this to decide whether it
+        # still needs to append the `/ranking` path itself.
+        self._has_custom_ranking_endpoint = False
         if isinstance(client, str):
             client = Client.from_str(client)
         validated_model: Model | None = None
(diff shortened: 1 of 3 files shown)


julian-risch (approved):
Looks good to me! Thank you for another contribution @SyedShahmeerAli12 !

PR #3485, opened 2026-06-22
feat(microsoft_sharepoint): Add MSSharePointFetcher

bogdankostic:
### Related Issues

- fixes https://github.com/deepset-ai/haystack-private/issues/424

### Proposed Changes:

Adds a content fetcher, `MSSharePointFetcher`, to the `microsoft-sharepoint-haystack` integration. It complements `MSSharePointRetriever` (which only returns Search snippets + metadata) by fetching the **full content** of each hit. Wire the retriever's `documents` (or a list of `web_url`s) plus a per-user delegated `access_token` into the fetcher, it dispatches on the entity type and returns `ByteStream`s, ready for a downstream converter (e.g. a `FileTypeRouter` in front of `PyPDFToDocument` / `DOCXToDocument` / `HTMLToDocument` / a JSON converter):

- **Files** (`driveItem`) → raw bytes, downloaded through the Graph `shares` endpoint
  (`/shares/{id}/driveItem/content`).
- **List items** (`listItem`) → a JSON `ByteStream` of the item's column values
  (`/sites/{site}/lists/{list}/items/{item}?$expand=fields`).
- **SharePoint pages / news** (`sitePage`) → an HTML `ByteStream` built from the page's web parts via the
  Pages API (`/sites/{site}/pages/{id}/microsoft.graph.sitePage?$expand=canvasLayout`), falling back to
  the list-item fields if the page can't be resolved.

Each stream's `meta` carries `url`, `file_name`, `content_type`, and a normalized `entity_type`. The component supports `run`/`run_async`, `to_dict`/`from_dict`, configurable `raise_on_failure`, and reuses the retriever's throttling/transient-error retry with `Retry-After` backoff. The `access_token` is a run input (typically from an upstream `OAuthResolver`) and accepts a `str` or a `Secret`.

Supporting changes:

- **Shared module** `haystack_integrations.common.microsoft_sharepoint` — extracted the errors, the Graph retry/backoff helpers, and the token resolution so both the retriever and the fetcher use them. The retriever's public API is unchanged (the error classes are re-exported from its package).
- **`MSSharePointRetriever` now surfaces SharePoint identifiers** in each `Document`'s meta (`site_id`, `list_id`, `list_item_id`, `list_item_unique_id`). The fetcher needs these to read list items and pages by ID. This is an additive, non-breaking change to the meta.

### How did you test it?

- **Unit tests** (all HTTP mocked) for the fetcher: per-entity dispatch (file / list item / page), page detection, JSON and HTML rendering, the page-id fallback, the raw-URL probe-then-fallback, skipping container hits and meta without a `web_url`, combined `documents` + `urls`, auth header and share-token encoding, error handling (401/403/404, 429 retry-then-succeed, give-up after max retries), `raise_on_failure=False`, the async paths, an end-to-end retriever → fetcher `Pipeline`, and `to_dict`/`from_dict` round-trips. Added a retriever test for the new ID meta. `mypy` and `ruff` are clean.
- **Manual live verification** against a real Microsoft 365 tenant with a delegated token (`Files.Read.All` + `Sites.Read.All`):

### Notes for the reviewer

- **Output is always `ByteStream`** so the existing converter ecosystem does the conversion. Design
  choices worth a look: list items are rendered as **JSON**, and pages use the **Pages API + `canvasLayout`** rather than the raw `CanvasContent1` field, yielding cleaner HTML.
- **Why the retriever changed:** files are resolvable from their `web_url` via `shares`, but list items and pages are not — they're read by ID, so the retriever now emits those IDs. The `shares` path is kept only as a fallback for raw `urls` input that carries no IDs.
- **Refactor callout:** the retriever's `errors.py` submodule was removed; its symbols now live in `common.microsoft_sharepoint` and are re-exported from the retriever package, so `from haystack_integrations.components.retrievers.microsoft_sharepoint import SharePointConfigError` still works.

### Checklist


integrations/microsoft_sharepoint/src/haystack_integrations/components/fetchers/microsoft_sharepoint/fetcher.py
```diff
@@ -0,0 +1,495 @@
...
+        self.timeout = timeout
+        self.max_retries = max_retries
+        self.raise_on_failure = raise_on_failure
+
+    @component.output_types(streams=list[ByteStream])
+    def run(
+        self,
+        access_token: str | Secret,
+        documents: list[Document] | None = None,
+        urls: list[str] | None = None,
```
sjrl: Hey @bogdankostic I'm a little worried about having `documents` and `urls` both be optional here. This is b/c if this component exists on a branch but still receives an access_token from before the branch it will still fire and run with `documents` and `urls` as empty even if it has connections made to them. 

For execution safety I think it would be better if we could do something like
```
docs_or_urls: list[Document] | list[str]
```
so we can make sure that this is a required variable for the component to run. I'm open to different names for the variable.

bogdankostic: Good catch! I merged `documents` and `urls` into a single mandatory `targets` parameter in 05e3579.

integrations/microsoft_sharepoint/tests/test_fetcher.py
```diff
@@ -0,0 +1,535 @@
...
+        assert [stream.meta["url"] for stream in streams] == [
+            "https://host/a.docx",
+            "https://host/b.docx",
+            "https://host/c.docx",
+        ]
+
+
+@pytest.mark.integration
+@pytest.mark.skipif(
+    not (os.environ.get("MS_SHAREPOINT_ACCESS_TOKEN") and os.environ.get("MS_SHAREPOINT_TEST_FILE_URL")),
```
sjrl, line 523: @bogdankostic do you need me to set these up in our secrets?

sjrl (approved):
Looks good! One minor comment about secrets.

integrations/microsoft_sharepoint/tests/test_fetcher.py
sjrl: Also I think the github workflow file needs to be updated to set these environment variables.

bogdankostic: Adapted the workflow in 5141199

PR #3488, opened 2026-06-23
feat(google_drive): Add Google Drive fetcher

bogdankostic:
### Related Issues

- fixes https://github.com/deepset-ai/haystack-private/issues/433

### Proposed Changes:

Adds a content fetcher, `GoogleDriveFetcher`, to the `google-drive-haystack` integration. It complements `GoogleDriveRetriever` (which returns only metadata / optionally exported snippet text) by fetching the **full content** of each file. 
Wire the retriever's `documents` (or a list of file ids / Drive URLs) plus a per-user delegated `access_token` into the fetcher. It dispatches on each file's mime type and always returns `ByteStream`s, ready for a downstream converter (for example a `FileTypeRouter` in front of `PyPDFToDocument` / `DOCXToDocument` / `XLSXToDocument` / `PPTXToDocument`):

- **Binary files** (PDF, DOCX, images, …) → downloaded as-is via `files.get?alt=media`.
- **Native Google Docs/Sheets/Slides** → exported via `files.export`, by default to the Office formats (Docs→DOCX, Sheets→XLSX, Slides→PPTX). Configurable via the `export_mime_types` init param. Drive caps a single export at 10 MB.
- **Folders** and other non-downloadable Google types (Forms, Sites, …) → skipped.

Each stream's `meta` carries `file_id`, `web_url`, `file_name`, and `content_type`. The component takes a single **mandatory** `targets` input accepting the retriever's `Document`s and/or raw file ids / Drive URLs. `supportsAllDrives` is set on every request so shared-drive files work. 
The fetcher supports `run`/`run_async`, `to_dict`/`from_dict`, configurable `raise_on_failure`, and reuses the retriever's throttling/transient-error retry with `Retry-After` backoff. The `access_token` is a run input (typically from an upstream `OAuthTokenResolver`) and accepts a `str` or a `Secret`.

Supporting change:

- **Shared module** `haystack_integrations.common.google_drive` — extracted the errors, the Drive API retry/backoff helpers, and the token resolution so both the retriever and the fetcher use them. 

### How did you test it?

- **Unit tests** (all HTTP mocked) for the fetcher: binary download, native export (default and configured mapping), folder / non-exportable Google type / missing-file-id skips, file-id parsing from Drive URLs and bare ids, the raw-string metadata lookup, the mandatory-input contract, an invalid item type, error handling (401/403/404, 429 retry-then-succeed, give-up after max retries), `raise_on_failure=False`, the async paths, an end-to-end retriever → fetcher `Pipeline`, and `to_dict`/`from_dict` round-trips. 
- **Manual live verification** against a real Google Drive with a delegated token (`https://www.googleapis.com/auth/drive.readonly`): 

### Notes for the reviewer

- **Output is always `ByteStream`** so the existing converter ecosystem does the conversion. 

### Checklist


integrations/google_drive/google_drive.md
```diff
@@ -0,0 +1,346 @@
+---
```
davidsbatista: I think this file doesn't belong here. This should go to haystack docs, maybe?

integrations/google_drive/src/haystack_integrations/components/fetchers/google_drive/fetcher.py
```diff
@@ -0,0 +1,428 @@
...
+        self.max_retries = max_retries
+        self.raise_on_failure = raise_on_failure
+        self.export_mime_types = export_mime_types
+        self._export_map = export_mime_types if export_mime_types is not None else _DEFAULT_EXPORT_MIME_TYPES
+
+    @component.output_types(streams=list[ByteStream])
+    def run(
+        self,
+        access_token: str | Secret,
+        targets: list[Document] | list[str],
```
davidsbatista: The code supports mixed lists, we can type this param like this `list[Document | str]`

integrations/google_drive/src/haystack_integrations/components/fetchers/google_drive/fetcher.py
```diff
@@ -0,0 +1,428 @@
...
+        :returns: A dictionary with a `streams` key holding the fetched content as `ByteStream` objects. Each
+            stream's `meta` carries `file_id`, `web_url`, `file_name`, and `content_type`.
+        :raises GoogleDriveConfigError: If an item is neither a `Document` nor a `str`, or if `access_token` is a
+            `Secret` that does not resolve to a string.
+        :raises GoogleDriveRequestError: If a fetch fails and `raise_on_failure` is `True`.
+        """
+        token = resolve_access_token(access_token)
+        resolved = self._collect_targets(targets)
+        streams: list[ByteStream] = []
+        async with httpx.AsyncClient(timeout=self.timeout, follow_redirects=True) as client:
```
davidsbatista, line 199: We could do it this way to have actually concurrent requests, using the `await asyncio.gather`

```python
import asyncio

  async with httpx.AsyncClient(timeout=self.timeout, follow_redirects=True) as client:
      tasks = [self._process_async(client, token, fid, hints) for fid, hints in resolved]
      results = await asyncio.gather(*tasks)
      streams = [s for s in results if s is not None]
```

or limting the number of concurrent request with a variable `max_concurrent_requests`

```python

 async with httpx.AsyncClient(timeout=self.timeout, follow_redirects=True) as client:
      semaphore = asyncio.Semaphore(self.max_concurrent_requests)

      async def fetch_one(file_id, hints):
          async with semaphore:  # blocks here if max concurrent requests are in-flight
              return await self._process_async(client, token, file_id, hints)

      results = await asyncio.gather(*[fetch_one(fid, hints) for fid, hints in resolved])
      streams = [s for s in results if s is not None]
```

and `max_concurrent_requests` could be specified on the init.

integrations/google_drive/src/haystack_integrations/components/fetchers/google_drive/fetcher.py
```diff
@@ -0,0 +1,428 @@
...
+        Deserialize this component from a dictionary.
+
+        :param data: The dictionary representation of this component.
+        :returns: The deserialized component instance.
+        """
+        return default_from_dict(cls, data)
+
+
+def _metadata_params() -> dict[str, Any]:
+    return {"fields": _METADATA_FIELDS, "supportsAllDrives": True}
```
davidsbatista, line 447: Maybe I'm missing some detail, just checking. 

Shouldn't the `"supportsAllDrives": True` should be conditioned on the Retriever init param `include_shared_drives`. 

If a user sets it to `False` this will still always be true, right?

integrations/google_drive/src/haystack_integrations/components/fetchers/google_drive/fetcher.py
```diff
@@ -0,0 +1,428 @@
...
+    ) -> ByteStream | None:
+        try:
+            return await self._dispatch_async(client, token, file_id, hints)
+        except (GoogleDriveRequestError, httpx.HTTPError) as error:
+            if self.raise_on_failure:
+                raise
+            logger.warning("Failed to fetch Google Drive file {file_id}: {error}", file_id=file_id, error=str(error))
+            return None
+
+    async def _dispatch_async(
```
davidsbatista, line 296: we can refactor this together with the sync version, extracting a helper function to avoid having duplicated code

davidsbatista (commented):
Looking good! Some initial comments/improvements

integrations/google_drive/google_drive.md
bogdankostic: removed in ad7d8a8.

integrations/google_drive/src/haystack_integrations/components/fetchers/google_drive/fetcher.py
bogdankostic: Changed the type in 4ff31a6.

integrations/google_drive/src/haystack_integrations/components/fetchers/google_drive/fetcher.py
bogdankostic: Added support for concurrency in 4ff31a6

integrations/google_drive/src/haystack_integrations/components/fetchers/google_drive/fetcher.py
bogdankostic: the always-True here is intentional. In the retriever, `include_shared_drives` toggles `includeItemsFromAllDrives` + `corpora=allDrives` to broaden the search scope. The fetcher never searches, it fetches explicit file ids (usually the ones the retriever already surfaced).

davidsbatista (approved):
looks good!

PR #3497, opened 2026-06-24
feat: add system_cachepoint_config to AmazonBedrockChatGenerator

tstadel:
### Related Issues

- coming back to bedrock prompt caching: 
  - cache order is tools -> system -> messages (according to [docs](https://docs.aws.amazon.com/bedrock/latest/userguide/prompt-caching.html#prompt-caching-get-started))
  - assuming system is mostly static, we should be able to include it into the cache without having to deal with ChatPromptBuilder or the message itself

### Proposed Changes:
- added `system_cachepoint_config` init param

### Checklist


sjrl (approved):
Looks good!

PR #3504, opened 2026-06-26
test: harden flaky LocalWhisperTranscriber integration test

julian-risch:
### Related Issues
None

Nightly failure: https://github.com/deepset-ai/haystack-core-integrations/actions/runs/28210530023/job/83570721059

### Proposed Changes:
- Switch both `LocalWhisperTranscriber` integration tests from the `tiny` model to the more robust `base` model (keeping `language="english"` forced). The `tiny` model is the most unstable Whisper model and on some platform / `torch` / `openai-whisper` version combinations it collapses into garbage output — even non-English characters — for clean English audio, which is what failed in nightly CI. Using `base` in both tests also means CI downloads a single shared model instead of two.
- Relax the exact-match assertion on the single-word `answer.wav` clip (`== "answer."`) to a non-empty check. A context-free single word is transcribed inconsistently across models (observed: `"add server"`, `"trans stay turned"`). That third source exists to exercise the `ByteStream` → temp-file branch, which is still asserted via `meta["audio_file"]`.

### How did you test it?
- Ran the whisper integration tests locally with the new model: both pass.

### Checklist

- I have read the [contributors guidelines](https://github.com/deepset-ai/haystack-core-integrations/blob/main/CONTRIBUTING.md) and the [code of conduct](https://github.com/deepset-ai/haystack-core-integrations/blob/main/CODE_OF_CONDUCT.md)
- I have updated the related issue with new insights and changes
- I added unit tests and updated the docstrings
- I've used one of the [conventional commit types](https://www.conventionalcommits.org/en/v1.0.0/) for my PR title: `fix:`, `feat:`, `build:`, `chore:`, `ci:`, `docs:`, `style:`, `refactor:`, `perf:`, `test:`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #3505, opened 2026-06-26
test: force CPU in LocalWhisperTranscriber integration tests

julian-risch:
### Related Issues
None

Follow-up to #3504. The `LocalWhisperTranscriber` integration test kept failing on the macOS CI runners even after #3504 switched `tiny` → `base`.

Failing run (macOS only): https://github.com/deepset-ai/haystack-core-integrations/actions/runs/28231824934

### Proposed Changes:
- **Pin the two local-model integration tests to CPU** (`device=ComponentDevice.from_str("cpu")`). This is the actual fix.
- Revert the model from `base` back to `tiny`: model size was a red herring (it failed with both), and `tiny` on CPU is reliable and faster to download in CI.
- Update the explanatory comments to reflect the real root cause.

**Root cause:** By default the component resolves to the best available device, which is **MPS on Apple Silicon**. Whisper inference on MPS is unreliable on the GitHub-hosted macOS runners — it produced long multilingual hallucinations for clean English audio (e.g. `), icon透 opinions guidance...достаточно美味...這是...`), regardless of model size. The matrix made this unmistakable:

| OS | result |
|----|--------|
| Linux (CPU) | ✅ pass |
| Windows | ✅ (test skipped, no ffmpeg) |
| macOS (MPS) | ❌ multilingual garbage with both `tiny` and `base` |

Because Whisper is deterministic per environment, the existing `--reruns 3 --reruns-delay 30` reproduced the exact same garbage on every attempt. Forcing CPU makes macOS use the same deterministic path that Linux already passes on.

### How did you test it?
- Reproduced locally: confirmed `ComponentDevice.resolve_device(None)` → `mps` on Apple Silicon, while the raw `whisper.load_model()` used in my earlier ad-hoc check defaulted to CPU (which is why #3504 looked fine locally but failed on CI).
- Ran the whisper integration tests locally with the CPU pin: both pass.
- Ran `hatch run fmt-check` (passed).

### Notes for the reviewer
Separately, it may be worth considering whether `LocalWhisperTranscriber` should default to MPS at all, given Whisper's MPS support is unreliable — but that's a component-level behavior change, out of scope for this test fix.

### Checklist

### DIFF
--- a/integrations/whisper/tests/test_whisper_local.py
+++ b/integrations/whisper/tests/test_whisper_local.py
@@ -235,7 +235,7 @@ class TestLocalWhisperTranscriber:
         path = test_files_path / "audio" / "the context for this answer is here.wav"
         assert path.absolute() == docs[1].meta["audio_file"]
 
-        # `answer.wav` is a single isolated word, which models transcribe inconsistently 
+        # `answer.wav` is a single isolated word, which models transcribe inconsistently
         # Therefore, we only assert that something was transcribed to a temporary file.
         assert docs[2].content.strip()
         # meta.audio_file should contain the temp path where we dumped the audio bytes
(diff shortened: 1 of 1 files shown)


davidsbatista (approved):

davidsbatista (approved):
I remember having this issue when running it locally.

PR #3515, opened 2026-07-01
feat: add dakera-haystack integration — self-hosted memory & document store for Haystack

ferhimedamine:
Closes #3516
Supersedes #3519 (the document-store integration is now included here; #3519 is closed as redundant)

## Summary

Adds `integrations/dakera/` — a single `dakera-haystack` package for [Dakera](https://dakera.ai), a self-hosted memory server. It exposes **two complementary integrations that share one Dakera server**:

- **Memory** — `DakeraMemoryStore`, `DakeraMemoryRetriever`, `DakeraMemoryWriter`: conversational, decay-weighted memory over Dakera's memory API (`POST /v1/memory/store`, `POST /v1/memory/recall`). `ChatMessage`-based; the server handles embedding and importance scoring.
- **Document store** — `DakeraDocumentStore`, `DakeraEmbeddingRetriever`: a standard Haystack `DocumentStore` over Dakera's vector-namespace API (via the official `dakera` Python SDK), for embedding-based document retrieval (RAG) with metadata filtering.

This follows the structure of `integrations/mem0/` and (for the document store) the `integrations/pinecone/` conventions.

## Components

| Class | Kind | Description |
|-------|------|-------------|
| `DakeraMemoryStore` | Client | Memory API client; `X-API-Key` auth |
| `DakeraMemoryWriter` | `@component` | Persists `ChatMessage` text as memories → `memories_written: int` |
| `DakeraMemoryRetriever` | `@component` | Decay-weighted recall → `memories: list[ChatMessage]` |
| `DakeraDocumentStore` | `DocumentStore` | Vector-namespace document store (sync + async) backed by the `dakera` SDK |
| `DakeraEmbeddingRetriever` | `@component` | Dense document retrieval with metadata filtering |

## Tests

- Unit tests use `respx` (memory) and `AsyncMock`/mocked SDK (document store) — no live server needed in CI.
- Integration tests (`test_memory_integration.py`, `test_document_integration.py`) run against a live Dakera server and are skipped unless `DAKERA_API_URL` / `DAKERA_URL` is set. The document-store suite reuses Haystack's `CountDocumentsTest` / `DeleteDocumentsTest` / `WriteDocumentsTest` mixins.

```bash
cd integrations/dakera
hatch run test:unit          # mocked, runs in CI
hatch run test:integration   # requires a live Dakera server
```

## Notes

- `dependencies = haystack-ai>=2.28.0, httpx>=0.27.0, dakera>=0.12.8`.
[...]

PR #3519, opened 2026-07-01
feat: add Dakera document store integration

ferhimedamine:
### Related Issues

New integration — no prior issue. Happy to open a tracking issue if the team prefers.

### Proposed Changes

This PR adds a new integration, **`dakera-haystack`** (`integrations/dakera/`), that lets Haystack pipelines store and retrieve documents from [Dakera](https://dakera.ai) — a self-hosted memory server that provides persistent, decay-weighted vector recall.

**Components**

- **`DakeraDocumentStore`** (sync + async) — implements the standard document-store surface (`count_documents`, `write_documents`, `filter_documents`, `delete_documents`, plus `_embedding_retrieval`) and their `_async` counterparts. It drives Dakera's raw vector-namespace API (`upsert` / `query` / `delete` / `get_namespace` / `configure_namespace`), so embeddings are supplied by any Haystack embedder while Dakera handles storage and similarity search — the same model as the other cloud/self-hosted vector-store integrations.
- **`DakeraEmbeddingRetriever`** — a `@component` for dense retrieval with `run` / `run_async` and full `FilterPolicy` support.

**Design notes**

- **Connection** is self-hosted: `url` (default `http://localhost:3000`, the port used by the [`dakera-deploy`](https://github.com/dakera-ai/dakera-deploy) docker-compose stack) plus an `api_key` `Secret` from `DAKERA_API_KEY`. No metered/hosted secret.
- **Namespace lifecycle**: on first use the store adopts the dimension of an existing namespace, or creates it via `configure_namespace` (upsert semantics) when absent.
- **Filters**: `filters.py` converts Haystack's filter syntax into Dakera's MongoDB-style DSL (`{"field": {"$eq": ...}}`, `$and`/`$or`, `$in`/`$nin`, comparison operators), mirroring the approach used by the Pinecone/Qdrant document stores.
- **Metadata-only filtering** reuses the established dummy-vector query pattern, and content is stored as a `content` metadata field.
- Registered in `.github/labeler.yml`, `.github/workflows/CI_coverage_comment.yml`, the root `README.md` inventory table, and a per-integration `.github/workflows/dakera.yml`.

### How did you test it?

- **48 unit tests** (mocked `DakeraClient`) covering init/lazy-init, namespace adoption vs. creation, `to_dict`/`from_dict` round-trips, metadata validation, document↔vector conversion, dummy-vector handling, the full filter converter, and the retriever. All green.
- **Integration tests** (`test_integration.py`) exercise a real write → count → embedding-retrieval → filter → delete round-trip against a live server; they auto-skip unless `DAKERA_URL` is set (start a server with `dakera-deploy`).
- Local run against `haystack-ai 2.30.2` and `dakera 0.12.8`: `hatch run test:unit` → 48 passed, `hatch run test:types` → clean, `ruff check`/`ruff format --check` → clean.

### Notes for the reviewer

- The integration folder is `dakera` (no hyphen) per the format check; the PyPI package name is `dakera-haystack`.
- Integration tests need the Dakera server + MinIO (via `dakera-deploy`), so the `dakera.yml` integration step is a no-op skip in CI until a `DAKERA_URL` is provided — unit tests, lint, and types run on every PR.

### Checklist

- [x] I've used a conventional commit type for my PR title (`feat:`)


ferhimedamine:
I have read the CLA Document and I hereby sign the CLA

julian-risch (changes requested):
Thank you for opening this PR @ferhimedamine !

Before we can merge this, could you please add more unit tests for async features? The following methods are not tested at the moment:
write_documents_async, count_documents_async, filter_documents_async, delete_documents_async, _embedding_retrieval_async, _initialize_async_client, and close_async

Let's use a name that is less prone to collisions with existing metadata names, for example `_dakera_content` instead of `content`. Otherwise a user's metadata might get overwritten and deleted with your `metadata.pop("content")` step

Tests preferrably should use Haystack's DocumentStoreBaseTests/FilterDocumentsTest mixins

ferhimedamine:
Thanks @julian-risch — all three points are addressed in `6c77ef83`.

**1. Async unit tests**
Added `tests/test_document_store_async.py` with unit tests (the SDK is mocked with `AsyncMock`, so they run in CI without a live server) covering every method you listed: `write_documents_async`, `count_documents_async`, `filter_documents_async`, `delete_documents_async`, `_embedding_retrieval_async`, `_initialize_async_client`, and `close_async` — including the namespace-adopt/create branches, batching, empty-input no-op, and the `NotFoundError` paths. Enabled pytest's asyncio auto mode.

**2. Metadata key collision**
Good catch. A Document's content is now stored under a reserved key `_dakera_content` instead of `content`, so a user's own `content` metadata field is no longer overwritten on write or dropped on read. Added a round-trip regression test (`test_user_metadata_named_content_is_not_clobbered`) that stores a document with `meta={"content": ...}` and asserts both the user value and the document body survive.

**3. DocumentStore test mixins**
The integration suite now has a `TestDakeraDocumentStore` class reusing Haystack's `CountDocumentsTest`, `DeleteDocumentsTest`, and `WriteDocumentsTest` mixins. Since Dakera upserts vectors by id, I skipped the duplicate-`FAIL`/`SKIP` cases and the lazy-namespace delete case — mirroring the approach the merged Pinecone integration takes for an overwrite-by-id vector store. I did not pull in `FilterDocumentsTest`: Dakera's server-side filter DSL doesn't cover the full mixin (no `NOT` operator, ISO-date, or DataFrame comparisons), so — as Pinecone also does — filter behaviour is covered by targeted tests instead. Happy to extend this if you'd prefer a different split.

These mixin tests run against a live Dakera server and are skipped in CI (no `DAKERA_URL`).

CI is green across the matrix (Python 3.10 & 3.13 on Linux, Windows, macOS): https://github.com/deepset-ai/haystack-core-integrations/actions/runs/30110294850 — ruff, mypy, docs build, and 61 unit tests at 91% coverage.

Ready for another look. 🙏

ferhimedamine:
We will merge this pr into https://github.com/deepset-ai/haystack-core-integrations/pull/3515

To avoid conflict and ensure integration is complete and more consistent 🙏

ferhimedamine:
Closing this in favour of #3515. Thanks a lot for the thorough review here, @julian-risch — it was very helpful.

We had two open PRs building the **same** `dakera-haystack` package in `integrations/dakera/` (this one for the document store, #3515 for the memory integration). Since only one PR can own that path, we've **consolidated both into #3515**, which now ships the document store *and* the memory components in a single package.

All of the feedback from this review has been carried over into #3515:
- async unit tests for `write_documents_async`, `count_documents_async`, `filter_documents_async`, `delete_documents_async`, `_embedding_retrieval_async`, `_initialize_async_client`, and `close_async`;
- the reserved `_dakera_content` metadata key (so a user's own `content` metadata is no longer clobbered), with a round-trip regression test;
- the `CountDocumentsTest` / `DeleteDocumentsTest` / `WriteDocumentsTest` mixins (duplicate-policy cases skipped, mirroring the Pinecone integration for an overwrite-by-id vector store).

Please continue the review on #3515 — the full document-store implementation (including this PR's `DakeraDocumentStore`, `DakeraEmbeddingRetriever`, and filters) lives there now. 🙏

PR #3520, opened 2026-07-02
fix(langfuse): replace noisy ToolInvoker input with focused tool call arguments

### LINKED ISSUE #3435: bug(langfuse): Observations missing `toolCalls`/`toolCallNames` fields; ToolInvoker misses tool arguments
**Describe the bug**

When using the Langfuse integration with a Haystack `Agent` (which uses `ToolInvoker` internally), `TOOL`-type observations are created with the correct `type` but the Langfuse-native `toolCalls`, `toolCallNames`, and `toolDefinitions` fields are always `null`. This has two practical consequences:

1. **Filtering by tool call name in the Langfuse UI does not work.** The "Tool Call Name" filter reads from `toolCallNames`, which is never populated. The tool name only appears in the observation's `name` string (e.g. `tool_invoker - ['read_pages']`), which is not indexed for that filter.

2. **The TOOL observation's `input` contains the entire conversation message history** passed to the `ToolInvoker`, not just the arguments of the specific tool call that was executed. This makes individual tool observations noisy and hard to read.

Example observation from a real trace (fields trimmed for brevity):

```json
{
  "type": "TOOL",
  "name": "tool_invoker - ['read_pages']",
  "toolDefinitions": null,
  "toolCalls": null,
  "toolCallNames": null
}
```

The same `null` values appear on `GENERATION` observations even when the LLM response contained tool call requests, so the `toolCalls` field is also missing there.

**Root cause**

In `tracer.py`, `LangfuseSpan.set_content_tag` stores the `ToolInvoker` input via:

```python
self._span.update(input=messages)  # `messages` is the full ChatMessage list
```

This writes the entire message list as a generic `input` blob. The Langfuse SDK fields `tool_calls`, `tool_definitions` etc. are never passed to `span.update(...)`, so Langfuse never indexes them.

`DefaultSpanHandler.handle` similarly never calls `span.raw_span().update(tool_calls=..., tool_definitions=...)` for either `ToolInvoker` or `ChatGenerator` components.

**Expected behavior**

- For `TOOL` observations: `input` should contain only the arguments of the specific tool call(s) that were executed; `output` should contain only the corresponding tool result(s). `toolCallNames` should be populated so the Langfuse filter works.
- For `GENERATION` observations: when the LLM reply contains tool call requests, `toolCalls` should be populated with the structured call data.

**To Reproduce**

1. Set up a Haystack `Agent` with at least one tool and wire up `LangfuseConnector`.
2. Run the agent on a query that triggers at least one tool call.
3. Open the resulting trace in the Langfuse UI.
4. Try to filter observations by "Tool Call Name"  (no results)

**Describe your environment:**

- OS: Linux (Ubuntu 24.04)
- Haystack version: `haystack-ai==2.30.1`
- Integration version: `langfuse-haystack==5.2.0`

chakshu-dhannawat:
## Summary

Fixes #3435.

When a Haystack `Agent` runs with `LangfuseConnector` enabled, `ToolInvoker` TOOL-type observations were writing the **entire ChatMessage history** as their `input` blob, making individual tool observations noisy and hard to read in the Langfuse UI.

This PR fixes `DefaultSpanHandler.handle()` so that, when content tracing is enabled, the ToolInvoker observation's `input` and `output` are replaced with structured, focused data:

- **`input`**: list of `{"tool_name": ..., "arguments": {...}}` — one entry per tool call extracted from the assistant messages
- **`output`**: list of `{"tool_name": ..., "result": ..., "error": bool}` — one entry per `ToolCallResult` from the tool response messages

The existing span name update (`tool_invoker - ['search_tool']`) is unchanged and still runs regardless of content tracing state.

## Changes

- `integrations/langfuse/src/.../tracer.py`: extend the `ToolInvoker` branch in `DefaultSpanHandler.handle()` to collect tool call arguments while building the name, then call `span.raw_span().update(input=..., output=...)` when content tracing is on
- `integrations/langfuse/tests/test_tracer.py`: two new tests:
  - `test_handle_tool_invoker_input_output_with_content_tracing` — verifies input/output are set to focused data
  - `test_handle_tool_invoker_no_content_tracing` — verifies no input/output update when content tracing is disabled

## Test plan

```
# 40 tests pass
python3 -m pytest integrations/langfuse/tests/test_tracer.py -x -q
```

- [x] ToolInvoker input is `[{"tool_name": "search_tool", "arguments": {...}}]` (not full message list)
- [x] ToolInvoker output is `[{"tool_name": "search_tool", "result": "...", "error": False}]`
- [x] No input/output update when `HAYSTACK_CONTENT_TRACING_ENABLED` is false
- [x] All existing tests still pass


integrations/langfuse/src/haystack_integrations/tracing/langfuse/tracer.py
```diff
@@ -357,6 +360,25 @@ def handle(self, span: LangfuseSpan, component_type: str | None) -> None:
                 formatted_names = [f"{name} (x{count})" if count > 1 else name for name, count in tool_counts.items()]
                 span.raw_span().update(name=f"{tool_invoker_name} - {sorted(formatted_names)}")
 
+            if tool_calls_input and proxy_tracer.is_content_tracing_enabled:
+                # Replace the noisy full message history with just the tool call arguments
+                span.raw_span().update(input=tool_calls_input)
+
+                output_messages = span.get_data().get(_COMPONENT_OUTPUT_KEY, {}).get("messages", [])
```
bogdankostic: `ToolInvoker.run` emits its results under `tool_messages`, not `messages`. So `.get("messages", [])` is always `[]`, `tool_results` stays empty, and `span.update(output=...)` is never called.

```suggestion
                output_messages = span.get_data().get(_COMPONENT_OUTPUT_KEY, {}).get("tool_messages", [])
```

integrations/langfuse/tests/test_tracer.py
```diff
@@ -748,6 +748,73 @@ def test_handle_tool_invoker(self):
...
+            "haystack.component.name": "tool_invoker",
+            "haystack.component.type": "ToolInvoker",
+            "haystack.component.input": {
+                "messages": [
+                    ChatMessage.from_user("what is RAG?"),
+                    ChatMessage.from_assistant(text="Calling search", tool_calls=[tool_call]),
+                ]
+            },
+            "haystack.component.output": {
+                "messages": [ChatMessage.from_tool("RAG stands for Retrieval-Augmented Generation", tool_call)]
```
bogdankostic: This is why the test passes despite the bug above: the fixture puts the tool response under a `"messages"` key, but a real `ToolInvoker` emits `tool_messages`. Please key it the way the component actually outputs so the test exercises the real path:

```suggestion
                "tool_messages": [ChatMessage.from_tool("RAG stands for Retrieval-Augmented Generation", tool_call)]
```

integrations/langfuse/tests/test_tracer.py
```diff
@@ -748,6 +748,73 @@ def test_handle_tool_invoker(self):
...
+        mock_span = Mock()
+        mock_span.raw_span.return_value = mock_span
+
+        tool_call = ToolCall(tool_name="weather_tool", arguments={"location": "Tokyo"})
+
+        span_data = {
+            "haystack.component.name": "tool_invoker",
+            "haystack.component.type": "ToolInvoker",
+            "haystack.component.input": {"messages": [ChatMessage.from_assistant(text="", tool_calls=[tool_call])]},
+            "haystack.component.output": {"messages": [ChatMessage.from_tool("Sunny, 28°C", tool_call)]},
```
bogdankostic: Same fixture-key issue here.

```suggestion
            "haystack.component.output": {"tool_messages": [ChatMessage.from_tool("Sunny, 28°C", tool_call)]},
```

bogdankostic (changes requested):
Thanks for raising this PR @chakshu-dhannawat! The `input` half works well. But the `output` rewrite needs to be adapted (see in-line code comments). Verified end-to-end against a real Langfuse trace from a `ToolInvoker` inside an `Agent` the TOOL observation's `input` becomes the focused `[{tool_name, arguments}]`, but its `output` is unchanged (still the full `{"tool_messages": [...], "state": {...}}` blob).

Also, the PR body says populating the Tool Call Name filter is out of scope because the SDK has no parameter for it. Half right: langfuse-sdk 4.12.0 indeed has no explicit field (`span.update()` has no tool-call parameter). But Langfuse computes `toolCallNames` **server-side at ingestion** by pattern-matching the observation's **output** for tool-call-like objects. A flat output item counts as a tool call iff it has `name` (not `tool_name`!) + `arguments` + a marker key (`id`/`index`/known `type`).

That means neither of this PR's shapes can ever populate the filter. `{"tool_name", "result", "error"}` has neither a recognized name key nor `arguments`. But a small rename/enrichment of the output items does:

```python
tool_results.append(
    {
        "id": tcr.origin.id or "",
        "name": tcr.origin.tool_name,
        "arguments": tcr.origin.arguments,
        "result": tcr.result,
        "error": tcr.error,
    }
)
```

chakshu-dhannawat:
Good catch, thank you @bogdankostic, you're right. `ToolInvoker.run` emits its results under `tool_messages` (`@component.output_types(tool_messages=...)`), so the previous `.get("messages", [])` always returned `[]` and the output was never set. The test passed only because its fixture used the same wrong key.

I've pushed a fix:
- `tracer.py`: read the output from `tool_messages`
- `test_tracer.py`: key the fixtures as the component actually outputs, so the test now exercises the real path (verified it fails without the tracer fix)

The `input` half is unchanged. Ready for another look.

chakshu-dhannawat:
Thanks for the detailed review, that was really helpful.

I've pushed the fixes:
- tracer now reads the output from the `tool_messages` key (it was reading `messages`, which is why the span output never populated)
- updated both test fixtures to key the output under `tool_messages` so the tests exercise the real path
- following your note on the Tool Call Name filter, the output items now carry `id`, `name` and `arguments` alongside `result`/`error`, so Langfuse detects them as tool calls at ingestion and the filter populates

Also merged main to keep it current. Full unit suite is green. Ready for another look when you get a chance.

bogdankostic (approved):
Looks good, thanks @chakshu-dhannawat!

PR #3522, opened 2026-07-02
fix(elasticsearch): remove duplicate delete/create in delete_all_documents(recreate_index=True)

### LINKED ISSUE #3521: bug(elasticsearch): delete_all_documents(recreate_index=True) calls delete+create twice, silently drops index settings
## Bug description

`ElasticsearchDocumentStore.delete_all_documents(recreate_index=True)` deletes the index and recreates it **twice**. The second recreation omits `settings`, so custom index settings (replica count, refresh interval, custom analyzers, etc.) are silently lost.

## Root cause

`document_store.py` lines 912–919:

```python
self._client.indices.delete(index=self._index)           # 1st delete ✓
self._client.indices.create(index=self._index,
    settings=settings, mappings=mappings)                 # 1st create with settings ✓

# delete index
self._client.indices.delete(index=self._index)           # 2nd delete ← redundant
# recreate with mappings
self._client.indices.create(index=self._index,
    mappings=mappings)                                    # 2nd create WITHOUT settings ← drops settings
```

The async version (`delete_all_documents_async`) does not have this issue — it performs exactly one delete and one create with both settings and mappings.

## Expected behaviour

After `delete_all_documents(recreate_index=True)`, the recreated index should have the same settings and mappings as the original — as the docstring states and as the async version already does.

## Impact

Any user relying on custom index settings (number of replicas, refresh interval, custom analyzers, index templates) will silently lose those settings every time `delete_all_documents(recreate_index=True)` is called. The index accepts documents but with default settings.

## Fix

Remove lines 915–919. Lines 912–913 already perform the correct delete+create with settings.

## Environment

- `elasticsearch-haystack` any version containing the sync `delete_all_documents` implementation

chakshu-dhannawat:
## Summary

Fixes #3521.

`ElasticsearchDocumentStore.delete_all_documents(recreate_index=True)` was calling `indices.delete` and `indices.create` **twice**. The second `create` omitted `settings`, silently dropping custom index configuration (replica count, custom analyzers, refresh interval, etc.) on every call.

**Before (buggy):**
```python
self._client.indices.delete(index=self._index)
self._client.indices.create(index=self._index, settings=settings, mappings=mappings)  # ✓
# delete index
self._client.indices.delete(index=self._index)   # ← redundant
# recreate with mappings
self._client.indices.create(index=self._index, mappings=mappings)  # ← drops settings
```

**After (fixed):**
```python
self._client.indices.delete(index=self._index)
self._client.indices.create(index=self._index, settings=settings, mappings=mappings)
```

The async version (`delete_all_documents_async`) already did this correctly — this brings the sync version in line.

## Test plan

New unit test `test_delete_all_documents_recreate_preserves_settings` verifies:
- `indices.delete` called exactly once
- `indices.create` called exactly once
- The single `create` call includes both `settings` and `mappings`

The existing integration test `test_delete_all_documents_index_recreation` (which runs against a real ES cluster) also covers this path end-to-end.

### DIFF
--- a/integrations/elasticsearch/src/haystack_integrations/document_stores/elasticsearch/document_store.py
+++ b/integrations/elasticsearch/src/haystack_integrations/document_stores/elasticsearch/document_store.py
@@ -912,12 +912,6 @@ class ElasticsearchDocumentStore:
             self._client.indices.delete(index=self._index)  # type: ignore
             self._client.indices.create(index=self._index, settings=settings, mappings=mappings)  # type: ignore
 
-            # delete index
-            self._client.indices.delete(index=self._index)  # type: ignore
-
-            # recreate with mappings
-            self._client.indices.create(index=self._index, mappings=mappings)  # type: ignore
-
         else:
             result = self._client.delete_by_query(**self._prepare_delete_all_request(is_async=False, refresh=refresh))  # type: ignore
             logger.info(
--- a/integrations/elasticsearch/tests/test_document_store.py
+++ b/integrations/elasticsearch/tests/test_document_store.py
@@ -709,6 +709,36 @@ async def test_sparse_vector_retrieval_async_builds_query_with_filters():
     )
 
 
+@patch("haystack_integrations.document_stores.elasticsearch.document_store.AsyncElasticsearch")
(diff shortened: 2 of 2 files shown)


sjrl (approved):
Thanks for the contribution!

Issue #1628, opened 2025-04-08
Document Store - closing resources


davidsbatista:
In most document stores, we establish a connection with some resource but never explicitly close it or destroy it; most of the clients we wrap in our document stores (e.g., ElasticSearch, Mongo, Chroma, etc.) have an `init()`/`connect()` and also a `close()` method.

We could add a method that is triggered when the doc store is destroyed (e.g., based on [del](https://docs.python.org/3/reference/datamodel.html#object.__del__)), and then release all the allocated resources and established connections.

davidsbatista:
This was already previously discussed for Weavite https://github.com/deepset-ai/haystack-core-integrations/issues/471

anakin87:
I encountered this topic when looking at #1634.

I think that we should discuss a common approach to handle this.
Here's my idea:
```python
class MyDocumentStore:
    def close(self):
        # Clean up sync resources
        ...

    async def aclose(self):
        # Clean up async resources
        ...

    def __del__(self):
        # not necessary, it might be nice to have
        try:
            self.close()
        except Exception:
            pass
```

In practice, especially for async, I would expect users to explicitly call `aclose`/`close_async`.

I would like to discuss this also with @mpangrazzi and @Amnah199.

anakin87:
Once every Document Store supports `close/close_async`, we can use the following idea to solve "closing resources when using Pipelines": if we provide retrievers (and other components using Document Storers) with `close/close_async` methods delegating to their Document Stores, we can use the new `Pipeline.close/close_async` methods.

anakin87:
## Recap

### Document Store

No `__del__`.

```python
class MyDocumentStore:
    def close(self):
        # Clean up sync resources
        if self._index is not None:
            with suppress(Exception):
                self._index.close()
            self._index = None

    async def close_async(self):
        # Clean up async resources
        ...
```

### Retriever
```python
class MyRetriever: # for a specific Document Store
    def close(self):
        self.document_store.close()

    async def close_async(self):
        await self.document_store.close_async()
```

- The Document Stores create the resources lazily with existing methods like `_initialize_index` in Pinecone (this is very similar to components' `warm_up`)
- Retrievers behave like other components, handling their resources (Document Stores). Running `Pipeline.close` on Pipelines containing Retrievers frees Document Stores resources.

[...]

PR #3523, opened 2026-07-02
feat(elasticsearch): add close() and close_async() to ElasticsearchDocumentStore

chakshu-dhannawat:
## What

Adds `close()` and `close_async()` to `ElasticsearchDocumentStore` so callers can release the HTTP connection pools when the store is no longer needed.

This aligns with the document store design discussed in #1628: split sync/async cleanup the same way the Pinecone store does, and reset the store on close so a later operation can lazily reinitialize it.

## Changes

- `close()` closes the sync client, drops the reference, and resets the store. `close_async()` does the same for the async client. No cross-closing, so the sync client no longer needs an `asyncio.to_thread` hack.
- After close, the next operation lazily reinitializes the store (close, run something, close again).
- Removed the context manager methods for now, per review.
- Fixed a latent bug that enabling reinit exposed: `_ensure_initialized` popped `headers` off `self._kwargs`, so user supplied headers were lost on the second init. It now copies kwargs locally.

## Tests

Sync and async close assert the client is dropped and state is reset, plus new tests for reinitialize-after-close and header preservation. Full unit suite green, ruff clean.


ErenAta16 (commented):
One inconsistency between the docstring and the implementation: `aclose()` says "Prefer this over `close` in async contexts to avoid blocking the event loop," but it still does `self._client.close()` synchronously (only the async client's close is awaited).

Checked whether that sync client actually exists in practice for async-only usage: `_ensure_initialized()` creates both `self._client` and `self._async_client` together in the same block, so `self._client` is populated even if the caller only ever uses the async API. That means `aclose()`'s sync branch isn't a rare edge case, it runs on every normal call.

Traced what that sync `close()` actually does: `Elasticsearch.close()` -> `Transport.close()` -> iterates the node pool and calls `Urllib3HttpNode.close()` -> `self.pool.close()`, which closes the underlying `urllib3` connection pool (closing idle sockets). That's real (if usually brief) blocking I/O executed directly inside an `async def`, which is exactly what the docstring says this method avoids.

Probably worth wrapping that call, e.g. `await asyncio.to_thread(self._client.close)`, so `aclose()` actually keeps the promise it makes in its own docstring. Not a big deal for a single document store instance, but it'd be a real (if small) event-loop stall if it happens on a hot path or during shutdown of many stores at once.

Rest of the change looks right: `close()` is a no-op before initialization (guarded by the `is not None` checks), and the context manager tests cover the sync path correctly.

chakshu-dhannawat:
Good catch @ErenAta16, you're right. Since `_ensure_initialized()` creates the sync client alongside the async one, that `self._client.close()` in `aclose()` runs on every call and does real blocking socket I/O (Transport.close closes the urllib3 pool), which is exactly what the docstring says the method avoids.

Fixed it by wrapping the sync close in `asyncio.to_thread` so `aclose()` keeps its non-blocking promise. Added a test that asserts the sync client's close is dispatched off the event loop and the async client is awaited directly. Thanks for the detailed trace.

ErenAta16 (commented):
Confirmed the fix. Wrapping the sync client's `close()` in `asyncio.to_thread` restores the non-blocking contract, and the new test checks both the dispatch (`to_thread` called with `mock_sync.close`) and the direct await of the async client's close. LGTM.

anakin87 (changes requested):
Thanks for the contribution and sorry for the long wait.

I studied the topic better and I've come up with a general idea: https://github.com/deepset-ai/haystack-core-integrations/issues/1628#issuecomment-5034357021

Could you please align this PR with that?
Also let's remove context managers for the moment.

chakshu-dhannawat:
Thanks, and no worries about the wait. I've aligned the PR with the recap in #1628:

- Renamed `aclose()` to `close_async()` and split responsibilities like the Pinecone store: `close()` handles the sync client, `close_async()` the async one. Since neither one closes the other's client anymore, the `asyncio.to_thread` workaround is gone.
- Both now drop the client reference and reset the store, so you can close it, run something that reinitializes it, and close again, as you described.
- Removed the context managers for now.

Two ES-specific notes for your input:

1. Unlike Pinecone's per-resource guards (`if self._index is not None: return`), this store creates both clients together behind a single `_initialized` flag, so I reset that flag on close to enable the reinit. Happy to move it toward per-client init tracking if you'd prefer that as a follow-up, but I kept the change minimal here.
2. Enabling reinit surfaced a latent bug: `_ensure_initialized` popped `headers` off `self._kwargs`, so user headers were dropped on the second init. I fixed it by copying kwargs locally and added a regression test.

Full unit suite is green and ruff is clean.

anakin87 (approved):
Thank you.

I aligned the PR a bit more to this new standard and added closing methods to the retrievers.

PR #3530, opened 2026-07-02
feat(mirage): Add Mirage integration as Agent tool

bogdankostic:
### Related Issues

- related to https://github.com/deepset-ai/haystack-private/issues/353

### Proposed Changes:

 
 

Adds a new `mirage-haystack` integration that exposes [Mirage](https://github.com/strukto-ai/mirage) — a unified virtual filesystem that mounts heterogeneous backends (object storage, databases, SaaS apps, local disk) as one filesystem. It lets an `Agent` operate over all of those backends with plain bash through a single tool. 

- `MirageMount` — a declarative, serializable description of one mounted backend (`path`, `resource` registry name, `config`, `read_only`). A single generic constructor covers all of Mirage's backends uniformly; `config` values may be plain values, Haystack `Secret`s, or an OAuth token source, and are resolved only when the live workspace is built.
- `MirageWorkspace` — holds the mount tree and cache configuration, lazily builds a live `mirage.Workspace` on first use via Mirage's resource registry, serializes cleanly (resolving `Secret`s / token sources only at build time), and exposes `run` / `run_async` that normalize and truncate output for an LLM.
- `MirageShellTool` — a Haystack `Tool` that gives an `Agent` one guarded bash command surface across every mount.

### How did you test it?

- Unit tests: 40 tests across `tests/test_workspace.py` and `tests/test_shell_tool.py`, covering mount/workspace construction and validation, `Secret` and OAuth-token-source serialization round-trips (no plaintext leakage), per-mount read-only enforcement, RAM/disk command execution, output truncation, tool shape and `to_dict`/`from_dict`, and the command guard — including the substitution/backtick/subshell bypass cases and fail-closed behavior for quoted or variable command names.
- Manual verification against live Google Drive and OneDrive backends (mounting, listing, cross-backend file copy, and an `Agent` driving the shell tool end to end).

### Notes for the reviewer

- Pins `mirage-ai==0.0.3`: Mirage is pre-1.0 and has shipped breaking changes across 0.0.x releases.

### Checklist


integrations/mirage/pyproject.toml
```diff
@@ -0,0 +1,181 @@
...
+check_untyped_defs = true
+disallow_incomplete_defs = true
+
+# Mirage ships no type stubs / py.typed marker, so treat its imports as untyped rather than erroring.
+[[tool.mypy.overrides]]
+module = ["mirage.*"]
+ignore_missing_imports = true
+
+[tool.ruff]
+target-version = "py310"
```
julian-risch:
```suggestion
target-version = "py311"
```
Let's keep in sync with `requires-python = ">=3.11"`

integrations/mirage/src/haystack_integrations/tools/mirage/shell_tool.py
```diff
@@ -0,0 +1,226 @@
...
+                if denied in command:
+                    msg = f"Command references a denied path '{denied}'."
+                    raise MirageCommandNotAllowedError(msg)
+
+        if self._allowed_commands is None:
+            return
+
+        allowed = set(self._allowed_commands)
+        names = _command_names_in(command)
+        for base in names:
```
julian-risch, line 178: When `allowed_commands` is set but `_command_names_in` returns an empty list then the loop does not run and the command passes the guard (but it doesn't actually run it seems). That can happen for lines like `> /data/x` or `>> /data/x` with no command name. Should we fail when there are no resolvable command names?

```suggestion
        names = _command_names_in(command)
        if not names:
            self._reject_command("<no resolvable command>", allowed)
        for base in names:
```

julian-risch: I tried to get mirage to write even if only `ls` command is allowed but didn't succed:

`'> /data/f.txt'  -> mirage_out='\n[exit code 1]\nunsupported tree-sitter node type: file_redirect' | file_mutated=False`

So it's secure as is right now with the pinned version but I still suggest to improve our integration.

integrations/mirage/src/haystack_integrations/tools/mirage/workspace.py
```diff
@@ -0,0 +1,337 @@
...
+    # --- lifecycle ---------------------------------------------------------------------------
+
+    def warm_up(self) -> None:
+        """Build the live `mirage.Workspace` eagerly. Idempotent."""
+        self._ensure_live()
+
+    def _ensure_live(self) -> _MirageWorkspace:
+        """Build (once) and return the live `mirage.Workspace`."""
+        if self._live is not None:
+            return self._live
```
julian-risch, line 220: It can happen that two Agent threads invoke the tool at the same time and then both see `self._live is None` and build resources leading to resource leaks.
AsyncExecutor has a `with cls._singleton_lock`. Can we do something similar here too?

```suggestion
        if self._live is not None:
            return self._live
        if self._live is not None:
            return self._live
        with self._build_lock:
            if self._live is None:
                ...  # build resources, assign self._live
        return self._live
```
and add `self._build_lock = threading.Lock() in __init__`

julian-risch (changes requested):
Looks good to me in general. I have two remarks on the more complex parts where I am not entirely sure if what I expect should be in scope. I believe it should.

julian-risch (approved):

PR #3534, opened 2026-07-02
chore: support sync streaming callbacks in async contexts for Haystack 2.x/3.x compatibility

julian-risch:
### Related Issues

Haystack 3.0 changes the semantics of `select_streaming_callback(requires_async=True)`: instead of raising `ValueError` for a sync callback in an async context, it now accepts it with a warning (the callback runs inline on the event loop). Accordingly, its return type widened from `AsyncStreamingCallbackT | None` to `StreamingCallbackT | None`, which fails mypy in every integration that passes the result to an async-only-typed internal handler.

### Proposed Changes:

Rather than casting the union away, this PR makes the nine affected chat generators actually support what 3.0 allows:

- Widen the async streaming handlers' callback parameter from `AsyncStreamingCallbackT` to `StreamingCallbackT` in **anthropic, cohere, google_genai, huggingface_api, litellm, ollama, transformers (`_AsyncHFTokenStreamingHandler`), vllm, watsonx**.
- Replace each `await callback(chunk)` with a call that awaits the result only if it is awaitable (the same pattern as haystack 3.0's internal `_invoke_streaming_callback`).

Haystack 2.x behavior is unchanged.

### How did you test it?

- Haystack 2.x tests pass.
- Haystack v3 branch (installed `git+https://github.com/deepset-ai/haystack.git@v3`): `hatch run test:types` is clean.

### Checklist

### DIFF
--- a/integrations/anthropic/src/haystack_integrations/components/generators/anthropic/chat/chat_generator.py
+++ b/integrations/anthropic/src/haystack_integrations/components/generators/anthropic/chat/chat_generator.py
@@ -1,3 +1,4 @@
+import inspect
 from datetime import datetime, timezone
 from typing import Any, ClassVar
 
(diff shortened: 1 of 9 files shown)


sjrl (approved):
Looks good!

PR #3535, opened 2026-07-02
test: guard ToolInvoker imports so chat-generator tests run under Haystack 3.0

julian-risch:
### Related Issues

- Part of deepset-ai/haystack-private#446

Haystack 3.0 removes `haystack.components.tools` (`ToolInvoker` has no drop-in replacement; tool execution moved into the `Agent`). In ten chat-generator test suites, the module-level `from haystack.components.tools import ToolInvoker` fails. As a consequence, none of the tests are run.

### Proposed Changes:

For **aimlapi, amazon_bedrock, cohere, cometapi, meta_llama, mistral, ollama, openrouter, orcarouter, togetherai**:

- Wrap the import in `try/except ImportError` with a `ToolInvoker = None` fallback (as prototyped in #3467).
- Add `@pytest.mark.skipif(ToolInvoker is None, ...)` to the tests that actually use `ToolInvoker` (one pipeline/live test per suite; two in ollama).

### How did you test it?

- Haystack 2.x: unit tests pass.
- Haystack v3 branch (installed `git+https://github.com/deepset-ai/haystack.git@v3` into the test envs): all ten suites now **collect and run**. The failures this unmasks belong to the other known v3-incompatibility categories (lazy OpenAI client init and serialization changes).

The failing tests are predominantly `test_init_*`/`test_from_dict_fail_wo_env_var` (v3's lazy OpenAI client init — no client, and no API-key error, at `__init__` anymore), `test_serde_in_pipeline`/`test_to_dict` (v3 serialization format changes), and the async `test_run_async*` suites in mistral/openrouter (same lazy-client root cause). Those are tracked separately in the compatibility effort.

cohere's `chat_generator.py` also mentions `haystack.components.tools` in its docstring usage example — that is not a runtime import and is left for a follow-up that reworks the example for 3.0.

### Notes for the reviewer

The skipped-on-3.0 tests are integration/live pipeline tests that chain generator → `ToolInvoker`. Once the 3.0 migration path for tool invocation in these examples is decided (e.g. `Agent`), they can be rewritten instead of skipped.

### Checklist

- I have read the [contributors guidelines](https://github.com/deepset-ai/haystack-core-integrations/blob/main/CONTRIBUTING.md) and the [code of conduct](https://github.com/deepset-ai/haystack-core-integrations/blob/main/CODE_OF_CONDUCT.md)
- I have updated the related issue with new insights and changes
- I added unit tests and updated the docstrings
[...]

PR #3536, opened 2026-07-02
fix: support Haystack 3.0's lazy OpenAI client initialization in OpenAI-inheriting generators

julian-risch:
### Related Issues

Haystack 3.0 no longer creates the OpenAI clients in `__init__`: `self.client`/`self.async_client` are `None` until `warm_up()`/`warm_up_async()`, the `_is_warmed_up` flag is gone, and a missing API key raises at warm-up instead of at init. This breaks the ten integrations that subclass `OpenAIChatGenerator`/`OpenAIGenerator`/`OpenAIResponsesChatGenerator`/the OpenAI embedders.

### Proposed Changes:

**Source changes**

- **mistral, openrouter** (`run`/`run_async` overrides): replace the removed `if not self._is_warmed_up: self.warm_up()` with an unconditional `self.warm_up()` (idempotent on both versions); in `run_async`, call `warm_up_async()` when it exists (3.0) so the async client is created on the running event loop; narrow the now-`Optional` clients for mypy with the same `assert`-after-warm-up idiom that haystack 3.0's own `run()` uses.
- **perplexity chat generator**: only wrap the clients with the Perplexity attribution headers in `__init__` when they exist (2.x); add `warm_up`/`warm_up_async` overrides that apply the headers when 3.0 creates the clients.
- **perplexity embedders**: the attribution header is injected via `http_client_kwargs`, but serialization must keep the user-provided value — under 3.0 the client is built from `self.http_client_kwargs` at warm-up, losing the header. Keep the enriched kwargs on `self.http_client_kwargs` (so both 2.x init and 3.0 warm-up build the clients with the header) and store the original user-provided value as `self._http_client_kwargs`, which `to_dict()` serializes.

**Test changes** (aimlapi, cometapi, meta_llama, mistral, nvidia, openrouter, orcarouter, perplexity, stackit, togetherai)

- `test_init_default`/`test_init_with_parameters`: no longer assert on client attributes; they check `component.api_key.resolve_value()` and the other init attributes. Client creation is covered by a new dedicated `test_warm_up` (calls `warm_up()`, harmless on 2.x, required on 3.0).
- `test_init_fail_wo_api_key`: runs init **and** `warm_up()` inside `pytest.raises(ValueError)`, accepting the error from either version's raise point. The error message is identical on both. `test_from_dict_fail_wo_env_var` is removed as redundant with this test.
- `test_init_default_async` → `test_warm_up_async`: async test that `await component.warm_up_async()` when available, since 3.0 only creates the async client there.

### How did you test it?

- **Haystack 2.x**: unit tests pass.
- **Haystack v3 branch** (installed `git+https://github.com/deepset-ai/haystack.git@v3` into the test envs; for the seven suites that also need the `ToolInvoker` import guard I verified on a local merge with #3535):
  - `test:types` is clean in all ten — the `_is_warmed_up`/`OpenAI | None` mypy errors from the sweep are gone.
  - All lazy-client test failures are fixed. The only remaining v3 unit failures are addressed in other opened PRs.

### Checklist

### DIFF
--- a/integrations/aimlapi/tests/test_aimlapi_chat_generator.py
+++ b/integrations/aimlapi/tests/test_aimlapi_chat_generator.py
@@ -88,12 +88,17 @@ class TestAIMLAPIChatGenerator:
     def test_init_default(self, monkeypatch):
         monkeypatch.setenv("AIMLAPI_API_KEY", "test-api-key")
         component = AIMLAPIChatGenerator()
-        component.warm_up()  # with haystack-ai >= 3.0 the client is created during warm-up
-        assert component.client.api_key == "test-api-key"
+        assert component.api_key.resolve_value() == "test-api-key"
         assert component.model == "openai/gpt-5-chat-latest"
         assert component.api_base_url == "https://api.aimlapi.com/v1"
         assert component.streaming_callback is None
         assert not component.generation_kwargs
+
+    def test_warm_up(self, monkeypatch):
+        monkeypatch.setenv("AIMLAPI_API_KEY", "test-api-key")
+        component = AIMLAPIChatGenerator()
+        component.warm_up()  # with haystack-ai >= 3.0 the client is created during warm-up
+        assert component.client.api_key == "test-api-key"
 
     def test_init_fail_wo_api_key(self, monkeypatch):
         monkeypatch.delenv("AIMLAPI_API_KEY", raising=False)
@@ -110,8 +115,7 @@ class TestAIMLAPIChatGenerator:
             api_base_url="test-base-url",
             generation_kwargs={"max_tokens": 10, "some_test_param": "test-params"},
         )
-        component.warm_up()  # with haystack-ai >= 3.0 the client is created during warm-up
-        assert component.client.api_key == "test-api-key"
+        assert component.api_key.resolve_value() == "test-api-key"
         assert component.model == "openai/gpt-5-chat-latest"
         assert component.streaming_callback is print_streaming_chunk
         assert component.generation_kwargs == {"max_tokens": 10, "some_test_param": "test-params"}
@@ -207,26 +211,6 @@ class TestAIMLAPIChatGenerator:
         assert component.extra_headers == {"test-header": "test-value"}
         assert component.timeout == 10
         assert component.max_retries == 10
-
-    def test_from_dict_fail_wo_env_var(self, monkeypatch):
-        monkeypatch.delenv("AIMLAPI_API_KEY", raising=False)
-        data = {
-            "type": ("haystack_integrations.components.generators.aimlapi.chat.chat_generator.AIMLAPIChatGenerator"),
-            "init_parameters": {
-                "api_key": {"env_vars": ["AIMLAPI_API_KEY"], "strict": True, "type": "env_var"},
-                "model": "openai/gpt-5-chat-latest",
-                "api_base_url": "test-base-url",
-                "streaming_callback": "haystack.components.generators.utils.print_streaming_chunk",
-                "generation_kwargs": {"max_tokens": 10, "some_test_param": "test-params"},
-                "extra_headers": {"test-header": "test-value"},
-                "timeout": 10,
-                "max_retries": 10,
-            },
-        }
-        with pytest.raises(ValueError, match=r"None of the .* environment variables are set"):
-            # haystack-ai 2.x raises at init; haystack-ai >= 3.0 raises when the client is created in warm_up
-            component = AIMLAPIChatGenerator.from_dict(data)
-            component.warm_up()
 
     def test_run(self, chat_messages, mock_chat_completion, monkeypatch):  # noqa: ARG002
         monkeypatch.setenv("AIMLAPI_API_KEY", "fake-api-key")
--- a/integrations/aimlapi/tests/test_aimlapi_chat_generator.py
+++ b/integrations/aimlapi/tests/test_aimlapi_chat_generator.py
@@ -88,6 +88,7 @@ class TestAIMLAPIChatGenerator:
     def test_init_default(self, monkeypatch):
         monkeypatch.setenv("AIMLAPI_API_KEY", "test-api-key")
         component = AIMLAPIChatGenerator()
+        component.warm_up()  # with haystack-ai >= 3.0 the client is created during warm-up
         assert component.client.api_key == "test-api-key"
         assert component.model == "openai/gpt-5-chat-latest"
         assert component.api_base_url == "https://api.aimlapi.com/v1"
@@ -97,7 +98,9 @@ class TestAIMLAPIChatGenerator:
     def test_init_fail_wo_api_key(self, monkeypatch):
         monkeypatch.delenv("AIMLAPI_API_KEY", raising=False)
         with pytest.raises(ValueError, match=r"None of the .* environment variables are set"):
-            AIMLAPIChatGenerator()
+            # haystack-ai 2.x raises at init; haystack-ai >= 3.0 raises when the client is created in warm_up
+            component = AIMLAPIChatGenerator()
+            component.warm_up()
 
     def test_init_with_parameters(self):
         component = AIMLAPIChatGenerator(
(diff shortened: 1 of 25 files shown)


integrations/openrouter/src/haystack_integrations/components/generators/openrouter/chat/chat_generator.py
```diff
@@ -354,8 +354,9 @@ def run(
             - `replies`: A list containing the generated responses as ChatMessage instances.
         """
         messages = _normalize_messages(messages)
-        if not self._is_warmed_up:
-            self.warm_up()
+        # warming up is idempotent on all supported haystack-ai versions; with haystack-ai >= 3.0 it also
+        # initializes the client
```
sjrl: Not sure we need this dev comment here. I think it would be fine to remove.

integrations/openrouter/src/haystack_integrations/components/generators/openrouter/chat/chat_generator.py
```diff
@@ -382,6 +383,7 @@ def run(
             tools_strict=tools_strict,
         )
         openai_endpoint = api_args.pop("openai_endpoint")
+        assert self.client is not None  # mypy: with haystack-ai >= 3.0 the client is built by warm_up above
         chat_completion = getattr(self.client.chat.completions, openai_endpoint)(**api_args)
```
sjrl: We've typically opted to use `# type: ignore[...]` instead of using asserts for this so I wonder about changing this to be consistent. Also fine to keep if you think that the assert is better.

integrations/perplexity/src/haystack_integrations/components/embedders/perplexity/document_embedder.py
sjrl: I wonder if instead of the new `warm_up` methods we should instead store the original user provided `http_client_kwargs` under a new key like `self._http_client_kwargs = http_client_kwargs` and then update the `to_dict` method to use this internal key. 

Then I would set `self.http_client_kwargs = _http_client_kwargs_with_attribution(http_client_kwargs)` which would avoid the need for the new `warm_up` methods. WDYT?

integrations/perplexity/src/haystack_integrations/components/embedders/perplexity/document_embedder.py
```diff
@@ -167,6 +205,10 @@ def _embed_batch(
                 "encoding_format": self.encoding_format,
             }
 
+            # with haystack-ai >= 3.0 the client is Optional and built by warm_up in run
+            if self.client is None:
+                msg = "The client is not initialized. Call warm_up() before running the component."
+                raise RuntimeError(msg)
```
sjrl: I think we should just use a `type: ignore` here since this is an internal method that's only called after the auto `warm_up` in run has been called. 

Also the error message isn't really accurate since we do auto warm_up the component at runtime.

integrations/openrouter/src/haystack_integrations/components/generators/openrouter/chat/chat_generator.py
julian-risch: removed

integrations/openrouter/src/haystack_integrations/components/generators/openrouter/chat/chat_generator.py
julian-risch: Replaced all four `assert self.client/async_client is not None` with `# type: ignore[union-attr]`

integrations/perplexity/src/haystack_integrations/components/embedders/perplexity/document_embedder.py
julian-risch: done

integrations/aimlapi/tests/test_aimlapi_chat_generator.py
```diff
@@ -107,6 +110,7 @@ def test_init_with_parameters(self):
             api_base_url="test-base-url",
             generation_kwargs={"max_tokens": 10, "some_test_param": "test-params"},
         )
+        component.warm_up()  # with haystack-ai >= 3.0 the client is created during warm-up
         assert component.client.api_key == "test-api-key"
```
anakin87: Since this test is about `init`, I'd avoid calling `warm_up` here.
We can instead assert something like `component.api_key.resolve_value() == "test-api-key"`

integrations/aimlapi/tests/test_aimlapi_chat_generator.py
```diff
@@ -220,7 +224,9 @@ def test_from_dict_fail_wo_env_var(self, monkeypatch):
             },
         }
         with pytest.raises(ValueError, match=r"None of the .* environment variables are set"):
-            AIMLAPIChatGenerator.from_dict(data)
+            # haystack-ai 2.x raises at init; haystack-ai >= 3.0 raises when the client is created in warm_up
+            component = AIMLAPIChatGenerator.from_dict(data)
+            component.warm_up()
```
anakin87: I'd remove this test. In my opinion, this does not provide value on top of `test_init_fail_wo_api_key`

integrations/aimlapi/tests/test_aimlapi_chat_generator.py
```diff
@@ -88,6 +88,7 @@ class TestAIMLAPIChatGenerator:
     def test_init_default(self, monkeypatch):
```
anakin87, line 88: This test and several tests in this PR now test both init and client creation. The name is also a bit misleading.
I'd prefer to split this into two: `test_init_default` (not covering client creation) and `test_warm_up`.
I'd apply the same change wherever needed in this PR.

anakin87 (changes requested):
I've left some general considerations on tests.

(I would probably have created separate PRs for each integration where we modify the source code.) Anyway, the important thing is releasing the integrations once merged to make them v3-ready)

integrations/aimlapi/tests/test_aimlapi_chat_generator.py
julian-risch: Done — split into `test_init_default` (init attributes only, no client assertions) and a dedicated `test_warm_up` in all ten integrations. The async `test_init_default_async` is renamed to `test_warm_up_async` accordingly, since it covers async client creation.

integrations/aimlapi/tests/test_aimlapi_chat_generator.py
julian-risch: Done — the init tests no longer call `warm_up()` and now assert `component.api_key.resolve_value() == "test-api-key"` instead of client attributes, applied across all integrations in this PR.

integrations/aimlapi/tests/test_aimlapi_chat_generator.py
julian-risch: Removed — in every integration this PR touches.

integrations/aimlapi/tests/test_aimlapi_chat_generator.py
anakin87: thank you!

integrations/perplexity/src/haystack_integrations/components/embedders/perplexity/document_embedder.py
anakin87: @julian-risch WDYT?

julian-risch: Done in c30b72d — the `warm_up`/`warm_up_async` overrides are gone. `self.http_client_kwargs` now keeps the attribution-enriched kwargs (so haystack-ai >= 3.0 builds the clients with the header at warm-up, and 2.x at init), and the original user-provided value is stored as `self._http_client_kwargs` and used in `to_dict()`. Applied to both embedders; verified unit tests + mypy on haystack-ai 2.30.2 and the v3 branch.

anakin87 (approved):
I've also aligned the Perplexity Chat Generator in https://github.com/deepset-ai/haystack-core-integrations/pull/3536/commits/1540d4125a798bd7f23f04dda8c427c63ee52060 and https://github.com/deepset-ai/haystack-core-integrations/pull/3536/commits/be173b78377f01ade1d56acf080c1c91ff3d3290

It's now ready to be merged

PR #3538, opened 2026-07-02
fix: adapt optimum, ragas, elasticsearch, opensearch to APIs removed in Haystack 3.0

julian-risch:
### Related Issues

Haystack 3.0 removes several APIs some of our integrations relied on.

### Proposed Changes:

- **optimum** (src): `HFModelType` and `check_valid_model` were removed from `haystack.utils.hf`, so the package failed to import at all under 3.0 (mypy errors plus collection errors in every embedder test module). Both are now vendored into `_backend.py` verbatim — they are small, HF-specific helpers, and vendoring keeps behavior identical on both versions with no conditional imports. The existing tests keep patching `_backend.check_valid_model` unchanged.
- **ragas** (tests): `AsyncPipeline` no longer exists at the `haystack` top level; 3.0 merged async execution into `Pipeline` (`run_async`). The test module now falls back to `AsyncPipeline = Pipeline` on `ImportError`.
- **elasticsearch, opensearch** (tests): the hybrid-retriever serde tests built a real `SentenceTransformersTextEmbedder`, which 3.0 removed from core. They now use `OpenAITextEmbedder` (in core on both versions) and build the expected embedder sub-dict with `embedder.to_dict()` at runtime instead of hard-coding it — which also removes the `device`/`revision` compat shims those hard-coded dicts had accumulated. An autouse fixture sets a fake `OPENAI_API_KEY`, since haystack-ai 2.x resolves the key eagerly at init.

### How did you test it?

- Haystack 2.x unit tests pass for all four integrations.
- Haystack v3 branch (installed `git+https://github.com/deepset-ai/haystack.git@v3` into the test envs): mypy is clean and unit tests pass 

### Notes for the reviewer

If we prefer not to vendor `check_valid_model` into optimum, then the alternative is dropping Hub-side model validation entirely and letting the model load fail naturally. The advantage of vendoring is that it preserves today's error messages for users.

The `SentenceTransformers*` mentions remaining in elasticsearch/opensearch **src** are docstring usage examples only (no runtime imports); updating those examples for the 3.0 component locations is a docs follow-up.

### Checklist

### DIFF
--- a/integrations/elasticsearch/tests/test_hybrid_retriever.py
+++ b/integrations/elasticsearch/tests/test_hybrid_retriever.py
@@ -8,7 +8,7 @@ from unittest.mock import Mock, patch
 
 import pytest
 from haystack import Document, Pipeline
-from haystack.components.embedders import SentenceTransformersTextEmbedder
+from haystack.components.embedders import OpenAITextEmbedder
 from haystack.components.joiners.document_joiner import JoinMode
 from haystack.core.component import component
 from haystack.document_stores.types import FilterPolicy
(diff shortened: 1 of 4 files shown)


integrations/optimum/src/haystack_integrations/components/embedders/optimum/_backend.py
```diff
@@ -37,6 +39,45 @@
...
+    api = HfApi()
+    try:
+        model_info = api.model_info(model_id, token=token.resolve_value() if token else None)
+    except RepositoryNotFoundError as e:
+        msg = f"Model {model_id} not found on HuggingFace Hub. Please provide a valid HuggingFace model_id."
+        raise ValueError(msg) from e
+
+    if model_type == HFModelType.EMBEDDING:
+        allowed_model = model_info.pipeline_tag in ["sentence-similarity", "feature-extraction"]
+        error_msg = f"Model {model_id} is not a embedding model. Please provide a embedding model."
```
bogdankostic:
```suggestion
        error_msg = f"Model {model_id} is not an embedding model. Please provide an embedding model."
```

integrations/optimum/src/haystack_integrations/components/embedders/optimum/_backend.py
```diff
@@ -37,6 +39,45 @@
...
+    elif model_type == HFModelType.GENERATION:
+        allowed_model = model_info.pipeline_tag in ["text-generation", "text2text-generation", "image-text-to-text"]
+        error_msg = f"Model {model_id} is not a text generation model. Please provide a text generation model."
```
bogdankostic: I don't think we need this branch given that we only support embedding models in optimum.

integrations/elasticsearch/tests/test_hybrid_retriever.py
```diff
@@ -25,93 +25,74 @@ def run(self, text: str, param_a: str = "default", param_b: str = "another_defau
...
+                        "index": "default",
+                        "api_key": {"type": "env_var", "env_vars": ["ELASTIC_API_KEY"], "strict": False},
+                        "api_key_id": {"type": "env_var", "env_vars": ["ELASTIC_API_KEY_ID"], "strict": False},
+                        "embedding_similarity_function": "cosine",
+                        "sparse_vector_field": None,
+                        "ingest_pipeline": None,
+                    },
                 },
+                # serialize the embedder with the installed haystack-ai version so the expected fields match it
+                "embedder": OpenAITextEmbedder().to_dict(),
```
bogdankostic: This makes the embedder sub-dict a tautology (expected is built from the same `.to_dict()` the retriever calls). I don't think there's a difference between v2 and v3 for `OpenAITextEmbedder` serializations.

bogdankostic (commented):
Looks good in general, just some minor comments.

integrations/elasticsearch/tests/test_hybrid_retriever.py
julian-risch: Addressed in https://github.com/deepset-ai/haystack-core-integrations/pull/3538/commits/3a5e3e2129e8111c3d0d8c9d3a1d728982bd0e01

integrations/optimum/src/haystack_integrations/components/embedders/optimum/_backend.py
julian-risch: Addressed in https://github.com/deepset-ai/haystack-core-integrations/pull/3538/commits/3a5e3e2129e8111c3d0d8c9d3a1d728982bd0e01

bogdankostic (approved):

PR #3533, opened 2026-07-02
test: make Tool/Agent serialization assertions version-agnostic for Haystack 2.x/3.x

julian-risch:
### Related Issues

- Part of deepset-ai/haystack-private#446

Serialization tests in five integrations compared `to_dict()` output against hard-coded expected dicts that pin the exact `Tool`/`Toolset`/`Agent` serialization format of one haystack-ai version. Haystack 3.0 adds fields (e.g. `Tool.to_dict()` now includes `async_function`; `Agent` gained `tool_concurrency_limit`, `hooks`, `tool_streaming_callback_passthrough`), which broke these tests. This PR makes them version-agnostic so they pass on both 2.x and the current `v3` branch:

### Proposed Changes:

- **anthropic, huggingface_api, transformers, watsonx**: the `to_dict` tests no longer pin the haystack-owned tools serialization format at all. They exclude the tools entries from the pinned-dict comparison and instead do a real `from_dict(component.to_dict())` round-trip, asserting the recovered tools equal the originals.
- **aimlapi, cometapi, meta_llama, mistral, openrouter, orcarouter, togetherai**: the same for their `test_serde_in_pipeline` tests — the tools entries are popped from the pinned pipeline dict; tool fidelity is covered by the `Pipeline.loads(pipeline.dumps())` round-trip (`Pipeline.__eq__` compares serialized forms) plus explicit attribute checks on the loaded generator.
- **github** (`test_pipeline_serialization`): the test pins the full pipeline dict including `Agent` and `OpenAIChatGenerator` init parameters, which belong to haystack-ai and change between versions. It now copies any init parameters the installed haystack-ai adds beyond the expected baseline before comparing, replacing the two existing single-key compat shims (`http_client_kwargs`/`tool_invoker_kwargs` pops and the `confirmation_strategies`/`required_variables`/`user_prompt` loop). The test keeps pinning what the integration owns: the `GitHubFileEditorTool` serialization. It also now asserts `Pipeline.from_dict(pipeline.to_dict()) == pipeline`.
- **github**: also changed `import unittest` to `import unittest.mock` — the test uses `unittest.mock.ANY` and only worked because haystack-ai 2.x happened to import `unittest.mock` transitively; with v3 installed it fails with `AttributeError: module 'unittest' has no attribute 'mock'`.

### How did you test it?

- Haystack 2.x: `fmt-check`, `test:types`, and full unit suites pass for all touched integrations.
[...]

PR #3542, opened 2026-07-03
fix(cometapi): make CometAPIChatGenerator deserializable in pipelines

julian-risch:
### Related Issues

- None (found while working on #3533)
- `CometAPIChatGenerator` could not be deserialized as part of a pipeline: `Pipeline.from_dict` / `Pipeline.loads` failed for any pipeline containing it. Two bugs, both fixed here:
- The class was never registered with the `@component` decorator. 
- The `to_dict` inherited from `OpenAIChatGenerator` serialized init parameters that `CometAPIChatGenerator.__init__` does not accept (`api_base_url`, `organization`), so even a registered component failed to deserialize with `TypeError: __init__() got an unexpected keyword argument 'api_base_url'`. 

### Proposed Changes:

- Added the decorator (and the two-argument `super(CometAPIChatGenerator, self).__init__` form it requires, since the decorator recreates the class — same pattern as `OpenRouterChatGenerator`).
Added a `to_dict` override that serializes exactly the component's own init parameters, mirroring openrouter.

### How did you test it?

The test now does a real `pipeline.dumps()` / `Pipeline.loads()` round-trip and asserts the loaded pipeline equals the original.

### Checklist

### DIFF
--- a/integrations/cometapi/src/haystack_integrations/components/generators/cometapi/chat/chat_generator.py
+++ b/integrations/cometapi/src/haystack_integrations/components/generators/cometapi/chat/chat_generator.py
@@ -1,5 +1,6 @@
 from typing import Any
 
+from haystack import component, default_to_dict
 from haystack.components.generators.chat import OpenAIChatGenerator
 from haystack.dataclasses import (
     StreamingCallbackT,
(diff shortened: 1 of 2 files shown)


sjrl (approved):
Thanks!

PR #3386, opened 2026-06-03
feat(anthropic): add anthropic_server_tools to AnthropicChatGenerator

Aftabbs:
## Summary

Closes #2724

`AnthropicChatGenerator` currently validates that all items in `tools` are `Tool` or `Toolset` instances, which prevents users from passing Anthropic's native server-side tools (web search, computer use, etc.). This PR adds an `anthropic_server_tools` parameter that accepts raw dicts and forwards them directly to the Anthropic API alongside any function tools.

## Root Cause

Anthropic's `messages.create` API accepts two categories of tools:
1. **Custom function tools** — validated and converted by haystack's `Tool`/`Toolset` machinery
2. **Built-in server-side tools** (e.g. `{"type": "web_search_20250305"}`) — these bypass function-call dispatch and are handled entirely by the Anthropic backend

The existing `tools` parameter only supports category 1. Users who tried to pass raw dicts got a `TypeError: Items in the tools list must be Tool or Toolset instances`.

## Changes

| File | What changed |
|------|-------------|
| `chat_generator.py` | Added `anthropic_server_tools: list[dict[str, Any]] \| None = None` parameter; stored, serialized, and appended to the tools list in `_prepare_request_params` |
| `test_chat_generator.py` | Updated 3 existing serde tests to include the new param; added 5 new unit tests |

## Usage

```python
from haystack_integrations.components.generators.anthropic import AnthropicChatGenerator

generator = AnthropicChatGenerator(
    model="claude-sonnet-4-5",
    anthropic_server_tools=[{"type": "web_search_20250305"}],
)
result = generator.run(messages=[...])
```

Function tools and server-side tools can also be combined:

```python
generator = AnthropicChatGenerator(
    tools=[my_function_tool],
    anthropic_server_tools=[{"type": "web_search_20250305"}],
)
```

## Testing

- [x] Existing tests pass (updated `test_to_dict_default`, `test_to_dict_with_parameters`, `test_serde_in_pipeline` for new param)
- [x] New test: `test_init_with_anthropic_server_tools` — verifies parameter is stored
- [x] New test: `test_to_dict_with_anthropic_server_tools` — verifies serialization round-trip
- [x] New test: `test_from_dict_with_anthropic_server_tools` — verifies deserialization
- [x] New test: `test_run_with_anthropic_server_tools` — verifies server tools are passed to `messages.create`
- [x] New test: `test_run_with_tools_and_anthropic_server_tools` — verifies mixed function + server tools

## Impact

[...]

PR #3555, opened 2026-07-06
Support Anthropic's native server-side tools in AnthropicChatGenerator

### LINKED ISSUE #2724: Support for Anthropic native web_search tool
The Anthropic chat endpoint support web_search tool that gives the model access to web directly. 
Here is more info https://platform.claude.com/docs/en/agents-and-tools/tool-use/web-search-tool

At the moment, I don't find a way how to use it with AnthropicChatGenerator because the "tools" parameter assume it is Tool that is function in my codebase.

The `OpenAIResponsesChatGenerator` support passing "native" tools like this and I think similar is needed for the AnthropicChatGenerator.
```python
OpenAIResponsesChatGenerator(tools=[{"type": "web_search"}])
```

Tried to ignore the Python typing but it fails to error `TypeError: Items in the tools list must be Tool or Toolset instances.`

ErenAta16:
### Related Issues

- Fixes #2724

### Proposed Changes:

Anthropic offers server-side tools like `web_search`, `code_execution`, and `bash` that are defined as plain dicts (e.g. `{"type": "web_search_20250305", "name": "web_search"}`) rather than Haystack `Tool` objects, since Anthropic runs them itself instead of calling back into user code. There was no way to pass these through `AnthropicChatGenerator`: the `tools` parameter only accepted `Tool`/`Toolset` instances, and passing a raw dict tripped the Tool/Toolset flattening and duplicate-name check with `TypeError: Items in the tools list must be Tool or Toolset instances.`

`OpenAIResponsesChatGenerator` already solves the equivalent problem for OpenAI's own native tools by accepting `list[dict]` alongside its normal `ToolsType` and branching on whether the first element is a dict. This mirrors that approach for Anthropic:

- `__init__`, `to_dict`/`from_dict`, and `_prepare_request_params` (used by both `run` and `run_async`) now detect a list of dicts and pass it straight through to the Anthropic API unchanged, instead of routing it through the Haystack tool machinery.
- `AnthropicVertexChatGenerator` and `AnthropicFoundryChatGenerator` mirror the same handling in their own `__init__`/`to_dict`/`from_dict` and `run`/`run_async` overrides, since they don't delegate to the base class for these.

Usage:
```python
generator = AnthropicChatGenerator(tools=[{"type": "web_search_20250305", "name": "web_search"}])
```

### How did you test it?

- Added unit tests covering initialization, serialization round-trip, that `run()` forwards native tool dicts unchanged, and that native tools correctly skip the duplicate-name check (which only makes sense for Haystack `Tool` objects with a `.name`).
- Verified locally: all existing + new unit tests pass, `ruff check` / `ruff format --check` are clean, and `mypy` reports no new errors (the two pre-existing `anthropic.Anthropic`/`AsyncAnthropic` attr-defined errors are unrelated to this change and present on `main` too).

### Notes for the reviewer

The Vertex and Foundry generators subclass `AnthropicChatGenerator` but reimplement their own `__init__`/`to_dict`/`from_dict` instead of delegating to it, so they needed the same native-tools guard applied separately. That part is in the second and third commits.

### Checklist


davidsbatista:
Could you please update the comment and fill in our PR template to help the review? Also, please sign the CLA and then ping me again for a review

ErenAta16:
> Could you please update the comment and fill in our PR template to help the review? Also, please sign the CLA and then ping me again for a review

Updated the description to follow the PR template. The CLA is signed. Ready for another look whenever you have time — thanks!

davidsbatista:
There's a bug on the `from_dict` function, see the code below:

```python
from haystack_integrations.components.generators.anthropic import AnthropicChatGenerator

gen = AnthropicChatGenerator(tools=[])
d = gen.to_dict()
AnthropicChatGenerator.from_dict(d)
---------------------------------------------------------------------------
IndexError                                Traceback (most recent call last)
Cell In[5], line 1
----> 1 AnthropicChatGenerator.from_dict(d)

File ~/haystack-core-integrations/integrations/anthropic/src/haystack_integrations/components/generators/anthropic/chat/chat_generator.py:253, in AnthropicChatGenerator.from_dict(cls, data)
    249 tools = data["init_parameters"].get("tools")
    250 is_haystack_toolset = isinstance(tools, dict) and tools.get("type") == "haystack.tools.toolset.Toolset"
    251 is_haystack_tool_list = (
    252     isinstance(tools, list)
--> 253     and isinstance(tools[0], dict)
    254     and tools[0].get("type") == "haystack.tools.tool.Tool"
    255 )
    256 if tools and (is_haystack_toolset or is_haystack_tool_list):
    257     deserialize_tools_or_toolset_inplace(data["init_parameters"], key="tools")

IndexError: list index out of range
```

- If a component was initialized with `tools=[]` `to_dict()` serializes it to `tools=[]` (haystack's serialize_tools_or_toolset([]) returns [], not None). 

- Feeding that back into `from_dict()` raises IndexError: list index out of range, since isinstance(tools, list) passes but tools[0] indexes an empty list.


- Also, please refactor the checks ((is_native_tools, is_haystack_toolset, is_haystack_tool_list)) into a single function in `utils.py` that can then be reused by the 3 generators.

- The bug above is duplicated, having one shared helper used by all three generators) fixes it once, and any future tweak (e.g. supporting more native-tool shapes) only needs to happen in one place.

ErenAta16:
Good catch, thanks. Confirmed the exact issue: `to_dict()` serializes an empty tools list as `[]` (since `serialize_tools_or_toolset([])` returns `[]`, not `None`), and `from_dict()` indexed into `tools[0]` to check for the Haystack-tool-list shape before checking the list was non-empty. The outer `if tools and (...)` guard didn't help because `is_haystack_tool_list` was evaluated eagerly above it, so the `IndexError` happened before that guard was even reached.

Pushed a fix. Also did the refactor you suggested: moved the native-tools-list check and the deserialize-or-not check into two shared helpers in `utils.py` (`_is_native_tools_list` and `_should_deserialize_tools_or_toolset`), used by all three generators (`AnthropicChatGenerator`, `AnthropicFoundryChatGenerator`, `AnthropicVertexChatGenerator`) in their `__init__`, `to_dict`, and `from_dict`. The bug was in fact duplicated identically across all three `from_dict` implementations, so this fixes it in one place instead of three.

Added `test_serde_with_empty_tools_list`, which round-trips `AnthropicChatGenerator(tools=[])` through `to_dict()`/`from_dict()`. Ran it against the previous code first to confirm it reproduces your `IndexError`, then against the fix to confirm it passes, and re-ran the existing native-tools tests (`test_init_with_native_tools`, `test_serde_with_native_tools`, `test_native_tools_skip_duplicate_name_check`, `test_from_dict`) to make sure nothing else moved.

Let me know if you'd rather this be two separate commits (bugfix, then refactor) instead of one — happy to split it if that's easier to review.

ErenAta16:
CI is green now (the two intermediate failures were a lint docstring formatting issue and a mypy type-narrowing regression from the refactor, both fixed in the follow-up commits).

davidsbatista:
@ErenAta16 thanks for changes 👍🏽  

I've added  roundtrip unit tests for the other generators + live a integration test (only we can test it)

davidsbatista:
@ErenAta16 Sorry for this, but we internally decided to go for another approach and have two parameters for different types of tools, for which a PR was already open https://github.com/deepset-ai/haystack-core-integrations/pull/3386.

I was also not aware until after starting to review your PR.

ErenAta16:
No worries at all - appreciate you flagging the direction change before more review time went into this, and thanks for catching the from_dict bug and adding the roundtrip tests for the other generators along the way. Closing this in favor of #3386.

davidsbatista:
I will, but also, keep in mind to see if a PR is already open before starting to work on something, the #3386 was already open since Jun 3rd

ErenAta16:
Fair point, will do. Thanks for the review time on this either way.

PR #2891, opened 2026-02-27
fix: Remove unnecessary connection test and add `close`/`close_async` methods to `WeaviateDocumentStore`

maxdswain:
### Related Issues

- fixes #470

### Proposed Changes:

 
 
- Weaviate's python clients now automatically attempt to connect when `self._client.connect()`/`self._async_client.connect()` is called allowing both `self._client.collections.list_all(simple=True)` and `self._async_client.collections.list_all(simple=True)` to be removed.
- Add `close` and `async_close` methods to `WeaviateDocumentStore` mirroring what is done in `ValkeyDocumentStore` and `PineconeDocumentStore`.
- Updated relevant tests/fixtures to properly close connections to Weaviate.

### How did you test it?

Modified existing integration tests.

### Checklist

- I have read the [contributors guidelines](https://github.com/deepset-ai/haystack-core-integrations/blob/main/CONTRIBUTING.md) and the [code of conduct](https://github.com/deepset-ai/haystack-core-integrations/blob/main/CODE_OF_CONDUCT.md)
- I have updated the related issue with new insights and changes
- I added unit tests and updated the docstrings
- I've used one of the [conventional commit types](https://www.conventionalcommits.org/en/v1.0.0/) for my PR title: `fix:`, `feat:`, `build:`, `chore:`, `ci:`, `docs:`, `style:`, `refactor:`, `perf:`, `test:`.

PR #3556, opened 2026-07-07
Add close() and close_async() to QdrantDocumentStore

ErenAta16:
Part of #1628

QdrantDocumentStore lazily creates a `QdrantClient` and `AsyncQdrantClient` on first use but never released either one, so long-running processes and test suites had no way to release the underlying HTTP connections.

`WeaviateDocumentStore`, `PineconeDocumentStore`, and `ValkeyDocumentStore` already expose this same `close()`/`close_async()` pair (introduced in #2891), so this follows the same convention: each method releases its respective client and resets it to `None`, letting it be lazily re-initialized on the next call.

Usage after this PR:

    document_store = QdrantDocumentStore(...)
    try:
        document_store.write_documents(docs)
    finally:
        document_store.close()

Test plan: four new tests covering the sync and async client, both the normal case (initialize, close, verify it's `None`, confirm lazy re-initialization still works) and the safety case (calling close before the client was ever created should be a no-op). All use Qdrant's `:memory:` mode, so no external server is required.

Verified locally: all 260 existing + new tests pass, `ruff check` / `ruff format --check` are clean, and `mypy` reports no issues.

### DIFF
--- a/integrations/qdrant/src/haystack_integrations/components/retrievers/qdrant/retriever.py
+++ b/integrations/qdrant/src/haystack_integrations/components/retrievers/qdrant/retriever.py
@@ -130,6 +130,18 @@ class QdrantEmbeddingRetriever:
             data["init_parameters"]["filter_policy"] = FilterPolicy.from_str(filter_policy)
         return default_from_dict(cls, data)
 
+    def close(self) -> None:
+        """
+        Release the synchronous resources of the underlying Document Store.
+        """
+        self._document_store.close()
+
+    async def close_async(self) -> None:
+        """
+        Release the asynchronous resources of the underlying Document Store.
+        """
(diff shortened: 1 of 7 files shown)


anakin87 (changes requested):
Hey, thanks for the contribution!

Your PR looks generally good but I finally found the time to study the topic better, and I've come up with a general idea: https://github.com/deepset-ai/haystack-core-integrations/issues/1628#issuecomment-5034357021

Could you please align this PR with that? Should be relatively easy...

anakin87:
@ErenAta16 I pushed the alignment myself but could you please sign the CLA, so we can merge this PR?

anakin87:
Closing in favor of #3671 due to CLA not signed.

PR #3563, opened 2026-07-08
fix(langfuse): stop relying on `haystack.utils.base_serialization` helpers

davidsbatista:
### Related Issues

- related to https://github.com/deepset-ai/haystack/pull/11905#pullrequestreview-4646995270

### Proposed Changes:

- `serialize_class_instance`/`deserialize_class_instance` in `haystack.utils.base_serialization` are being removed from `haystack-ai` as dead code (see deepset-ai/haystack#11905). 

- `to_dict`: serialize `span_handler` with `component_to_dict` (from `haystack.core.serialization`) instead of `serialize_class_instance`.

- `from_dict`: deserialize `span_handler` with `deserialize_component_inplace` (from `haystack.utils`) instead of `deserialize_class_instance`

### How did you test it?

- Updated `test_langfuse_connector.py` fixtures to match the new (flatter) `span_handler` serialization format.
- Ran the full unit test suite, ruff, and mypy for the `langfuse` integration.

### Checklist

### DIFF
--- a/integrations/langfuse/tests/test_langfuse_connector.py
+++ b/integrations/langfuse/tests/test_langfuse_connector.py
@@ -104,10 +104,7 @@ class TestLangfuseConnector:
                 },
                 "span_handler": {
                     "type": "tests.test_langfuse_connector.CustomSpanHandler",
-                    "data": {
-                        "type": "tests.test_langfuse_connector.CustomSpanHandler",
-                        "init_parameters": {},
-                    },
+                    "init_parameters": {},
                 },
                 "host": "https://example.com",
                 "langfuse_client_kwargs": {"timeout": 30.0},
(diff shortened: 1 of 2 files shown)


integrations/langfuse/tests/test_langfuse_connector.py
```diff
@@ -199,10 +196,7 @@ def test_from_dict_with_params(self, monkeypatch):
                 },
                 "span_handler": {
                     "type": "tests.test_langfuse_connector.CustomSpanHandler",
-                    "data": {
-                        "type": "tests.test_langfuse_connector.CustomSpanHandler",
-                        "init_parameters": {},
-                    },
+                    "init_parameters": {},
```
anakin87, line 199: Would it be possible to keep the old serialization format working (handling this in `from_dict`)? Otherwise, this would break serialized pipelines

davidsbatista: What do you mean by the "old serialization format working"? Using the 

```python
def serialize_class_instance(obj: Any) -> dict[str, Any]:
```

The purpose of the PR is to avoid using that.

davidsbatista: or wait, you mean still being able to deserialise old formats?

anakin87: I mean: if a user has a serialized Pipeline with `LangfuseConnector` and `langfuse-haystack==5.0.2`, the new `from_dict` method would not be able to load it.

So, to ease the transition, we could modify the `from_dict` method so it does not break deserialization and add a test covering this case.

anakin87 (approved):
Thank you!

Please release 5.1.0 after merging

Issue #2340, opened 2025-09-19
MariaDB Vector DocumentStore Integration for Haystack


UmaisMu:
Title: RFC: MariaDB Vector DocumentStore Integration for Haystack

##Summary
I'm participating in the MariaDB Python Hackathon and planning to build a MariaDB Vector DocumentStore integration for Haystack 2.x. I'd appreciate feedback from the community.

## Background
- MariaDB 11.7+ now supports vector operations with VECTOR datatype and HNSW indexing
- Currently no Haystack DocumentStore implementation exists for MariaDB
- This would complete MariaDB's AI framework ecosystem alongside existing LangChain/LlamaIndex integrations

## Proposed Implementation
- Full DocumentStore protocol compliance (write_documents, filter_documents, delete_documents, count_documents)
- Vector similarity search using MariaDB's VEC_DISTANCE functions
- Metadata filtering combining SQL queries with vector search
- Production-ready features: connection pooling, error handling, batch operations

## Questions for Community
1. Would this integration be valuable for Haystack users?
2. Any specific implementation requirements or preferences?
3. Should this be part of haystack-integrations or separate package?
4. Any guidance on testing or documentation standards?

## Timeline
October 2025 development as part of MariaDB hackathon, targeting production-ready PyPI package.

Looking forward to your thoughts and guidance!

ShubhamGond105:
Hi @julian-risch ,
I'd like to take this up. I noticed @GunaPalanivel  expressed interest in March but no PR has been opened yet. I plan to follow the same approach — scaffold with python scripts/create_new_integration.py --name mariadb --type document_stores, model it after pgvector, and use MariaDB's VEC_DISTANCE and MHNSW indexing. I'll open a draft PR within the week.
Please let me know if this is still available!

ShubhamGond105:
Hi @julian-risch 
I've already the implementation and pushed an initial scaffold + document store skeleton to my fork:
https://github.com/ShubhamGond105/haystack-core-integrations/tree/feature/mariadb-document-store

My approach:
Modeled after the AlloyDB integration (same SQL + vector pattern, just merged into upstream)
Using mariadb Python connector with lazy connection and reconnect handling
VECTOR datatype with MHNSW index for vector similarity search
VEC_DISTANCE_COSINE and VEC_DISTANCE_EUCLIDEAN for similarity functions
JSON metadata (vs JSONB in PostgreSQL)

[...]

PR #3565, opened 2026-07-08
Add MariaDB Vector DocumentStore integration

SyedShahmeerAli12:
## Summary

Part of #2340

Implements a complete MariaDB document store integration using MariaDB 11.7+ native `VECTOR` support.

- `MariaDBDocumentStore`  full DocumentStore protocol (`write_documents`, `filter_documents`, `delete_documents`, `count_documents`)
- Vector similarity search using `VEC_DISTANCE_COSINE` / `VEC_DISTANCE_EUCLIDEAN` with `MHNSW` indexing
- Full-text keyword search via `MATCH ... AGAINST (IN NATURAL LANGUAGE MODE)` on a `FULLTEXT` index
- Haystack metadata filtering converted to `JSON_UNQUOTE(JSON_EXTRACT(...))` SQL expressions with parameterized queries
- `MariaDBEmbeddingRetriever` and `MariaDBKeywordRetriever` with `FilterPolicy` support
- `DuplicatePolicy` support: `FAIL` (INSERT), `OVERWRITE` (upsert via `ON DUPLICATE KEY UPDATE`), `SKIP` (`INSERT IGNORE`)
- Lazy connection with reconnect on ping failure

## Tests

80 tests total  all passing:
- 68 unit tests (mocked DB, no external dependency)
- 12 integration tests verified against a real **MariaDB 11.7** Docker container

## CI

Added `.github/workflows/mariadb.yml` with a MariaDB 11.7 service container, matching the pattern used by `pgvector`.

## How to run locally

```bash
docker run -d --name mariadb-haystack \
  -e MARIADB_ROOT_PASSWORD=password \
  -e MARIADB_DATABASE=haystack \
  -p 3306:3306 mariadb:11.7

cd integrations/mariadb
pip install -e .
MARIADB_USER=root MARIADB_PASSWORD=password pytest tests/ -m integration
```


integrations/mariadb/CHANGELOG.md
```diff
@@ -0,0 +1,10 @@
+# Changelog
```
anakin87: this file is auto-generated and updated at release time.
Remove it.

integrations/mariadb/README.md
```diff
@@ -0,0 +1,77 @@
+# MariaDB Document Store for Haystack
```
anakin87: This should be minimal, like other READMEs in this repo.
See for example https://github.com/deepset-ai/haystack-core-integrations/tree/main/integrations/pgvector

integrations/mariadb/pyproject.toml
```diff
@@ -0,0 +1,169 @@
...
+tag-pattern = 'integrations\/mariadb-v(?P<version>.*)'
+
+[tool.hatch.version.raw-options]
+root = "../.."
+git_describe_command = 'git describe --tags --match="integrations/mariadb-v[0-9]*"'
+
+[tool.hatch.envs.default]
+installer = "uv"
+# skip-install prevents hatch from installing the project (and thus the mariadb C extension)
+# in the docs environment. The pydoc search_path points to src/ so modules are found directly.
```
anakin87: what's the problem here?

integrations/mariadb/tests/test_document_store.py
```diff
@@ -0,0 +1,401 @@
+# SPDX-FileCopyrightText: 2023-present deepset GmbH <info@deepset.ai>
```
anakin87, line 1: For integration tests, we should use classes from `haystack.testing.document_store` ( `DocumentStoreBaseTests` and `DocumentStoreExtendedTests` or their inner classes). Check Pgvector for an example.

integrations/mariadb/src/haystack_integrations/document_stores/mariadb/document_store.py
```diff
@@ -0,0 +1,482 @@
...
+    def _ensure_connection(self) -> None:
+        """Lazily establish the DB connection and initialize the table."""
+        if self._connection is not None and self._cursor is not None:
+            try:
+                self._connection.ping()
+                return
+            except Exception:
+                self._close_connection()
+
+        import mariadb  # noqa: PLC0415
```
anakin87: `mariadb` should be imported at the top of the module

integrations/mariadb/src/haystack_integrations/document_stores/mariadb/document_store.py
```diff
@@ -0,0 +1,482 @@
...
+
+VECTOR_FUNCTION_TO_SQL = {
+    "cosine_similarity": "VEC_DISTANCE_COSINE",
+    "l2_distance": "VEC_DISTANCE_EUCLIDEAN",
+}
+
+CREATE_TABLE_STATEMENT = """
+CREATE TABLE IF NOT EXISTS `{table_name}` (
+    id VARCHAR(128) PRIMARY KEY,
+    embedding VECTOR({embedding_dimension}) COMMENT 'MHNSW(M=16)',
```
anakin87: this does not seem to actually use HNSW. See https://mariadb.com/docs/server/reference/sql-structure/vectors/create-table-with-vectors

 Making it a real index hits a constraint
> Note that there can be only one vector index in the table, and the indexed vector column must be NOT NULL.

So if we want to use a vector index, every Document `embedding` must be not null, but this can conflict with users who are only interested in keyword search.

I propose adding a `create_vector_index` flag to init, to make users explicitly choose what they want.

Please investigate this topic and discuss approaches with me.

integrations/mariadb/src/haystack_integrations/document_stores/mariadb/document_store.py
```diff
@@ -0,0 +1,482 @@
...
+from haystack.dataclasses import ByteStream, Document
+from haystack.document_stores.errors import DocumentStoreError, DuplicateDocumentError
+from haystack.document_stores.types import DocumentStore, DuplicatePolicy
+from haystack.utils.auth import Secret, deserialize_secrets_inplace
+
+from .filters import _convert_filters_to_where_clause_and_params, _validate_filters
+
+logger = logging.getLogger(__name__)
+
+VALID_VECTOR_FUNCTIONS = ["cosine_similarity", "l2_distance"]
```
anakin87: I'd use `cosine` and `euclidean` as per MariaDB docs. Or are there reasons to diverge?

integrations/mariadb/src/haystack_integrations/document_stores/mariadb/document_store.py
```diff
@@ -0,0 +1,482 @@
...
+CREATE TABLE IF NOT EXISTS `{table_name}` (
+    id VARCHAR(128) PRIMARY KEY,
+    embedding VECTOR({embedding_dimension}) COMMENT 'MHNSW(M=16)',
+    content LONGTEXT,
+    blob_data LONGBLOB,
+    blob_meta JSON,
+    blob_mime_type VARCHAR(255),
+    meta JSON,
+    FULLTEXT KEY content_ft_idx (content)
+) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4
```
anakin87, line 42: We should also set the distance function at table creation.

Also, let's make distance not changeable after table creation:
> Declare DISTANCE explicitly. The default is euclidean, and a query using a different distance function than the one the index was built for cannot use the index, it falls back to a full table scan. 

See https://mariadb.com/docs/server/reference/sql-structure/vectors/create-table-with-vectors

integrations/mariadb/src/haystack_integrations/document_stores/mariadb/document_store.py
```diff
@@ -0,0 +1,482 @@
...
+        written = 0
+        for doc in documents:
+            row = _document_to_row(doc)
+            try:
+                self._cursor.execute(sql, row)
+                written += 1
+            except mariadb.IntegrityError as e:
+                if policy == DuplicatePolicy.FAIL:
+                    msg = f"Document with id '{doc.id}' already exists"
+                    raise DuplicateDocumentError(msg) from e
```
anakin87:
```python
    try:
        self._connection.begin()
        self._cursor.executemany(sql, rows)
        written = self._cursor.rowcount
        self._connection.commit()
    except mariadb.IntegrityError as e:
        self._connection.rollback()
        msg = "Some documents already exist and policy is FAIL"
        raise DuplicateDocumentError(msg) from e
    except mariadb.Error as e:
        self._connection.rollback()
        raise DocumentStoreError("Failed to write documents") from e
```

Doing something like this would ensure both atomic writes and exact numbers for `DuplicatePolicy.SKIP`

integrations/mariadb/src/haystack_integrations/document_stores/mariadb/filters.py
```diff
@@ -0,0 +1,193 @@
...
+    return clause, [val]
+
+
+def _build_field_expr(field: str, value: Any) -> str:
+    """Return a safe SQL expression for a document field or meta JSON path."""
+    if field.startswith("meta."):
+        field_name = field.split(".", 1)[1]
+        # JSON_UNQUOTE(JSON_EXTRACT(meta, '$.field')) returns the value as a string.
+        # Cast to numeric types when the comparison value is numeric.
+        base = f"JSON_UNQUOTE(JSON_EXTRACT(meta, '$.{field_name}'))"
```
anakin87, line 106: this might be prone to SQL injection. Let's make it more robust

integrations/mariadb/src/haystack_integrations/document_stores/mariadb/document_store.py
```diff
@@ -0,0 +1,482 @@
...
+    meta = VALUES(meta)
+"""
+
+KEYWORD_QUERY = """
+SELECT *, MATCH(content) AGAINST(? IN NATURAL LANGUAGE MODE) AS score
+FROM `{table_name}`
+{where_clause}
+ORDER BY score DESC
+LIMIT ?
+"""
```
anakin87, line 91: this might also return irrelevant documents if relevant ones are less than `top_k`
we should also apply matching on `WHERE` or filtering by score

integrations/mariadb/src/haystack_integrations/document_stores/mariadb/document_store.py
```diff
@@ -0,0 +1,482 @@
...
+            self._connection = None
+        self._table_initialized = False
+
+    def close(self) -> None:
+        """Close the database connection."""
+        self._close_connection()
+
+    def __del__(self) -> None:
+        self._close_connection()
+
```
anakin87: I'd avoid this. `close` is sufficient.

anakin87 (changes requested):
I took a first look and found some points to address and some others to discuss.

integrations/mariadb/pyproject.toml
SyedShahmeerAli12: mariadb==1.1.14 uses distutils internally, which is removed in Python 3.14 setuptools provides the shim to keep the test env working.

integrations/mariadb/src/haystack_integrations/document_stores/mariadb/document_store.py
SyedShahmeerAli12: The lazy import is intentional  top-level import triggers the C extension which fails on Python 3.14 due to the distutils removal; placing it after the early-return ping check means unit tests with mocked connections never hit it.

integrations/mariadb/src/haystack_integrations/document_stores/mariadb/document_store.py
SyedShahmeerAli12: Agreed .... will add a create_vector_index flag (default False) so users opt-in; when False the column stays nullable and no MHNSW index is created, keeping keyword-only use cases working., or what you suggest ?

SyedShahmeerAli12:
Addressed all the straightforward review comments. The HNSW index design (`create_vector_index` flag) and the lazy import approach are still open for discussion

integrations/mariadb/src/haystack_integrations/document_stores/mariadb/document_store.py
anakin87: Yes, seems to make sense

integrations/mariadb/tests/test_document_store.py
anakin87: `FilterDocumentsTest` should also be used

integrations/mariadb/pyproject.toml
anakin87: This seems like a workaround. Can we use the solution suggested in https://mariadb.com/docs/connectors/mariadb-connector-python/faq#modulenotfounderror-no-module-named-packaging?

integrations/mariadb/src/haystack_integrations/document_stores/mariadb/document_store.py
anakin87: See https://github.com/deepset-ai/haystack-core-integrations/pull/3565#discussion_r3636706088

anakin87 (changes requested):
There are still some comments to address.

Please request my review when they are fixed. In the meantime, ask questions if needed.

SyedShahmeerAli12:
ruff auto-detects `mariadb `and haystack as separate import sections, so `import mariadb` must go in its own block after all `from haystack` imports.
 Every attempt I made grouped them together, which ruff kept rejecting until I let` --fix` show me the exact layout it expected.

SyedShahmeerAli12:
I mistakently Referenced some PRs here ,  Well I have cleaned the mess the Code is opened for review

integrations/mariadb/src/haystack_integrations/document_stores/mariadb/document_store.py
```diff
@@ -0,0 +1,510 @@
...
+SELECT *, {vec_func}(embedding, ?) AS score
+FROM `{table_name}`
+WHERE embedding IS NOT NULL
+{extra_where}
+ORDER BY score ASC
+LIMIT ?
+"""
+
+
+class MariaDBDocumentStore(DocumentStore):
```
anakin87:
```suggestion
class MariaDBDocumentStore:
```
`DocumentStore` is a Python protocol. We generally don't inherit from it.
To stick to the protocol, only implementing the necessary methods is required (which you're already doing).

integrations/mariadb/pyproject.toml
anakin87: Since users will need `packaging` to use `mariadb`, I'd add this package to main `dependencies` (not `test`). Does it make sense?

integrations/mariadb/src/haystack_integrations/document_stores/mariadb/document_store.py
```diff
@@ -0,0 +1,510 @@
...
+    id VARCHAR(128) PRIMARY KEY,
+    embedding VECTOR({embedding_dimension}) NOT NULL,
+    content LONGTEXT,
+    blob_data LONGBLOB,
+    blob_meta JSON,
+    blob_mime_type VARCHAR(255),
+    meta JSON,
+    FULLTEXT KEY content_ft_idx (content),
+    VECTOR INDEX vec_idx (embedding) COMMENT 'MHNSW(distance={distance})'
+) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4
```
anakin87, line 56: this does not seem to match the syntax in https://mariadb.com/docs/server/reference/sql-structure/vectors/create-table-with-vectors

integrations/mariadb/src/haystack_integrations/document_stores/mariadb/document_store.py
anakin87: unattended

integrations/mariadb/src/haystack_integrations/document_stores/mariadb/document_store.py
```diff
@@ -0,0 +1,510 @@
...
+            raise DocumentStoreError(msg) from e
+
+    def _embedding_retrieval(
+        self,
+        query_embedding: list[float],
+        *,
+        filters: dict[str, Any] | None = None,
+        top_k: int = 10,
+        score_threshold: float | None = None,
+        vector_function: str | None = None,
```
anakin87: > Searches using a different distance function will not be able to use a vector index

I'd simply not allow users to change the distance function at runtime, in other words by removing this parameter

integrations/mariadb/src/haystack_integrations/document_stores/mariadb/document_store.py
```diff
@@ -0,0 +1,510 @@
...
+    """
+
+    def __init__(
+        self,
+        *,
+        host: str = "localhost",
+        port: int = 3306,
+        database: str = "haystack",
+        user: Secret | str = Secret.from_env_var("MARIADB_USER"),
+        password: Secret | str = Secret.from_env_var("MARIADB_PASSWORD"),
```
anakin87:
```suggestion
        user: Secret = Secret.from_env_var("MARIADB_USER"),
        password: Secret = Secret.from_env_var("MARIADB_PASSWORD"),
```

We don't want to support strings

integrations/mariadb/tests/test_document_store.py
```diff
@@ -0,0 +1,361 @@
...
+from haystack_integrations.document_stores.mariadb.document_store import (
+    _bytes_to_embedding,
+    _document_to_row,
+    _embedding_to_bytes,
+    _rows_to_documents,
+)
+
+# ---------------------------------------------------------------------------
+# Helper: build a store with a mocked DB connection
+# ---------------------------------------------------------------------------
```
anakin87:
```suggestion
```

Remove all these separation comments

anakin87 (changes requested):
There are still a few rough edges to address in this PR, including some points that were left unresolved from the previous review.

Since this is a new integration, I'd like to see a more thorough pass: try it yourself, refine the implementation, and make sure to address the existing feedback.
 Otherwise, I may need to deprioritize further review.

SyedShahmeerAli12:
Hi @anakin87 , I've addressed all of your review comments:

* Removed `DocumentStore` protocol inheritance
* Fixed the import order
* Updated the vector index SQL to use `VECTOR INDEX (embedding)` (removed `COMMENT 'MHNSW(...)'`)
* Added `HAVING score > 0` to the keyword search to filter out zero-score results
* Changed `user`/`password` to use `Secret` only (removed the plain string fallback)
* Removed the `vector_function` parameter from `_embedding_retrieval` (and the retriever)
* Moved `packaging>=21.3` to the main dependencies
* Removed all section separator comments from the tests
* Updated all test fixtures to use `Secret.from_token()`

I also verified everything against a real MariaDB 11.7 Docker instance all 120 tests are passing (65 unit + 55 integration).

integrations/mariadb/pyproject.toml
```diff
@@ -0,0 +1,169 @@
...
+skip-install = true
+dependencies = ["haystack-pydoc-tools", "ruff", "haystack-ai>=2.28.0"]
```
anakin87:
```suggestion
dependencies = ["haystack-pydoc-tools", "ruff"]
```

integrations/mariadb/src/haystack_integrations/document_stores/mariadb/document_store.py
anakin87: Sorry for the misunderstanding. I meant:
- let's make `distance` configurable in `MariaDBDocumentStore.__init__` and take this parameter into account only when the table is created
- do not expose `distance` in `_embedding_retrieval`

integrations/mariadb/src/haystack_integrations/document_stores/mariadb/filters.py
```diff
@@ -0,0 +1,206 @@
...
+
+    operator: str = condition["operator"]
+    if operator not in COMPARISON_OPERATORS:
+        msg = f"Unknown comparison operator '{operator}'. Valid operators are: {list(COMPARISON_OPERATORS.keys())}"
+        raise FilterError(msg)
+
+    value: Any = condition["value"]
+    sql_field = _build_field_expr(field, value)
+    clause, val = COMPARISON_OPERATORS[operator](sql_field, value)
+    # _in/_not_in return the value list directly; flatten it so params stay flat
```
anakin87, line 88: currently boolean meta filters seem to match nothing.
Try for example the `condition` `{"field": "meta.flag", "operator": "==", "value": True}`

integrations/mariadb/src/haystack_integrations/document_stores/mariadb/document_store.py
```diff
@@ -0,0 +1,496 @@
...
+        """
+        Retrieve documents by vector similarity.
+
+        :param query_embedding: Query vector.
+        :param filters: Optional Haystack filters.
+        :param top_k: Maximum results.
+        :param score_threshold: Minimum score to include a document. Documents below this score are excluded.
+        :returns: List of Documents ordered by similarity (most similar first).
+        """
+        _validate_filters(filters)
```
anakin87, line 425: MariaDB does not raise an error if the `query_embedding` does not match the default embedding dimension but just returns records with null score. I propose raising an error ourselves.

anakin87 (changes requested):

SyedShahmeerAli12:
I did all the testing against MariaDB 11.7via  Docker

integrations/mariadb/pyproject.toml
anakin87: sorry, I now understand the issue

anakin87 (approved):
I pushed some minor changes and cleaned up tests.

Time to merge!

PR #3571, opened 2026-07-09
feat(google_genai): add google_server_tools to GoogleGenAIChatGenerator

Aftabbs:
## Summary

- Adds `google_server_tools: list[dict[str, Any]] | None = None` parameter to `GoogleGenAIChatGenerator.__init__`
- Enables users to pass Google's built-in server-side tools (e.g. `google_search`, `code_execution`) to the Gemini API
- Server tools are merged with Haystack `Tool`/`Toolset` at request time and passed in `GenerateContentConfig.tools`
- Serializes/deserializes cleanly via `to_dict()` / `from_dict()`

## Motivation

Google's Gemini API supports built-in tools that are not Haystack `Tool` objects and cannot be expressed as function declarations. For example:

```python
from google.genai import types
# Native Google search tool
types.Tool(google_search=types.GoogleSearch())

# Code execution tool
types.Tool(code_execution=types.ToolCodeExecution())
```

Currently there is no way to pass these through `GoogleGenAIChatGenerator` because the `tools` parameter only accepts `ToolsType` (Haystack Tools/Toolsets). This PR adds `google_server_tools` to fill that gap.

**Usage:**

```python
generator = GoogleGenAIChatGenerator(
    model="gemini-2.5-flash",
    google_server_tools=[{"google_search": {}}],
)
```

Or combined with Haystack tools:

```python
generator = GoogleGenAIChatGenerator(
    model="gemini-2.5-flash",
    tools=[my_haystack_tool],
    google_server_tools=[{"google_search": {}}, {"code_execution": {}}],
)
```

## Relation to existing work

Follows the same pattern as `anthropic_server_tools` introduced in #3386 for `AnthropicChatGenerator`.

## Test plan

- [ ] `test_init_with_google_server_tools` — verifies `_google_server_tools` is stored on init
- [ ] `test_to_dict_with_google_server_tools` — verifies serialization round-trip
- [ ] `test_from_dict_with_google_server_tools` — verifies deserialization round-trip
- [ ] `test_to_dict_default_google_server_tools_is_none` — verifies default is `None`
- [ ] `test_run_with_google_server_tools` — verifies native tool is passed to `GenerateContentConfig`
- [ ] `test_run_merges_haystack_and_google_server_tools` — verifies Haystack + native tools are merged

### DIFF
--- a/integrations/google_genai/src/haystack_integrations/components/generators/google_genai/chat/chat_generator.py
+++ b/integrations/google_genai/src/haystack_integrations/components/generators/google_genai/chat/chat_generator.py
@@ -201,6 +201,7 @@ class GoogleGenAIChatGenerator:
         safety_settings: list[dict[str, Any]] | None = None,
         streaming_callback: StreamingCallbackT | None = None,
         tools: ToolsType | None = None,
+        google_server_tools: list[dict[str, Any]] | None = None,
         timeout: float | None = None,
         max_retries: int | None = None,
     ) -> None:
@@ -236,6 +237,10 @@ class GoogleGenAIChatGenerator:
         :param streaming_callback: A callback function that is called when a new token is received from the stream.
         :param tools: A list of Tool and/or Toolset objects, or a single Toolset for which the model can prepare calls.
             Each tool should have a unique name.
+        :param google_server_tools: A list of Google server-side (built-in) tools passed directly to the API.
+            Use this for native Google tools such as `{"google_search": {}}` or `{"code_execution": {}}`.
+            Each entry must be a dict that matches the `google.genai.types.Tool` schema.
+            These are merged with any Haystack tools at request time.
         :param timeout:
             Timeout for Google GenAI client calls. If not set, it defaults to the default set by the Google GenAI
             client.
@@ -263,6 +268,7 @@ class GoogleGenAIChatGenerator:
         self._safety_settings = safety_settings or []
         self._streaming_callback = streaming_callback
         self._tools = tools
+        self._google_server_tools = google_server_tools
         self._timeout = timeout
         self._max_retries = max_retries
 
@@ -291,6 +297,7 @@ class GoogleGenAIChatGenerator:
             safety_settings=self._safety_settings,
             streaming_callback=callback_name,
             tools=serialized_tools,
+            google_server_tools=self._google_server_tools,
             timeout=self._timeout,
             max_retries=self._max_retries,
         )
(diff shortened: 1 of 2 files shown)


integrations/google_genai/src/haystack_integrations/components/generators/google_genai/chat/chat_generator.py
```diff
@@ -236,6 +237,10 @@ def __init__(
         :param streaming_callback: A callback function that is called when a new token is received from the stream.
         :param tools: A list of Tool and/or Toolset objects, or a single Toolset for which the model can prepare calls.
             Each tool should have a unique name.
+        :param google_server_tools: A list of Google server-side (built-in) tools passed directly to the API.
```
anakin87, line 240: Could you explain how this syntax works?

We should support all [Google built-in tools](https://ai.google.dev/gemini-api/docs/tools#build-in-tools) if possible, also allowing to pass potential parameters they accept.

So let's verify that this syntax is functional for this goal or choose another one; also provide users with an example in docstring and a link to Google tools.

integrations/google_genai/tests/test_chat_generator.py
```diff
@@ -299,6 +299,33 @@ def test_serde_with_response_format(self, monkeypatch):
         assert restored._generation_kwargs["response_format"] == schema
         assert restored._generation_kwargs["temperature"] == 0.5
 
+    def test_init_with_google_server_tools(self, monkeypatch):
```
anakin87: Let's incorporate these new tests into existing tests or at least move them close to other similar ones

integrations/google_genai/tests/test_chat_generator.py
```diff
@@ -365,6 +392,34 @@ def test_run_with_tools_passes_tools_to_config(self, monkeypatch, mock_response,
...
+
+        component.run([ChatMessage.from_user("Search for AI news")], tools=tools)
+
+        call_kwargs = component._client.models.generate_content.call_args
+        config = call_kwargs.kwargs.get("config") or call_kwargs[1].get("config")
+        assert config.tools is not None
+        # _convert_tools_to_google_genai_format bundles all function declarations into one Tool object,
+        # plus one Tool object for google_search
+        assert len(config.tools) == 2
+
```
anakin87, line 438: Let's add a live integration test in the appropriate section.

anakin87 (changes requested):
Thanks for the contribution.

I found some opportunities for improvement.

Quick question: have you tried this feature locally yourself?

Aftabbs:
Yes — I tested the parameter wiring and serialization locally (unit tests pass). For the live API path, I've added `test_live_run_with_google_server_tools` to `TestGoogleGenAIChatGeneratorInference` which exercises the full end-to-end flow when `GOOGLE_API_KEY` is present. Since integration tests require an API key (which fork PRs don't have in CI), the live test is guarded and will only run with a real key. The unit tests cover the serialization, deserialization, and that the `google_server_tools` list reaches the request params correctly.

PR #3580, opened 2026-07-11
fix(cohere): include citations in streaming ChatMessage meta

Aftabbs:
## Summary

`CohereChatGenerator` exposes citations in non-streaming responses via
`message.meta["citations"]` (from `chat_response.message.citations`), but
the streaming path silently discards this data.

**Root cause**: `_convert_cohere_chunk_to_streaming_chunk` had no handler
for `citation-start` chunks, so citation data was lost. Additionally,
`_parse_streaming_response` and `_parse_async_streaming_response` never
aggregated citation data into the final `ChatMessage.meta`.

**Fix**:
1. Add a `citation-start` branch in `_convert_cohere_chunk_to_streaming_chunk`
   that captures the `Citation` object from `chunk.delta.message.citations`
   into `meta["citations"]` on the per-chunk `StreamingChunk`.
2. After `_convert_streaming_chunks_to_chat_message`, both streaming parsers
   now collect those per-chunk citations and write them to
   `message.meta["citations"]` (set to `None` when absent) — matching the
   non-streaming behaviour.

## How to reproduce the bug

```python
from haystack_integrations.components.generators.cohere import CohereChatGenerator

chunks = []
gen = CohereChatGenerator(model="command-a-03-2025", streaming_callback=chunks.append)
result = gen.run(messages=[ChatMessage.from_user("Who invented the telephone?")])
# Before fix: result["replies"][0].meta has no "citations" key
# After fix:  result["replies"][0].meta["citations"] == [Citation(...), ...]
```

## Testing

- Updated `test_parse_response` to include `"citations": None` in expected
  meta for streams with no citation events (keeps existing assertion exact).
- Added `test_convert_citation_start_chunk`: verifies the chunk converter
  stores the `Citation` object in `meta["citations"]`.
- Added `test_parse_streaming_response_with_citations`: end-to-end check
  that `_parse_streaming_response` surfaces citations in the final
  `ChatMessage.meta`.

All 8 unit tests pass.


davidsbatista (approved):
looks good, thanks for the contribution

PR #3586, opened 2026-07-13
fix(pgvector): parenthesize `not in` clause so it does not escape an AND filter

chuenchen309:
### The bug

In the pgvector filter translation, `_not_in` renders:

```python
return SQL("{} IS NULL OR {} != ALL(%s)").format(field, field), [value]
```

with no surrounding parentheses. `not in` is the only comparison operator that produces an internal `OR`. When `_parse_logical_condition` joins sibling conditions with ` AND `, that ungrouped `OR` escapes the `AND` scope, because in SQL `AND` binds tighter than `OR`.

So this filter:

```python
{
    "operator": "AND",
    "conditions": [
        {"field": "meta.x", "operator": "==", "value": 5},
        {"field": "meta.author", "operator": "not in", "value": ["John", "Jane"]},
    ],
}
```

renders as:

```sql
WHERE ((meta->>'x')::integer = %s AND meta->>'author' IS NULL OR meta->>'author' != ALL(%s))
```

which SQL parses as `(x = 5 AND author IS NULL) OR (author != ALL(...))`. A document with `x = 999, author = 'Bob'` **wrongly matches** — the `not in` branch is OR'd against the entire rest of the `AND`, instead of being one required conjunct. The intended semantics are `x = 5 AND author not-in-list`.

This stayed hidden because the existing `test_logical_condition_nested` only nests `not in` under an `OR` (where the missing parentheses are harmless), and the base filter tests only AND-combine `==`.

### The fix

Wrap the expression in parentheses so the internal `OR` stays grouped:

```python
return SQL("({} IS NULL OR {} != ALL(%s))").format(field, field), [value]
```

### Tests

- Added `test_logical_condition_not_in_is_parenthesized_when_and_combined`, which AND-combines `not in` with another condition and asserts the `not in` clause is parenthesized. It fails before the change (the `OR` escapes the `AND`) and passes after.
- Updated `test_logical_condition_nested`'s expected SQL to include the now-parenthesized `not in` clause (semantically unchanged — it was already inside an `OR`).

Verified locally: `hatch run test:unit` passes (71 passed), `hatch run fmt` and `hatch run test:types` are clean.

---

### DIFF
--- a/integrations/pgvector/src/haystack_integrations/document_stores/pgvector/filters.py
+++ b/integrations/pgvector/src/haystack_integrations/document_stores/pgvector/filters.py
@@ -229,7 +229,7 @@ def _not_in(field: Composable, value: Any) -> tuple[Composed, list]:
     if not isinstance(value, list):
         msg = f"{field}'s value must be a list when using 'not in' comparator in Pinecone"
         raise FilterError(msg)
-    return SQL("{} IS NULL OR {} != ALL(%s)").format(field, field), [value]
+    return SQL("({} IS NULL OR {} != ALL(%s))").format(field, field), [value]
 
 
 def _in(field: Composable, value: Any) -> tuple[Composed, list]:
--- a/integrations/pgvector/tests/test_filters.py
+++ b/integrations/pgvector/tests/test_filters.py
@@ -228,11 +228,30 @@ def test_logical_condition_nested():
         "("
         "(meta->>'domain' IS DISTINCT FROM %s OR meta->>'chapter' = ANY(%s))"
         " AND "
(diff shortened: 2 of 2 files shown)


sjrl:
Hey @chuenchen309 thanks for opening the PR! 

Please sign the CLA https://github.com/deepset-ai/haystack-core-integrations/pull/3586#issuecomment-4957879224 and then I'll go ahead with a review.

chuenchen309:
Hi @sjrl — looks like the CLA check is passing now ("Contributor License Agreement is signed", https://cla-assistant.io/deepset-ai/haystack-core-integrations?pullRequest=3586). Let me know if there's anything else needed on my end for review.

sjrl (approved):
Thanks for the fix!

PR #3590, opened 2026-07-13
fix(azure_ai_search): escape single quotes in OData string literals

chuenchen309:
### The bug

The Azure AI Search filter builders `_eq`, `_ne`, and `_in` interpolate string filter values directly into single-quoted OData literals, without escaping:

```python
def _eq(field, value):
    if isinstance(value, str) and value != "null":
        return f"{field} eq '{value}'"

def _in(field, value):
    ...
    values = ",".join(map(str, value))
    return f"search.in({field},'{values}',',')"
```

Any metadata value that contains an apostrophe — names (`O'Brien`), titles, possessives — produces a malformed `$filter`:

```python
{"field": "meta.author", "operator": "==", "value": "O'Brien"}
# -> author eq 'O'Brien'      (the embedded ' closes the literal early)

{"field": "meta.name", "operator": "in", "value": ["O'Brien", "d'Arc"]}
# -> search.in(name,'O'Brien,d'Arc',',')
```

Azure AI Search rejects these as syntactically invalid OData (HTTP 400), so a perfectly ordinary equality/`in` filter fails whenever a value contains a single quote.

### The fix

In OData, a single quote inside a string literal is escaped by doubling it (`'` → `''`). Apply that in all three affected operators (via a small shared helper), so `O'Brien` becomes the valid literal `'O''Brien'`.

I fixed the three operators that interpolate string literals because reproducing showed they share the exact same defect. (The separate `search.in` comma-as-delimiter limitation — values containing a literal comma — is a different issue and out of scope here.)

### Test

Extended the parametrized `test_normalize_filters_comparison_conditions` with `==`, `!=`, and `in` cases whose values contain apostrophes. They fail before the change (unescaped quotes) and pass after.

Verified locally: `hatch run test:unit` passes (87 passed), `hatch run fmt` and `hatch run test:types` are clean.

---


davidsbatista (approved):
looks good, thanks for the contribution

PR #3591, opened 2026-07-14
chore: move agent pack to core integrations

anakin87:
### Proposed Changes:
- move agent pack to core integrations - part of https://github.com/deepset-ai/haystack-private/issues/466

### How did you test it?
CI; also tested separately in a Colab env

### Notes for the reviewer
I don't plan to release a new package until we have finally decided on a name but we can merge this PR in the meantime when it looks good.

### Checklist


integrations/agent_pack/src/agent_pack_haystack/__init__.py
```diff
@@ -0,0 +1,9 @@
+# SPDX-FileCopyrightText: 2022-present deepset GmbH <info@deepset.ai>
+#
+# SPDX-License-Identifier: Apache-2.0
+
+"""Short-import alias for the agent pack. Allows `from agent_pack_haystack import create_deep_research_agent`."""
+
```
anakin87: This diverges from our standard (Docling is the only other exception), but does not break the general namespace package assumption of Core Integrations and `from agent_pack_haystack import create_deep_research_agent` seems convenient to me. WDYT?

sjrl: Yeah I like it

integrations/agent_pack/pydoc/config_docusaurus.yml
```diff
@@ -0,0 +1,13 @@
+loaders:
+  - modules:
+      - haystack_integrations.components.agents.agent_pack.deep_research.agent
```
sjrl: I'd think we'd also want to surface the tools and hooks at least in the API docs as well. What do you think?

integrations/agent_pack/src/haystack_integrations/agent_pack/deep_research/hooks.py
```diff
@@ -0,0 +1,103 @@
...
+        state.set("brief", brief)
+        state.set("messages", [ChatMessage.from_user(brief)])
+
+
+class WriteHook:
+    """Write the final report from the brief and notes, once. Runs as an `on_exit` hook."""
+
+    # using an `after_run` hook point would be better and would avoid producing empty reports if the orchestrator
+    # exhausts its step budget
+    allowed_hook_points = ("on_exit",)
```
sjrl: For a different PR but before_run and after_run have been added as hook points to haystack main so feel free to use them now.

integrations/agent_pack/src/agent_pack_haystack/__init__.py
```diff
@@ -0,0 +1,9 @@
...
+from haystack_integrations.components.agents.agent_pack import create_deep_research_agent
```
sjrl: Feels like a kind of long import to have. Could we also not just do something like `haystack_integrations.agent_pack`? Just wondering if we really need to nest it all the way down into `components.agents.agent_pack`. WDYT?

integrations/agent_pack/src/haystack_integrations/components/agents/agent_pack/deep_research/tools.py
```diff
@@ -0,0 +1,150 @@
...
+
+def make_web_search_tool(*, max_search_results: int) -> ComponentTool:
+    """
+    `web_search` = ComponentTool over TavilyWebSearch (async-capable).
+
+    :param max_search_results: Results returned per `web_search` call.
+    """
+    tavily_import.check()
+    return ComponentTool(
+        component=TavilyWebSearch(top_k=max_search_results),
```
sjrl: Sorry this didn't come up before in my review, but I wonder if this would be better as a new class. Something like 
```
class TavilyWebSearchTool(top_k=max_search_results):
    ...
```
we could then export this as a pre-made tool for others to easily reuse potentially. Perhaps even in the tavily integration. WDYT?

integrations/agent_pack/pydoc/config_docusaurus.yml
anakin87: My first thought was to offer factory functions as primary entry points.
Why would you like to surface tools and hooks? To make them reusable, to document how the Agent works...?

integrations/agent_pack/src/haystack_integrations/agent_pack/deep_research/hooks.py
anakin87: I'll do in another PR

integrations/agent_pack/src/agent_pack_haystack/__init__.py
anakin87: I initially did this to somehow respect the conventions in this repo. But I realize we are already deviating as per https://github.com/deepset-ai/haystack-core-integrations/pull/3591#discussion_r3578513089.

I think it's OK to deviate since this integration does not offer regular components but component factories.
So I'd propose just choosing an import path structure we like and sticking with it. Something like
- `from haystack_integrations.agent_pack import create_deep_research_agent` (two levels)
- `from agent_pack_haystack import create_deep_research_agent` (one level)

WDYT?

integrations/agent_pack/pydoc/config_docusaurus.yml
sjrl: My initial thought was mostly to better document how the agents work. 

But also in the case if the Tavily Web Search tool it was for reuse. But if we do that it may be better to put that tool elsewhere like in the Tavily integration.

integrations/agent_pack/src/agent_pack_haystack/__init__.py
sjrl: Good points. Let's go with `from haystack_integrations.agent_pack import create_deep_research_agent`

integrations/agent_pack/pydoc/config_docusaurus.yml
sjrl: But if you think the readme and other docs are better for explanation I think that's also fine.

integrations/agent_pack/src/haystack_integrations/components/agents/agent_pack/deep_research/tools.py
anakin87: Implemented in https://github.com/deepset-ai/haystack-core-integrations/pull/3591/commits/aafc8705018f152166208773efdc13fbcf5abbc7

Should I improve it/put it in Tavily? I'm not sure since this is something users can easily build themselves with a few lines of code

integrations/agent_pack/pydoc/config_docusaurus.yml
anakin87: Ah, yes. I'd not expose them in the API reference.

The README is something to remove when we have proper docs.

I think that docs could be structured with two levels:
- for users who just want to import and use an Agent, high-level documentation for the factory function.
- for users who want to take inspiration, copy the code, etc. provide lower-level details like the README.

Makes sense?

sjrl: Yeah that makes sense to me!

integrations/agent_pack/src/haystack_integrations/components/agents/agent_pack/deep_research/tools.py
sjrl: I think it could still be worth adding to Tavily. Even though it's technically not much work for a user to do, it's still nicer to just have an import and a one line set up.

integrations/agent_pack/src/agent_pack_haystack/__init__.py
anakin87: I restructured the package in https://github.com/deepset-ai/haystack-core-integrations/pull/3591/commits/852021f0dfe1b9a8964c668704bd3ae7837ccf16

integrations/agent_pack/src/haystack_integrations/agent_pack/deep_research/hooks.py
anakin87: Since it is simple, I did the change in https://github.com/deepset-ai/haystack-core-integrations/pull/3591/commits/70e93e7fd4b3d35313df4b191840b546bfe51d0f

integrations/agent_pack/src/haystack_integrations/agent_pack/deep_research/hooks.py
```diff
@@ -0,0 +1,96 @@
...
+from haystack.core.serialization import default_from_dict, default_to_dict
+from haystack.dataclasses import ChatMessage
+
+logger = logging.getLogger(__name__)
+
+
+class ScopeHook:
+    """Turn the user query into a research brief, once. Runs as a `before_run` hook."""
+
+    allowed_hook_points = "before_run"
```
sjrl: This is expected to be a tuple currently in the agent code in haystack

integrations/agent_pack/src/haystack_integrations/components/agents/agent_pack/deep_research/tools.py
anakin87: Here it is: https://github.com/deepset-ai/haystack-core-integrations/pull/3597

integrations/agent_pack/src/haystack_integrations/agent_pack/deep_research/hooks.py
anakin87: thank you. Fixed in https://github.com/deepset-ai/haystack-core-integrations/pull/3591/commits/b5d6dc8ecdec1dd5a3927600fe487d22edeb6c90

sjrl (approved):
Looks good!

anakin87:
In https://github.com/deepset-ai/haystack-core-integrations/pull/3591/commits/a529f97b1805589496c71711833d11ec8bd91e04, I am using the `TavilyWebSearchTool` from the Tavily integration instead of redefining it here.

PR #3593, opened 2026-07-14
fix(pgvector): NO_VALUE sentinel collides with the literal string 'no_value'

chuenchen309:
### The bug

`NO_VALUE` is the string `"no_value"`, used internally to mark a param slot as "already handled inline (e.g. `IS NULL`), don't add this to the params list":

```python
NO_VALUE = "no_value"
```

`_convert_filters_to_where_clause_and_params` filters params by equality against this sentinel (`value != NO_VALUE`). So a legitimate equality filter whose value happens to be the literal string `"no_value"` gets its param silently dropped too:

```python
filters = {"field": "meta.tag", "operator": "==", "value": "no_value"}
where_clause, params = _convert_filters_to_where_clause_and_params(filters)
# where_clause still has one `%s` placeholder for the value
# params == ()  <- but the placeholder has no matching param!
```

That placeholder/param-count mismatch raises `psycopg.ProgrammingError` at `execute()` time — a valid, unremarkable filter value crashes the query.

### The fix

Make `NO_VALUE` a unique `object()` sentinel instead of a string, so it can never collide with a real user-supplied value:

```python
NO_VALUE = object()  # unique sentinel; can never equal a user-supplied filter value
```

Identity/`!=` comparison still works exactly the same for the internal use (only 3 reference sites: definition, the filter, and the `IS NULL` branch that returns it).

### Testing

Added `test_convert_filters_to_where_clause_and_params_value_equal_to_no_value_sentinel`, asserting `params == ("no_value",)` for the equality filter above. Confirmed it fails against the pre-fix code (`params == ()`, the mismatch) and passes after.

- `hatch run test:unit tests/test_filters.py` — 13/13 passed.
- `hatch run ruff check`/`ruff format --check` on the changed files — clean.
- `hatch run test:types` — no issues.

### AI disclosure

Developed with AI assistance (Claude Code), which located the sentinel-collision pattern while auditing `filters.py` for defects. I reviewed the diff, independently reproduced the placeholder/param mismatch and the fix by executing both against the actual function, and ran the tests/linters/type-checker myself before opening this PR.

### DIFF
--- a/integrations/pgvector/src/haystack_integrations/document_stores/pgvector/filters.py
+++ b/integrations/pgvector/src/haystack_integrations/document_stores/pgvector/filters.py
@@ -48,7 +48,7 @@ def _convert_filters_to_where_clause_and_params(
         query, values = _parse_logical_condition(filters)
 
     where_clause = SQL(f" {operator} ") + query
-    params = tuple(value for value in values if value != NO_VALUE)
+    params = tuple(value for value in values if value is not NO_VALUE)
 
     return where_clause, params
 
--- a/integrations/pgvector/src/haystack_integrations/document_stores/pgvector/filters.py
+++ b/integrations/pgvector/src/haystack_integrations/document_stores/pgvector/filters.py
@@ -20,7 +20,7 @@ PYTHON_TYPES_TO_PG_TYPES = {
     bool: "boolean",
 }
 
-NO_VALUE = "no_value"
+NO_VALUE = object()  # unique sentinel; can never equal a user-supplied filter value
 
 
 def _validate_filters(filters: dict[str, Any] | None = None) -> None:
--- a/integrations/pgvector/tests/test_filters.py
+++ b/integrations/pgvector/tests/test_filters.py
@@ -265,6 +265,18 @@ def test_convert_filters_to_where_clause_and_params_handle_null():
     assert params == ("intro",)
 
 
+def test_convert_filters_to_where_clause_and_params_value_equal_to_no_value_sentinel():
+    """A legitimate equality filter whose value is the literal string the internal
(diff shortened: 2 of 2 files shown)


anakin87:
Hey @chuenchen309, thank you for contributing.

I see you currently have 7 different PRs open across Haystack ecosystem. I recommend pausing this work for a bit, as it may be difficult for the team to keep up with reviews.

chuenchen309:
Thanks @anakin87, that's fair. I'll hold off on opening new PRs across the Haystack ecosystem and let the current ones move at your team's review pace — no rush on any of them. Thanks for the heads-up.

anakin87 (changes requested):
Good idea but I recommend using `is not` instead of `!=` below:
https://github.com/deepset-ai/haystack-core-integrations/blob/eca1e6ec766796912c48508417564042aa757c8a/integrations/pgvector/src/haystack_integrations/document_stores/pgvector/filters.py#L51

chuenchen309:
You're right, and the reason is stronger than style: `!=` calls the user-supplied value's `__ne__`, so a filter value whose `__ne__` returns something falsy would be dropped from `params` and desync the placeholders — the exact failure this PR is fixing. Identity can't be intercepted. Changed in 337979a5.

One thing I noticed while tracing the callers, purely as a heads-up: `integrations/alloydb/.../filters.py` looks like a copy of this file and has the same `NO_VALUE = "no_value"` / `!=` shape at the same lines (23/51). Diffing it against pgvector before this PR, the only differences are 12 lines of error-message text. I'm not opening a PR for it — I said I'd hold off on new PRs across the ecosystem until the current ones clear, and that still stands. Just flagging it in case it's useful to you.

anakin87 (approved):
Looks good.

Feel free to open another PR for the Alloydb fix.

PR #3594, opened 2026-07-14
feat(stackit): add dimensions parameter to text and document embedders

sebastianh-deepset:
### Related Issues

- N/A — no related issue, this was identified directly as a missing feature.

### Proposed Changes:

The Qwen embedding model (`Qwen/Qwen3-VL-Embedding-8B`) hosted on STACKIT supports Matryoshka-style variable-dimension embeddings, but `STACKITTextEmbedder` and `STACKITDocumentEmbedder` hardcoded `dimensions=None` when calling the underlying `OpenAITextEmbedder`/`OpenAIDocumentEmbedder`, so there was no way to request a reduced embedding size.

This PR exposes `dimensions` as a keyword-only init parameter on both embedders and forwards it to the parent OpenAI-compatible client, following the same pattern already used for `timeout`/`max_retries`/`http_client_kwargs`. It's added as keyword-only (rather than inserted positionally after `model`) so existing positional-argument usage of these components isn't broken. `to_dict`/`from_dict` were updated so the new parameter round-trips through serialization.

### How did you test it?

- Unit tests: constructor defaults/overrides, `to_dict`/`from_dict` round-trip, and a mocked-client test asserting `dimensions` is actually forwarded to `client.embeddings.create(...)`.
- Integration tests (require `STACKIT_API_KEY`): requesting `dimensions=256` against `Qwen/Qwen3-VL-Embedding-8B` returns embeddings of that length; requesting `dimensions` against a model that doesn't support it (`intfloat/e5-mistral-7b-instruct`) surfaces the API's 400 error, with the two embedders asserting on their differing failure behavior (`STACKITDocumentEmbedder` logs and continues, `STACKITTextEmbedder` raises).
- Ran locally against the real STACKIT API with my own key: `hatch run test:unit`, `hatch run test:integration`, `hatch run test:types` (mypy), `hatch run default:fmt-check` (ruff) — all green.

### Notes for the reviewer

- `dimensions` is keyword-only in both constructors (after the `*`), not inserted positionally after `model`, to avoid a backward-compatibility break for any code calling these components with positional args.
- The integration tests need `STACKIT_API_KEY` to run; they'll be skipped in CI for this fork PR since repo secrets aren't forwarded to `pull_request` runs from forks, but they will run once merged to `main` and in the nightly job.

### Checklist

- [x] I've used one of the [conventional commit types](https://www.conventionalcommits.org/en/v1.0.0/) for my PR title: `feat:`

### DIFF
--- a/integrations/stackit/src/haystack_integrations/components/embedders/stackit/document_embedder.py
+++ b/integrations/stackit/src/haystack_integrations/components/embedders/stackit/document_embedder.py
@@ -51,6 +51,7 @@ class STACKITDocumentEmbedder(OpenAIDocumentEmbedder):
         meta_fields_to_embed: list[str] | None = None,
         embedding_separator: str = "\n",
         *,
+        dimensions: int | None = None,
         timeout: float | None = None,
         max_retries: int | None = None,
         http_client_kwargs: dict[str, Any] | None = None,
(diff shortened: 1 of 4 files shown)


julian-risch (approved):
Looks good to me! Just made sure that the tests are compatible with 2.x and 3.x. 
By the way, no need for you to fork with your permissions. If you don't fork that also means that the integration tests run on your PR, which is better for us.
I ran the integration tests locally.

PR #3599, opened 2026-07-15
chore: Update litellm dependency version in pyproject.toml

julian-risch:
### Related Issues

There is a security advisory regarding litellm pypi package < 1.84.0 https://github.com/advisories/GHSA-4xpc-pv4p-pm3w

SDK usage and thus usage of our Haystack integration is unaffected. Only proxy deployments must patch. Our integration's constraint is litellm>=1.60.0,<2.0 right now, which allows users to upgrade. Most users won't need to. No need for further action on our side.

### Proposed Changes:

- Tighten the litellm dependency constraint so that unpatched versions are excluded

### Checklist

### DIFF
--- a/integrations/litellm/pyproject.toml
+++ b/integrations/litellm/pyproject.toml
@@ -22,7 +22,6 @@ classifiers = [
   "Programming Language :: Python :: 3.14",
   "Programming Language :: Python :: Implementation :: CPython",
 ]
-# litellm 1.82.7 and 1.82.8 were malicious releases and are excluded.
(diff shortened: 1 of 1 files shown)


integrations/litellm/pyproject.toml
```diff
@@ -23,7 +23,7 @@ classifiers = [
   "Programming Language :: Python :: Implementation :: CPython",
 ]
 # litellm 1.82.7 and 1.82.8 were malicious releases and are excluded.
```
anakin87: can we also remove or update this comment?

anakin87 (approved):

PR #3600, opened 2026-07-15
feat: Add advanced rag agent

sjrl:
### Related Issues

- fixes https://github.com/deepset-ai/haystack/issues/11000

### Proposed Changes:

Adds an **Advanced RAG Agent** to the `agent_pack` integration (`haystack_integrations.agent_pack.advanced_rag`): a metadata-aware RAG agent that inspects the document store's metadata and constructs Haystack filters for its retrieval.

- **Metadata inspection tools** (`Tool` subclasses over any `DocumentStore` implementing the opt-in introspection methods): 
  - `ListMetadataFieldsTool`
  - `GetMetadataFieldValuesTool` (with a client-side `search_term` — store-native implementations have incompatible semantics across integrations)
  - `GetMetadataFieldRangeTool`
- **`FetchDocumentsByFilterTool`**: direct, unscored fetching by metadata filter, results in reading order (grouped by parent file, sorted by split/page via `MetaFieldGroupingRanker`)
  - capped per fetch, with a `count_documents_by_filter` guard that refuses over-broad filters before fetching.
- **`DocumentStoreToolset`**: the four store-backed tools bundled as one serializable unit.
- **`create_advanced_rag_agent()`** factory: wires the toolset plus a `search_documents` tool built from a user-provided retriever (`TextRetriever` protocol) or retrieval pipeline (e.g. hybrid). 
  - The filter grammar from #11001 is embedded in the retrieval tools' `filters` parameter description. 
  - Retrieved documents accumulate into the agent's `documents` state output and answers cite them via short `[doc <id-prefix>]` references. **NOTE:** I'm open to other ideas here.
  - A `BackupAnswerHook` (`after_run`) produces a best-effort answer when `max_agent_steps` cuts the run off.

### How did you test it?

- unit tests and integration tests
- Eval harness runs: 8/8 (small corpus) and 6/6 (~150k-doc corpus) on both `InMemoryDocumentStore` and a real `OpenSearch` instance — verifying the model reliably calls `list_metadata_fields` before retrieving, builds correct `meta.`-prefixed filters across keyword/numeric/boolean/date fields

[advanced_rag_eval.py](https://github.com/user-attachments/files/30076458/advanced_rag_eval.py)

### Notes for the reviewer

Sorry for such a large diff. To get started I might suggest running the provided example script which is a mini-evaluation harness for testing how well the RAG agent does for a small corpus and a large corpus. 

Also I'm open to moving the tools and potentially the Hook to Haystack main since they could be useful outside of this Agent. Let me know what you think. 

### Checklist


integrations/agent_pack/src/haystack_integrations/agent_pack/advanced_rag/agent.py
```diff
@@ -0,0 +1,168 @@
...
+    :param retrieval_pipeline: A custom scoring-based retrieval pipeline (e.g. embedder -> retriever, or hybrid
+        retrieval) to use instead of a single retriever. Mutually exclusive with `retriever`.
+    :param retrieval_pipeline_input_mapping: Required with `retrieval_pipeline`: maps the tool inputs to pipeline input
+        sockets; must have exactly the keys "query" and "filters",
+        e.g. `{"query": ["embedder.text"], "filters": ["retriever.filters"]}`.
+    :param retrieval_pipeline_output_mapping: Optional with `retrieval_pipeline`: maps pipeline output sockets to tool
+        outputs, e.g. `{"retriever.documents": "documents"}`.
+    :param llm: LLM that drives the agent loop. Defaults to `OpenAIResponsesChatGenerator("gpt-5.4")` with low
+        reasoning effort.
+    :param system_prompt: Overrides the pre-made system prompt
```
anakin87: About all customization parameters below. Ok to keep them in this first version, but for the future, let's discuss how to allow users to customize Agents in a consistent way.

integrations/agent_pack/examples/advanced_rag_eval.py
```diff
@@ -0,0 +1,537 @@
...
+Run from the integration directory (`integrations/agent_pack`) with `OPENAI_API_KEY` set:
+
+    hatch run test:python examples/advanced_rag_eval.py small
+    hatch run test:python examples/advanced_rag_eval.py large   # needs `datasets`, downloads reviews from HF
+
+To validate against a real database instead of `InMemoryDocumentStore`, pass `--store opensearch`
+(needs `opensearch-haystack`). Set `OPENSEARCH_URL` if not http://localhost:9200, and
+`OPENSEARCH_USERNAME` / `OPENSEARCH_PASSWORD` for a security-enabled instance (use an https:// URL
+then). Start one locally with e.g.
+`docker run -p 9200:9200 -e discovery.type=single-node -e DISABLE_SECURITY_PLUGIN=true opensearchproject/opensearch:2`:
```
anakin87: nit: would it also work with opensearch 3?

integrations/agent_pack/examples/advanced_rag_eval.py
```diff
@@ -0,0 +1,537 @@
+# SPDX-FileCopyrightText: 2022-present deepset GmbH <info@deepset.ai>
+#
+# SPDX-License-Identifier: Apache-2.0
+
+"""
+Mini evaluation harness for the advanced RAG agent.
```
anakin87, line 6: I'd run a check over this comment to verify that everything is correct, maybe srhink it a bit and remove references like `deepset-ai/haystack#11000`

integrations/agent_pack/src/haystack_integrations/agent_pack/advanced_rag/hooks.py
```diff
@@ -0,0 +1,89 @@
...
+
+    def warm_up(self) -> None:
+        """Prepare the hook's generator for use; called from the Agent's `warm_up`."""
+        if hasattr(self.generator, "warm_up"):
+            self.generator.warm_up()
+
+    def close(self) -> None:
+        """Release the hook's generator resources; called from the Agent's `close`."""
+        if hasattr(self.generator, "close"):
+            self.generator.close()
```
anakin87: Are `warm_up` and `close` methods used? How does it work?

integrations/agent_pack/src/haystack_integrations/agent_pack/advanced_rag/agent.py
```diff
@@ -0,0 +1,168 @@
...
+        max_retries=_DEFAULT_MAX_RETRIES,
+        generation_kwargs={"reasoning": {"effort": "low"}},
+    )
+
+
+def create_advanced_rag_agent(
+    *,
+    document_store: DocumentStore,
+    retriever: TextRetriever | None = None,
+    retrieval_pipeline: Pipeline | None = None,
```
anakin87: Just a curiosity. Have you considered `retriever: TextRetriever | Pipeline | None`? WDYT?

integrations/agent_pack/src/haystack_integrations/agent_pack/advanced_rag/tools.py
```diff
@@ -0,0 +1,715 @@
...
+    """
+    sockets = getattr(retriever, "__haystack_input__", None)
+    socket_names = set(sockets._sockets_dict.keys()) if sockets is not None else set()
+    missing = {"query", "filters"} - socket_names
+    if missing:
+        msg = (
+            f"{type(retriever).__name__} does not accept the required inputs {sorted(missing)}. "
+            f"The retriever must take `query` and `filters`. Wrap embedding retrievers that take a "
+            f"`query_embedding` in haystack's `TextEmbeddingRetriever(retriever=..., text_embedder=...)`, "
+            f"or use `make_retrieval_pipeline_tool` for a custom retrieval pipeline."
```
anakin87: I don't think users are expected to directly use `make_retrieval_pipeline_tool`. Correct?

integrations/agent_pack/pydoc/config_docusaurus.yml
```diff
@@ -1,6 +1,9 @@
 loaders:
   - modules:
       - haystack_integrations.agent_pack.deep_research.agent
+      - haystack_integrations.agent_pack.advanced_rag.agent
+      - haystack_integrations.agent_pack.advanced_rag.tools
```
anakin87, line 5: could you check that only the expected symbols are exposed by running `hatch run docs`?
If some unexpected symbols appear, let's try to hide them, e.g. by making them private.

integrations/agent_pack/src/haystack_integrations/agent_pack/advanced_rag/tools.py
```diff
@@ -0,0 +1,715 @@
...
+        msg = (
+            f"{type(document_store).__name__} does not implement `{method}`, which this tool requires. "
+            f"Use a document store that supports metadata introspection."
+        )
+        raise ValueError(msg)
+
+
+def _strip_meta_prefix(field: str) -> str:
+    """
+    Normalize a field name: filters use 'meta.year' but the store methods expect 'year'.
```
anakin87: Document Stores should take care of this internally. No?

integrations/agent_pack/src/haystack_integrations/agent_pack/advanced_rag/tools.py
```diff
@@ -0,0 +1,715 @@
...
+
+    def _get_metadata_field_values(self, field: str, search_term: str | None = None) -> str:
+        """
+        List the distinct values of a metadata field, optionally narrowed by a substring search.
+
+        The search is applied CLIENT-SIDE, on purpose: the stores' own `search_term` parameter is split roughly
+        half/half between two incompatible semantics. OpenSearch, Elasticsearch, Weaviate, Chroma, pgvector and
+        InMemory match it against document CONTENT and aggregate the surviving documents' values;
+        ArcadeDB, Astra, Azure AI Search, MongoDB Atlas, Valkey and FalkorDB match the VALUES themselves
+        (FalkorDB case-sensitively); Oracle matches both. Filtering the fetched values ourselves gives every store
```
anakin87: Mmmhh, don't you think we should fix this sooner or later?

integrations/agent_pack/src/haystack_integrations/agent_pack/advanced_rag/tools.py
```diff
@@ -0,0 +1,715 @@
...
+        elif "limit" in signature_params:
+            kwargs["limit"] = fetch_size
+
+        result = method(field, **kwargs)
+        # Return shapes: a bare list of values (FAISS, AlloyDB, IBM DB), a (values, total_count) tuple
+        # (InMemory, Chroma, Weaviate, Pinecone, Astra, ...), or a (values, cursor) tuple from
+        # cursor-paginating stores (OpenSearch, Elasticsearch, FalkorDB) — a non-None cursor dict in
+        # place of the count means more pages exist.
+        values = list(result[0]) if isinstance(result, tuple) else list(result)
+        count = result[1] if isinstance(result, tuple) and isinstance(result[1], int) else len(values)
```
anakin87: I found it hard to understand this method, and I probably found some bugs.
One of these is: here, when the store sends no int, we get `count=len(values)`
then in the line after, we have `count > len(values)` -> always `False`.

For a store that takes `size`/`limit` and returns a plain list, we stop the fetch ourselves and then report the page as the whole field. Check Qdrant for example (and also what distinguishes Qdrant from FAISS?)

I recommend asking an Agent to sweep all Document Stores because my impression is that this method is problematic or not honest for several of them.

Let's also add more tests for this Tool.

Of course, in the long run, I'd just hope to have a reliable `get_metadata_field_unique_values`.

integrations/agent_pack/src/haystack_integrations/agent_pack/advanced_rag/tools.py
```diff
@@ -0,0 +1,715 @@
...
+    of `_GROUP_FIELDS`, via `MetaFieldGroupingRanker`) and sorted by their position within it (first present field of
+    `_SORT_FIELDS`).
+
+    :param documents: The fetched documents.
+    :returns: The documents, grouped and sorted when the meta fields allow it.
+    """
+    group_by = next((f for f in _GROUP_FIELDS if any(f in d.meta for d in documents)), None)
+    sort_by = next((f for f in _SORT_FIELDS if any(f in d.meta for d in documents)), None)
+    if group_by:
+        ranker = MetaFieldGroupingRanker(group_by=group_by, sort_docs_by=sort_by)
```
anakin87: also this can trigger a `TypeError`?

integrations/agent_pack/src/haystack_integrations/agent_pack/advanced_rag/agent.py
```diff
@@ -0,0 +1,168 @@
...
+    llm = llm or _default_llm("gpt-5.4")
+
+    tools: list[Tool | Toolset] = [
+        DocumentStoreToolset(document_store, max_fetched_docs=max_fetched_docs),
+        retrieval_tool,
+    ]
+    if extra_tools is not None:
+        tools.extend([extra_tools] if isinstance(extra_tools, Toolset) else list(extra_tools))
+
+    merged_hooks: dict[HookPoint, list[Hook]] = {"after_run": [BackupAnswerHook(generator=llm)]}
```
anakin87: maybe better not to use the same `llm`? A copy would be better, I think...

integrations/agent_pack/src/haystack_integrations/agent_pack/advanced_rag/tools.py
```diff
@@ -0,0 +1,715 @@
...
+        "properties": {
+            "filters": {"type": "object", "description": prompts.FILTER_GRAMMAR, "additionalProperties": True},
+            "max_docs": {
+                "type": "integer",
+                "minimum": 1,
+                "maximum": max_docs,
+                "description": f"How many documents to return at most. Defaults to the maximum, {max_docs}.",
+            },
+        },
+        "required": ["filters"],
```
anakin87, line 564: I see that filters are required but this conflicts with the prompt

integrations/agent_pack/src/haystack_integrations/agent_pack/advanced_rag/prompts.py
```diff
@@ -0,0 +1,82 @@
...
+2. Before filtering on a field, verify what it contains: use `get_metadata_field_values` for
+   keyword/boolean fields (filter values must match exactly) and `get_metadata_field_range` for
+   numeric or date-like fields.
+3. Call `search_documents` with a focused query and, when metadata can narrow the question, a
+   filter (the filter syntax is described in its `filters` parameter; always prefix field names
+   with "meta."). Not every question needs a filter — omit it when metadata cannot help.
+   When you can identify the exact documents you need by their metadata alone (e.g. a specific
+   title, source or file name), fetch them directly with `fetch_documents_by_filter` instead of
+   searching. The same applies when you need the COMPLETE set of documents matching a filter:
+   make one `fetch_documents_by_filter` call with that filter — do not repeat searches with
```
anakin87: here we say to do one `fetch_documents_by_filter` call. And above we say
> every document matching the filter is returned.

So, let's say 25 docs match the filter. We only get 10 and the LLM is discouraged from making another call. Right?

anakin87 (changes requested):
Nice work.

I have a few questions and may have found some potential bugs.

If by chance you could remove some code, I'd be happier :-)

integrations/agent_pack/examples/advanced_rag_eval.py
sjrl: Fair question, to help reduce code size of this, I'm going to drop this example for now and leave it as an upload on the PR description. If we think it adds value we can add it back later.

integrations/agent_pack/src/haystack_integrations/agent_pack/advanced_rag/hooks.py
sjrl: Yes they are! A lot like tools. When I added Hooks to Agent I gave them lifecycle methods. You can see how they are being used in Agent's [warm_up](https://github.com/deepset-ai/haystack/blob/bfb7a2896a393f031ccc7669b6923a32ca27918d/haystack/components/agents/agent.py#L772-L777) and [warm_up_async](https://github.com/deepset-ai/haystack/blob/bfb7a2896a393f031ccc7669b6923a32ca27918d/haystack/components/agents/agent.py#L779-L786) 

And same with `close` also called by the Agent.

integrations/agent_pack/src/haystack_integrations/agent_pack/advanced_rag/agent.py
sjrl: Good idea! I'll switch it to that.

integrations/agent_pack/src/haystack_integrations/agent_pack/advanced_rag/tools.py
sjrl: Good point, let me double check.

integrations/agent_pack/src/haystack_integrations/agent_pack/advanced_rag/tools.py
sjrl: Yes I agree, we should fix this soon. I'll go ahead and drop the search_term from the tool for now and will consider adding it back once this is fixed. I'll open an issue for this.

integrations/agent_pack/src/haystack_integrations/agent_pack/advanced_rag/tools.py
sjrl: Thanks for looking into this. I'll go ahead and make this a separate method so its easier to test and make sure we then cover all different scenarios from the different doc stores. 

I'll also include this as part of the issue for fixing/standardizing `get_metadata_field_unique_values`

integrations/agent_pack/src/haystack_integrations/agent_pack/advanced_rag/agent.py
sjrl: I don't think it's an issue to reuse the same LLM right? I'm happy to expose another llm in the init method for the agent factory but we have no easy way to deep copy a class.

integrations/agent_pack/src/haystack_integrations/agent_pack/advanced_rag/tools.py
sjrl: Done in https://github.com/deepset-ai/haystack-core-integrations/pull/3600/commits/76015844f648ef86224cafe9ef83a53a0145ae67

integrations/agent_pack/src/haystack_integrations/agent_pack/advanced_rag/agent.py
sjrl: Done in https://github.com/deepset-ai/haystack-core-integrations/pull/3600/commits/76015844f648ef86224cafe9ef83a53a0145ae67

integrations/agent_pack/examples/advanced_rag_eval.py
sjrl: Done in https://github.com/deepset-ai/haystack-core-integrations/pull/3600/commits/e3cb252a6f822dafe1631594fad7788010411e5e 

Eval script is now in the PR description

integrations/agent_pack/src/haystack_integrations/agent_pack/advanced_rag/tools.py
sjrl: Analysis from Claude Code

> `meta.` prefix handling also diverges: all implementations of `get_metadata_field_min_max` strip/normalize it, but in `get_metadata_field_unique_values` **MongoDB Atlas** (builds `$meta.{metadata_field}` unstripped), **Pinecone** and **Qdrant** (exact dict-key lookups) do not — a `meta.`-prefixed field name silently returns no values.

I'll include this in the issue I'm planning to open.

sjrl: Removed function in https://github.com/deepset-ai/haystack-core-integrations/pull/3600/commits/0f13a42100754a49552db7f73c8f111256416484

integrations/agent_pack/src/haystack_integrations/agent_pack/advanced_rag/tools.py
sjrl: Updated in https://github.com/deepset-ai/haystack-core-integrations/pull/3600/commits/d1b9e77bb216e5b14ec0e597fdc31a2a0cc6b5e8 and explained why we suppress it

integrations/agent_pack/src/haystack_integrations/agent_pack/advanced_rag/tools.py
sjrl: Good catch! Fixed in https://github.com/deepset-ai/haystack-core-integrations/pull/3600/commits/d1b9e77bb216e5b14ec0e597fdc31a2a0cc6b5e8

integrations/agent_pack/pydoc/config_docusaurus.yml
sjrl: Fixed in https://github.com/deepset-ai/haystack-core-integrations/pull/3600/commits/d1b9e77bb216e5b14ec0e597fdc31a2a0cc6b5e8 and made more methods private

integrations/agent_pack/src/haystack_integrations/agent_pack/advanced_rag/agent.py
sjrl: Sounds good!

integrations/agent_pack/src/haystack_integrations/agent_pack/advanced_rag/agent.py
sjrl: Added a new llm config option in https://github.com/deepset-ai/haystack-core-integrations/pull/3600/commits/271daf7b5011d9c6f99532ee8d9a486f67f2161d

integrations/agent_pack/src/haystack_integrations/agent_pack/advanced_rag/prompts.py
sjrl: Good catch! Addressed in https://github.com/deepset-ai/haystack-core-integrations/pull/3600/commits/626e86f070265b2ede1cd708576880fc60242ed1 added new prompt instructions and also added an `offset` param so the LLM can page through results up to the `fetch_limit`.

integrations/agent_pack/src/haystack_integrations/agent_pack/advanced_rag/tools.py
sjrl: Done in https://github.com/deepset-ai/haystack-core-integrations/pull/3600/commits/5d0a76b425581162c9f03f5a297a4bd4f048db06 but I'll be doing some more manual checking to ensure the test lines up with the actual doc stores

sjrl:
@anakin87 this is ready for another review, I'm still working on opening issues to fix the document stores

integrations/agent_pack/src/haystack_integrations/agent_pack/advanced_rag/hooks.py
anakin87: Thank you!

integrations/agent_pack/src/haystack_integrations/agent_pack/advanced_rag/tools.py
anakin87: Sounds good

integrations/agent_pack/src/haystack_integrations/agent_pack/advanced_rag/agent.py
```diff
@@ -0,0 +1,172 @@
...
+        e.g. `{"query": ["embedder.text"], "filters": ["retriever.filters"]}`.
+    :param retrieval_pipeline_output_mapping: Optional when `retriever` is a `Pipeline`: maps pipeline output sockets
+        to tool outputs, e.g. `{"retriever.documents": "documents"}`.
+    :param llm: LLM that drives the agent loop. Defaults to `OpenAIResponsesChatGenerator("gpt-5.4")` with low
+        reasoning effort.
+    :param backup_answer_llm: LLM the built-in `BackupAnswerHook` uses to write a best-effort answer when the run
+        is cut off by `max_agent_steps`. Defaults to a new instance with the same settings as the default `llm` —
+        deliberately a separate instance, never `llm` itself, so the agent loop's generator is not shared with a hook.
+        The default is an OpenAI generator and requires `OPENAI_API_KEY` as soon as the agent warms up, even when a
+        custom `llm` is passed — when not using OpenAI, pass a `backup_answer_llm` too.
```
anakin87: I'd make this description more compact for users (similar to `llm`)

integrations/agent_pack/src/haystack_integrations/agent_pack/advanced_rag/tools.py
```diff
@@ -0,0 +1,698 @@
...
+        inner_data = data["data"]
+        deserialize_component_inplace(inner_data, key="document_store")
+        return cls(**inner_data)
+
+
+class GetMetadataFieldValuesTool(Tool):
+    """Tool that returns the distinct values of a metadata field from a document store."""
+
+    # Cap on the number of values listed per call, protecting the agent's context window from high-cardinality fields
+    # (the total count is always reported).
```
anakin87:
```suggestion
```
no longer true, right?

integrations/agent_pack/src/haystack_integrations/agent_pack/advanced_rag/tools.py
```diff
@@ -0,0 +1,698 @@
...
+
+    Filter-fetched documents carry no relevance order, so they are grouped by their parent file (first present field
+    of `_GROUP_FIELDS`, via `MetaFieldGroupingRanker`) and sorted by their position within it (first present field of
+    `_SORT_FIELDS`).
+
+    :param documents: The fetched documents.
+    :returns: The documents, grouped and sorted when the meta fields allow it.
+    """
+    group_by = next((f for f in _GROUP_FIELDS if any(f in d.meta for d in documents)), None)
+    sort_by = next((f for f in _SORT_FIELDS if any(f in d.meta for d in documents)), None)
```
anakin87: Would it be better to select fields that are present in all documents for grouping and sorting? With any, we risk using a field that only a document has. (All might be too strict?)
WDYT?

anakin87 (approved):
Thank you for responding to all my comments! 😃 

I have a few minor ones left. Feel free to merge this PR when they are addressed.

In addition:
- let's open the issue about `get_metadata_field_unique_values` standardization
- feel free to re-add the example back, maybe just cleaning up some comments. Sorry for pushing to removal, I think it's still helpful.

integrations/agent_pack/src/haystack_integrations/agent_pack/advanced_rag/agent.py
sjrl: Done in https://github.com/deepset-ai/haystack-core-integrations/pull/3600/commits/8ce5a4c74ef7f383aefe73860e247b019e014bed

integrations/agent_pack/src/haystack_integrations/agent_pack/advanced_rag/tools.py
sjrl: Done in https://github.com/deepset-ai/haystack-core-integrations/pull/3600/commits/8ce5a4c74ef7f383aefe73860e247b019e014bed

integrations/agent_pack/src/haystack_integrations/agent_pack/advanced_rag/tools.py
sjrl: Updated it to prefer a field present in all docs metas. If that isn't possible then we fall back to using `any` so we can do some amount of mixed ordering.

You can see the new function and explanation in https://github.com/deepset-ai/haystack-core-integrations/pull/3600/commits/8ce5a4c74ef7f383aefe73860e247b019e014bed

sjrl:
Example script with refined comments added back in https://github.com/deepset-ai/haystack-core-integrations/pull/3600/commits/7d6c7228ec3c00c0a0840c0bf5a4ee201a0517df

sjrl:
> let's open the issue about get_metadata_field_unique_values standardization

@anakin87 opened parent issue here https://github.com/deepset-ai/haystack-private/issues/473

PR #3480, opened 2026-06-22
feat: add TwelveLabs integration (Marengo embeddings + Pegasus video converter)

mohit-twelvelabs:
partially addresses https://github.com/deepset-ai/haystack-core-integrations/issues/3481

## Description

Hi! I’m Mohit, and I work at TwelveLabs.

This PR adds a `twelvelabs` integration with three components, following the project’s standard integration layout:

* Secret-based `api_key`
* `to_dict` / `from_dict` serialization
* `@component.output_types`
* `httpx`
* Apache-2.0 licensing
* Hatch-vcs versioning

### `TwelveLabsTextEmbedder` and `TwelveLabsDocumentEmbedder`

These components provide TwelveLabs **Marengo** (`marengo3.0`) multimodal embeddings.

Marengo embeds text, images, audio, and video into a shared 512-dimensional space, enabling cross-modal retrieval workflows such as text-to-image or text-to-video search.

### `TwelveLabsVideoConverter`

This component provides TwelveLabs **Pegasus** on-the-fly video analysis.

Pegasus analyzes both visual content and audio using its built-in ASR, then converts the result into `Document`s. The converter supports both public video URLs and local video files.

## Implementation details

The integration is:

* Built with pure `httpx`
* Free of any TwelveLabs SDK dependency
* Fully typed
* Configured through `TWELVELABS_API_KEY`

## Tests and quality

Validation completed:

* 10 unit tests pass
* `ruff check` passes
* `ruff format` is clean

I also live-verified the integration end-to-end against the TwelveLabs API:

* Both embedders return 512-dimensional vectors
* The converter produced an accurate Pegasus analysis `Document`

Happy to split this into separate PRs for the embedders and converter, or adjust naming and scope to better match the project’s preferences.

PR #3617, opened 2026-07-20
feat: add TwelveLabs Marengo multimodal embedders (image/audio/video)

mohit-twelvelabs:
## Description

Follow-up to #3480 (the initial TwelveLabs integration). That PR shipped **text** embeddings
via Marengo (`TwelveLabsTextEmbedder`, `TwelveLabsDocumentEmbedder`) plus the Pegasus
`TwelveLabsVideoConverter`. But Marengo embeds **text, images, audio, and video into one shared
512-dimensional space**, and only the text path was exposed. This PR adds the image/audio/video
embedding capabilities so the shared space can actually be used for cross-modal retrieval.

Modeled on the Voyage integration's multimodal embedder, and kept consistent with the existing
Text/Document embedder pairing in this integration.

### `TwelveLabsMultimodalEmbedder`
Embeds a single image, audio, or video (a local file path **or** a public URL) into Marengo's
shared space and returns `{"embedding": [...], "meta": {...}}`. This is the cross-modal query
companion to `TwelveLabsTextEmbedder` — e.g. embed a text query with the text embedder and search
a store of video/image embeddings, or vice-versa. Modality is inferred from the file
extension / MIME type and can be set explicitly via `modality=`.

```python
from haystack_integrations.components.embedders.twelvelabs import TwelveLabsMultimodalEmbedder

embedder = TwelveLabsMultimodalEmbedder()  # reads TWELVELABS_API_KEY
result = embedder.run(source="https://example.com/cat.jpg")
print(len(result["embedding"]))
```

### `TwelveLabsDocumentMultimodalEmbedder`
The indexing counterpart to `TwelveLabsDocumentEmbedder` (which embeds text `content`). It reads
each Document's media path from `meta["file_path"]` (configurable via `file_path_meta_field` /
`root_path`), embeds the media, and writes `Document.embedding` — following the same convention as
the existing `CohereDocumentImageEmbedder` / `JinaDocumentImageEmbedder`. Modality is inferred
per-Document (override with `meta["modality"]`).

```python
from haystack import Document
from haystack_integrations.components.embedders.twelvelabs import TwelveLabsDocumentMultimodalEmbedder

docs = [Document(meta={"file_path": "clip.mp4"})]
docs = TwelveLabsDocumentMultimodalEmbedder().run(documents=docs)["documents"]
```

## Implementation notes

- Images and audio are embedded synchronously via `client.embed.create`.
- Video uses the asynchronous TwelveLabs video-embedding task: the component submits the task,
  polls `embed.tasks.status` until it is `ready`, then retrieves the embedding (whole-video scope).
- Both components provide `run` and `run_async`.
- No new dependency — uses the `twelvelabs` SDK already declared by the integration.

## Tests and quality

- 22 new unit tests (56 total pass); `ruff check`, `ruff format --check`, and `mypy` all clean.
- Gated integration tests (`@pytest.mark.integration`, skipped without `TWELVELABS_API_KEY`) cover
  a live image embedding and a live video-Document embedding.

## Notes for reviewers

- Naming: went with `TwelveLabsMultimodalEmbedder` (query-side) + `TwelveLabsDocumentMultimodalEmbedder`
  (document-side) to mirror the existing `Text` / `Document` pairing and the repo's image-embedder
  convention. Happy to rename or split further if you'd prefer.
- Opening as a **draft** — feedback on scope, naming, and the media-input interface welcome before
  I finalize.

### DIFF
--- a/integrations/twelvelabs/README.md
+++ b/integrations/twelvelabs/README.md
@@ -8,6 +8,38 @@
 
 ---
 
+## Components
+
+TwelveLabs **Marengo** embeds text, images, audio, and video into a single shared vector space, so
+embeddings from any modality are directly comparable — enabling cross-modal retrieval.
+
+- `TwelveLabsTextEmbedder` — embeds a text query.
+- `TwelveLabsDocumentEmbedder` — embeds the text `content` of Documents.
+- `TwelveLabsMultimodalEmbedder` — embeds a single image, audio, or video (local path or URL) into
+  the shared Marengo space; the cross-modal companion to `TwelveLabsTextEmbedder`.
+- `TwelveLabsDocumentMultimodalEmbedder` — embeds the media referenced by each Document
+  (via `meta["file_path"]`) for indexing.
+- `TwelveLabsVideoConverter` — analyzes video with **Pegasus** and converts the result to Documents.
+
+```python
+from haystack import Document
+from haystack_integrations.components.embedders.twelvelabs import (
+    TwelveLabsMultimodalEmbedder,
+    TwelveLabsDocumentMultimodalEmbedder,
+)
+
+# Set the TWELVELABS_API_KEY environment variable
+
+# Embed a single media query (image / audio / video, local path or URL).
+query_embedding = TwelveLabsMultimodalEmbedder().run(source="cat.jpg")["embedding"]
+
+# Embed media Documents for indexing (modality inferred from the file extension).
+docs = [Document(meta={"file_path": "clip.mp4"})]
+docs = TwelveLabsDocumentMultimodalEmbedder().run(documents=docs)["documents"]
+```
+
+---
+
 ## Contributing
 
 Refer to the general [Contribution Guidelines](https://github.com/deepset-ai/haystack-core-integrations/blob/main/CONTRIBUTING.md).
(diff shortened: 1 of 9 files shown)


julian-risch (changes requested):
Hi @mohit-twelvelabs and thank you for extending the TwelveLabs integration! 

I noticed an inconsistency between run and run_async that I would like you to have another look at and implement in a more consistent way:
- run_async resolves all sources and modalities up front before any embed_media_async call. Saves the user costs.
- run interleaves resolution with embedding inside the loop, so a bad modality/missing path on document only raises after other documents have already been embedded and billed.

My initial thinking is that run should also do sth like:
```python
self._validate(documents)
key = self.api_key.resolve_value() or ""
sources = [self._resolve_source(d) for d in documents]
modalities = [self._resolve_modality(d, s) for d, s in zip(documents, sources, strict=True)]
# then loop embed_media(source, modality, ...) over the pre-resolved lists
```

However, I realized that `TwelveLabsVideoConverter` takes a different approach than other components in the integration. It only has run (no run_async at all) and each source is wrapped in try/except that logs a warning and continues, which means no fail-fast one bad source doesn't fail the whole batch.

PR #3618, opened 2026-07-21
fix(valkey): report accurate written count when async batch writes partially fail

### LINKED ISSUE #3579: ValkeyDocumentStore.write_documents_async under-reports written_count and can mask partial batch writes when one document fails
### Bug Description

`ValkeyDocumentStore.write_documents_async` writes each batch concurrently with `asyncio.gather`, wrapped in a single try/except that only increments `written_count` after the whole batch succeeds:

```python
written_count = 0
for i in range(0, len(documents), self._batch_size):
    batch = documents[i : i + self._batch_size]
    try:
        await asyncio.gather(*[write_single_doc(doc) for doc in batch])
        written_count += len(batch)
    except Exception as e:
        msg = f"Failed to write batch starting at index {i}: {e}"
        raise ValkeyDocumentStoreError(msg) from e
```

Each `write_single_doc` call is an independent `glide_json.set(...)` against Valkey, so it durably persists the document the moment Valkey acknowledges it, regardless of what the other concurrent calls in the same `gather` do. If one document in a batch fails (a transient connection blip, a bad value, anything) while its siblings in the same batch already got a successful reply from the server, `gather` still raises for the whole batch, and `written_count` is never incremented for that batch at all. The caller sees a `ValkeyDocumentStoreError` and a document count that implies nothing in that batch was written, when in fact some documents in it already are, durably, in the store.

Verified the mechanics with a stand-in store (write becomes durable the instant the underlying coroutine completes, exactly like a `glide_json.set()` server ack):

```python
store = {}

async def write_single_doc(doc_id, should_fail, delay):
    await asyncio.sleep(delay)
    if should_fail:
        raise ConnectionError(f"simulated transient failure writing {doc_id}")
    store[doc_id] = "written"
    return doc_id

async def write_documents_async(batch):
    written_count = 0
    try:
        await asyncio.gather(*[write_single_doc(doc_id, fail, delay) for doc_id, fail, delay in batch])
        written_count += len(batch)
    except Exception as e:
        raise RuntimeError(f"Failed to write batch: {e}") from e
    return written_count

batch = [("doc_1", False, 0.05), ("doc_2", True, 0.01), ("doc_3", False, 0.05)]
```

Output:
```
write_documents_async raised: Failed to write batch: simulated transient failure writing doc_2
reported progress for this batch: 0 (exception propagated before increment)
documents actually persisted in the store despite the reported failure: ['doc_1', 'doc_3']
```

`doc_2` legitimately failed, but `doc_1` and `doc_3` were already written before the exception surfaced. The function reports 0 progress for the batch and raises, even though two thirds of it succeeded.

### Impact

Callers reacting to `ValkeyDocumentStoreError` by retrying the whole batch, or by trusting the returned `written_count` to reconcile what made it into the store, will be working from incorrect information: some "failed" documents are already present. Depending on the write path taken for `doc.id`-keyed documents, a retry of an already-written document can hit duplicate-key behavior instead of a clean insert, in addition to just being wasted work.

### Affected code

`integrations/valkey/src/haystack_integrations/document_stores/valkey/document_store.py`, `write_documents_async`.

### Expected Behavior

`written_count` should reflect what was actually persisted, not what was attempted. Gathering with `return_exceptions=True`, counting the documents whose write actually succeeded, and only then raising (or returning partial success with the details of which document ids failed) would keep the return value accurate regardless of which individual writes in a batch fail.

### System

deepset-ai/haystack-core-integrations main branch, `valkey` integration, `ValkeyDocumentStore.write_documents_async`.

HussainNizamani:
### Related Issues

- fixes #3579

### Proposed Changes:

`write_documents_async` counted writes per batch. When one document in a
batch failed, the siblings that were already written were not counted, and
the error message said nothing was written. That is not true, because each
`glide_json.set` call is durable on its own.

Now the batch is gathered with `return_exceptions=True`, successes are
counted per document, and the error message reports the real number of
written documents. Batches after a failure are not started.

The second commit makes `write_single_doc` async, so errors from
`_prepare_document_dict` (for example invalid metadata) go through the
same per-document accounting and are wrapped in `ValkeyDocumentStoreError`,
the same way the sync path wraps them.

### How did you test it?

Four new unit tests (mocked glide client, no live Valkey needed):
all-success count, failure in a middle batch, failure in the first batch
with a successful sibling, and a document preparation error. Full run:
`hatch run test:unit` (116 passed), `hatch run test:types`,
`hatch run fmt` — all clean.

### Notes for the reviewer

The error message now also says how many documents were written before
the failure, for example: "3 document(s) were written before this failure."

### Checklist

### DIFF
--- a/integrations/valkey/src/haystack_integrations/document_stores/valkey/document_store.py
+++ b/integrations/valkey/src/haystack_integrations/document_stores/valkey/document_store.py
@@ -705,12 +705,18 @@ class ValkeyDocumentStore(DocumentStore):
         written_count = 0
         for i in range(0, len(documents), self._batch_size):
             batch = documents[i : i + self._batch_size]
-            try:
-                await asyncio.gather(*[write_single_doc(doc) for doc in batch])
-                written_count += len(batch)
-            except Exception as e:
-                msg = f"Failed to write batch starting at index {i}: {e}"
-                raise ValkeyDocumentStoreError(msg) from e
+            results = await asyncio.gather(*[write_single_doc(doc) for doc in batch], return_exceptions=True)
+            for result in results:
+                if not isinstance(result, BaseException):
+                    written_count += 1
+
+            for doc, result in zip(batch, results, strict=True):
+                if isinstance(result, BaseException):
+                    msg = (
+                        f"Failed to write document {doc.id}: {result}. "
+                        f"{written_count} document(s) were written before this failure."
+                    )
+                    raise ValkeyDocumentStoreError(msg) from result
 
         return written_count
 
--- a/integrations/valkey/src/haystack_integrations/document_stores/valkey/document_store.py
+++ b/integrations/valkey/src/haystack_integrations/document_stores/valkey/document_store.py
@@ -697,10 +697,10 @@ class ValkeyDocumentStore(DocumentStore):
         client = await self._get_connection_async()
         await self._create_index_async()
 
-        def write_single_doc(doc: Document) -> Any:
+        async def write_single_doc(doc: Document) -> Any:
             doc_dict = self._prepare_document_dict(doc)
             key = f"{self._index_name}:{doc.id}"
-            return glide_json.set(client, key, "$", json.dumps(doc_dict))
+            return await glide_json.set(client, key, "$", json.dumps(doc_dict))
 
         written_count = 0
         for i in range(0, len(documents), self._batch_size):
--- a/integrations/valkey/src/haystack_integrations/document_stores/valkey/document_store.py
+++ b/integrations/valkey/src/haystack_integrations/document_stores/valkey/document_store.py
@@ -709,13 +709,16 @@ class ValkeyDocumentStore(DocumentStore):
                 if not isinstance(result, BaseException):
                     written_count += 1
 
-            for doc, result in zip(batch, results, strict=True):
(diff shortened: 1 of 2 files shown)


davidsbatista:
@HussainNizamani thanks for the contribution

One comment: when more than one document in the same batch fails, only the first failure is reported.

```python
  for doc, result in zip(batch, results, strict=True):
      if isinstance(result, BaseException):
          msg = (
              f"Failed to write document {doc.id}: {result}. "
              f"{written_count} document(s) were written before this failure."
          )
          raise ValkeyDocumentStoreError(msg) from result
```

We should consider collecting all failed docs into the message before raising, e.g.:

```python
  failures = [(doc, result) for doc, result in zip(batch, results, strict=True) if isinstance(result, BaseException)]
  if failures:
      ids = ", ".join(doc.id for doc, _ in failures)
      msg = f"Failed to write document(s) {ids}. {written_count} document(s) were written before this failure."
      raise ValkeyDocumentStoreError(msg) from failures[0][1]
```

HussainNizamani:
Thanks for the review! Good point — done. All failed documents in a batch are now collected into one error message, with the accurate written count, and I added a test with two failures in the same batch checking both ids are reported.

davidsbatista (approved):
Looks good! Thanks!

PR #3619, opened 2026-07-21
fix(microsoft_sharepoint,google_drive): cancel and drain in-flight fetches when one fails

### LINKED ISSUE #3577: Shared httpx.AsyncClient closed mid-flight when one fetch fails in MSSharePointFetcher/GoogleDriveFetcher run_async, aborting unrelated concurrent fetches
### Bug Description

`MSSharePointFetcher.run_async` and `GoogleDriveFetcher.run_async` both open a single shared `httpx.AsyncClient` and fan out one `fetch_one` coroutine per target through `asyncio.gather(...)` (no `return_exceptions=True`), inside the `async with httpx.AsyncClient(...) as client:` block:

```python
async with httpx.AsyncClient(timeout=self.timeout, follow_redirects=True) as client:
    results = await asyncio.gather(*(fetch_one(client, url, hints) for url, hints in resolved))
```

With the default `raise_on_failure=True`, a single failed fetch (one bad URL, one permission error, one transient API error) makes `_process_async` re-raise, which propagates straight through `gather`. Since the other `fetch_one` calls are still using the very same shared `client`, exiting the `async with` block closes that client while their requests are still in flight. This does not just leak an orphaned task, it actively tears down the connection those other in-flight requests depend on, so a single failing target can make otherwise-healthy fetches fail too.

Confirmed with a live client/server, not just theory. Ran an `httpx.AsyncClient` shared between one coroutine that raises after a short delay and one that is mid-request against a real local aiohttp server with an artificial 1s delay:

```python
async with httpx.AsyncClient(timeout=10) as client:
    try:
        results = await asyncio.gather(fast_fail(), slow_fetch())
    except ValueError as e:
        print("gather raised:", e)
# client.__aexit__ already ran here, closing the client
```

Output, then draining the still-pending task that gather left behind:
```
gather raised: simulated failure on first fetch
pending tasks after client closed: 1 (plus the aiohttp server-side tasks handling the still-open connection)
orphaned task outcome: ReadError('')
```

`slow_fetch()`'s request, which had nothing to do with the failure, is aborted with `httpx.ReadError` purely because the shared client got closed underneath it. Applied to the fetchers, this means fetching a batch of SharePoint or Google Drive documents where one target is bad can produce spurious `ReadError`/connection-closed failures on unrelated, otherwise-fetchable documents in the same batch, not just a clean failure for the one bad target.

### Affected code

- `integrations/microsoft_sharepoint/src/haystack_integrations/components/fetchers/microsoft_sharepoint/fetcher.py`, `run_async` / `fetch_one` / `_process_async` (re-raises `SharePointRequestError`/`httpx.HTTPError` when `raise_on_failure=True`, the default)
- `integrations/google_drive/src/haystack_integrations/components/fetchers/google_drive/fetcher.py`, same shape (`GoogleDriveRequestError`/`httpx.HTTPError`, same default)

### Expected Behavior

A failure fetching one target with `raise_on_failure=True` should not affect the outcome of other targets in the same batch. Either gather with `return_exceptions=True` and only raise after all fetches have actually finished (so the client stays open for the duration of every in-flight request), or give each concurrent fetch its own client/connection so tearing one down on error can't disturb the others.

### System

deepset-ai/haystack-core-integrations main branch, `microsoft_sharepoint` and `google_drive` integrations, current `run_async` implementations.

HussainNizamani:
### Related Issues

- fixes #3577

### Proposed Changes:

In both fetchers, `run_async` shares one `httpx.AsyncClient` between all
downloads. When one download failed, the exception left the `async with`
block and closed the client while other downloads were still using it. So
one bad file could break healthy fetches too, and tasks were left orphaned.

Now the tasks are created explicitly, and a small helper cancels and
awaits the remaining tasks before the exception is re-raised, while the
client is still open. Same idea as deepset-ai/haystack#11967.
`raise_on_failure=False` behavior is unchanged.

The second commit catches `BaseException` in the helper, so the drain also
runs when `run_async` itself is cancelled (for example by a caller timeout).

### How did you test it?

New unit test in each integration: one slow download and one failing
download share the client; the test checks the slow one gets cancelled
cleanly and no orphaned tasks remain. The new tests fail on the old code.
Full runs in both integrations: `hatch run test:unit` (80 + 90 passed),
`hatch run test:types`, `hatch run fmt` — all clean.

### Notes for the reviewer

The helper is duplicated in both integrations because each package is
self-contained. Happy to move it to a shared place if you prefer.

### Checklist

### DIFF
--- a/integrations/google_drive/src/haystack_integrations/common/google_drive/utils.py
+++ b/integrations/google_drive/src/haystack_integrations/common/google_drive/utils.py
@@ -2,6 +2,9 @@
 #
 # SPDX-License-Identifier: Apache-2.0
 
+import asyncio
+from typing import TypeVar
+
 import httpx
 from haystack import logging
 from haystack.utils import Secret
@@ -18,10 +21,29 @@ from .errors import GoogleDriveConfigError, GoogleDriveRequestError
 
 logger = logging.getLogger(__name__)
 
+T = TypeVar("T")
+
 DEFAULT_API_BASE_URL = "https://www.googleapis.com/drive/v3"
 _RETRYABLE_STATUS = frozenset({429, 500, 502, 503, 504})
 # Fallback backoff when a throttled response carries no `Retry-After` header: 1s, 2s, 4s, ...
 _EXPONENTIAL_BACKOFF = wait_exponential()
+
+
+async def _gather_tasks_with_cancel(tasks: list[asyncio.Task[T]]) -> list[T]:
+    """
+    Wait for all tasks, cancelling and draining unfinished siblings if one fails.
+
+    :param tasks: Tasks to wait for.
+    :returns:
+        The task results in input order.
+    """
+    try:
+        return await asyncio.gather(*tasks)
+    except Exception:
+        for task in tasks:
+            task.cancel()
+        await asyncio.gather(*tasks, return_exceptions=True)
+        raise
 
 
 def resolve_access_token(access_token: str | Secret) -> str:
--- a/integrations/google_drive/src/haystack_integrations/common/google_drive/utils.py
+++ b/integrations/google_drive/src/haystack_integrations/common/google_drive/utils.py
@@ -39,7 +39,7 @@ async def _gather_tasks_with_cancel(tasks: list[asyncio.Task[T]]) -> list[T]:
     """
     try:
         return await asyncio.gather(*tasks)
-    except Exception:
+    except BaseException:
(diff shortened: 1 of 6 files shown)


anakin87:
@bogdankostic could you please take a look?

bogdankostic (approved):
Looks good, thanks @HussainNizamani! Integration tests pass locally.

PR #3633, opened 2026-07-21
refactor: Elasticsearch - split sync/async paths

anakin87:
### Related Issues

Elasticsearch is the only remaining Document Store where we mix sync and async clients creation.
This will be problematic once we implement the approach described in https://github.com/deepset-ai/haystack-core-integrations/issues/1628#issuecomment-5034357021
 
### Proposed Changes:
- Elasticsearch: create the 2 clients separately and adapt the code
- minor change in Opensearch: remove the `self._initialized` field from the document store, which is unused but partly hides the fact that sync and async handling is separate

### How did you test it?
CI, adapt existing tests a bit

### Checklist


davidsbatista (approved):
LGTM!

PR #3640, opened 2026-07-22
feat: Opensearch - add methods to close resources

anakin87:
### Related Issues

- part of https://github.com/deepset-ai/haystack-core-integrations/issues/1628

### Proposed Changes:
- add closing methods to Document Store and retrievers as described in the issue

### How did you test it?
CI, new tests

### Checklist

### DIFF
--- a/integrations/opensearch/tests/test_document_store.py
+++ b/integrations/opensearch/tests/test_document_store.py
@@ -752,6 +752,31 @@ async def test_delete_all_documents_async_recreate_works_for_concrete_index(_moc
     mock_client.indices.create.assert_called_once()
 
 
+def test_close():
+    store = OpenSearchDocumentStore(hosts="testhost", http_auth=("a", "b"))
+    mock_client = MagicMock()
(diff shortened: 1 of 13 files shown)


integrations/opensearch/tests/test_document_store.py
```diff
@@ -752,6 +752,31 @@ async def test_delete_all_documents_async_recreate_works_for_concrete_index(_moc
     mock_client.indices.create.assert_called_once()
 
 
+def test_close():
```
sjrl, line 755: Do you think it would make sense to extend the doc store mixin tests to cover this in the future?

sjrl (approved):
Looks good! One minor comment

integrations/opensearch/tests/test_document_store.py
anakin87: Nowadays, the mixin contains only integration tests, so I'd not include this unit test.

We could instead include a test like `test_close_and_reopen` below but make it more generic

PR #3641, opened 2026-07-22
fix: `PineCone` normalisation of .meta usage in `get_metadata_field_unique_values()`

davidsbatista:
### Related Issues

- fixes #https://github.com/deepset-ai/haystack-private/issues/477

### Proposed Changes:

- prefix check inside `_get_metadata_field_unique_values_impl()`

### How did you test it?

- added a unit tests

### Checklist

### DIFF
--- a/integrations/pinecone/src/haystack_integrations/document_stores/pinecone/document_store.py
+++ b/integrations/pinecone/src/haystack_integrations/document_stores/pinecone/document_store.py
@@ -893,10 +893,11 @@ class PineconeDocumentStore:
         documents: list[Document], metadata_field: str, search_term: str | None, from_: int, size: int
     ) -> tuple[list[str], int]:
         """Helper method to get unique values for a metadata field with search and pagination."""
+        field_name = metadata_field.removeprefix("meta.")
         unique_values: set[str] = set()
         for doc in documents:
-            if doc.meta and metadata_field in doc.meta:
-                value = doc.meta[metadata_field]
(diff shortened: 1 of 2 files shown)


sjrl:
@davidsbatista looks good! Do you think it's worth updating the test in the mixin that tests this method rather than adding a specific test here?

davidsbatista:
We can, but then we have to wait until the next Haystack release for integrations to use those new MixIn tests.

davidsbatista:
We can add them already to haystack MixIn and then update only the tests in integrations after a new release.

sjrl:
> We can, but then we have to wait until the next Haystack release for integrations to use those new MixIn tests.

We also run tests against Haystack main so those should automatically be picked up as long as we add them to an existing mixin right?

davidsbatista:
https://github.com/deepset-ai/haystack-core-integrations/blob/main/.github/workflows/pinecone.yml#L129

It only runs against `main` in the nightly run and **only** for the unit tests. Which in this particular case is fine, but still, we are opening PRs without seeing it pick up the MixIn tests from main.

sjrl (approved):
Thanks for the explanation! Didn't realize we only ran the unit tests against main in our nightly tests. Looks good!

davidsbatista:
I've created an issue to tackle this: https://github.com/deepset-ai/haystack/issues/12126

We could also then discuss other alternatives for the Mixin tests so that we don't run into this restriction again.

PR #3645, opened 2026-07-22
fix: `Oracle` match `search_term` against metadata field value only, drop content match

davidsbatista:
### Related Issues

- fixes https://github.com/deepset-ai/haystack-private/issues/481

### Proposed Changes:

- match metadata value only

### How did you test it?

- added unit tests

### Checklist

### DIFF
--- a/integrations/oracle/src/haystack_integrations/document_stores/oracle/document_store.py
+++ b/integrations/oracle/src/haystack_integrations/document_stores/oracle/document_store.py
@@ -773,7 +773,7 @@ class OracleDocumentStore:
 
         :param metadata_field: Metadata field name. May be prefixed with ``"meta."``
             (e.g. ``"meta.lang"`` or ``"lang"``).
-        :param search_term: Optional substring filter applied to both the document text and the field value.
+        :param search_term: Optional substring filter applied to the metadata field's own value.
         :param from_: Zero-based offset for pagination. Defaults to ``0``.
         :param size: Maximum number of values to return. When ``None`` all values from ``from_`` onward
             are returned.
@@ -786,7 +786,7 @@ class OracleDocumentStore:
         base_sql = f"FROM {self.table_name} WHERE JSON_VALUE(metadata, '$.{field_path}') IS NOT NULL"
         params: dict[str, Any] = {}
         if search_term:
-            base_sql += f" AND (text LIKE :search OR JSON_VALUE(metadata, '$.{field_path}') LIKE :search)"
+            base_sql += f" AND JSON_VALUE(metadata, '$.{field_path}') LIKE :search"
             params["search"] = f"%{search_term}%"
 
         sql_count = f"SELECT COUNT(DISTINCT JSON_VALUE(metadata, '$.{field_path}')) {base_sql}"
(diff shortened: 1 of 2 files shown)


integrations/oracle/src/haystack_integrations/document_stores/oracle/document_store.py
```diff
@@ -786,7 +786,7 @@ def get_metadata_field_unique_values(
         base_sql = f"FROM {self.table_name} WHERE JSON_VALUE(metadata, '$.{field_path}') IS NOT NULL"
         params: dict[str, Any] = {}
         if search_term:
-            base_sql += f" AND (text LIKE :search OR JSON_VALUE(metadata, '$.{field_path}') LIKE :search)"
+            base_sql += f" AND JSON_VALUE(metadata, '$.{field_path}') LIKE :search"
```
anakin87: https://github.com/deepset-ai/haystack-private/issues/481 mentions case-insensitivity. This is not case-insensitive, right?

integrations/oracle/src/haystack_integrations/document_stores/oracle/document_store.py
```diff
@@ -773,7 +773,7 @@ def get_metadata_field_unique_values(
 
         :param metadata_field: Metadata field name. May be prefixed with ``"meta."``
             (e.g. ``"meta.lang"`` or ``"lang"``).
-        :param search_term: Optional substring filter applied to both the document text and the field value.
+        :param search_term: Optional substring filter applied to the metadata field's own value.
         :param from_: Zero-based offset for pagination. Defaults to ``0``.
         :param size: Maximum number of values to return. When ``None`` all values from ``from_`` onward
```
anakin87: https://github.com/deepset-ai/haystack-private/issues/481 mentions "Align the size default (10) with the spec."
WDYT?

integrations/oracle/tests/test_document_store.py
```diff
@@ -47,6 +47,36 @@ def _uid(suffix: str = "") -> str:
...
+    cursor.fetchone.return_value = (1,)
+    cursor.fetchall.return_value = [("bar",)]
+
+    patched_store.get_metadata_field_unique_values("category", search_term="bar")
+
+    count_sql, count_params = cursor.execute.call_args_list[0][0]
+    vals_sql, vals_params = cursor.execute.call_args_list[1][0]
+
+    for sql in (count_sql, vals_sql):
+        assert "text LIKE" not in sql
```
anakin87: I would not test against the previous wrong behavior. The positive assertion should be enough.

integrations/oracle/tests/test_document_store.py
```diff
@@ -47,6 +47,36 @@ def _uid(suffix: str = "") -> str:
...
+
+def test_get_metadata_field_unique_values_no_search_term_no_like_clause(patched_store, mock_pool):
+    _, _, cursor = mock_pool
+    cursor.fetchone.return_value = (0,)
+    cursor.fetchall.return_value = []
+
+    patched_store.get_metadata_field_unique_values("category")
+
+    sql = cursor.execute.call_args_list[0][0][0]
+    assert "LIKE" not in sql
```
anakin87: This seems just a regression test that can be removed. This is already covered by https://github.com/deepset-ai/haystack/blob/22117f90dcfae4a3a23c0d1ca0bd2a4026ad9797/haystack/testing/document_store.py#L1212. Right?

anakin87 (changes requested):
Left a few comments

integrations/oracle/src/haystack_integrations/document_stores/oracle/document_store.py
davidsbatista: Good catch! Oracle doesn't have the "ILIKE" operator, and I thought this was case-insensitive, but it's not.

https://docs.oracle.com/cd/B13789_01/server.101/b10759/conditions016.htm

anakin87 (approved):
Looks good!

PR #3646, opened 2026-07-22
fix: `ElasticSearch` match `search_term` against metadata field value, not content

davidsbatista:
### Related Issues

- fixes partially https://github.com/deepset-ai/haystack-private/issues/488

### Proposed Changes:

- search_term now matches the metadata field value

### Checklist

### DIFF
--- a/integrations/elasticsearch/src/haystack_integrations/document_stores/elasticsearch/document_store.py
+++ b/integrations/elasticsearch/src/haystack_integrations/document_stores/elasticsearch/document_store.py
@@ -3,7 +3,6 @@
 # SPDX-License-Identifier: Apache-2.0
 
 import copy
-import re
 
 # ruff: noqa: FBT002, FBT001    boolean-type-hint-positional-argument and boolean-default-value-positional-argument
 # ruff: noqa: B008              function-call-in-default-argument
@@ -1888,10 +1887,21 @@ class ElasticsearchDocumentStore:
         }
         if search_term:
             # Composite aggregation terms sources don't support `include`/`exclude` (that's only valid on
-            # standalone `terms` aggregations), so the substring match on the field's own value is applied
-            # as a query-level filter instead. Case-sensitive: there is no normalizer-based case-insensitivity
-            # infrastructure elsewhere in this document store to hook into.
-            body["query"] = {"regexp": {field_name: f".*{re.escape(search_term)}.*"}}
+            # standalone `terms` aggregations), and a `regexp` query only works on keyword/text fields, not
+            # numeric ones. A doc-value script query works uniformly across field types by stringifying the
+            # field's value, matching the case-sensitive substring semantics documented above.
+            body["query"] = {
+                "script": {
+                    "script": {
+                        "source": (
+                            "def v = doc[params.field]; "
+                            "if (v.size() == 0) { return false; } "
+                            "return v.value.toString().contains(params.term)"
+                        ),
+                        "params": {"field": field_name, "term": search_term},
+                    }
+                }
+            }
 
         result = self.client.search(index=self._index, body=body)
         aggregations = result.get("aggregations", {})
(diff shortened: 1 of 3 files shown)


anakin87:
I'd prefer @sjrl to review this PR since https://github.com/deepset-ai/haystack-private/issues/488 is not totally clear to me.
In any case, I'll follow his review to better understand...

sjrl:
@anakin87 I believe the OpenSearch issue https://github.com/deepset-ai/haystack-private/issues/482 better explains what needs to be changed. But basically the issue is for updating **search_term** to work on the metadata value and that the spec of the function should be updated to 
```python
    self, metadata_field: str, search_term: str | None = None, from_: int = 0, size: int = 10
) -> tuple[list[str], int]:
```
if possible. 

So it looks like the changes of switching `after` to `from_` and the return type haven't been done yet. @davidsbatista are you planning to do those changes in this PR or a separate one?

davidsbatista:
yes, those will come in another PR

integrations/elasticsearch/src/haystack_integrations/document_stores/elasticsearch/document_store.py
```diff
@@ -1845,19 +1845,20 @@ def get_metadata_field_unique_values(
...
 
         Uses composite aggregations for proper pagination beyond 10k results.
 
         See: https://www.elastic.co/docs/reference/aggregations/search-aggregations-bucket-composite-aggregation
 
         :param metadata_field: The metadata field to get unique values for.
-        :param search_term: Optional search term to filter documents by matching in the content field.
-        :param size: The number of unique values to return per page. Defaults to 10000.
+        :param search_term: Optional term to filter the returned values by, matching as a case-sensitive substring
+            of the metadata field's own value (not the document content).
```
sjrl: No need to mention the previous implementation
```suggestion
            of the metadata field's own value.
```

integrations/elasticsearch/src/haystack_integrations/document_stores/elasticsearch/document_store.py
```diff
@@ -1912,19 +1923,21 @@ async def get_metadata_field_unique_values_async(
...
+        Optionally filtered by a substring match on the field's own value.
         Uses composite aggregations for proper pagination beyond 10k results.
 
         See: https://www.elastic.co/docs/reference/aggregations/search-aggregations-bucket-composite-aggregation
 
         :param metadata_field: The metadata field to get unique values for.
-        :param search_term: Optional search term to filter documents by matching in the content field.
-        :param size: The number of unique values to return per page. Defaults to 10000.
+        :param search_term: Optional term to filter the returned values by, matching as a case-sensitive substring
+            of the metadata field's own value (not the document content).
```
sjrl:
```suggestion
            of the metadata field's own value.
```

integrations/elasticsearch/src/haystack_integrations/document_stores/elasticsearch/document_store.py
```diff
@@ -1949,15 +1956,31 @@ async def get_metadata_field_unique_values_async(
...
+            # standalone `terms` aggregations), and a `regexp` query only works on keyword/text fields, not
+            # numeric ones. A doc-value script query works uniformly across field types by stringifying the
+            # field's value, matching the case-sensitive substring semantics documented above.
+            body["query"] = {
+                "script": {
+                    "script": {
+                        "source": (
+                            "def v = doc[params.field]; "
+                            "if (v.size() == 0) { return false; } "
+                            "return v.value.toString().contains(params.term)"
```
sjrl: could we update this to be case-insensitive like in the other integrations? I think we can update to `v.value.toString().toLowerCase().contains(params.term.toLowerCase())`

davidsbatista: I was reviewing it and just found this bug as well. On it 👍🏽  (same fix for OpenSearch)

integrations/elasticsearch/src/haystack_integrations/document_stores/elasticsearch/document_store.py
```diff
@@ -1912,19 +1923,21 @@ async def get_metadata_field_unique_values_async(
...
 
+        Optionally filtered by a substring match on the field's own value.
         Uses composite aggregations for proper pagination beyond 10k results.
 
         See: https://www.elastic.co/docs/reference/aggregations/search-aggregations-bucket-composite-aggregation
 
         :param metadata_field: The metadata field to get unique values for.
-        :param search_term: Optional search term to filter documents by matching in the content field.
-        :param size: The number of unique values to return per page. Defaults to 10000.
+        :param search_term: Optional term to filter the returned values by, matching as a case-sensitive substring
```
sjrl: Since we need to use a script to accomplish the substring matching on the value of the field this operation is quite expensive for a large corpus. Could we add that as a warning in the docstrings here and in the sync method?

integrations/elasticsearch/src/haystack_integrations/document_stores/elasticsearch/document_store.py
```diff
@@ -1845,19 +1845,20 @@ def get_metadata_field_unique_values(
...
 
         Uses composite aggregations for proper pagination beyond 10k results.
 
         See: https://www.elastic.co/docs/reference/aggregations/search-aggregations-bucket-composite-aggregation
 
         :param metadata_field: The metadata field to get unique values for.
-        :param search_term: Optional search term to filter documents by matching in the content field.
-        :param size: The number of unique values to return per page. Defaults to 10000.
+        :param search_term: Optional term to filter the returned values by, matching as a case-insensitive substring
+            of the metadata field's own value.
```
sjrl: Could we add the warning about the potentially slow performance of using search term here as well?

sjrl (approved):
One remaining minor comment, otherwise looks good!

PR #3655, opened 2026-07-23
fix(alloydb): parenthesize `not in` clause so it does not escape an AND filter

sjrl:
### Related Issues

- found by reviewing similar issue in pgvector https://github.com/deepset-ai/haystack-core-integrations/pull/3586

### Proposed Changes:

 
 

Same change as in https://github.com/deepset-ai/haystack-core-integrations/pull/3586

Wrap the expression in parentheses so the internal `OR` stays grouped:

```python
return SQL("({} IS NULL OR {} != ALL(%s))").format(field, field), [value]
```

### How did you test it?

- Added `test_logical_condition_not_in_is_parenthesized_when_and_combined`, which AND-combines `not in` with another condition and asserts the `not in` clause is parenthesized. It fails before the change (the `OR` escapes the `AND`) and passes after.
- Also added some additional unit tests that test filter errors that can occur

### Checklist


davidsbatista (approved):
good catch 👍🏽

PR #3668, opened 2026-07-27
feat: add Celeris integration

tomhamer:
### Related Issues

None — opening this directly as a new integration. Happy to file one first if you'd prefer.

### Proposed Changes:

Adds a `celeris` integration providing `CelerisChatGenerator`, a subclass of `OpenAIChatGenerator`.

[Celeris](https://celeris.ai) serves diffusion-based language models over an OpenAI-compatible chat-completions API, aimed at low-latency short responses — classification, extraction, judging, query rewriting. `celeris-1` is the only model served today.

**Disclosure:** I work with Celeris — this is a vendor-submitted integration and I'll maintain it. Flagging that up front so you can weight the review accordingly. The code was drafted with AI assistance and then reviewed and tested by hand against the live API.

A plain `OpenAIChatGenerator` pointed at their base URL does not work, which is why this is a subclass rather than a docs note. The API has two constraints that ordinary usage trips over immediately:

**`max_tokens` must be `1` or a positive multiple of 256.**

```
400 max_tokens must be 1 (warm ping) or a positive multiple of 256, got 100
```

The component rounds the caller's value up to the next multiple of 256. `1` is passed through, since the API accepts it as a warm ping.

**Prompt and completion share one 8,192-token budget.** Not 8K in plus 8K out — a single combined budget:

```
400 This model's maximum context length is 8192 tokens. However, you requested
1024 output tokens and your prompt contains at least 7169 input tokens, for a
total of at least 8193 tokens.
```

So `max_tokens` is additionally capped downward — still on a 256 boundary — to fit alongside the estimated prompt. If the prompt leaves under 256 tokens of room, the component raises `ValueError` rather than letting the API return a 400.

Parameters the API genuinely doesn't implement raise `ValueError` instead of failing at request time: `response_format` (no JSON mode, no JSON schema — structured output has to route through tool calling), image content, and `tool_choice="required"`.

### How did you test it?

`hatch run test:unit-cov-retry` — **52 passed, 5 deselected**, 100% coverage on `chat_generator.py`. `hatch run fmt-check` and `hatch run test:types` (mypy) both clean.

Tests are fully mocked, so CI needs no API key. I verified that by running the suite with `CELERIS_API_KEY` unset and all HTTP forced through a dead proxy. The integration-marked tests skip cleanly without the secret (`5 skipped, 50 deselected`, exit 0), so the workflow won't fail before a secret is configured.

Also verified on the other end of your CI matrix — Python 3.10 with `haystack-ai==2.22.0` — to confirm the lowest-direct-dependencies job passes. Every API the component relies on (`ToolsType`, `flatten_tools_or_toolsets`, `_prepare_api_call`, `ChatMessage.images`) exists in 2.22.0.

Manual verification against the live API covered basic completion, streaming, tool calling, and both 400 paths above.

### Notes for the reviewer

- `.github/workflows/celeris.yml` is `mistral.yml` with the name, path, secret and coverage ID substituted — mistral was the closest model since it also subclasses `OpenAIChatGenerator`. It includes the nightly Haystack-main job.
- The API base URL is pinned to `https://inference.celeris.ai/celeris-1/v1` because Celeris namespaces the endpoint by model rather than serving many models from one base.
- The prompt-size estimate is a character-based heuristic, deliberately biased to over-count. Celeris publishes no tokenizer, and over-counting only shortens the completion; under-counting would produce a 400.
- One quirk worth knowing: Celeris's own API reference lists `logprobs` as unsupported, but the endpoint accepts it and returns per-token values. The component passes it through rather than rejecting it, and the docstring says so. I've asked them to correct their reference.
- Running the workflow requires a `CELERIS_API_KEY` secret. Happy to sort that out with whoever handles it.

### Checklist

- I've used a conventional commit type for my PR title


anakin87:
Hi @tomhamer and thanks for submitting this PR. 

We recommend publishing and releasing this integration yourself using [our repository template](https://github.com/deepset-ai/custom-component).

Once you have released it, your integration can be added to this [repo](https://github.com/deepset-ai/haystack-integrations) so that it gets listed on our [web page ](https://haystack.deepset.ai/integrations).

This approach gives you full ownership of the integration while still making it discoverable to the wider Haystack community.

Sounds good?

anakin87:
Closing as per https://github.com/deepset-ai/haystack-core-integrations/pull/3668#issuecomment-5102053498

PR #3677, opened 2026-07-28
fix: `AlloyDBDocumentStore` match case insensitive metadata value

davidsbatista:
### Related Issues

- fixes partially https://github.com/deepset-ai/haystack-private/issues/485

### Proposed Changes:

- match case insensitive metadata value

### How did you test it?

- added two more tests, for case insensitive and pagination since there was o test yet

### Checklist

### DIFF
--- a/integrations/alloydb/src/haystack_integrations/document_stores/alloydb/document_store.py
+++ b/integrations/alloydb/src/haystack_integrations/document_stores/alloydb/document_store.py
@@ -1262,12 +1262,18 @@ class AlloyDBDocumentStore(DocumentStore):
         return {"min": result.get("min_value"), "max": result.get("max_value")}
 
     def _build_unique_values_queries(
-        self, normalized_field: str, search_term: str | None, from_: int, size: int
+        self,
+        normalized_field: str,
+        filters: dict[str, Any] | None,
+        search_term: str | None,
+        from_: int,
+        size: int,
     ) -> tuple[Composed, Composed, tuple]:
         """
         Builds SQL queries for getting unique metadata field values.
 
         :param normalized_field: The normalized metadata field name.
+        :param filters: Optional filters to restrict the documents considered.
         :param search_term: Optional search term to filter unique values by a case-insensitive substring match
             against the metadata field's own value.
         :param from_: The offset for pagination (0-based).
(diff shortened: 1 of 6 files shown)


integrations/alloydb/tests/test_document_store.py
```diff
@@ -82,6 +82,44 @@ def test_write_blob(self, document_store: AlloyDBDocumentStore):
...
+    def test_get_metadata_field_unique_values_search_term(self, document_store: AlloyDBDocumentStore):
+        docs = [
+            Document(content="This document mentions Python explicitly", meta={"language": "Python"}),
+            Document(content="Unrelated content about recipes", meta={"language": "Java"}),
+            Document(content="Another one", meta={"language": "JavaScript"}),
+        ]
+        document_store.write_documents(docs)
+
+        # case-insensitive substring match against the metadata field's OWN value, not content
+        values, total = document_store.get_metadata_field_unique_values("meta.language", "python", 0, 10)
```
sjrl: It looks like this call is being made incorrectly. The `get_metadata_field_unique_values` function still has the `filters` in its method signature and the call here assumes its not present. Perhaps for safety we could call this function using keyword arguments instead of positional?
```python
        self,
        field: str,
        filters: dict[str, Any] | None = None,
        from_: int = 0,
        size: int = 10,
```

sjrl: I think this applies to all current tests.

davidsbatista: good catch - thanks

integrations/alloydb/src/haystack_integrations/document_stores/alloydb/document_store.py
```diff
@@ -1261,33 +1261,106 @@ def get_metadata_field_min_max(self, field: str) -> dict[str, Any]:
...
+        :param records: The list of records from the select query.
+        :returns: A tuple containing (unique_values, total_count).
+        """
+        total_count = count_result.get("total", 0) if count_result else 0
+        unique_values = [str(record.get("value", "")) for record in records if record.get("value") is not None]
+        return unique_values, total_count
+
+    def get_metadata_field_unique_values(
+        self,
+        field: str,
```
sjrl: Sorry missed this in the issue but we are converging on using the name `metadata_field` for this parameter right? If so probably worth changing here as well.

integrations/alloydb/src/haystack_integrations/document_stores/alloydb/document_store.py
```diff
@@ -1261,33 +1261,106 @@ def get_metadata_field_min_max(self, field: str) -> dict[str, Any]:
...
+        filters: dict[str, Any] | None = None,
```
sjrl: If we want to add `filters` as an optional param, it should probably move to the end of the function def so if were to define the protocol as 
```python
    self, metadata_field: str, search_term: str | None = None, from_: int = 0, size: int = 10
) -> tuple[list[str], int]:
```
then this doc store would still obey it even with the additional `filters` param

integrations/alloydb/src/haystack_integrations/document_stores/alloydb/document_store.py
davidsbatista: yes we are, I missed this one - I'm fixing it and will check the other open PRs as well

integrations/alloydb/src/haystack_integrations/document_stores/alloydb/document_store.py
davidsbatista: same for here, moving as last arg and checking other PRs - should be only Qdrant

sjrl (approved):
Looks good!

PR #3678, opened 2026-07-28
feat(litellm): include provider cache tokens in usage metadata

Aftabbs:
## Summary

`_extract_usage` in `LiteLLMChatGenerator` only captured the three base token counts (`prompt_tokens`, `completion_tokens`, `total_tokens`) and silently dropped provider-specific cache fields that LiteLLM already surfaces on its `Usage` object:

- `cache_creation_input_tokens` / `cache_read_input_tokens` for **Anthropic** (`anthropic/*` and `bedrock/anthropic.*` models with `cache_control` blocks)
- `prompt_tokens_details.cached_tokens` for **OpenAI** (cached prompt tokens served from KV cache)

Without these fields, users routing to Anthropic or OpenAI via LiteLLM have no way to inspect prompt-caching costs from the `ChatMessage` metadata produced by the generator.

## Changes

- Extended `_extract_usage` to include cache token fields when non-zero (zero values are omitted to keep metadata clean for providers that never use caching)
- Updated the return type annotation from `dict[str, int]` to `dict[str, Any]` to reflect the actual content
- Added three new unit tests:
  - `test_anthropic_cache_tokens_in_meta` — non-streaming path with Anthropic cache fields
  - `test_openai_cached_tokens_in_meta` — non-streaming path with OpenAI `prompt_tokens_details`
  - `test_streaming_anthropic_cache_tokens_in_meta` — streaming path (final usage-only chunk) with Anthropic cache fields

## Why now

Anthropic prompt caching is widely used to reduce costs on long system prompts. Without surfacing `cache_read_input_tokens` in message metadata, pipeline operators can't tell whether caching is working or estimate savings. LiteLLM already provides these values — the gap was purely in our extraction layer.

## Test plan

All existing unit tests continue to pass (the change is additive). New tests verify that:
1. Anthropic cache tokens appear in `reply.meta["usage"]` for both streaming and non-streaming paths
2. OpenAI `cached_tokens` appear via `prompt_tokens_details` in the non-streaming path
3. Non-caching responses (zero cache fields) produce unchanged metadata with no extra keys

### DIFF
--- a/integrations/litellm/src/haystack_integrations/components/generators/litellm/chat/chat_generator.py
+++ b/integrations/litellm/src/haystack_integrations/components/generators/litellm/chat/chat_generator.py
@@ -266,16 +266,31 @@ _FINISH_REASON_MAPPING: dict[str, FinishReason] = {
 }
 
 
-def _extract_usage(obj: Any) -> dict[str, int]:
+def _extract_usage(obj: Any) -> dict[str, Any]:
     """Pull token usage off a litellm response or chunk, tolerating its absence."""
     usage = getattr(obj, "usage", None)
     if not usage:
         return {}
-    return {
+    result: dict[str, Any] = {
         "prompt_tokens": getattr(usage, "prompt_tokens", 0),
         "completion_tokens": getattr(usage, "completion_tokens", 0),
         "total_tokens": getattr(usage, "total_tokens", 0),
     }
+    # Anthropic prompt-caching fields (present when using anthropic/* or bedrock/anthropic.* models
+    # with cache_control blocks in the request)
+    cache_creation = getattr(usage, "cache_creation_input_tokens", None)
+    if cache_creation:
+        result["cache_creation_input_tokens"] = cache_creation
+    cache_read = getattr(usage, "cache_read_input_tokens", None)
+    if cache_read:
+        result["cache_read_input_tokens"] = cache_read
+    # OpenAI cached prompt tokens (present when the prompt was served from OpenAI's KV cache)
+    prompt_details = getattr(usage, "prompt_tokens_details", None)
+    if prompt_details:
+        cached = getattr(prompt_details, "cached_tokens", None)
+        if cached:
+            result["cached_tokens"] = cached
+    return result
 
 
 def _build_chat_message(response: Any, choice: Any) -> ChatMessage:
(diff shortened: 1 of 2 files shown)


integrations/litellm/src/haystack_integrations/components/generators/litellm/chat/chat_generator.py
```diff
@@ -266,16 +266,31 @@ def from_dict(cls, data: dict[str, Any]) -> "LiteLLMChatGenerator":
...
+    cache_read = getattr(usage, "cache_read_input_tokens", None)
+    if cache_read:
+        result["cache_read_input_tokens"] = cache_read
+    # OpenAI cached prompt tokens (present when the prompt was served from OpenAI's KV cache)
+    prompt_details = getattr(usage, "prompt_tokens_details", None)
+    if prompt_details:
+        cached = getattr(prompt_details, "cached_tokens", None)
+        if cached:
+            result["cached_tokens"] = cached
+    return result
```
anakin87: for simplicity, we can replace this block with
```python
if hasattr(usage, "model_dump"):
    return usage.model_dump(exclude_none=True)
```

Ok to serialize everything inside `usage`

integrations/litellm/tests/test_litellm_chat_generator.py
```diff
@@ -236,6 +236,64 @@ def test_usage_in_meta(self):
...
+        with mock.patch.dict(sys.modules, {"litellm": fake_litellm}):
+            result = gen.run(
+                messages=[ChatMessage.from_user("hi")],
+                streaming_callback=lambda _c: None,
+            )
+            reply_usage = result["replies"][0].meta["usage"]
+            assert reply_usage["cache_creation_input_tokens"] == 1200
+            assert reply_usage["cache_read_input_tokens"] == 800
+            assert reply_usage["total_tokens"] == 15
+
```
anakin87, line 258: If we go with my simplification suggestion above, I recommend a simple unit test like

```python
from litellm.types.utils import Usage

def test_cache_tokens_in_meta(self):
    gen = LiteLLMChatGenerator(model="anthropic/claude-sonnet-4-20250514")
    mock_resp = _make_mock_response(model="anthropic/claude-sonnet-4-20250514")
    mock_resp.usage = Usage(
        prompt_tokens=2010,
        completion_tokens=5,
        total_tokens=2015,
        cache_creation_input_tokens=1200,
        cache_read_input_tokens=800,
    )

    fake_litellm = types.ModuleType("litellm")
    fake_litellm.completion = MagicMock(return_value=mock_resp)

    with mock.patch.dict(sys.modules, {"litellm": fake_litellm}):
        usage = gen.run(messages=[ChatMessage.from_user("hi")])["replies"][0].meta["usage"]
        assert usage["cache_creation_input_tokens"] == 1200
        assert usage["cache_read_input_tokens"] == 800
        assert usage["prompt_tokens"] == 2010
```

anakin87 (changes requested):

anakin87 (approved):
Pushed a small fix

PR #3683, opened 2026-07-29
fix(twelvelabs): avoid orphaned async embed tasks when a batch call fails

### LINKED ISSUE #3578: TwelveLabsDocumentEmbedder.run_async leaks orphaned embed_text_async tasks when one call in a batch fails
### Bug Description

`TwelveLabsDocumentEmbedder.run_async` embeds each batch concurrently with `asyncio.gather` over `embed_text_async` calls, without `return_exceptions=True`:

```python
batch_embeddings = await asyncio.gather(*(embed_text_async(text, self.model, key) for text in batch))
```

`embed_text_async` (in `_embed.py`) has no exception handling of its own, it awaits the real TwelveLabs API client directly and can raise from either the network call itself or from `_extract_vector` when a response comes back with no usable embedding segment:

```python
async def embed_text_async(text: str, model: str, api_key: str) -> list[float]:
    client = AsyncTwelveLabs(api_key=api_key)
    response = await client.embed.create(model_name=model, text=text)
    return _extract_vector(response)
```

If one document in a batch triggers a rate limit, transient network error, or a malformed embedding response, that exception propagates through `gather` immediately, while the sibling `embed_text_async` calls for the rest of the batch are still awaiting their own in-flight requests to the TwelveLabs API. Those requests are left running with nothing awaiting them: the caller has already moved on (the exception has propagated out of `run_async`), so their eventual result is discarded and any later exception they raise is unretrieved.

Verified the underlying `asyncio.gather` behavior directly (independent of the TwelveLabs client, since hitting the real API isn't practical here): one coroutine that fails quickly and one that is still sleeping (standing in for an in-flight HTTP call) when gathered together leave the second one `pending` and running after `gather` has already raised out to the caller, exactly the shape this embedder's batch loop relies on:

```python
async def _run_one(should_fail, delay):
    if should_fail:
        await asyncio.sleep(0.01)
        raise ValueError("failed")
    await asyncio.sleep(delay)
    return "ok"

before = asyncio.all_tasks()
try:
    await asyncio.gather(_run_one(True, 0), _run_one(False, 2.0))
except ValueError:
    pass
print(len(asyncio.all_tasks() - before))  # -> 1
```

### Affected code

`integrations/twelvelabs/src/haystack_integrations/components/embedders/twelvelabs/document_embedder.py`, `run_async`, and the `embed_text_async` helper it calls in `_embed.py`.

### Expected Behavior

A single failed embedding call within a batch shouldn't leave the rest of that batch's in-flight requests running unsupervised after the component has already raised to its caller. Gathering with `return_exceptions=True`, collecting/logging the per-item failures, and only then raising (or returning a partial result, depending on the desired failure semantics) would keep the batch's lifetime bounded to the `run_async` call itself.

### System

deepset-ai/haystack-core-integrations main branch, `twelvelabs` integration, `TwelveLabsDocumentEmbedder.run_async`.

RahilOp:
### Related Issues

- fixes #3578

### Proposed Changes

- Use `asyncio.gather(..., return_exceptions=True)` inside `TwelveLabsDocumentEmbedder.run_async`.
- Inspect the gathered results and raise the first exception that occurred.
- This prevents sibling `embed_text_async` coroutines from becoming orphaned / unretrieved when one call in a batch raises.

### How did you test it?

- Added `test_run_async_propagates_exception_and_waits_for_all_tasks`, which simulates one failing and one succeeding embed call and asserts both are awaited and the original exception is raised.
- Ran the unit tests for the integration:

```bash
cd integrations/twelvelabs
hatch run test:unit
```

Result: `27 passed, 5 deselected`.
- `hatch run fmt` and `hatch run test:types` are green.

### Notes for the reviewer

N/A

### DIFF
--- a/integrations/twelvelabs/src/haystack_integrations/components/embedders/twelvelabs/document_embedder.py
+++ b/integrations/twelvelabs/src/haystack_integrations/components/embedders/twelvelabs/document_embedder.py
@@ -176,6 +176,11 @@ class TwelveLabsDocumentEmbedder:
             desc="Calculating embeddings",
         ):
             batch = texts_to_embed[i : i + self.batch_size]
-            batch_embeddings = await asyncio.gather(*(embed_text_async(text, self.model, key) for text in batch))
-            embeddings.extend(batch_embeddings)
+            batch_embeddings = await asyncio.gather(
+                *(embed_text_async(text, self.model, key) for text in batch), return_exceptions=True
+            )
+            for embedding in batch_embeddings:
+                if isinstance(embedding, BaseException):
+                    raise embedding
+            embeddings.extend(batch_embeddings)  # type: ignore[arg-type]
         return self._build_result(documents, embeddings, self.model)
--- a/integrations/twelvelabs/tests/test_document_embedder.py
+++ b/integrations/twelvelabs/tests/test_document_embedder.py
@@ -98,6 +98,24 @@ async def test_run_async_rejects_non_documents():
         await embedder.run_async(documents="a string")
 
 
+@pytest.mark.asyncio
+async def test_run_async_propagates_exception_and_waits_for_all_tasks():
+    embedder = TwelveLabsDocumentEmbedder(api_key=Secret.from_token("tlk_test"))
+    docs = [Document(content="first"), Document(content="second")]
+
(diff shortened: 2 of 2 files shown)


davidsbatista (approved):
thanks for the contribution!

PR #3688, opened 2026-07-29
fix: `FalkorDB` changing pagination to standard method

davidsbatista:
### Related Issues

- fixes https://github.com/deepset-ai/haystack-private/issues/483

### Proposed Changes:

- returns (values, count) and has params: from_: int = 0, size: int = 10,

### How did you test it?

- updated the existing tests

### Checklist

### DIFF
--- a/integrations/falkordb/src/haystack_integrations/document_stores/falkordb/document_store.py
+++ b/integrations/falkordb/src/haystack_integrations/document_stores/falkordb/document_store.py
@@ -607,25 +607,24 @@ SET d.{self.embedding_field} = vecf32(doc.emb)
         self,
         metadata_field: str,
         search_term: str | None = None,
-        size: int | None = 10,
-        after: dict[str, Any] | None = None,
-    ) -> tuple[list[Any], dict[str, Any] | None]:
+        from_: int = 0,
+        size: int = 10,
+    ) -> tuple[list[Any], int]:
         """
         Return distinct values for the given metadata field with optional filtering and pagination.
 
         :param metadata_field: Metadata field name. May include or omit the `meta.` prefix.
         :param search_term: Optional case-insensitive substring filter applied to the metadata
             field's own value.
-        :param size: Maximum number of values to return per page. Defaults to 10 000.
-        :param after: Pagination cursor returned by a previous call. Pass `None` for the first page.
-        :returns: Tuple of `(values, next_cursor)`. `next_cursor` is `None` on the last page.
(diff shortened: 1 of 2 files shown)


integrations/falkordb/tests/test_document_store.py
```diff
@@ -371,26 +371,26 @@ def test_get_metadata_field_min_max(self, mock_falkordb, rows, expected):
...
 
     def test_get_metadata_field_unique_values_pagination(self, mock_falkordb):
         _, _, graph = mock_falkordb
-        graph.query.side_effect = [_result([]), _result([]), _result([["A"], ["B"], ["C"]])]
-        values, cursor = FalkorDBDocumentStore().get_metadata_field_unique_values("category", size=2)
+        graph.query.side_effect = [_result([]), _result([]), _result([[["A", "B"], 3]])]
+        values, total = FalkorDBDocumentStore().get_metadata_field_unique_values("category", size=2)
         assert values == ["A", "B"]
-        assert cursor == {"offset": 2}
+        assert total == 3
```
sjrl: Could we update this test to a non-zero `from_` since thats what pagination should be testing right?

integrations/falkordb/src/haystack_integrations/document_stores/falkordb/document_store.py
```diff
@@ -607,25 +607,24 @@ def get_metadata_field_unique_values(
         self,
         metadata_field: str,
         search_term: str | None = None,
-        size: int | None = 10,
-        after: dict[str, Any] | None = None,
-    ) -> tuple[list[Any], dict[str, Any] | None]:
+        from_: int = 0,
+        size: int = 10,
+    ) -> tuple[list[Any], int]:
```
sjrl: Is it possible to update this to 
```suggestion
    ) -> tuple[list[str], int]:
```
or is there a reason it must be Any?

sjrl: If so lets leave a note in the returns docstring what types of values can be returned.

davidsbatista: not particular reason - I changed to string

integrations/falkordb/tests/test_document_store.py
davidsbatista: I've changed to a more robust test

sjrl (approved):
Looks good!

PR #3697, opened 2026-07-30
fix: `Qdrant` adding offset based pagination to `get_metadata_field_unique_values()`

davidsbatista:
### Related Issues

- fixes https://github.com/deepset-ai/haystack-private/issues/484

### Proposed Changes:

- adding offset based pagination 

### How did you test it?

- updated existing tests

### Checklist

### DIFF
--- a/integrations/qdrant/src/haystack_integrations/document_stores/qdrant/document_store.py
+++ b/integrations/qdrant/src/haystack_integrations/document_stores/qdrant/document_store.py
@@ -1288,11 +1288,14 @@ class QdrantDocumentStore:
         from_: int = 0,
         size: int = 10,
         filters: dict[str, Any] | None = None,
-    ) -> tuple[list[str], int]:
+    ) -> tuple[list[Any], int]:
         """
         Returns unique values for a metadata field, with optional filters, search term and pagination.
 
         Unique values are ordered by first occurrence during scroll.
+
+        **Note**: This operation can be expensive for metadata fields with many unique values, since all
+        matching documents must be scrolled through to compute the total count.
 
         :param metadata_field: The metadata field key (inside ``meta``) to get unique values for.
         :param search_term: Optional case-insensitive substring filter applied to the metadata field's own value.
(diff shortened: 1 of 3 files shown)


integrations/qdrant/tests/test_document_store.py
```diff
@@ -269,20 +269,18 @@ def test_process_records_count_unique(self):
...
-    def test_process_records_unique_values_stops_when_filled(self):
+    def test_process_records_unique_values_collects_all(self):
         records = [SimpleNamespace(payload={"meta": {"v": i}}) for i in range(10)]
         values: list = []
         values_set: set = set()
-        done = QdrantDocumentStore._process_records_unique_values(records, "v", values, values_set, offset=0, limit=3)
-        assert done is True
-        assert values[:3] == [0, 1, 2]
+        QdrantDocumentStore._process_records_unique_values(records, "v", values, values_set)
+        assert values == list(range(10))
```
sjrl, line 277: Hey it seems like this test implies that the unique values returned by `_process_records_unique_values` and then also returned by `get_metadata_field_unique_values` are not guaranteed to be a `list[str]`. Instead in this case it shows that the values come back as a list of integers right?

integrations/qdrant/src/haystack_integrations/document_stores/qdrant/document_store.py
```diff
@@ -1371,7 +1364,7 @@ async def get_metadata_field_unique_values_async(
 
         try:
             next_offset = None
-            while len(unique_values) < offset + limit:
+            while True:
```
sjrl, line 1373: I think its worth mentioning in the docstrings that this operation can now be expensive if a metadata field has many unique values since we iterate over all of them to be able to calculate the total amount.

davidsbatista: this will also be the same for ElasticSearch and OpenSearch - I will add a docstring

integrations/qdrant/tests/test_document_store.py
davidsbatista: should be `Any`

sjrl: Okay but this violates the protocol we established for this method right? Should the protocol be updated?

davidsbatista: I think we should update the protocol to return `Any`, to avoid losing information/types. I overlooked that we were casting everything to `str`

sjrl: Okay sounds good! Could you double check if any already merged PRs on existing docstores should be updated to update the type and avoid unnecessarily casting to string?

davidsbatista: Most of them are casting to string - I will open individual PRs for each one that needs this update

sjrl (approved):
Looks good!

PR #3698, opened 2026-07-30
fix(docling): stringify binary_hash to prevent OpenSearch long overflow

### LINKED ISSUE #3604: DoclingConverter: binary_hash metadata can exceed signed 64-bit integer range
**Describe the bug**
The default DoclingConverter includes dl_meta.origin.binary_hash in the document metadata. binary_hash is an unsigned 64-bit integer (uint64), but some document stores (e.g. deepset Cloud/OpenSearch) store numeric metadata as a signed 64-bit long.

When binary_hash is greater than 2^63 - 1, document indexing fails with an error similar to:

failed to parse field [dl_meta.origin.binary_hash] of type [long]
Numeric value (...) out of range of long

Since binary_hash is primarily implementation metadata, it might make sense to either omit it from the default metadata extraction or export it as a string instead.

**To Reproduce**
1. Create a pipeline using the default DoclingConverter.
2. Index a document whose dl_meta.origin.binary_hash exceeds the signed 64-bit integer range.
3. Observe that indexing fails because the metadata field cannot be mapped to an OpenSearch long.

For example, the metadata contains:

{
  "dl_meta": {
    "origin": {
      "binary_hash": 9768961288489567249
    }
  }
}

which results in:

failed to parse field [dl_meta.origin.binary_hash] of type [long]
Numeric value (9768961288489567249) out of range of long

As a workaround, removing binary_hash (or converting it to a string field such as binary_hash_text) before writing documents avoids the issue.

* current HEP builder env

SyedShahmeerAli12:
## Related Issues

- fixes #3604

## Proposed Changes

`MetaExtractor` previously stored Docling's `origin.binary_hash` as an integer in document metadata. Since `binary_hash` is an unsigned 64-bit integer, values greater than `2^63 - 1` break document stores that map numeric metadata to signed 64-bit types (e.g. OpenSearch `long`), causing:

```
failed to parse field [dl_meta.origin.binary_hash] of type [long]
Numeric value (9768961288489567249) out of range of long
```

Following @julian-risch's guidance on #3604, `binary_hash` is now always stored as a string via a new `_stringify_binary_hash()` recursive helper applied in both `extract_dl_doc_meta` and `extract_chunk_meta`.

## Reproduction

Before fix â€” `binary_hash` returned as `int`, overflows OpenSearch `long`:
```python
# model_dump() returns: {binary_hash: 9768961288489567249}  â† int, rejected by OpenSearch
```

After fix â€” stored as `str`, no overflow:
```python
# _stringify_binary_hash() returns: {binary_hash: 9768961288489567249}  â† safe keyword
```

## Testing

- Added 3 new unit tests: oversized hash in doc meta, chunk meta, and nested structures
- All 8 `TestMetaExtractor` tests pass
- `ruff check` clean

## Notes

- Breaking metadata-type change for callers that assumed `binary_hash` was `int` â€” aligns with maintainer suggestion for a major release
- Both small and large hashes are stringified for consistency


SyedShahmeerAli12:
Closing  reopening with a clean branch that doesn't carry unrelated MariaDB commits.

PR #3717, opened 2026-08-02
fix(anthropic): include cached tokens in the OpenAI-compatible prompt_tokens

arthi-arumugam-git:
## The bug

`_get_openai_compatible_usage` renames Anthropic's `input_tokens` to `prompt_tokens`:

```python
if "input_tokens" in usage:
    usage["prompt_tokens"] = usage.pop("input_tokens")
```

The two do not mean the same thing, and the difference is exactly the thing the function exists to normalise.

- **OpenAI's** `prompt_tokens` **includes** tokens served from the prompt cache. The cached portion is reported separately in `prompt_tokens_details.cached_tokens` as a subset.
- **Anthropic's** `input_tokens` is the opposite: it is **net** of the cache, with the cached tokens in `cache_read_input_tokens` and `cache_creation_input_tokens`.

So the conversion claims OpenAI semantics for a number that does not have them. All three are billed, and `input_tokens` is by definition only the part that was *not* cached, so the warmer the cache the smaller the reported figure gets:

```
input_tokens                   3
cache_read_input_tokens        20000
cache_creation_input_tokens    1500

prompt_tokens (before)         3
actually billed                21503
```

This matters here specifically because prompt caching is a supported, documented feature of this integration, with its own `example/prompt_caching.py`. The integration is at its most useful precisely when this number is at its most wrong.

## The fix

Fold the cached tokens into `prompt_tokens` so it means what OpenAI means by it. The Anthropic-native keys are deliberately **left in the dict untouched**, so anything already reading `cache_read_input_tokens` or `cache_creation_input_tokens` directly, including `example/prompt_caching.py`, is unaffected.

`or 0` rather than a default in `.get`, because Anthropic sends these as `null` rather than omitting them when unused. Both shapes already appear in your fixtures (`cache_creation_input_tokens=0` in one test, `=None` in others), and both are covered.

## Tests

Eight tests in `integrations/anthropic/tests/test_cached_prompt_tokens.py`.

**Three fail on `main`:**

```
FAILED test_cache_read_tokens_are_included_in_prompt_tokens
FAILED test_cache_creation_tokens_are_included
FAILED test_both_cache_fields_are_included
```

**Five pass on both**, and they are the point: the uncached case, explicit zeros, explicit nulls, that the Anthropic-native keys survive, and that empty usage is untouched.

All 29 existing tests in `tests/test_utils.py` still pass.

## Scope note

I first also changed the streaming path in `chat_generator.py`, on the assumption that it dropped the cache keys because it only assigns `input_tokens` into `meta["usage"]`. That was wrong: Haystack core's chunk converter already propagates the full usage dict, so the cache keys are present there, and your own test asserting `usage["cache_creation_input_tokens"] is None` proves it. I reverted that change. This PR touches one file.

## Related

The same defect in three other frameworks, two already merged: [`pipecat-ai/pipecat#5163`](https://github.com/pipecat-ai/pipecat/pull/5163), [`livekit/agents#6663`](https://github.com/livekit/agents/pull/6663), and [`run-llama/llama_index#22548`](https://github.com/run-llama/llama_index/pull/22548). Four independent implementations treating Anthropic's `input_tokens` as the whole input. The shape of the mistake seems easier to make than to notice, which is why the tests here sit on the boundary rather than the happy path.


integrations/anthropic/src/haystack_integrations/components/generators/anthropic/chat/utils.py
```diff
@@ -360,11 +360,26 @@ def _convert_messages_to_anthropic_format(
...
+
+    OpenAI's `prompt_tokens` INCLUDES tokens served from the prompt cache, with the cached
+    portion reported separately in `prompt_tokens_details.cached_tokens`. Anthropic's
+    `input_tokens` is the opposite: it is net of the cache, and the cached tokens live in
+    `cache_read_input_tokens` and `cache_creation_input_tokens`.
+
+    Renaming `input_tokens` to `prompt_tokens` therefore claimed OpenAI semantics for a
+    number that did not have them. All three are billed, so anything downstream reading
+    `prompt_tokens` as an OpenAI-compatible total under-counted by the whole cached portion,
+    which on a warm cache is most of the prompt.
```
anakin87:
```suggestion
```

this section only describes old code, so I'd remove it

integrations/anthropic/tests/test_cached_prompt_tokens.py
```diff
@@ -0,0 +1,115 @@
+"""Cached prompt tokens are billed and must survive the OpenAI-compatible conversion.
```
anakin87: 1. Remove this new file
2. Port `TestOpenAICompatibleUsage` test class into https://github.com/deepset-ai/haystack-core-integrations/blob/main/integrations/anthropic/tests/test_utils.py
3. Remove/parametrize some tests to reduce code without losing coverage

anakin87 (changes requested):
Thank you!

I found some opportunities to improve this PR

arthi-arumugam-git:
Thanks for the review, all three are done.

The docstring now describes only the current behaviour. The paragraph explaining what the function used to do is gone, since the reader does not need the history.

`test_cached_prompt_tokens.py` is deleted and its coverage moved into `tests/test_utils.py` as `TestOpenAICompatibleUsage`, parametrized down to three test functions covering the same nine cases.

Local run of `tests/test_utils.py`: 38 passed.

anakin87 (approved):

arthi-arumugam-git:
Follow-up on this one, with two things that came out of running the check back over this repo.

First, a correction on my side. The Bedrock chat integration reads `cacheReadInputTokens` and `cacheWriteInputTokens` correctly, and an early version of my checker called that a defect, because it only knew Anthropic's `cache_creation_input_tokens` spelling of the same counter. My bug, now fixed. `haystack-core-integrations` at HEAD reports nothing across 657 files.

Second, this PR's shape earned its own rule. `usage["prompt_tokens"] = usage.pop("input_tokens")` has no arithmetic in it at all, so every pattern I had was blind to it: nothing is added up, a field is just renamed across a boundary where the name changes meaning. That is probably the most common way this defect survives review, and it was the one thing the checker could not see.

The same shape was in four other codebases, none of which share code. https://github.com/arthi-arumugam-git/cachecheck, `uvx cachecheck .`, no dependencies, exit code 1 on a finding.

PR #3539, opened 2026-07-02
ci: scope integration-test secrets to the steps that use them and address SAST findings

FahimaGold:
Addresses a set of high-severity findings raised by an automated SAST scan of this repository.

### Related Issues

No linked issue — these came from an automated security scan rather than a reported bug. Happy to open a tracking issue if you'd prefer one.

### Proposed Changes:

**`gha-workflow-env-secret` — secrets exposed to every step of a workflow (47 workflows)**

Workflow-level `env:` blocks make each secret visible to every step in the workflow, including third-party actions that have no need for them (checkout, setup-python, coverage comment, artifact upload). Each secret is now scoped to the step(s) that actually run integration tests.

`e2b.yml` is the one exception: its secret is scoped to the *job* rather than the step, because the "Run integration tests" step gates on the secret in its own `if:` condition, and a step's own step-level `env:` is not in scope for that same step's `if:`.

**`run-shell-injection` — `CI_pypi_release.yml`**

`${{ github.ref_name }}` was interpolated directly into a `run:` script (once into a shell command, once into an inline Python script). Both now pass it through an intermediate `env:` var and read it from the environment, so tag contents can't terminate the surrounding script.

**AQL injection — `arangodb` document store**

`collection_name` is interpolated directly into AQL queries because ArangoDB collection names cannot be supplied as bind parameters. It is now validated against an allowlist matching ArangoDB's "traditional" naming rules (leading letter, then letters/digits/underscores/dashes, max 256 chars) before use, mirroring the existing identifier validation in the Oracle integration. Invalid names raise `ValueError` from `__init__`, and the docstring documents the constraint.

**`dependabot-missing-cooldown` / `uv-missing-dependency-cooldown`**

Raise the dependabot cooldown `default-days` from 1 to 7, and `exclude-newer` in `uv.toml` from 24 hours to 7 days, widening the window in which a compromised release is likely to be caught and yanked before it reaches CI.

These last two are a policy trade-off rather than a defect fix — a longer window also delays legitimate dependency updates by up to a week. I'm happy to drop them from this PR if you'd rather set that policy yourselves; the rest of the changes are independent of them.

### How did you test it?

- The 47 workflow edits are mechanical `env:`-relocations; every changed YAML file was verified to parse, and `Core / Github workflows linter` passes.
[...]

PR #3721, opened 2026-08-03
ci: scope integration-test secrets to the steps that use them

julian-risch:
### Related Issues

- Split out of #3539, which bundled these CI changes with unrelated `arangodb` and `nvidia` source changes and had maintainer edits disabled (maintainerCanModify: false)

### Proposed Changes:

**`gha-workflow-env-secret`: secrets exposed to every step of a workflow (51 workflows)**

Workflow-level `env:` blocks make each secret visible to every step in the workflow, including third-party actions that have no need for them (checkout, setup-python, coverage comment, artifact upload). Each secret is now scoped to the step(s) that actually run integration tests.

`e2b.yml`: its secret is scoped to the *job* rather than the step, because the "Run integration tests" step gates on the secret in its own `if:` condition, and a step's own step-level `env:` is not in scope for that same step's `if:`.

`mirage.yml` keeps its job-level `HAS_GOOGLE_DRIVE_SECRET` as is because it holds a boolean, not a secret value, and the `Authenticate to Google Cloud` step's `if:` depends on it.

**`run-shell-injection` — `CI_pypi_release.yml`**

`${{ github.ref_name }}` was interpolated directly into a `run:` script (once into a shell command, once into an inline Python script) which is bad practice because it allows things like `integrations/anthropic-v1.0.0"+__import__("os").system("id")+"` as a tag to be passed on. Tags can only be pushed by someone who already has write access so it's not an actual security issue for us. Still, both now pass it through an intermediate `env:` var and read it from the environment, so tag contents can't terminate the surrounding script.

### Notes for the reviewer
For this PR, nearly every integration's test suite runs with live API keys on each push here. Worth being deliberate about how many times we push to it.

Authored by @FahimaGold and split out of #3539 so the CI hardening can merge independently of the source changes; commit authorship is preserved. 
Three workflows that #3539 missed are included here: `agent_pack.yml` (`OPENAI_API_KEY`, `TAVILY_API_KEY`), `linkup.yml` (`LINKUP_API_KEY`), `mirage.yml` (`OPENAI_API_KEY`).

I won't rerun failing tests.

### Checklist

- [ ] I have updated the related issue with new insights and changes — n/a, no linked issue
- [ ] I added unit tests and updated the docstrings — n/a, CI-only change with no source or docstring changes

### DIFF
--- a/.github/workflows/brave.yml
+++ b/.github/workflows/brave.yml
@@ -29,7 +29,6 @@ concurrency:
 env:
   PYTHONUNBUFFERED: "1"
   FORCE_COLOR: "1"
-  BRAVE_API_KEY: ${{ secrets.BRAVE_API_KEY }}
   TEST_MATRIX_OS: '["ubuntu-latest", "windows-latest", "macos-latest"]'
   TEST_MATRIX_PYTHON: '["3.10", "3.14"]'
 
@@ -104,6 +103,8 @@ jobs:
           path: python-coverage-comment-action-brave.txt
 
       - name: Run integration tests
+        env:
+          BRAVE_API_KEY: ${{ secrets.BRAVE_API_KEY }}
         run: hatch run test:integration-cov-append-retry
 
       - name: Store combined coverage
@@ -125,6 +126,8 @@ jobs:
 
       - name: Nightly - run tests with Haystack main branch
         if: github.event_name == 'schedule'
+        env:
+          BRAVE_API_KEY: ${{ secrets.BRAVE_API_KEY }}
         run: |
           hatch env prune
           hatch -e test env run -- uv pip install git+https://github.com/deepset-ai/haystack.git@main
--- a/.github/workflows/e2b.yml
+++ b/.github/workflows/e2b.yml
@@ -29,7 +29,6 @@ concurrency:
 env:
   PYTHONUNBUFFERED: "1"
   FORCE_COLOR: "1"
-  E2B_API_KEY: ${{ secrets.E2B_API_KEY }}
   TEST_MATRIX_OS: '["ubuntu-latest", "windows-latest", "macos-latest"]'
   TEST_MATRIX_PYTHON: '["3.10", "3.14"]'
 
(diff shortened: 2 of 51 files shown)


.github/workflows/brave.yml
```diff
@@ -125,6 +126,8 @@ jobs:
 
       - name: Nightly - run tests with Haystack main branch
         if: github.event_name == 'schedule'
+        env:
+          BRAVE_API_KEY: ${{ secrets.BRAVE_API_KEY }}
```
anakin87, line 130: I'm now realizing that here we run integration tests nightly with Haystack main branch.
We usually run integration tests with Haystack main only for OpenAI-inherited Chat Generators and this is not the case.
Could you please fix this?

.github/workflows/e2b.yml
```diff
@@ -61,6 +60,12 @@ jobs:
         os: ${{ fromJSON(needs.compute-test-matrix.outputs.os) }}
         python-version: ${{ fromJSON(needs.compute-test-matrix.outputs.python-version) }}
 
+    # Scoped to this job (not the workflow-level env) so the secret is not exposed to other
+    # jobs, while remaining available to the "Run integration tests" step's `if:` condition
+    # (a step's own step-level env is not in scope for that same step's `if`).
+    env:
+      E2B_API_KEY: ${{ secrets.E2B_API_KEY }}
```
anakin87, line 67: why this diverges? Sorry but the comment above is hard to understand for me

anakin87 (changes requested):
Looks good overall.

Just two minor comments

.github/workflows/e2b.yml
julian-risch: Sorry about that. :D I'll see how to better phrase this. In short, the difference here is that we use the secret in an if statement later on: `if: env.E2B_API_KEY != ''`. Let me check if I can clean this up somehow.

anakin87 (approved):
Let's address the small comments in another PR

.github/workflows/e2b.yml
anakin87: I see. In any case, I'd try to modify this workflow to match the standard:
- in the workflow, always run integration tests
- in test files, skip them if the env var is not present

PR #3722, opened 2026-08-03
feat/microsoft-sharepoint: add region parameter to MSSharePointRetriever

### LINKED ISSUE #3673: MSSharePointFetcher - Allow specifying region as request parameter (needed when using application permissions).
Currently, the MSSharePoint integration only supports Microsoft on-behalf-of OAuth workflows (using the current [OAuth integration components](https://haystack.deepset.ai/integrations/oauth) available in Haystack). In the case you want to use app-only authentication and authorization, you need to acquire the token accessing the api directly, e.g.
```python
import requests
token_url = f"https://login.microsoftonline.com/{TENANT_ID}/oauth2/v2.0/token"
data = {
    "client_id": CLIENT_ID,
    "scope": "https://graph.microsoft.com/.default",
    "client_secret": CLIENT_SECRET,
    "grant_type": "client_credentials"
}
response = requests.post(token_url, data=data)
response.raise_for_status()
token = response.json().get("access_token")
```

but when running the pipeline using the `MSSharePointRetriever` setting the token, I receive the following error:
```
haystack Component type: 'MSSharePointRetriever'
haystack Error: Microsoft Graph search request failed with status 400: {"error":{"code":"BadRequest","message":"SearchRequest Invalid (Region is required when request with application permission.)","target":"","details":[{"code":"Microsoft.SubstrateSearch.Api.ErrorReporting.ResourceBasedExceptions.BadRequestException","message":"Region is required when request with application permission.","target":"","httpCode":400}],"httpCode":400}
```

Testing the API directly, I see the same error, and it only works when we add the field "region": "US" in the request body.

Allowing to define the region as init parameter for the MSSharePointFetcher object and adding it to the [request body](https://github.com/deepset-ai/haystack-core-integrations/blob/main/integrations/microsoft_sharepoint/src/haystack_integrations/components/retrievers/microsoft_sharepoint/retriever.py#L210-L217).

**Additional context**
When not including the region in the request, the following error is thrown from the API:
```json
  "error": {
    "code": "BadRequest",
    "message": "SearchRequest Invalid (Region is required when request with application permission.)",
    "target": "",
    "details": [
      {
        "code": "Microsoft.SubstrateSearch.Api.ErrorReporting.ResourceBasedExceptions.BadRequestException",
        "message": "Region is required when request with application permission.",
        "target": "",
        "httpCode": 400
      }
    ],
    "httpCode": 400
  },
```

This component is using the searchEntity API: 
- https://learn.microsoft.com/en-us/graph/api/search-query?view=graph-rest-1.0&tabs=http
- https://learn.microsoft.com/en-us/graph/api/resources/searchrequest?view=graph-rest-1.0

SyedShahmeerAli12:
## Summary

Closes #3673

- Added `region: str | None = None` init parameter to `MSSharePointRetriever`
- When set, the `region` field is included in the Microsoft Search API request body
- Required when using app-only (client credentials) authentication  without it the API returns `400: Region is required when request with application permission`
- Omitted from the request body when `None` (default), preserving existing delegated-auth behavior
- Serialized correctly via `to_dict` / `from_dict`

## Test plan

- [x] `test_defaults`  asserts `region is None` by default
- [x] `test_to_dict`  asserts `region: None` in serialized output
- [x] `test_to_dict_with_region`  asserts `region: "US"` serializes correctly
- [x] `test_region_included_in_request_body_when_set`  verifies `region` appears in the HTTP POST body
- [x] `test_region_omitted_from_request_body_when_none` —verifies `region` key is absent when not set
- [x] `TestLiveAppOnly`  integration test class for real app-only auth (skipped unless `MS_SHAREPOINT_TENANT_ID`, `MS_SHAREPOINT_CLIENT_ID`, `MS_SHAREPOINT_CLIENT_SECRET` env vars are set)
- [x] All 81 unit tests pass
- [x] Lint clean (`ruff check` + `ruff format`)

### DIFF
--- a/integrations/microsoft_sharepoint/src/haystack_integrations/components/retrievers/microsoft_sharepoint/retriever.py
+++ b/integrations/microsoft_sharepoint/src/haystack_integrations/components/retrievers/microsoft_sharepoint/retriever.py
@@ -97,7 +97,7 @@ class MSSharePointRetriever:
             `'{searchTerms} path:"https://contoso.sharepoint.com/sites/Team"'`. The literal `{searchTerms}`
             placeholder is replaced by the run-time query. The template uses
             [Keyword Query Language (KQL)](https://learn.microsoft.com/en-us/sharepoint/dev/general-development/keyword-query-language-kql-syntax-reference).
-        :param region: The region code for the Microsoft Search index, for example `"US"`, `"EU"`, or `"APAC"`.
+        :param region: The region code for the Microsoft Search index, for example `"NAM"`, `"EUR"`, or `"APC"`.
             Required when using application permissions (app-only / client-credentials auth); omit for delegated
             (on-behalf-of) tokens. See the `region` property of the
             [searchRequest resource](https://learn.microsoft.com/en-us/graph/api/resources/searchrequest).
@@ -140,9 +140,9 @@ class MSSharePointRetriever:
             operators directly in the query, for example `filetype:docx`, `author:"Jane Doe"`, or
             `path:"https://contoso.sharepoint.com/sites/Team"`. See the
             [KQL syntax reference](https://learn.microsoft.com/en-us/sharepoint/dev/general-development/keyword-query-language-kql-syntax-reference).
-        :param access_token: A delegated Microsoft Graph bearer token for the user whose content is searched,
-            typically wired from an upstream `OAuthResolver` (which emits a plain `str`). A `Secret` is also
-            accepted and resolved internally.
+        :param access_token: A Microsoft Graph bearer token. For delegated auth, pass a per-user token wired from
+            an upstream `OAuthResolver`. For app-only (client-credentials) auth, pass an application token and set
+            `region` at init time. A `Secret` is also accepted and resolved internally.
         :param top_k: Overrides the `top_k` configured at initialization for this run.
         :returns: A dictionary with a `documents` key holding the list of retrieved `Document` objects.
         :raises SharePointConfigError: If `access_token` is a `Secret` that does not resolve to a string.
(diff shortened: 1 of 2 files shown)


integrations/microsoft_sharepoint/tests/test_retriever.py
```diff
@@ -410,3 +431,65 @@ def test_run_against_microsoft_graph(self):
...
+
+
+@pytest.mark.integration
+@pytest.mark.skipif(
+    not all(
+        os.environ.get(v) for v in ("MS_SHAREPOINT_TENANT_ID", "MS_SHAREPOINT_CLIENT_ID", "MS_SHAREPOINT_CLIENT_SECRET")
+    ),
+    reason="MS_SHAREPOINT_TENANT_ID / MS_SHAREPOINT_CLIENT_ID / MS_SHAREPOINT_CLIENT_SECRET not set",
+)
+class TestLiveAppOnly:
```
bogdankostic, line 441: Let's change these environment variables to `MS_GRAPH_TENANT_ID`, `MS_GRAPH_CLIENT_ID`, `MS_GRAPH_CLIENT_SECRET`, the names we already use in our CI.

integrations/microsoft_sharepoint/tests/test_retriever.py
```diff
@@ -410,3 +431,65 @@ def test_run_against_microsoft_graph(self):
...
+
+    def test_app_only_search_requires_region(self):
+        token = self._get_app_token()
+
+        # Without region, the Search API must return 400.
+        retriever_no_region = MSSharePointRetriever(top_k=3, max_retries=0)
+        with pytest.raises(SharePointRequestError) as exc_info:
+            retriever_no_region.run(query="test", access_token=token)
+        assert exc_info.value.status_code == 400
+        assert "Region is required" in str(exc_info.value)
```
bogdankostic: I don't think we need this test.

integrations/microsoft_sharepoint/src/haystack_integrations/components/retrievers/microsoft_sharepoint/retriever.py
```diff
@@ -96,6 +97,10 @@ def __init__(
             `'{searchTerms} path:"https://contoso.sharepoint.com/sites/Team"'`. The literal `{searchTerms}`
             placeholder is replaced by the run-time query. The template uses
             [Keyword Query Language (KQL)](https://learn.microsoft.com/en-us/sharepoint/dev/general-development/keyword-query-language-kql-syntax-reference).
+        :param region: The region code for the Microsoft Search index, for example `"US"`, `"EU"`, or `"APAC"`.
```
bogdankostic: Based on the [MS docs](https://learn.microsoft.com/en-us/sharepoint/dev/solution-guidance/multigeo-discovery), I think we need to use `"NAM"`, `"EUR"`, and `"APC"`
```suggestion
        :param region: The region code for the Microsoft Search index, for example `"NAM"`, `"EUR"`, or `"APC"`.
```

integrations/microsoft_sharepoint/src/haystack_integrations/components/retrievers/microsoft_sharepoint/retriever.py
bogdankostic, line 35: Let's add to the doc strings that an app-only token works when region is set. (Not only delegated Microsoft Graph bearer tokens as before.)

bogdankostic: Let's add to the doc strings that an app-only token works when region is set. (Not only delegated Microsoft Graph bearer tokens as before.)

bogdankostic (changes requested):
Thanks @SyedShahmeerAli12! Looks good in general, just a few minor comments.

bogdankostic (approved):
Looks good, thanks for addressing the feedback so quickly @SyedShahmeerAli12!

PR #3731, opened 2026-08-06
docs: fix TypeError in FastembedSparseDocumentEmbedder usage example

### LINKED ISSUE #3729: Usage example in FastembedSparseDocumentEmbedder docstring raises TypeError
**Describe the bug**
The usage example in the `FastembedSparseDocumentEmbedder` docstring ends with

    print(f"Sparse Embedding Dimension: {len(result['documents'][0].sparse_embedding)}")

`SparseEmbedding` does not define `__len__`, so running the example as written fails:

    TypeError: object of type 'SparseEmbedding' has no len()

Happy to open a PR with the one-line fix.

**To Reproduce**
Run the usage example from the `FastembedSparseDocumentEmbedder` docstring as is
  (`integrations/fastembed/src/haystack_integrations/components/embedders/fastembed/fastembed_sparse_document_embedder.py`).

 - OS: Ubuntu 24.04 LTS
 - Haystack version: 3.0.0

keosung:
### Related Issues

- fixes #3729 

### Proposed Changes:

The usage example in the `FastembedSparseDocumentEmbedder` docstring calls `len()` on a `SparseEmbedding`, which has no `__len__`, so the example fails with a TypeError. This changes the last line to print `len(sparse_embedding.indices)`, the number of non-zero elements.

### How did you test it?

Ran the example with `prithivida/Splade_PP_en_v1` before and after the change. The original fails at the `len()` line, the fixed one completes.

### Notes for the reviewer

One-line docstring change.

### Checklist

- [x] I updated the docstrings (docs-only change, no unit tests needed)


integrations/fastembed/src/haystack_integrations/components/embedders/fastembed/fastembed_sparse_document_embedder.py
```diff
@@ -53,7 +53,7 @@ class FastembedSparseDocumentEmbedder:
     result = sparse_doc_embedder.run(document_list)
     print(f"Document Text: {result['documents'][0].content}")
     print(f"Document Sparse Embedding: {result['documents'][0].sparse_embedding}")
-    print(f"Sparse Embedding Dimension: {len(result['documents'][0].sparse_embedding)}")
+    print(f"Sparse Embedding Non-Zero Elements: {len(result['documents'][0].sparse_embedding.indices)}")
```
sjrl: Correct me if I'm wrong this still prints out a length so the "Non-Zero Elements" description doesn't seem correct.

keosung: Thanks for the review! Since `SparseEmbedding` stores only non-zero values, `len(indices)` gives the number of non-zero elements. Haystack core uses the same convention in `Document.__repr__` ("vector with N non-zero elements"), and "Dimension" would technically be the vocab size, so I changed the label.

That said, I agree it could be clearer since the printed value is a count. Would `Number of non-zero elements: ...` work, or did you have different wording in mind?

sjrl: Your suggested message sounds good!

sjrl (approved):

PR #3742, opened 2026-08-06
fix: Chroma adding optional parameter `filters` to `get_metadata_field_unique_values()`

davidsbatista:
### Proposed Changes:

- adding optional parameter `filters` to `get_metadata_field_unique_values() `

### How did you test it?

- added one more test to check filters usage

### DIFF
--- a/integrations/chroma/src/haystack_integrations/document_stores/chroma/document_store.py
+++ b/integrations/chroma/src/haystack_integrations/document_stores/chroma/document_store.py
@@ -1302,12 +1302,31 @@ class ChromaDocumentStore:
         result = await self._async_collection.get(include=["metadatas"])
         return self._compute_field_min_max(result.get("metadatas", []), field_name)
 
+    @staticmethod
+    def _prepare_get_metadatas_kwargs(filters: dict[str, Any] | None = None) -> dict[str, Any]:
+        """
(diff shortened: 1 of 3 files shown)


integrations/chroma/src/haystack_integrations/document_stores/chroma/document_store.py
```diff
@@ -1302,12 +1302,31 @@ async def get_metadata_field_min_max_async(self, metadata_field: str) -> dict[st
         result = await self._async_collection.get(include=["metadatas"])
         return self._compute_field_min_max(result.get("metadatas", []), field_name)
 
+    @staticmethod
+    def _prepare_get_metadatas_kwargs(filters: dict[str, Any] | None = None) -> dict[str, Any]:
```
anakin87: Instead of adding a new method, could we parametrize the similar `_prepare_get_kwargs`?

anakin87 (changes requested):

integrations/chroma/src/haystack_integrations/document_stores/chroma/document_store.py
davidsbatista: yes, good catch 👍🏽

anakin87 (approved):
👍

PR #3744, opened 2026-08-06
feat: add Python 3.14 support for WatsonX integration

### LINKED ISSUE #3036: WatsonX integration: bring Python 3.14 support when implemented in the original library
Currently, WatsonX does not support Python 3.14, due to a relatively simple bug: https://github.com/IBM/watsonx-ai-python-sdk/issues/11

I opened the above issue, but I'm not sure if someone will pick it up because they only host docs on GitHub and don't know if they monitor the repo.

In any case, when WatsonX starts supporting Python 3.14, we should
- update our pin in pyproject
- test with 3.14 (instead of 3.13)
- add 3.14 information to `classifiers` in pyproject.toml

SyedShahmeerAli12:
`ibm-watsonx-ai>=1.6.0` now supports Python 3.14. This updates the dependency pin, adds the 3.14 classifier, and switches CI to test against 3.14.

Closes #3036

### DIFF
--- a/integrations/watsonx/pyproject.toml
+++ b/integrations/watsonx/pyproject.toml
@@ -22,7 +22,7 @@ classifiers = [
     "Programming Language :: Python :: 3.13",
     "Programming Language :: Python :: 3.14",
 ]
-dependencies = ["haystack-ai>=2.30.0", "ibm-watsonx-ai>=1.6.0", "pandas>=2.2.3"]
+dependencies = ["haystack-ai>=2.30.0", "ibm-watsonx-ai>=1.5.13", "pandas>=2.2.3"]
 
 [project.urls]
 Documentation = "https://github.com/deepset-ai/haystack-core-integrations/tree/main/integrations/watsonx#readme"
--- a/integrations/watsonx/pyproject.toml
+++ b/integrations/watsonx/pyproject.toml
@@ -21,8 +21,9 @@ classifiers = [
     "Programming Language :: Python :: 3.11",
     "Programming Language :: Python :: 3.12",
     "Programming Language :: Python :: 3.13",
+    "Programming Language :: Python :: 3.14",
 ]
-dependencies = ["haystack-ai>=2.30.0", "ibm-watsonx-ai>=1.3.26", "pandas>=2.2.3"]
+dependencies = ["haystack-ai>=2.30.0", "ibm-watsonx-ai>=1.6.0", "pandas>=2.2.3"]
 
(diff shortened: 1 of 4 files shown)


integrations/watsonx/pyproject.toml
```diff
@@ -17,12 +17,12 @@ classifiers = [
...
     "Development Status :: 4 - Beta",
     "Programming Language :: Python",
-    "Programming Language :: Python :: 3.10",
     "Programming Language :: Python :: 3.11",
     "Programming Language :: Python :: 3.12",
     "Programming Language :: Python :: 3.13",
+    "Programming Language :: Python :: 3.14",
 ]
-dependencies = ["haystack-ai>=2.30.0", "ibm-watsonx-ai>=1.3.26", "pandas>=2.2.3"]
+dependencies = ["haystack-ai>=2.30.0", "ibm-watsonx-ai>=1.6.0", "pandas>=2.2.3"]
```
sjrl: It seems that the first version that added python 3.14 support is v1.5.13 https://ibm.github.io/watsonx-ai-python-sdk/v1.6.0a/changelog.html#id3 

could we change it to that version?

integrations/watsonx/pyproject.toml
```diff
@@ -17,12 +17,12 @@ classifiers = [
     "License :: OSI Approved :: Apache Software License",
     "Development Status :: 4 - Beta",
     "Programming Language :: Python",
-    "Programming Language :: Python :: 3.10",
```
sjrl, line 20: Also looking at their changelog https://ibm.github.io/watsonx-ai-python-sdk/v1.6.0a/changelog.html I don't see them mentioning that they dropped python 3.10 support. Could you point me to where they say they no longer support python 3.10?

integrations/watsonx/pyproject.toml
SyedShahmeerAli12: Updated to `>=1.5.13`  thanks!

integrations/watsonx/pyproject.toml
SyedShahmeerAli12: The changelog doesn't mention it explicitly, but `ibm-watsonx-ai==1.5.13` (the first version with 3.14 support) has `Requires-Python: <3.15,>=3.11` in its wheel metadata  so the library itself stopped installing on 3.10 from that version onwards. That's what forced the drop here.

sjrl (approved):
Thanks!

PR #3750, opened 2026-08-07
feat(cohere): add CohereAzureRanker for rerank models served via Azure AI

### LINKED ISSUE #2795: Add Azure Ranker Component
We have OpenAIChatGenerator and AzureOpenAIDocumentEmbedder components in Haystack. In addition we have Retriever components in the Azure AI Search Haystack integration. However, we don't have a component that covers ranking with models served via Azure AI.

We should add a new component that enables users to rerank documents with cross-encoder models served via Microsoft Azure. I believe this new component could become part of the existing Azure AI Search integration or become a separate integration depending on dependencies it might require. If it does not require azure-search-documents then it could become a separate integration where we have a Ranker Component and in future also an Embedder component.

**Additional context**
My understanding is that the [Azure AI Search service](https://learn.microsoft.com/en-us/azure/search/search-create-service-portal) feature of a [semantic ranker](https://learn.microsoft.com/en-us/azure/search/semantic-how-to-enable-disable) as described in https://learn.microsoft.com/en-us/azure/search/search-get-started-semantic?pivots=python is completely separate from this GitHub issue. We are looking for a ranker component that works independent of how the documents are stored and what index they use.

Aryan-Pardeshi:
Closes #2795.

Adds a ranker for Cohere rerank models served through Azure AI, in the existing `cohere` integration rather than a new one — per the conclusion in the issue thread that these models are reached without any Azure SDK dependency.

Built on the reference code @sjrl posted on 2026-04-22, using `httpx` rather than the Cohere SDK because of cohere-ai/cohere-python#661, which @bglearning flagged on the issue as blocking the SDK route.

Adapted rather than copied — the reference had an import-time `os.getenv` default, passed a raw `Secret` through `to_dict`, defined custom exception classes, and carried a `__del__` with a bare except. This follows `CohereRanker` instead: `Secret` handling and `to_dict`/`from_dict` round-trip, `ValueError` for validation, and `dataclasses.replace` so documents are not mutated in place.

Unit tests mock the transport; no credentials or network needed.

### DIFF
--- a/integrations/cohere/pydoc/config_docusaurus.yml
+++ b/integrations/cohere/pydoc/config_docusaurus.yml
@@ -6,6 +6,7 @@ loaders:
       - haystack_integrations.components.embedders.cohere.utils
       - haystack_integrations.components.generators.cohere.generator
       - haystack_integrations.components.generators.cohere.chat.chat_generator
+      - haystack_integrations.components.rankers.cohere.azure_ranker
       - haystack_integrations.components.rankers.cohere.ranker
     search_path: [../src]
 processors:
--- a/integrations/cohere/src/haystack_integrations/components/rankers/cohere/__init__.py
+++ b/integrations/cohere/src/haystack_integrations/components/rankers/cohere/__init__.py
@@ -1,3 +1,4 @@
+from .azure_ranker import CohereAzureRanker
 from .ranker import CohereRanker
 
-__all__ = ["CohereRanker"]
+__all__ = ["CohereAzureRanker", "CohereRanker"]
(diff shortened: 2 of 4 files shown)


bogdankostic:
Thanks a lot for this @Aryan-Pardesh! 

I tried the existing `CohereRanker` against a live Cohere rerank deployment on Azure AI Foundry and it works as-is. I added a working example in #2795. Therefore, I'm closing this PR given that a separate `CohereAzureRanker` is not needed anyore.

Either way, thanks for the effort you put in here.

PR #3760, opened 2026-08-11
feat: add You.com web search integration (YouComWebSearch)

### LINKED ISSUE #3649: New integration: You.com web search (YouComWebSearch)
## Summary and motivation

Add a You.com web-search integration (`youcom-haystack`, `YouComWebSearch`) alongside the existing websearch packages (serperdev, searchapi, brave, tavily, linkup, ddgs, firecrawl).

What it adds over the existing providers: **zero-configuration web search**. You.com exposes a keyless free tier (IP-rate-limited, 100 searches/day) in addition to its keyed Search API. The component defaults to `Secret.from_env_var("YDC_API_KEY", strict=False)` — when the variable is unset it transparently uses the keyless endpoint, so a getting-started RAG pipeline runs with no signup step at all; setting `YDC_API_KEY` switches it to the keyed API with higher limits. Web search appears in most getting-started pipelines, and every current websearch integration requires provisioning an API key first — this removes that friction entirely (useful for docs, tutorials, and workshops).

The component is implemented, tested, and live at **https://github.com/brainsparker/youcom-haystack** (structure mirrors `integrations/brave/` so it can be adopted into this monorepo as `integrations/youcom/` with minimal changes). I work at You.com — we're happy to provide an API key for this repo's CI secrets (`YDC_API_KEY`) for the integration test workflow, and to maintain the integration.

## Detailed design

Mirrors `BraveWebSearch` anatomy:

- Package `youcom-haystack`, import `from haystack_integrations.components.websearch.youcom import YouComWebSearch`
- `@component` class with `run(query, top_k=None)` and `run_async(...)`, both returning `{"documents": list[Document], "links": list[str]}`
- Init params: `api_key: Secret = Secret.from_env_var("YDC_API_KEY", strict=False)`, `top_k=10`, `freshness`, `country`, `search_lang`, `safesearch`, `extra_params`, `timeout=10`, `max_retries=3`
- Endpoint selection at request time: key resolvable → `https://ydc-index.io/v1/search` with `X-API-Key`; otherwise → keyless `https://api.you.com/v1/agents/search`. Both return the same response shape ([OpenAPI spec](https://you.com/specs/openapi_search_v1.yaml))
- Documents carry `title`, `url`, `source` (`web`/`news`), `page_age` in meta; content is the joined text snippets (fallback: description). Web results precede news results
- HTTP via `haystack.utils.requests_utils.request_with_retry` / `async_request_with_retry`; keyless 402/429 responses raise an enriched error with upgrade guidance
- Serialization via Haystack's default component serialization (Secret-aware); round-trip covered in tests
- 17 unit tests (mocked HTTP: endpoint selection, env fallback, param mapping, parsing, error enrichment, serialization) + 3 live integration tests marked `integration`
- Component sends a `youcom-haystack/<version>` User-Agent so You.com can attribute framework traffic (same idea as Tavily's `client_name="haystack"`)

Usage:

```python
from haystack_integrations.components.websearch.youcom import YouComWebSearch

websearch = YouComWebSearch(top_k=5)  # works immediately, no API key
result = websearch.run(query="What is Haystack by deepset?")
result["documents"]  # list[Document]
result["links"]      # list[str]
```

## Checklist

### Tasks
- [ ] A new label named like `integration:youcom` has been added to the list of labels for this [repository](https://github.com/deepset-ai/haystack-core-integrations/labels)
- [ ] An integration tile with a usage example has been added to https://github.com/deepset-ai/haystack-integrations

brainsparker:
### Related Issues

- fixes #3649

### Proposed Changes:

Adds `youcom-haystack` with a `YouComWebSearch` component, alongside the existing websearch integrations. The anatomy mirrors `integrations/brave/`.

What it adds over the existing websearch providers is **zero-configuration web search**. `api_key` defaults to `Secret.from_env_var("YDC_API_KEY", strict=False)`, and when the variable is unset the component transparently uses You.com's keyless free tier (IP-rate-limited) instead of raising — so a getting-started RAG pipeline runs with no signup step. Setting `YDC_API_KEY` switches it to the keyed [You.com Search API](https://you.com/docs/api-reference/search/v1-search) with higher limits. Both endpoints return the same response shape.

```python
from haystack_integrations.components.websearch.youcom import YouComWebSearch

websearch = YouComWebSearch(top_k=5)  # works immediately, no API key
result = websearch.run(query="What is Haystack by deepset?")
result["documents"]  # list[Document]
result["links"]      # list[str]
```

Details:

- `run(query, top_k=None)` and `run_async(...)`, both returning `{"documents": list[Document], "links": list[str]}`
- Init params: `api_key`, `top_k=10`, `freshness`, `country`, `search_lang`, `safesearch`, `extra_params`, `timeout=10`, `max_retries=3`
- HTTP via `haystack.utils.requests_utils.request_with_retry` / `async_request_with_retry`
- Documents carry `title`, `url`, `source` (`web`/`news`) and `page_age` in meta; content is the joined text snippets, falling back to the description. Web results precede news results
- Keyless 402/429 responses are re-raised with upgrade guidance pointing at `YDC_API_KEY`
- Serialization uses Haystack's default Secret-aware component serialization
- Sends a `youcom-haystack/<version>` User-Agent so You.com can attribute framework traffic (same idea as Tavily's `client_name="haystack"`)

Generated with `python scripts/create_new_integration.py --name youcom --type websearch`, so the CI workflow, `labeler.yml` rules, `CI_coverage_comment.yml` registration and the README inventory row are all included.

### How did you test it?

- `hatch run fmt-check` and `hatch run test:types` — clean
- `hatch run test:unit-cov-retry` — 19 mocked unit tests pass, 97% coverage (endpoint selection, env fallback, param mapping, parsing, error enrichment, serialization)
- `hatch run test:integration` — 3 live tests pass (keyless, keyed sync, keyed async)
- Also ran the CI's lowest-direct-dependency step (`haystack-ai==2.24.1`, `httpx==0.27.0`) — unit tests pass
- `hatch run docs` + `npx docusaurus-mdx-checker` — API reference builds and compiles cleanly

### Notes for the reviewer

- **Repository secret needed:** the integration test job reads `YDC_API_KEY` for the two keyed tests. I work at You.com and we're happy to provide a key for this repo's secrets — just say where to send it. Until then those two tests skip via `skipif`, and CI stays green: the keyless live test needs no key.
- The keyless live test skips (rather than fails) on a 402/429, since the free tier is rate limited per IP and a shared runner can legitimately hit the limit. That keeps the nightly run from paging you over a rate limit.
- One deliberate difference from `BraveWebSearch`: no second `response.raise_for_status()` after `request_with_retry`, since that helper already raises. The component instead catches `httpx.HTTPStatusError` to attach the keyless upgrade hint.
- Happy to follow up with the integration tile PR against [deepset-ai/haystack-integrations](https://github.com/deepset-ai/haystack-integrations) once this lands, and we'll maintain the integration going forward.

### Checklist

- I've used one of the [conventional commit types](https://www.conventionalcommits.org/en/v1.0.0/) for my PR title: `feat:`


brainsparker:
Pushed one follow-up after watching the first CI run: the keyed live tests reported `PASSED` on this fork PR, which was misleading. GitHub sets a secret-backed env var to an *empty string* when the secret isn't available (as on fork PRs), so `"YDC_API_KEY" not in os.environ` was false, the tests ran, and the component quietly fell back to the keyless endpoint. They now skip on a falsy value, so the run log tells the truth about what was exercised.

Net effect on CI today: 19 unit tests pass everywhere, 1 keyless live search passes, 2 keyed live tests skip until the `YDC_API_KEY` secret is added.

brainsparker:
One more correction, pushed just now: the keyed endpoint is `https://api.you.com/v1/search`, not `ydc-index.io/v1/search`. The `ydc-index.io` host still answers and is what our published OpenAPI spec advertises, but it's the legacy host — `api.you.com` is the one we're keeping, so better to ship that than have you inherit a redirect later. Verified live before pushing: HTTP 200 with the same response shape, and both keyed live tests pass against it. Keyless and keyed now share a single host.

No other behavior changed; unit tests, mypy and ruff are all still clean.

brainsparker:
Small attribution follow-up, pushed: the component's `User-Agent` now carries a second product token naming this repo —

```
youcom-haystack/<version> youdotcom-integration/deepset-ai-haystack-core-integrations
```

Space-separated product tokens per RFC 9110, and still only sent to You.com hosts. On our side this is what lets us attribute framework traffic to Haystack specifically; it's the same intent as Tavily's `client_name="haystack"`, just carried in the UA. Asserted in `test_run_sends_user_agent` so it survives refactors.

Happy to drop or shorten it if you'd rather integrations not name the repo in the UA — no functional dependency on it.

.github/workflows/youcom.yml
```diff
@@ -0,0 +1,143 @@
...
+        if: matrix.python-version == '3.10' && runner.os == 'Linux' && github.event_name == 'pull_request' && steps.coverage_comment.outputs.COMMENT_FILE_WRITTEN == 'true'
+        uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v7.0.1
+        with:
+          name: coverage-comment-youcom
+          path: python-coverage-comment-action-youcom.txt
+
+      - name: Run integration tests
+        if: matrix.python-version == '3.10' && runner.os == 'Linux'
+        env:
+          YDC_API_KEY: ${{ secrets.YDC_API_KEY }}
```
davidsbatista: For readability purposes for the maintainers of all keys, I suggest:

```suggestion
          YOUDOTCOM_API_KEY: ${{ secrets.YOUDOTCOM_API_KEY }}
```

integrations/youcom/src/haystack_integrations/components/websearch/youcom/youcom_websearch.py
```diff
@@ -0,0 +1,243 @@
...
+    websearch = YouComWebSearch(top_k=5)  # no API key needed to get started
+    result = websearch.run(query="What is Haystack by deepset?")
+    documents = result["documents"]
+    links = result["links"]
+    ```
+    """
+
+    def __init__(
+        self,
+        api_key: Secret = Secret.from_env_var("YDC_API_KEY", strict=False),
```
davidsbatista: When the env var isn't set, `resolve_value()` returns None, and the component falls through to the keyless branch. There's no log line or other signal when this happens.

Could this be made explicit (e.g., a keyless_fallback: bool = True (or similar) constructor param, so the choice shows up in the signature and in to_dict()/serialized pipelines.

Can you also add a link to keyless search docs?

integrations/youcom/src/haystack_integrations/components/websearch/youcom/youcom_websearch.py
```diff
@@ -0,0 +1,243 @@
...
+        safesearch: str | None = None,
+        extra_params: dict[str, Any] | None = None,
+        timeout: int = 10,
+        max_retries: int = 3,
+    ) -> None:
+        """
+        Initialize the YouComWebSearch component.
+
+        :param api_key:
+            You.com API key. Defaults to the `YDC_API_KEY` environment variable, resolved
```
davidsbatista:
```suggestion
            You.com API key. Defaults to the `YOUDOTCOM_API_KEY` environment variable, resolved
```

integrations/youcom/README.md
```diff
@@ -0,0 +1,30 @@
...
+[![PyPI - Python Version](https://img.shields.io/pypi/pyversions/youcom-haystack.svg)](https://pypi.org/project/youcom-haystack)
+
+- [Changelog](https://github.com/deepset-ai/haystack-core-integrations/blob/main/integrations/youcom/CHANGELOG.md)
+
+---
+
+## Usage
+
+`YouComWebSearch` runs without any configuration: with no API key available it uses the You.com keyless
+free tier (rate limited per IP). Set the `YDC_API_KEY` environment variable to use the keyed
```
davidsbatista:
```suggestion
free tier (rate limited per IP). Set the `YOUDOTCOM_API_KEY` environment variable to use the keyed
```

integrations/youcom/README.md
```diff
@@ -0,0 +1,30 @@
...
+result["documents"]  # list[Document]
+result["links"]  # list[str]
+```
+
+## Contributing
+
+Refer to the general [Contribution Guidelines](https://github.com/deepset-ai/haystack-core-integrations/blob/main/CONTRIBUTING.md).
+
+The unit tests are fully mocked and need no credentials. The integration tests include one live keyless
+search that needs no key; to also run the keyed tests, export the `YDC_API_KEY` environment variable.
```
davidsbatista:
```suggestion
search that needs no key; to also run the keyed tests, export the `YOUDOTCOM_API_KEY` environment variable.
```

integrations/youcom/src/haystack_integrations/components/websearch/youcom/youcom_websearch.py
```diff
@@ -0,0 +1,243 @@
...
+        if self.search_lang is not None:
+            params["language"] = self.search_lang
+        if self.safesearch is not None:
+            params["safesearch"] = self.safesearch
+        if self.extra_params:
+            params.update(self.extra_params)
+        return params
+
+    @staticmethod
+    def _enrich_error(error: httpx.HTTPStatusError, headers: dict[str, str]) -> httpx.HTTPStatusError:
```
davidsbatista: This re-raises a `httpx.HTTPStatusError` with the upgrade hint appended to the message string.

Works and is tested, but the hint is only recoverable by string-matching, and callers can't distinguish "YouCom failed" from any other httpx.HTTPStatusError. 

`SerperDevWebSearch` uses its own SerperDevError(ComponentError), a dedicated exception would be a cleaner pattern here too.

davidsbatista:
@brainsparker thanks for this PR!

It looks good overall; I just left a few comments. I'm happy to iterate with you and bring it to our repo.

.github/workflows/youcom.yml
brainsparker: Done — the workflow now reads `secrets.YOUDOTCOM_API_KEY`. Worth flagging for whoever adds it: the secret does not exist yet, and GitHub sets a secret-backed env var to an *empty string* rather than leaving it unset, so the two keyed live tests gate on a truthy value and skip until it is added. The keyless live test needs no key, so CI stays green either way.

integrations/youcom/src/haystack_integrations/components/websearch/youcom/youcom_websearch.py
brainsparker: Applied.

integrations/youcom/README.md
brainsparker: Applied.

integrations/youcom/README.md
brainsparker: Applied.

integrations/youcom/src/haystack_integrations/components/websearch/youcom/youcom_websearch.py
brainsparker: Good catch — agreed that a silent fall-through was the wrong default. Added `keyless_fallback: bool = True`, so the choice is in the signature and in `to_dict()`:

- `True` (default): search the keyless tier, and log at debug which endpoint was selected and why.
- `False`: raise `YouComError` naming the env vars, so a missing key fails fast instead of quietly degrading to a rate-limited tier — the case I think you were pointing at for production pipelines.

Docs link added in three places: the class docstring, the `keyless_fallback` param docs, and the README.

On the env var: I kept `YDC_API_KEY` as a second accepted name alongside `YOUDOTCOM_API_KEY` — `Secret.from_env_var` takes a list, resolves in order, and serializes both. `YDC_API_KEY` is what our own docs and other integrations have shipped for a while, so dropping it outright would silently downgrade existing users to the keyless tier. Happy to make it primary-only if you would rather not carry two names.

integrations/youcom/src/haystack_integrations/components/websearch/youcom/youcom_websearch.py
brainsparker: Agreed, and thanks for the pointer to `SerperDevWebSearch` — that is a better pattern. Added `YouComError(ComponentError)`, exported from the package, and both `run` and `run_async` now raise it with the original exception as `__cause__`.

Two details beyond a literal swap, tell me if you would rather I narrow them:

- I catch `httpx.HTTPError` rather than just `HTTPStatusError`, so transport failures and timeouts surface as `YouComError` too instead of leaking httpx types — matching what serperdev does with its second `except`.
- The upgrade hint is still in the message, but only for a keyless 402/429. It now reads `An error occurred while querying the You.com Search API. Error: …, Response: …`, with the hint appended, so the type carries the "who failed" signal and the message carries the "what to do".

brainsparker:
Thanks for the review @davidsbatista — all three points are addressed in `3e6a85c2`, with replies in each thread.

- **Key env var** → `YOUDOTCOM_API_KEY` everywhere (workflow secret, docstring, README). I kept `YDC_API_KEY` as a second accepted name so existing configurations do not silently drop to the keyless tier; `Secret.from_env_var` takes a list and serializes both. Say the word if you would rather carry only one name.
- **Keyless fallback is now explicit** — `keyless_fallback: bool = True` in the signature and in `to_dict()`, a debug log naming the selected endpoint, and `keyless_fallback=False` raises instead of degrading. Docs link added in the class docstring, the param docs and the README.
- **`YouComError(ComponentError)`** following `SerperDevWebSearch`, exported from the package, raised from both `run` and `run_async` with the original httpx error as `__cause__`.

Verified locally: ruff and mypy clean, 25 unit tests (98% coverage), 3 live integration tests pass, and `hatch run docs` + `docusaurus-mdx-checker` build cleanly.

**One thing I cannot clear myself:** the new-integration bot asks for a linked issue labeled `new integration`. That is #3649 — the proposal you endorsed — but it carries only `P1`, and the gate arrived in the `main` merge you pushed to this branch, so it postdates your review there. Could you add the label to #3649? Happy to file a fresh proposal with the Adoption signals section filled in instead if you would rather have it on the current template — just did not want to open a duplicate of an issue you already approved.

davidsbatista:
I've adjusted the tests a bit

davidsbatista (approved):
Looks good! Thanks for the contribution @brainsparker!

PR #3761, opened 2026-08-12
fix(vespa): handle 404 errors in Vespa document store

bogdankostic:
### Related Issues

- [Failing nightly vespa tests](https://github.com/deepset-ai/haystack-core-integrations/actions/workflows/vespa.yml)

### Proposed Changes:

 
 
Two independent nightly problems, both of which made the vespa nightly fail silently every night since 2026-07-22 (22 consecutive failures).

**1. Vespa integration tests fail (`httpr.HTTPStatusError: Client error '404 Not Found'`).** An upstream regression, nothing changed on our side. pyvespa's `raise_for_status()` branches on whether the response object has a `raise_for_status` method: the requests/httpx branch catches `requests.exceptions.HTTPError`, and only the other branch honours `raise_on_not_found`. httpr **0.5.2** (released 2026-07-20) added `Response.raise_for_status`, so the first branch is now taken and the `httpr.HTTPStatusError` it raises escapes uncaught — `raise_on_not_found=False` stopped suppressing 404s. Both of our `get_data()` call sites relied on that flag, so `write_documents()` raised for every *new* document under the default `FAIL` policy and `get_documents_by_id()` raised instead of skipping missing ids. This was a runtime bug for users, not only a CI failure.

Fixed upstream in vespa-engine/pyvespa#1330 (merged 2026-07-29), which is not yet released. This PR therefore this the transitive dependency: `httpr>=0.4.0,<0.5.2`. The pin is documented inline in `pyproject.toml` and tracked for removal in #3764 together with raising the pyvespa floor once a release containing vespa-engine/pyvespa#1330 is out.

**2. `notify-slack-on-failure` never posted.** The job did trigger, but died before sending: `The action slackapi/slack-github-action@v2 is not allowed in deepset-ai/haystack-core-integrations because all actions must be pinned to a full-length commit SHA.` Three workflows still pinned `deepset-ai/notify-slack-action@3cda73b77a…` (`# v1`), whose `action.yml` references the floating tag `slackapi/slack-github-action@v2`. 

This PR bumps `vespa.yml`, `amazon_textract.yml` and `litellm.yml`.

### How did you test it?

Unit and integration tests pass now.

### Checklist

### DIFF
--- a/.github/workflows/amazon_textract.yml
+++ b/.github/workflows/amazon_textract.yml
@@ -134,6 +134,6 @@ jobs:
     if: failure() && github.event_name == 'schedule'
     runs-on: ubuntu-slim
     steps:
-      - uses: deepset-ai/notify-slack-action@3cda73b77a148f16f703274198e7771340cf862b # v1
+      - uses: deepset-ai/notify-slack-action@a65def0c8bf91d6520286ab34280151c76a5a008 # v1.1.0
         with:
           slack-webhook-url: ${{ secrets.SLACK_WEBHOOK_URL_NOTIFICATIONS }}
--- a/.github/workflows/litellm.yml
+++ b/.github/workflows/litellm.yml
@@ -139,6 +139,6 @@ jobs:
     if: failure() && github.event_name == 'schedule'
     runs-on: ubuntu-slim
     steps:
-      - uses: deepset-ai/notify-slack-action@3cda73b77a148f16f703274198e7771340cf862b # v1
+      - uses: deepset-ai/notify-slack-action@a65def0c8bf91d6520286ab34280151c76a5a008 # v1.1.0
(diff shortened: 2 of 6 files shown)


anakin87 (changes requested):
Thank you, especially for the fix on actions...

**About Vespa**
The bug has already been fixed upstream in pyvespa in https://github.com/vespa-engine/pyvespa/pull/1330 but a new version has not been released

What I'd suggest doing:
- pin  `"httpr<0.5.2"`
- release a patch version of this integration
- create an issue on this repo to remove the httpr pin and bump pyvespa when pyvespa releases a new version

I think this would be cleaner and would not require workaround code.

anakin87 (approved):

PR #3766, opened 2026-08-13
fix: ValidationException on invalid bedrock tool result json

tstadel:
### Related Issues

```python
AmazonBedrockInferenceError
Could not perform inference for Amazon Bedrock model eu.anthropic.claude-sonnet-4-6 due to: · An error occurred (ValidationException) when calling the ConverseStream operation: The format of the value at messages.28.content.0.toolResult.content.0.json is invalid. Provide a json object for the field and try again.
```

Whenever a tool returns its result as a plain string, Haystack tries json.loads() on it. If that succeeds, it wraps the parsed value under {"json": ...} and sends it to Bedrock's Converse API as toolResult.content[0].json.

The problem: json.loads() succeeds not just for JSON objects, but for any valid JSON value — numbers, booleans, null, strings, and arrays. For example, if a tool's string result is:

"42" → parses to 42 (an int)
"true" → parses to True
"[1, 2, 3]" → parses to a list
"null" → parses to None
...then content becomes [{"json": 42}] or [{"json": [1, 2, 3]}], etc. But Bedrock's toolResult.content[].json field (the Document type) specifically requires a JSON object, not an arbitrary JSON value.

### Proposed Changes:

_format_tool_result_message now only wraps a parsed JSON string as a Bedrock {"json": ...} content block when it's actually a JSON object (dict). If a tool's string result parses to a number, boolean, null, or array, it falls back to a {"text": ...} block — avoiding the ValidationException.

### How did you test it?
- Added two regression tests in test_chat_generator_utils.py covering the JSON-object case (still uses json) and the non-object cases ("42", "true", "null", "[1, 2, 3]" — now fall back to text).
- besides unit tests, I ran into this issue having a tool that returns a list of strings. Tested against this too

### Checklist

### DIFF
--- a/integrations/amazon_bedrock/src/haystack_integrations/components/generators/amazon_bedrock/chat/utils.py
+++ b/integrations/amazon_bedrock/src/haystack_integrations/components/generators/amazon_bedrock/chat/utils.py
@@ -197,10 +197,16 @@ def _format_tool_result_message(tool_call_result_message: ChatMessage) -> dict[s
     # Assuming tool call result messages will only contain tool results
     tool_results = []
     for tool_call_result in tool_call_result_message.tool_call_results:
+        content: list[dict[str, Any]]
         if isinstance(tool_call_result.result, str):
             try:
                 json_result = json.loads(tool_call_result.result)
-                content = [{"json": json_result}]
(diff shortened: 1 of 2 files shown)


sjrl (approved):
Thanks!

PR #3772, opened 2026-08-14
test: `ElasticSearchDocumentStore` removing duplicated tests already covered by haystack Mixin tests + overriding

davidsbatista:
### Related Issues

### Proposed changes

- `test_get_metadata_field_unique_values_distinct_types` expects int, float, str and bool values sharing the same metadata field name to all come back as distinct. 

- Elasticsearch's dynamic field mapping fixes a field's type from the first document written to it, so writing an int and then a bool to the same field fails outright. Not fixable in code.

- Override Mixin test

### Checklist


julian-risch (approved):
Looks good overall but we're actually losing some coverage of things:

1. The deleted block tested from_=10 against 3 values. That was the only test of early exit in [_skip_unique_values] (if not buckets and len(buckets) < step). The mixin's pagination test uses from_=2 of 5 values, so remaining reaches 0 and neither branch executes. 

2. The mixin's search-term test only uses strings but we should test Elasticsearch's search_term against a non-string field too. The deleted block asserted search_term="1" on a long meta.priority returns [1], exercising v.value.toString().toLowerCase() for a numeric doc value. 

That's borderline. In my opinion, it's okay to merge in this case but I do see at least some risk of dropping that coverage.

PR #3774, opened 2026-08-14
fix: `ArcadeDBDocumentStore` removing duplicated tests already covered by haystack Mixin tests + fix dedup

davidsbatista:
### Related Issues

### Proposed changes

- removed deuplicated tests, now using Mixin
- Made ArcadeDB's `SELECT DISTINCT` type-aware. The Python-side dedup alone wasn't enough: the server also collapsed the int `1` and the float `1.0` before the rows ever reached us. Projecting the value's type alongside it (`meta[field].type()`) makes DISTINCT compare the pair, so both survive. `count_unique_metadata_by_filter()` had the same collapse and gets the same fix.
- Fixed sorting of mixed-type fields. Values are grouped into mutually comparable types before sorting, so a field holding both numbers and strings no longer raises `TypeError`, and numbers keep numeric rather than string order (`[1, 2, 10]`, not `[1, 10, 2]`) — the same choice pgvector and alloydb make. Ties between values that compare equal but differ in type (`1 == 1.0 == True`) are broken by type name so pagination is stable.

### Checklist


julian-risch (approved):
Looks good to me now! 👍 
I found a sorting issue that I fixed directly and a simplification for the `SELECT DISTINCT` statement. Now, ints and floats are compared correctly and numbers keep numeric order instead of degrading to string order ("10" before "2"). That's consistent with a number of other document stores that we have though not all.

davidsbatista:
Thanks for the review and extra detailed fixed - I will take not of this and check which other doc stores could benefit from the same fix.

PR #3778, opened 2026-08-14
test: `FAISSDocumentStore` removing duplicated tests already covered by haystack Mixin tests

davidsbatista:
### Related Issues

## Why this change

  `get_metadata_field_unique_values()` used a plain `set()` to dedupe unique
  metadata values. Python treats `1 == True == 1.0`, so values of different
  types sharing the same field silently collapsed:

  ```python
  >>> {1, True, 1.0}
  {1}

  FAISS keeps documents purely in memory, so the underlying values were already
  correctly typed — this was a pure Python bug, not a backend limitation.

  Fix: dedupe by (type(value), value) instead of value alone.

### Checklist


integrations/faiss/src/haystack_integrations/document_stores/faiss/document_store.py
bogdankostic: Should we add the same logic here as in `get_metadata_field_unique_values`?

bogdankostic (approved):
Looks good, just found one place where we might want to apply the same logic regarding unique values.

integrations/faiss/src/haystack_integrations/document_stores/faiss/document_store.py
davidsbatista: nice, good catch 👍🏽  fixed it

PR #3780, opened 2026-08-14
test: `OpenSearchDocumentStore` removing duplicated tests already covered by haystack Mixin tests

davidsbatista:
### Related Issues

### Checklist


davidsbatista:
@sjrl wait, something was not commited, going to put it back in draft, sorry

sjrl (approved):
Looks good! Does it make sense to update the mixin to not reuse the same metadata key which is the reason we need the override?

davidsbatista:
> Does it make sense to update the mixin to not reuse the same metadata key which is the reason we need the override?

It does, I will open a PR soon

PR #3781, opened 2026-08-14
fix: `MongoDBDocumentStore` removing duplicated tests already covered by haystack Mixin tests

davidsbatista:
### Related Issues

### Proposed Changes

- updated the test_get_metadata_field_unique_values_distinct_types due to how MongoDB handles primitve types in metadata fields

### Checklist


julian-risch (changes requested):
Test changes look good to me. The changes to document_store change behavior though and therefore the PR title (fix:...) and description should be adjusted. After that, it's ready to be merged.

integrations/mongodb_atlas/src/haystack_integrations/document_stores/mongodb_atlas/document_store.py
julian-risch: Let's document both sync and async methods as in your other PR.
```suggestion
       Asynchronously retrieves unique values for a metadata field, optionally filtered by a search term.
       
       **Note**: values of different types are kept distinct even when they compare equal in Python
       (e.g. the int `1`, the bool `True` and the str `"1"` are returned as three separate values), with
       one exception: MongoDB's aggregation `$group` compares numeric values across BSON subtypes, so a
       whole-number float (e.g. `1.0`) is grouped together with a numerically equal int (`1`) and only
       one of the two survives - regardless of whether they were written to the same metadata field.
       Floats with a fractional part (e.g. `1.5`) are unaffected and stay distinct from ints.

       :param metadata_field: The metadata field to retrieve unique values for.
```

integrations/mongodb_atlas/src/haystack_integrations/document_stores/mongodb_atlas/document_store.py
```diff
@@ -585,7 +600,16 @@ async def get_metadata_field_unique_values_async(
         """
         Asynchronously retrieves unique values for a metadata field, optionally filtered by a search term.
 
-        :param metadata_field: The metadata field to retrieve unique values for.
+       Asynchronously retrieves unique values for a metadata field, optionally filtered by a search term.
+
```
julian-risch:
```suggestion

```

julian-risch (approved):

PR #3784, opened 2026-08-14
test: `PineConeDocumentStore` removing duplicated tests already covered by haystack Mixin tests

davidsbatista:
### Related Issues

## Proposed Changes

- Removed duplicated tests, now relying on Mixin

### Checklist


integrations/pinecone/src/haystack_integrations/document_stores/pinecone/document_store.py
```diff
@@ -894,15 +894,26 @@ def _get_metadata_field_unique_values_impl(
     ) -> tuple[list[Any], int]:
         """Helper method to get unique values for a metadata field with search and pagination."""
         field_name = metadata_field.removeprefix("meta.")
-        unique_values: set[Any] = set()
+        # Values of different types can be equal in Python (`1 == True`, `1 == 1.0`), so a plain set
+        # would silently merge them. Dedupe by (type, value) instead to keep them distinct.
```
bogdankostic, line 898: Today I learned something new, good to know that 😁

bogdankostic (approved):
Looks good!

integrations/pinecone/src/haystack_integrations/document_stores/pinecone/document_store.py
davidsbatista: yeap :) I learned this (8 times) yesterday when tests failed :)

PR #3785, opened 2026-08-14
fix: `QdrantDocumentStore` removing duplicated tests already covered by haystack Mixin tests + fixing dedup + adding missing Async test

davidsbatista:
### Related Issues

### Proposed changes

- Removed duplicataed tests, now relying on Mixng

- Python treats `1 == True == 1.0`, so the plain `set()` used to dedupe unique metadata values  collapsed them into one when a field held mixed types across documents.   Now dedupes by (type(value), value) instead of value alone, so values of different types stay distinct.

- The sync test class already overrides test_get_metadata_field_unique_values_pagination  with an order-tolerant check (page size / total count / no overlap), since Qdrant's scroll() doesn't  guarantee value order. The async class was missing the equivalent override

### Checklist


julian-risch (changes requested):
Let's drop the mixin pagination test overrides (sync and async) and instead sort values, for example with:
`values.sort(key=lambda v: (str(v), type(v).__name__))`. That would be the same as in InMemoryDocumentStore (`sorted(unique_values, key=lambda k: (k[1], k[0]))`)
docstrings and test docstrings need to be updated then.

Independent of that, _process_records_count_unique still has the dedup bug and I suggest you apply your fix there too

integrations/qdrant/tests/test_document_store.py
```diff
@@ -629,47 +629,3 @@ def test_get_metadata_field_unique_values_pagination(self, document_store: Qdran
 
         # Values should not overlap
         assert set(values_page_1) != set(values_page_2)
```
julian-risch: `isdisjoint` is what we want to check here according to the comment

integrations/qdrant/tests/test_document_store_async.py
```diff
@@ -235,6 +235,30 @@ async def test_write_documents_async(self, document_store: QdrantDocumentStore):
...
+
+        values_page_1, total_1 = await document_store.get_metadata_field_unique_values_async("value", from_=0, size=2)
+        assert len(values_page_1) == 2
+        assert total_1 == 5
+
+        values_page_2, total_2 = await document_store.get_metadata_field_unique_values_async("value", from_=2, size=2)
+        assert len(values_page_2) == 2
+        assert total_2 == 5
+
+        assert set(values_page_1) != set(values_page_2)
```
julian-risch: `isdisjoint` is what we want to check here according to the comment

davidsbatista:
> Let's drop the mixin pagination test overrides (sync and async) and instead sort values, for example with:
> values.sort(key=lambda v: (str(v), type(v).__name__)). That would be the same as in InMemoryDocumentStore (sorted(unique_values, key=lambda k: (k[1], k[0]))) docstrings and test docstrings need to be updated then.

fixed

> Independent of that, _process_records_count_unique still has the dedup bug and I suggest you apply your fix there too

Good catch! It's fixed now, using the same strategy of (type(value)

julian-risch (approved):

PR #3786, opened 2026-08-14
test: `ValkeyDocumentStore` removing duplicated tests already covered by haystack Mixin tests + allow non-schema-typed metadata values in Valkey get_metadata_field_unique_value

davidsbatista:
### Related Issues

### Proposed Changes:

- removed duplicated tests, now relying on Mixin

- Unknown field no longer raises — get_metadata_field_unique_values()/_async used to raise ValueError when queried on a field not pre-configured for filtering.

### How did you test it?

- full local suite (make all: lint, mypy, unit, integration against a real Valkey container + CI

### Notes for the reviewer

- No changes to the shared Haystack DocumentStore test mixin — only to ValkeyDocumentStore and Valkey-local tests.

### Checklist


sjrl (approved):
Thanks!

PR #3787, opened 2026-08-14
tests: `WeaviateDocumentStore` removing duplicated tests already covered by haystack Mixin tests

davidsbatista:
### Related Issues

### Checklist


integrations/weaviate/tests/test_document_store.py
```diff
@@ -1144,123 +1144,61 @@ def test_count_unique_metadata_by_filter_field_not_found(self, document_store):
...
-        assert len(values2) == 2
+    # --- Overrides of mixin tests to account for Weaviate-specific behaviour ---
 
-        assert set(values).isdisjoint(set(values2))
+    def test_get_metadata_field_unique_values_missing_field(self, document_store):
+        # Override: Weaviate validates the field against the collection schema and raises
+        # ValueError for an unknown field, instead of returning an empty result as the
+        # generic mixin expects.
+        with pytest.raises(ValueError, match="not found in collection schema"):
+            document_store.get_metadata_field_unique_values("nonexistent_field")
```
bogdankostic: Should we create an issue to adapt the `WeaviateDocumentStore` to follow the expected behavior, i.e. returning an empty result instead of throwing an error?

integrations/weaviate/tests/test_document_store.py
```diff
@@ -1144,123 +1144,61 @@ def test_count_unique_metadata_by_filter_field_not_found(self, document_store):
...
+    @pytest.mark.xfail(
+        reason=(
+            "WeaviateDocumentStore.get_metadata_field_unique_values() returns aggregated numeric "
+            "values that are not strictly the original int type (values compare equal but "
+            "isinstance(value, int) is False) - needs a decision on whether this is fixable in "
+            "the store implementation or an inherent limitation of Weaviate's GroupByAggregate."
+        ),
+        strict=True,
+    )
+    def test_get_metadata_field_unique_values_preserves_type(self, document_store):
```
bogdankostic: Should we create an issue to also investigate this?

integrations/weaviate/tests/test_document_store.py
```diff
@@ -1144,123 +1144,61 @@ def test_count_unique_metadata_by_filter_field_not_found(self, document_store):
...
         document_store.write_documents(docs)
 
-        values, total_count = document_store.get_metadata_field_unique_values("number")
-        assert total_count == 2
-        assert set(values) == {1, 2}
+        values, total_count = document_store.get_metadata_field_unique_values(metadata_field="priority")
 
-    # --- Overrides of mixin tests to account for Weaviate-specific behaviour ---
+        assert total_count == 4
+        assert len(values) == 4
```
bogdankostic: Instead of using xfail, should we here follow the approach we took for the OracleDocumentStore?
https://github.com/deepset-ai/haystack-core-integrations/pull/3782/changes#diff-650d18de4574fc2c51dca5ab757501687e3c4a403acaf3b50ab9e6970d627468R349

bogdankostic (commented):
Looks good in general, just added some questions on creating issues to align the behavior of WeaviateDocumentStore with the expected behavior and on using the same approach for `test_get_metadata_field_unique_values_missing_field` as we do for `OracleDocumentStore`.

integrations/weaviate/tests/test_document_store.py
davidsbatista: But an empty result can mislead the user, thinking that there are no matches when in fact the problem is that the metadata field he's passing does not exist

bogdankostic: `test_get_metadata_field_unique_values_missing_field` in the `GetMetadataFieldUniqueValuesTest` mixin expects the return value to be an empty list though, so I guess we would have this problem in all other document stores?

integrations/weaviate/tests/test_document_store.py
davidsbatista: yes, good catch 👍🏽

integrations/weaviate/tests/test_document_store.py
davidsbatista: ah ok! I did not check that, better to replicate the same behaviour.

integrations/weaviate/tests/test_document_store.py
davidsbatista: I think we can do it here, similar we did for new dedup strategy. the `_compute_field_unique_values` alreay access the index config to check if metadata field exits, there we can also get the internal Weavite type and do the proper casting

davidsbatista: this will remove the override and pass the Mixin tests

davidsbatista:
- reverted get_metadata_field_unique_values()/_async() in document_store.py: unknown field now returns ([], 0) instead of raising ValueError, matching the mixin's expectation and removed override

- is now an override test_get_metadata_field_unique_values_distinct_types (sync+async) instead a xfail-wrapped

davidsbatista:
@bogdankostic here's the update on the #2 issue - thanks for raising it 👍🏽 

There's a limitation in weaviate-client:

### Write side
- when inserting a document, weaviate-client only preserves int vs. float for list properties (e.g. meta={"tags": [1, 2]}). A scalar value (e.g. meta={"priority": 1}) gets packed into a generic google.protobuf.Struct, a JSON-like type whose only numeric slot is number_value (a double). JSON itself has no int/float distinction, so there's no int_value to use. The 1 becomes 1.0 before it even leaves the client.

### Read side
- Reading unique values back via GroupByAggregate goes through the same kind of decoding — it returns float, even if we explicitly declare the field as DataType.INT in the schema. 

- Bottom line: the int is lost on both ends, before our code ever sees the value, and there's no schema signal left anywhere to recover it.

### Proposal

I thought about adding a skip instead of the xfail, or an override:

- Skip
  - Pros: simplest, clearly communicates "not going to happen."
  - Cons: no signal if it ever gets fixed upstream — stays silently skipped forever.

- xfail(strict=True)
  - Pros: automatic canary — if the a new client supports this test flips forcing someone to notice.
  - Cons: keeps the mixin's "ideal" assertion (asserts int), so it reads as "broken" rather than documenting behavior; 

- Override
  - Pros: documents actual behavior in plain code (asserts float, not int); consistent with how every other store in this PR/session was handled; still catches an upstream fix (assertion just starts failing normally instead of XPASS).
  - Cons: less visible as "known limitation" at a glance — buried in a docstring rather than surfaced in pytest's xfailed summary.
  
I would vote for the override + docstring, WDTY?

bogdankostic:
> I would vote for the override + docstring, WDTY?

That's also fine for me, let's do it this way.

bogdankostic (approved):
LGTM

PR #3788, opened 2026-08-14
tests: `FalkorDBDocumentStore` removing duplicated tests already covered by haystack Mixin tests + overriding one

davidsbatista:
### Related Issues

### Proposed changes

- `test_get_metadata_field_unique_values_distinct_types` expects int, float, str and bool values sharing the same metadata field name to all come back as distinct. FalkorDB fails this because of a Cypher limitation:

  ```cypher
  WITH DISTINCT d.priority AS val   -- 1 and 1.0 collapse into one value
    ```

  Verified live: a per-node property read correctly preserves each value's own
  type, but Cypher's DISTINCT treats the int 1 and the float 1.0 as equal,
  collapsing them into a single value (3 distinct values instead of 4). Not
  fixable in code.

  Fix: adapted the test to use one metadata field per type instead of one
  shared field (matching the same pattern applied to other document stores
  hitting this limitation), and used a fractional float (1.5) to sidestep the
  int/float ambiguity. Documented the limitation in the docstring and an inline
  comment.

### Checklist

### DIFF
--- a/integrations/falkordb/src/haystack_integrations/document_stores/falkordb/document_store.py
+++ b/integrations/falkordb/src/haystack_integrations/document_stores/falkordb/document_store.py
@@ -614,6 +614,12 @@ SET d.{self.embedding_field} = vecf32(doc.emb)
         """
         Return distinct values for the given metadata field with optional filtering and pagination.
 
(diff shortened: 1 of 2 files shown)


sjrl (approved):
Looks good!

PR #3790, opened 2026-08-16
fix(oauth): rebuild the OAuthRefreshTokenSource refresh lock per event loop

### LINKED ISSUE #3789: oauth: OAuthRefreshTokenSource caches its asyncio.Lock across event loops, breaking reuse in a new loop
### Description

`OAuthRefreshTokenSource.resolve_async` creates its `asyncio.Lock` lazily and then keeps it for the lifetime of the source:

```python
# integrations/oauth/src/haystack_integrations/utils/oauth/sources.py
if self._async_lock is None:
    self._async_lock = asyncio.Lock()
async with self._async_lock:
```

An `asyncio.Lock` binds itself to the loop that first awaits it **under contention** and raises `RuntimeError` when awaited from any other one. A source is a plain, serializable object that routinely outlives the loop that first used it — one `asyncio.run(...)` per request is a common deployment, and it is also how a lot of test code is written — so the second loop's first *contended* refresh fails.

Worth stressing that contention is not an edge case here. Collapsing a burst of concurrent callers into a single network refresh is the only reason the lock exists, and the class docstring says so.

The lazy creation itself is not a data race within one loop: there is no `await` between the `is None` check and the assignment, so a single event loop cannot interleave there. Two threads each driving their own loop can, but the loop-binding failure below is the one that reproduces deterministically.

### Reproduction

```python
import asyncio
from unittest.mock import patch

import httpx
from haystack_integrations.utils.oauth import OAuthRefreshTokenSource

source = OAuthRefreshTokenSource(
    token_url="https://idp.example.com/oauth2/token",
    client_id="client-1",
    refresh_token=Secret.from_token("rt"),
)


async def slow_post(*_args, **_kwargs):
    # Yield to the loop so the second caller reaches the lock while the first holds it.
    await asyncio.sleep(0)
    return httpx.Response(200, json={"access_token": "acc", "expires_in": 0})


async def two_concurrent_callers():
    with patch("httpx.AsyncClient.post", side_effect=slow_post):
        return await asyncio.gather(source.resolve_async(), source.resolve_async())


print(asyncio.run(two_concurrent_callers()))  # ['acc', 'acc']
print(asyncio.run(two_concurrent_callers()))  # RuntimeError
```

```text
RuntimeError: <asyncio.locks.Lock object at 0x...> is bound to a different event loop
```

Reproduced on Python 3.12.3, `oauth-haystack` at `main`.

Note that dropping the `asyncio.sleep(0)` makes the script pass: an uncontended `Lock.acquire()` returns before it ever calls `_get_loop()`, so the loop is never bound. That is why the existing `test_resolve_async_success_and_caches` does not catch this — it awaits one caller at a time.

### Expected Behavior

A source reused from a new event loop should refresh normally. The lock is per-loop runtime state, so it should be rebuilt when the running loop changes rather than cached for the lifetime of the object.

### Additional Context

`OAuthTokenExchangeSource` is not affected: it uses a `threading.Lock` held only around cache access and never across the network call, which its own comment calls out.

I have a fix and a regression test ready and will open a PR referencing this issue.

---

Disclosure: I used AI assistance while investigating and writing this up. I ran the reproduction and confirmed the failure and the root cause myself.

pcbeingused333:
### Related Issues

- fixes #3789

### Proposed Changes:

`OAuthRefreshTokenSource.resolve_async` built its `asyncio.Lock` lazily and then kept it for the lifetime of the source. An `asyncio.Lock` binds to the loop that first awaits it under contention and raises `RuntimeError` when awaited from another one, so a source reused across event loops — one `asyncio.run` per request is a common deployment — failed on the second loop's first contended refresh:

```text
RuntimeError: <asyncio.locks.Lock object> is bound to a different event loop
```

Contention is not an edge case here: collapsing a burst of concurrent callers into a single network refresh is the only reason the lock exists.

`_get_async_lock` now returns the lock for the running loop and rebuilds it when the loop changes. A dedicated `_async_lock_guard` (a `threading.Lock`) covers the check and the assignment, never an await, so two threads driving separate loops cannot each install a lock and lose mutual exclusion between them.

`_sync_lock` was deliberately not reused for that guard: `resolve` holds it across a blocking network call, so an async caller waiting on it would stall its own event loop for the duration of a token refresh.

`OAuthTokenExchangeSource` needed no change — it holds a `threading.Lock` only around cache access and never across the network call, as its own comment says.

### How did you test it?

Two tests added to `TestOAuthRefreshTokenSource`:

- `test_resolve_async_works_in_a_second_event_loop` is the regression pin: it drives two concurrent callers under `asyncio.run`, then does it again on a fresh loop with the same source.
- `test_concurrent_resolve_async_refreshes_once` pins the property the lock exists for — four concurrent callers, one network refresh. It passes before and after; it is there so a future change to the locking cannot quietly drop the coalescing.

I checked the regression test is not vacuous: with `sources.py` reverted to `main` it fails with the `RuntimeError` above, raised from `asyncio/mixins.py:20` — the bug itself, not an incidental assertion mismatch. The standalone reproduction in #3789 was run both ways too (fails on `main`, passes on this branch).

Both halves of the test have to contend for the lock or it proves nothing: an uncontended `Lock.acquire()` returns before it ever calls `_get_loop()`, so the loop is never bound. That is exactly why the existing `test_resolve_async_success_and_caches` did not catch this, and the `_slow_post` helper exists to force the overlap.

`hatch run test:unit tests/test_sources.py` → 33 passed. `hatch run test:types` and `hatch run fmt-check .` clean.

### Notes for the reviewer

Two judgement calls worth a look:

1. **Rebuilding on loop change rather than keeping a lock per loop.** If two event loops in two threads use the same source *simultaneously*, each one rebuilds the lock and mutual exclusion between them is lost. A `WeakKeyDictionary` keyed by loop would hold in that case too. I did not do it because the class is documented as single-identity and "use a single instance in either sync or async mode, not both", so the realistic pattern is sequential loops, not simultaneous ones — but say the word and I will switch it.

2. **The extra `threading.Lock`.** It is one more attribute on a class that already has one lock. Without it, the check-and-set is not atomic across threads. Happy to drop it if you consider cross-thread use out of scope.

### Checklist

### DIFF
--- a/integrations/oauth/src/haystack_integrations/utils/oauth/sources.py
+++ b/integrations/oauth/src/haystack_integrations/utils/oauth/sources.py
@@ -121,7 +121,10 @@ class OAuthRefreshTokenSource:
 
         # Runtime state — never serialized.
         self._sync_lock = Lock()
+        # `_async_lock_guard` only guards the creation of `_async_lock`; see `_get_async_lock`.
+        self._async_lock_guard = Lock()
         self._async_lock: asyncio.Lock | None = None
+        self._async_lock_loop: asyncio.AbstractEventLoop | None = None
         self._cached_token: str | None = None
         self._expires_at: float = 0.0
         self._current_refresh_token: str | None = None
@@ -178,13 +181,29 @@ class OAuthRefreshTokenSource:
             )
             return self._handle_response(response)
 
+    def _get_async_lock(self) -> asyncio.Lock:
+        """
+        Return the refresh lock for the running event loop, building it if there isn't one yet.
+
+        The lock cannot be created in `__init__`, which is synchronous and may run with no loop. It also cannot be
+        created once and kept for the lifetime of the source: an `asyncio.Lock` binds itself to the loop that first
+        awaits it under contention, and raises `RuntimeError` when awaited from any other one. A source routinely
+        outlives the loop that first used it — one `asyncio.run` per request is a common deployment — so the lock is
+        rebuilt whenever the running loop changes.
+
+        `_async_lock_guard` is held for the check and the assignment only, never across an await, so that two threads
+        driving separate event loops cannot each install a lock and lose mutual exclusion between them.
+        """
+        loop = asyncio.get_running_loop()
+        with self._async_lock_guard:
+            if self._async_lock is None or self._async_lock_loop is not loop:
+                self._async_lock = asyncio.Lock()
+                self._async_lock_loop = loop
+            return self._async_lock
+
     async def resolve_async(self) -> str:
         """Asynchronous counterpart of `resolve`. Use a single instance in either sync or async mode, not both."""
-        # Create the asyncio.Lock lazily (not in __init__): it must bind to the running event loop, but __init__
-        # is sync and may run without one.
-        if self._async_lock is None:
-            self._async_lock = asyncio.Lock()
-        async with self._async_lock:
+        async with self._get_async_lock():
             cached = self._cached_if_valid()
             if cached is not None:
                 return cached
--- a/integrations/oauth/tests/test_sources.py
+++ b/integrations/oauth/tests/test_sources.py
@@ -2,6 +2,7 @@
 #
 # SPDX-License-Identifier: Apache-2.0
 
+import asyncio
 from unittest.mock import patch
 
 import httpx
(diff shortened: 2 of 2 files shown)


bogdankostic (approved):

PR #3793, opened 2026-08-16
fix: LibreOffice - serialize output_file_type and increase unit tests coverage

julian-risch:
### Related Issues

- Related to closed issue https://github.com/deepset-ai/haystack-private/issues/275 about increasing unit test coverage. Similar to PR https://github.com/deepset-ai/haystack-core-integrations/pull/3794. Here from 58% to 100%.

I also found two small problems to fix so this PR does not just increase unit test coverage.

### Proposed Changes:

Fixes:
- `to_dict()` dropped `output_file_type`, so it did not survive a `from_dict()` round trip. Pass it through.
- `run_async` discarded the soffice exit code, while `run` passes `check=True`. Since every `ByteStream` in one call is written to the same `input` temp file, a failed async conversion could return the previous source's converted bytes rather than raising. Now, we check the exit code and raise the same `CalledProcessError` the sync path already raises

Tests:
- Add unit tests to increase coverage, mostly by patching the soffice call instead of requiring a LibreOffice installation: `run_async` (was 0%), the `run` ByteStream branch, `_get_conversion_args`, every `_validate_args` reject branch, and the `output_file_type` resolution
- Split the suite into focused classes (as suggested in https://github.com/deepset-ai/haystack-core-integrations/pull/3794) `TestInit`, `TestSerde`, `TestGetConversionArgs`, `TestValidateArgs`, `TestConvert`, `TestResolveMimeType`, `TestRunIntegration`. `TestConvert` is parametrized over `run` and `run_async`
- Move fixtures to `tests/conftest.py`
- Rewrite four pre-existing integration tests into one parametrized case

### Checklist

- I have updated the related issue with new insights and changes — n/a


sjrl (approved):
Looks good!

PR #3807, opened 2026-08-17
fix: KreuzbergConverter config round-trips as YAML dict instead of JSON string

### LINKED ISSUE #3806: KreuzbergConverter: YAML Config Serialization Issue
**Describe the bug**
`KreuzbergConverter.from_dict()` only accepts `config` as a JSON string. Any system that stores or round-trips the pipeline YAML — loading it into a Python dict and dumping it back — will naturally produce a YAML mapping for `config`. When that happens, `from_dict()` silently skips deserialization and passes the raw dict into `__init__()` as the `config` argument.

Python does not enforce the `ExtractionConfig` type annotation at runtime, so no error is raised on init. The failure is deferred until the component is first used, when `_build_config()` calls `config_to_json(self.config)` on a plain dict instead of an `ExtractionConfig` object.

**To Reproduce**
The same failure occurs when a user authors `config` as a YAML mapping directly:

```yaml
config:
  disable_ocr: true
  pdf_options:
    extract_images: false
```

This is the natural and expected way to write structured config in YAML, yet it silently produces a broken component.

The only form that currently works is a raw JSON string, which any YAML load/dump cycle will re-emit as an escaped blob:

```yaml
config: "{\n  \"disable_ocr\": true,\n  \"pdf_options\": {\n    \"extract_images\": false\n\
  \  }\n}"
```

deep-rloebbert:
Fixes #3806

## What

`to_dict()` now emits `config` as a plain Python dict instead of a raw JSON string. PyYAML serializes dicts as native YAML mappings, so the config is readable after any load/dump cycle and can be authored naturally by hand.

`from_dict()` gains a `dict` branch to accept config in the natural YAML mapping form. The existing `str` branch is kept for backward compatibility with pipelines already carrying the JSON string form.

## Changes

**`converter.py`**
- Added `import json`
- `to_dict()`: `config_to_json(self.config)` → `json.loads(config_to_json(self.config))`
- `from_dict()`: added `elif isinstance(config_data, dict)` branch

**`tests/test_converter.py`**
- Updated `test_serialization_roundtrip_all_params` assertions from `str`/`json.loads` to direct dict access
- Added `test_serialization_from_dict_config_as_dict`: verifies dict input is correctly deserialized
- Added `test_serialization_from_dict_config_as_str_backward_compat`: verifies legacy JSON string input still works


davidsbatista (approved):
Looks good! Thanks for this Robert 👍🏽

PR #3808, opened 2026-08-17
fix: serialize missing init params in to_dict (transformers, huggingface_api, ragas)

pcbeingused333:
### Related Issues

### Proposed Changes:

Three components accept an `__init__` parameter, store it, and use it at run time, but leave it out of `to_dict`. In each case the value silently falls back to its default after a `to_dict` / `from_dict` round trip, and nothing in the serialized dict shows that the setting was ever there. `from_dict` needed no change in any of them — `default_from_dict` passes `init_parameters` straight through to `__init__`.

**`TransformersZeroShotTextRouter.multi_label`** — passed to the Hugging Face pipeline in `run`:

```python
self.pipeline([text], candidate_labels=self.labels, multi_label=self.multi_label)
```

The flag decides whether label scores are normalized to sum to 1 across labels, or whether each label is scored independently against its own contradiction score. The router picks its output branch from those scores, so the same text can route to a different branch after a pipeline is saved and reloaded. `model` and `device` are absent from `to_dict` too, but correctly so: `_resolve_hf_pipeline_kwargs` folds them into `huggingface_pipeline_kwargs`, which is serialized. `multi_label` is the only one that goes nowhere. The sibling component in this same integration, `TransformersZeroShotDocumentClassifier`, has the same shape and already serializes it — which is what suggests an oversight rather than a decision.

**`HuggingFaceTEIRanker.raw_scores`** — goes directly into the request payload, on both the sync and async paths:

```python
payload: dict[str, Any] = {"query": query, "texts": texts, "raw_scores": self.raw_scores}
```

After a reload the ranker silently stops asking the TEI endpoint for raw relevance scores: the request body changes with nothing to show why. Everything else `__init__` takes (`url`, `top_k`, `timeout`, `token`, `max_retries`, `retry_status_codes`) is already serialized; `raw_scores` is the only one missing.

**`RagasEvaluator.concurrency_limit`** — sizes the semaphore bounding how many metrics are evaluated at once in `run_async`:

```python
sem = Semaphore(max(1, self.concurrency_limit))
```

`to_dict` serialized only `ragas_metrics`, so an evaluator built with `concurrency_limit=16` comes back at the default of `4`. Every one of those concurrent evaluations is an LLM call, so the value decides whether an evaluation run finishes in a quarter of the time or hits the judge model's rate limit.

### How did you test it?

One new round-trip regression test per component, each asserting the value survives `to_dict` → `from_dict`:

| Component | Test | Suite |
|---|---|---|
| `TransformersZeroShotTextRouter` | `test_multi_label_survives_a_serialization_round_trip` | `hatch run test:pytest tests/test_zero_shot_text_router.py` — 9 passed |
| `HuggingFaceTEIRanker` | `test_raw_scores_survives_a_serialization_round_trip` | `hatch run test:pytest tests/test_ranker.py` — 16 passed |
| `RagasEvaluator` | `test_concurrency_limit_survives_a_serialization_round_trip` | `hatch run test:pytest tests/test_evaluator.py` — 27 passed, 8 skipped (integration tests needing an API key) |

`hatch run test:types` and `hatch run fmt-check` are clean in all three integrations.

I checked each regression test is not vacuous — with the production file reverted to `main`, each fails for the reason the bug describes rather than on an incidental assertion mismatch:

- router: `assert False is True` on the restored component's `multi_label`
- ranker: `assert False is True` on the restored component's `raw_scores`
- evaluator: `assert 4 == 16` — the round trip falling back to the default

`test_to_dict` in the transformers and ragas suites compares the whole dict, so both needed the new key added in the same commit. The huggingface_api `test_to_dict` asserts key by key, which is also why that gap survived: a full-dict assertion would have flagged the missing key when `raw_scores` was added.

### Notes for the reviewer

This supersedes #3809 (`raw_scores`) and #3810 (`concurrency_limit`), which I have closed in favour of this one. Those PRs said I would keep the fixes separate because each integration ships on its own; I have grouped them here instead, since it is one class of bug found by one sweep and it reads better reviewed as a whole than as three near-identical PRs. Releases are unaffected — `CI_pypi_release.yml` generates each changelog with `git-cliff --include-path "<integration>/**/*"`, so the entry is partitioned by file path, not by the commit scope in the title.

The three commits are kept separate and per-integration, so the change can still be split back apart if you would rather take them one at a time.

`concurrency_limit` is documented as only affecting `run_async`, so that round trip is silent on the sync path — part of why it is easy to miss.

### Checklist

### DIFF
--- a/integrations/huggingface_api/src/haystack_integrations/components/rankers/huggingface_api/ranker.py
+++ b/integrations/huggingface_api/src/haystack_integrations/components/rankers/huggingface_api/ranker.py
@@ -108,6 +108,7 @@ class HuggingFaceTEIRanker:
             self,
             url=self.url,
             top_k=self.top_k,
+            raw_scores=self.raw_scores,
             timeout=self.timeout,
             token=self.token,
             max_retries=self.max_retries,
--- a/integrations/huggingface_api/tests/test_ranker.py
+++ b/integrations/huggingface_api/tests/test_ranker.py
@@ -60,6 +60,14 @@ class TestHuggingFaceTEIRanker:
         assert data["init_parameters"]["max_retries"] == 4
         assert data["init_parameters"]["retry_status_codes"] == [500, 502]
 
+    def test_raw_scores_survives_a_serialization_round_trip(self, del_hf_env_vars_if_empty):
+        """raw_scores goes into the request payload, so losing it changes what the API returns."""
+        ranker = HuggingFaceTEIRanker(url="https://api.my-tei-service.com", raw_scores=True)
+
+        restored = HuggingFaceTEIRanker.from_dict(ranker.to_dict())
+
+        assert restored.raw_scores is True
+
     def test_from_dict(self, del_hf_env_vars_if_empty):
         """Test deserialization from dict with environment variable token"""
         data = {
--- a/integrations/ragas/src/haystack_integrations/components/evaluators/ragas/evaluator.py
+++ b/integrations/ragas/src/haystack_integrations/components/evaluators/ragas/evaluator.py
@@ -85,7 +85,11 @@ class RagasEvaluator:
         :returns:
             Dictionary with serialized data.
         """
-        return default_to_dict(self, ragas_metrics=[_serialize_metric(m) for m in self.metrics])
+        return default_to_dict(
+            self,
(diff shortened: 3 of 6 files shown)


bogdankostic (approved):

PR #3809, opened 2026-08-18
fix(huggingface_api): serialize raw_scores in HuggingFaceTEIRanker

pcbeingused333:
### Related Issues

### Proposed Changes:

`HuggingFaceTEIRanker.__init__` accepts `raw_scores`, stores it, and puts it directly into the request payload — on both paths:

```python
payload: dict[str, Any] = {"query": query, "texts": texts, "raw_scores": self.raw_scores}
```

`to_dict` does not serialize it, so a ranker built with `raw_scores=True` comes back from `from_dict` with the `False` default. Save a pipeline containing that ranker, reload it, and it silently stops asking the TEI endpoint for raw relevance scores — the request body changes, and nothing in the serialized dict shows why.

Everything else `__init__` takes (`url`, `top_k`, `timeout`, `token`, `max_retries`, `retry_status_codes`) is already serialized; `raw_scores` is the only one missing.

`from_dict` needed no change — `default_from_dict` passes `init_parameters` through to `__init__`.

### How did you test it?

- New unit test `test_raw_scores_survives_a_serialization_round_trip`: builds the ranker with `raw_scores=True`, runs it through `to_dict` then `from_dict`, and asserts the flag is still `True`.
- I checked it is not vacuous: with `ranker.py` reverted to `main` it fails with `assert False is True` on the restored component — the round trip losing the flag, not an incidental assertion mismatch.
- `hatch run test:pytest tests/test_ranker.py` — 16 passed.
- `hatch run test:types` clean (10 source files), `hatch run fmt-check` clean.

### Notes for the reviewer

The existing `test_to_dict` asserts key by key rather than comparing the whole dict, so it did not need changing — which is also why the gap survived: a full-dict assertion would have flagged the missing key when `raw_scores` was added.

This came out of a sweep comparing every component's `to_dict` against its `__init__` signature. It found the same class of gap in a few other packages, and since each integration ships separately I am opening those as their own PRs rather than one cross-package change.

### Checklist

### DIFF
--- a/integrations/huggingface_api/src/haystack_integrations/components/rankers/huggingface_api/ranker.py
+++ b/integrations/huggingface_api/src/haystack_integrations/components/rankers/huggingface_api/ranker.py
@@ -108,6 +108,7 @@ class HuggingFaceTEIRanker:
             self,
             url=self.url,
             top_k=self.top_k,
+            raw_scores=self.raw_scores,
             timeout=self.timeout,
             token=self.token,
             max_retries=self.max_retries,
--- a/integrations/huggingface_api/tests/test_ranker.py
+++ b/integrations/huggingface_api/tests/test_ranker.py
@@ -60,6 +60,14 @@ class TestHuggingFaceTEIRanker:
         assert data["init_parameters"]["max_retries"] == 4
         assert data["init_parameters"]["retry_status_codes"] == [500, 502]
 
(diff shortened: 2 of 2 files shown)


pcbeingused333:
Folding this into #3808 together with the same fix in `TransformersZeroShotTextRouter` and `RagasEvaluator` — it is one class of bug from one sweep, and it reads better reviewed as a whole than as three near-identical PRs. The commit for `raw_scores` is carried over unchanged; closing here to keep the review in one place.

PR #3812, opened 2026-08-18
fix: confine S3Downloader downloads to file_root_path

julian-risch:
### Related Issues

- Fixes deepset-ai/haystack-private#533

`S3Downloader._download_file` builds the local destination in a way that does not confine it to `file_root_path`

### Proposed Changes:
- Added `_resolve_local_file_path()`, which resolves the destination and requires it to be `is_relative_to()` the resolved `file_root_path`. It also refuses a file name that resolves to the root directory itself (e.g. `"sub/.."`), which would otherwise be handed to `os.utime`/`os.replace`.
- A document that fails the check is logged at warning level and skipped, exactly like a document missing the file name metadata key. 
- Docstrings for `file_root_path`, `file_name_meta_key`, `run()` and `_download_file()` now state the confinement and the skip behaviour.
- Updated some `:raises:`:
  - `run()` documented `:raises ValueError: If the path where files will be downloaded is not set.`, but `file_root_path` is validated in `__init__` and `run()` can never raise for it. 
  - `__init__` now documents the `AWSConfigurationError` that `get_aws_session` raises on invalid credentials, matching what sibling Bedrock components already document.
  - `warm_up()` documented nothing, but raises `ValueError` when the bucket environment variable is unset and `S3ConfigurationError` when the S3 client cannot be created.

### How did you test it?

- Added two unit tests

### Checklist

- I have updated the related issue with new insights and changes — n/a, issue already describes the fix

### DIFF
--- a/integrations/amazon_bedrock/src/haystack_integrations/components/downloaders/s3/s3_downloader.py
+++ b/integrations/amazon_bedrock/src/haystack_integrations/components/downloaders/s3/s3_downloader.py
@@ -153,7 +153,8 @@ class S3Downloader:
             - `documents`: The downloaded `Document`s; each has `meta['file_path']`. Documents whose file name
               is missing, or resolves outside of `file_root_path`, are logged and skipped.
         :raises S3Error: If a download attempt fails or the file does not exist in the S3 bucket.
-        :raises ValueError: If the path where files will be downloaded is not set.
+        :raises ValueError: If the component has not been warmed up yet and the environment variable naming the
+            S3 bucket (`s3_bucket_name_env`, by default `S3_DOWNLOADER_BUCKET`) is not set.
         """
 
         if self._storage is None:
--- a/integrations/amazon_bedrock/src/haystack_integrations/components/downloaders/s3/s3_downloader.py
+++ b/integrations/amazon_bedrock/src/haystack_integrations/components/downloaders/s3/s3_downloader.py
@@ -160,9 +160,6 @@ class S3Downloader:
             - `documents`: The downloaded `Document`s; each has `meta['file_path']`. Documents whose file name
               is missing, or resolves outside of `file_root_path`, are logged and skipped.
         :raises S3Error: If a download attempt fails or the file does not exist in the S3 bucket.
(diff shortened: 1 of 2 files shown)


integrations/amazon_bedrock/src/haystack_integrations/components/downloaders/s3/s3_downloader.py
```diff
@@ -181,14 +192,42 @@ def _filter_documents_by_extensions(self, documents: list[Document]) -> list[Doc
             if Path(doc.meta.get(self.file_name_meta_key, "")).suffix.lower() in self.file_extensions
         ]
 
+    def _resolve_local_file_path(self, file_name: str) -> Path:
```
julian-risch, line 193: This could become a helper in `haystack/utils/` as `path_validation.py` similar to `url_validation.py` but I don't want to add a dependency on latest Haystack release for that here.

integrations/amazon_bedrock/src/haystack_integrations/components/downloaders/s3/s3_downloader.py
```diff
@@ -148,9 +157,12 @@ def run(
...
         :param documents: Document containing the name of the file to download in the meta field.
         :returns: A dictionary with:
-            - `documents`: The downloaded `Document`s; each has `meta['file_path']`.
+            - `documents`: The downloaded `Document`s; each has `meta['file_path']`. Documents whose file name
+              is missing, or resolves outside of `file_root_path`, are logged and skipped.
         :raises S3Error: If a download attempt fails or the file does not exist in the S3 bucket.
-        :raises ValueError: If the path where files will be downloaded is not set.
+        :raises ValueError: If the environment variable naming the S3 bucket (`s3_bucket_name_env`, by default
+            `S3_DOWNLOADER_BUCKET`) is not set. `run()` warms the component up on first use, so this surfaces
+            from the implicit `warm_up()` call.
```
julian-risch:
```suggestion
```

bogdankostic (approved):

PR #3823, opened 2026-08-20
fix: fix finish reason mapping for streaming Gemini models

sjrl:
### Related Issues

- fixes #issue-number

### Proposed Changes:

 
 

Some Gemini models stream their finish reason in a separate chunk from the last content part. This broke our existing finish reason mapping in our streaming code which assumed a streamed tool call chunk also carried its finish reason. This isn't guaranteed so we now support both cases where a finish reason could be sent on its own or with a tool call content part. 

### How did you test it?

New and existing tests. 

### Checklist

### DIFF
--- a/integrations/google_genai/src/haystack_integrations/components/generators/google_genai/chat/utils.py
+++ b/integrations/google_genai/src/haystack_integrations/components/generators/google_genai/chat/utils.py
@@ -590,6 +590,8 @@ def _convert_google_chunk_to_streaming_chunk(
     index: int,
     component_info: ComponentInfo,
     model: str,
+    *,
+    tool_calls_seen: bool = False,
 ) -> StreamingChunk:
     """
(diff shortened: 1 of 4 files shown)


integrations/google_genai/src/haystack_integrations/components/generators/google_genai/chat/utils.py
```diff
@@ -627,7 +631,7 @@ def _convert_google_chunk_to_streaming_chunk(
     if cached_content_token_count is not None:
         usage["cached_content_token_count"] = cached_content_token_count
 
-    if candidate.content and candidate.content.parts:
+    if chunk.candidates and candidate.content and candidate.content.parts:
```
sjrl, line 634: This is to guard against the case where `candidate` is not defined.

anakin87 (approved):
Thank you!

PR #3829, opened 2026-08-21
feat(fastembed): add Python 3.14 support

### LINKED ISSUE #3035: FastEmbed integration: bring Python 3.14 support when implemented in the original library
Currently, FastEmbed does not support Python 3.14: https://github.com/qdrant/fastembed/issues/576

We should monitor the above issue and, when FastEmbed becomes compatible with 3.14:
- update our pin in pyproject
- test with 3.14 (instead of 3.13)
- add 3.14 information to `classifiers` in pyproject.toml

SyedShahmeerAli12:
FastEmbed 0.8.0 added Python 3.14 support (tracked in qdrant/fastembed#576). This PR adds the 3.14 classifier to `pyproject.toml` and updates the CI matrix to test on 3.14 instead of 3.13.

Closes #3035

### DIFF
--- a/.github/workflows/fastembed.yml
+++ b/.github/workflows/fastembed.yml
@@ -24,7 +24,7 @@ env:
   PYTHONUNBUFFERED: "1"
   FORCE_COLOR: "1"
   TEST_MATRIX_OS: '["ubuntu-latest", "windows-latest", "macos-latest"]'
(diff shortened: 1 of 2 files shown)


sjrl:
@SyedShahmeerAli12 thanks for the contribution! It looks like one test is failing could you take a look and address that?

sjrl (approved):
Thanks!

PR #3846, opened 2026-08-25
fix(azure_ai_search): use a value-safe delimiter for search.in filters

### LINKED ISSUE #3844: AzureAISearchDocumentStore's 'in' filter joins values with the same comma Azure uses as its search.in() delimiter
**Describe the bug**

`AzureAISearchDocumentStore`'s `in` filter operator builds a `search.in(field, valueList, delimiter)` OData expression by joining values with a comma (`_in()` in `filters.py`), the same character Azure uses as the delimiter argument. Any string metadata value that itself contains a literal comma gets split apart by Azure's own server side parsing of that value list, so the filter never matches the value you passed in. A value like `"New York, NY"` becomes three fragments (`New York`, ` NY`, `Austin` in a two-value list), none of which equal the original string.

`_escape_odata_literal` only escapes single quotes for the `eq`/`ne` operators; `_in` calls it too, but escaping a quote does nothing for a comma, since the comma is consumed by `search.in`'s own delimiter parsing, not by OData string-literal parsing.

**To Reproduce**

```python
from haystack_integrations.document_stores.azure_ai_search.filters import _normalize_filters

_normalize_filters({"operator": "in", "field": "meta.city", "value": ["New York, NY", "Austin"]})
# "search.in(city,'New York, NY,Austin',',')"
```

Per Azure's documented `search.in(variable, valueList, delimiter)` semantics, the service splits the value-list string on the delimiter before comparing, so the query above resolves to matching against `["New York", " NY", "Austin"]`, never `"New York, NY"`. I reproduced the query-construction output above in a clean container against current `main` (`39ba76eb3`); I do not have a live Azure Search index to confirm the server-side split independently, but it follows Azure's own documented behavior for `search.in`.

A fix would need a delimiter that cannot appear in the values (or per-value escaping of the chosen delimiter before joining), plus a regression test alongside the existing `_in`/`_escape_odata_literal` tests.

**Describe your environment**

 - Integration: `azure-ai-search-haystack`
 - File: `integrations/azure_ai_search/src/haystack_integrations/document_stores/azure_ai_search/filters.py`, `_in()` (around line 91)
 - Reproduced against `main` at `39ba76eb3`

chakshu-dhannawat:
Fixes #3844

Azure's `search.in(field, valueList, delimiter)` splits `valueList` on the supplied delimiter before comparing. The previous implementation always joined values with a comma and passed `,` as the delimiter, so any metadata value that itself contained a comma (e.g. `"New York, NY"`) was split apart by Azure's server-side parsing and never matched.

Changes:
- `_in()` now picks a delimiter that does not occur in any of the supplied values (candidates: U+001F, U+001E, U+001D).
- If every candidate delimiter happens to be present in the values, it falls back to an OR chain of `eq` comparisons.
- Updated existing unit-test expectations and added a regression test for comma-containing values.

Verified with `hatch run test:unit`, `hatch run test:types`, and `hatch run fmt` in `integrations/azure_ai_search`.


integrations/azure_ai_search/src/haystack_integrations/document_stores/azure_ai_search/filters.py
```diff
@@ -92,8 +92,18 @@ def _in(field: str, value: Any) -> str:
...
         raise AzureAISearchDocumentStoreFilterError(msg)
-    values = ",".join(_escape_odata_literal(v) for v in value)
-    return f"search.in({field},'{values}',',')"
+
+    # Azure's search.in() splits the value-list string on the supplied delimiter.
+    # If the delimiter also appears inside a value, the value is split apart.
+    # Pick a delimiter that does not occur in any of the values, so commas (or
+    # any other character) inside metadata values are preserved. If every
+    # candidate delimiter is present, fall back to an OR chain of eq clauses.
+    candidates = ["\u001f", "\u001e", "\u001d"]
```
sjrl: Lets expand the candidate list to include common delimiters since its possible that the values don't contain the common ones
`candidates = [",", "|", ";", "\u001f", "\u001e", "\u001d"]`

integrations/azure_ai_search/src/haystack_integrations/document_stores/azure_ai_search/filters.py
```diff
@@ -92,8 +92,18 @@ def _in(field: str, value: Any) -> str:
...
+    for delimiter in candidates:
+        if all(delimiter not in v for v in value):
+            values = delimiter.join(_escape_odata_literal(v) for v in value)
+            return f"search.in({field},'{values}','{delimiter}')"
+    return " or ".join(f"{field} eq '{_escape_odata_literal(v)}'" for v in value)
```
sjrl, line 106: Lets make sure to expand the unit tests to cover this branch, currently it doesn't look like its covered.

sjrl (approved):
Thanks!

PR #3848, opened 2026-08-25
fix(tika): update tests to use Tika 3 server and pin sdk to v3

sjrl:
### Related Issues

- fixes failing tika tests

### Proposed Changes:

 
 

We pin the tika python sdk to less than 4 (a new major version), and restrict the tests to use the latest 3.X server image. 

Tika has released a new 4.X server image which is now what `:latest` points to. This is not compatible with any tika python sdk version which is why we are restricting the tika docker image to 3.X and adding api docstrings to warn users not to use the 4.X docker image. 

### How did you test it?

Existing tests

### Checklist

### DIFF
--- a/integrations/tika/pyproject.toml
+++ b/integrations/tika/pyproject.toml
@@ -23,7 +23,7 @@ classifiers = [
   "Programming Language :: Python :: Implementation :: CPython",
   "Programming Language :: Python :: Implementation :: PyPy",
 ]
-dependencies = ["haystack-ai>=2.9.0", "tika==3.1.0"]
+dependencies = ["haystack-ai>=2.9.0", "tika>=3.1.0,<4.0"]
 
 [project.urls]
 Documentation = "https://github.com/deepset-ai/haystack-core-integrations/tree/main/integrations/tika#readme"
--- a/.github/workflows/tika.yml
+++ b/.github/workflows/tika.yml
@@ -106,9 +106,8 @@ jobs:
       - name: Start Tika server
         if: runner.os == 'Linux'
         run: |
-          # Pinned to the last 3.x image: the tika client is pinned to 3.x, which supports the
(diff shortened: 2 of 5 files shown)


anakin87:
I was looking into the failing test and I found this PR, that I'd be happy to review.

I'd take a more minimal approach
- use the 3.x Docker container
- inform our users through docs
- open an issue to track the release of a version of https://github.com/chrismattmann/tika-python supporting Tika 4 server and then adapt our integration

WDYT?

sjrl:
> I was looking into the failing test and I found this PR, that I'd be happy to review.
> 
> I'd take a more minimal approach
> 
> * use the 3.x Docker container
> * inform our users through docs
> * open an issue to track the release of a version of https://github.com/chrismattmann/tika-python supporting Tika 4 server and then adapt our integration
> 
> WDYT?

Yeah that sounds good to me!

sjrl:
Opened issue for it here https://github.com/deepset-ai/haystack-core-integrations/issues/3881

sjrl:
@anakin87 this should be ready for review

integrations/tika/pyproject.toml
```diff
@@ -23,7 +23,7 @@ classifiers = [
   "Programming Language :: Python :: Implementation :: CPython",
   "Programming Language :: Python :: Implementation :: PyPy",
 ]
-dependencies = ["haystack-ai>=2.9.0", "tika==3.1.0"]
+dependencies = ["haystack-ai>=2.9.0", "tika>=3.1.0,<4.0"]
```
sjrl, line 26: I also decided to unpin the SDK but restrict it to be less than the next major version

anakin87 (approved):
👍 

I'd just change the title of this PR so that the generated changelog is correct

PR #3852, opened 2026-08-25
fix: `SupaBase` override test_get_metadata_field_unique_values_distinct_types

davidsbatista:
### Proposed Changes:

- `test_get_metadata_field_unique_values_distinct_types` failed because PostgreSQL's JSONB treats `1` and `1.0` as equal, so `pgvector`'s `SELECT DISTINCT meta->'field'` collapses them. 

- This store just subclasses `PgvectorDocumentStore` from the published `pgvector-haystack` package. Adapted the test to use one field per type (and a fractional float) instead of one shared field.

- added pgvector-haystack = false to [exclude-newer-package]. pgvector-haystack is a first-party package published just like haystack-ai which is already excluded from the 24h supply-chain safety rule

### How did you test it?

- Override integration tests, verified locally against a running Postgres/pgvector  instance.


sjrl (approved):
Looks good!

PR #3745, opened 2026-08-07
fix: `FastEmbedder` reordered the top_k fallback to happen after validation

davidsbatista:
### Related Issues

- fixes https://github.com/deepset-ai/haystack-private/issues/506

### Proposed Changes:

- reordered the top_k fallback to happen after validation

### How did you test it?

- one new test

### Checklist

- I have read the [contributors guidelines](https://github.com/deepset-ai/haystack-core-integrations/blob/main/CONTRIBUTING.md) and the [code of conduct](https://github.com/deepset-ai/haystack-core-integrations/blob/main/CODE_OF_CONDUCT.md)
- I have updated the related issue with new insights and changes
- I added unit tests and updated the docstrings
- I've used one of the [conventional commit types](https://www.conventionalcommits.org/en/v1.0.0/) for my PR title: `fix:`, `feat:`, `build:`, `chore:`, `ci:`, `docs:`, `style:`, `refactor:`, `perf:`, `test:`.

PR #3855, opened 2026-08-26
fix: respect falsy run() overrides in `SentenceTransformersSimilarityRanker`

### LINKED ISSUE #3854: Falsy run() overrides are silently ignored in SentenceTransformersSimilarityRanker
**Describe the bug**
`run()` resolves its overrides with `or`:

    top_k = top_k or self.top_k
    scale_score = scale_score or self.scale_score
    score_threshold = score_threshold or self.score_threshold

`or` falls back on any falsy value, not just `None`, so valid falsy arguments are silently replaced by the init values, although the docstring says "If set, overrides the value set at initialization":

- `run(scale_score=False)` is ignored: with the default `scale_score=True`, `False or True` is `True`
- `run(score_threshold=0.0)` is replaced by the init value, or by no filtering at all
- `run(top_k=0)` silently uses the init `top_k` instead of raising the documented ValueError

**To Reproduce**

    ranker = SentenceTransformersSimilarityRanker()   # scale_score=True by default
    ranker.run(query=query, documents=docs, scale_score=False)
    # the cross encoder still runs with the Sigmoid activation

 - OS: Ubuntu 24.04 LTS
 - Haystack version: 3.0.0

**Additional context**
#3745 fixed the same `or` pattern for `top_k` in the fastembed rankers.
Happy to open a PR with the same fix and regression tests.

keosung:
### Related Issues

- fixes #3854

### Proposed Changes:

`run()` resolved its overrides with `or`, which falls back on any falsy value, not just `None`. So `run(scale_score=False)` never disabled scaling, `run(score_threshold=0.0)` was replaced by the init value, and `run(top_k=0)` silently used the init `top_k` instead of raising the documented ValueError. This validates `top_k` before the fallback and switches the three fallbacks to `None` checks, the same change #3745 and #3746 made for the fastembed rankers. Behavior now matches the run() docstring, which already says "If set, overrides the value set at initialization."

### How did you test it?

Added 3 regression tests, one per parameter. They fail on main and pass with the fix. Also verified with a real cross-encoder model that `run(scale_score=False)` now returns raw logits instead of sigmoid-scaled scores, and that the unit and integration suites both pass (23 + 3 tests).

### Notes for the reviewer

Behavior change: `run(top_k=0)` now raises instead of silently using the init value, same as the fastembed rankers fix.

### Checklist

### DIFF
--- a/integrations/sentence_transformers/src/haystack_integrations/components/rankers/sentence_transformers/sentence_transformers_similarity.py
+++ b/integrations/sentence_transformers/src/haystack_integrations/components/rankers/sentence_transformers/sentence_transformers_similarity.py
@@ -250,13 +250,13 @@ class SentenceTransformersSimilarityRanker:
         if not documents:
             return {"documents": []}
 
-        top_k = top_k or self.top_k
-        scale_score = scale_score or self.scale_score
-        score_threshold = score_threshold or self.score_threshold
-
-        if top_k <= 0:
+        if top_k is not None and top_k <= 0:
             msg = f"top_k must be > 0, but got {top_k}"
             raise ValueError(msg)
+
+        top_k = self.top_k if top_k is None else top_k
+        scale_score = self.scale_score if scale_score is None else scale_score
+        score_threshold = self.score_threshold if score_threshold is None else score_threshold
 
         deduplicated_documents = _deduplicate_documents(documents)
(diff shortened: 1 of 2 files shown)


bogdankostic (approved):
Looks good, thanks for the PR @keosung!

PR #3860, opened 2026-08-26
feat: add Apache Solr integration

julian-risch:
### Related Issues

- Related to #3863
- Closes https://github.com/deepset-ai/haystack-private/issues/246

### Proposed Changes:

Adds `solr-haystack`, an [Apache Solr](https://solr.apache.org/) integration. Solr has shipped
`DenseVectorField` and the `{!knn}` query parser since 9.0, so it can serve both halves of a RAG
pipeline

- `SolrDocumentStore`
- `SolrBM25Retriever`, `SolrEmbeddingRetriever`, `SolrHybridRetriever`
- Scaffolded with `scripts/create_new_integration.py`, so `.github/labeler.yml`,
  `CI_coverage_comment.yml` and `README.md` are added/updated

### How did you test it?

- New unit tests
- Integration tests inherit `DocumentStoreBaseExtendedTests`, `CountDocumentsByFilterTest`,
  `CountUniqueMetadataByFilterTest`, `GetMetadataFieldsInfoTest`, `GetMetadataFieldMinMaxTest` and
  `GetMetadataFieldUniqueValuesTest`, plus async.
- Verified against **Solr 10.0.0** and **Solr 9.10.1**
- Each test gets its own core, which is what makes `-n 4` safe.

### Notes for the reviewer

Design decisions to check in the review:

- **`httpx` rather than a Solr client.** `pysolr` (the maintained one) has no async support at all,
  `aiosolr` has no delete method, and `solrpy` is 4-Beta with no license file. Solr's surface here is
  small and entirely JSON, so one `httpx.Client`/`AsyncClient` pair gives sync and async from a
  single code path.
- **Metadata field names encode the value's Python type** (`meta_s_page` for `"100"`, `meta_l_page`
  for `100`). Solr fixes a field's type when the field is created, while Haystack metadata types are
  only known at write time. This keeps the round-trip exact, and keeps values that merely share a
  string form — the int `1`, the str `"1"`, the float `1.0` and `True` — as four distinct values.
  The type code is a *prefix* because Solr dynamic fields only allow a leading or trailing wildcard,
  so `meta_*_s` is not a legal pattern.
- `field:-10` is a **Lucene syntax error**, and `field:(-10 OR 100)` silently parses as "100 but not
  10". Negative numerics are therefore escaped in `filters.py`, with regression tests covering equality, ranges and grouped `in` clauses.

### Checklist

- [ ] I have updated the related issue with new insights and changes — n/a, no linked issue yet


integrations/solr/tests/conftest.py
```diff
@@ -0,0 +1,96 @@
...
+from haystack_integrations.document_stores.solr import SolrDocumentStore
+
+SOLR_URL = os.environ.get("SOLR_URL", "http://localhost:8983/solr")
+
+#: Dimension used by `haystack.testing.document_store.create_filterable_docs`.
+#:
+#: Every test must use this one value. Test cores are created from the shared `_default` configset, and
+#: such cores reference the configset's managed schema instead of owning one, so the `embedding` field
+#: is shared across all of them. A store built with a different dimension would write a conflicting
+#: field type into that shared schema and every later test would be refused.
```
julian-risch:
```suggestion
```

integrations/solr/tests/test_solr_hybrid_retriever.py
```diff
@@ -0,0 +1,200 @@
...
+        assert retriever.pipeline.get_component("bm25_retriever")._fuzziness == 2
+        assert retriever.pipeline.get_component("embedding_retriever")._top_k == 7
+
+
+class TestSerialization:
+    @pytest.fixture(autouse=True)
+    def openai_api_key(self, monkeypatch):
+        # These tests build a real OpenAITextEmbedder, which resolves its key at init. A component
+        # defined in the test module itself cannot be used here: Haystack only deserializes classes
+        # from allowlisted modules, and `tests.*` is not one of them.
```
julian-risch:
```suggestion
```

davidsbatista:
Overall looks good to me, just some minor adjustments: 

### Core sync methods
  - write_documents docstring doesn't state that DuplicatePolicy.NONE resolves to FAIL
   
### Retriever class
  - No top_k <= 0 validation  (only 3 of ~20+ retriever integrations do this repo-wide; worth adding already since it's a new integration)
 
### Tests
  - No test that an invalid filter_policy string raises ValueError

### Documentation
  - SolrEmbeddingRetriever docstring example uses a raw embedding list rather than the DocumentEmbedder/TextEmbedder pipeline pattern

### Extended methods 
  - `delete_by_filter`/`delete_by_filter_async` count matching documents then issue a separate delete-by-query — a concurrent write between the two calls could make the returned count stale; not documented as a caveat
  
  
I'm OK with being merged as is, but since it's a new integration lets raise the bar, some of those should also be applied to existing doc stores

davidsbatista (approved):
looks good, just left some comments that can improve the quality overall

julian-risch:
> looks good, just left some comments that can improve the quality overall

Thanks for the feedback. All five points addressed. Merging and releasing now! 🚀

PR #3861, opened 2026-08-27
feat: add Synthorai integration

### LINKED ISSUE #3853: New Integration Proposal: Synthorai
## Summary and motivation

Requesting a new Haystack integration for Synthorai (https://synthorai.io), an OpenAI/Anthropic-compatible LLM gateway routing to 113 models across 11 upstream providers (Claude, GPT, Gemini, GLM, Kimi, DeepSeek, Qwen, etc.) at direct upstream pricing, no markup. Docs: https://synthorai.io/docs. Use case: one API key/base URL for both Claude-family and GPT-family models plus a longer tail of Chinese-origin models, without juggling multiple provider SDKs in a Haystack pipeline.

## Adoption signals

Being upfront that Synthorai is early-stage, so these numbers are modest right now — flagging this as a heads-up request rather than expecting immediate prioritization, in case it's useful to have on record:

- **GitHub stars**: N/A (Synthorai itself isn't an open-source project; it's a hosted gateway)
- **PyPI downloads**: N/A, no Python SDK published yet — plain HTTP/OpenAI-SDK-compatible
- **Release activity**: Live production service, continuously updated
- **Maintenance**: Actively developed and maintained
- **Haystack community demand**: None yet that I'm aware of — this is a proactive request, not in response to existing community asks
- **Anything else**: Comparable integrations already exist for similar gateways (OpenRouter is already integrated at #1723; Requesty's attempt at #3581 was closed only for a missing CLA signature, not a rejection of the concept)

## Detailed design

Following the OpenRouter integration (#1723) as a template: a new `integrations/synthorai/` package with a `SynthoraiChatGenerator` wrapping the existing OpenAI-compatible generator machinery, defaulting `api_base_url` to `https://synthorai.io/v1` and reading `SYNTHORAI_API_KEY`. Plain OpenAI-compatible surface, no custom request/response transformation needed.

I'm not attempting the PR myself yet given the adoption-signal bar described above — happy to if the team thinks it's worth building despite the early-stage numbers, and can use the scaffolding script per CONTRIBUTING.md at that point.

cuihuan:
Adds a Synthorai integration. Closes #3853.

**Disclosure: I'm affiliated with Synthorai** — vendor adding its own integration, not a user request. Not claiming demand; nobody else has asked for this as far as I know.

### How it was built

Generated with `python scripts/create_new_integration.py --name synthorai --type generators`, exactly as CONTRIBUTING describes, then filled in following **`integrations/cometapi`** — the closest precedent, since that's also an OpenAI-compatible gateway wrapping `OpenAIChatGenerator`. The file set matches it one for one (no `CHANGELOG.md`, which is generated on release).

### What Synthorai is

An LLM gateway serving Anthropic, OpenAI, Google, DeepSeek, Qwen, Moonshot and Z.ai models behind one endpoint and one key. One thing worth knowing for this integration: **model ids are bare, not vendor-prefixed** — `claude-opus-5`, not `anthropic/claude-opus-5` — which differs from most gateways, so the docstring says so explicitly.

### One deliberate difference from the scaffold's default

`api_base_url` is **pinned**, not exposed as a parameter. Every model is served from the same host, so a configurable endpoint would have no legitimate use and would only be a way to send the API key elsewhere. `to_dict` therefore doesn't serialise it either.

### Two things the scaffold leaves to the contributor, both done

- `pydoc/config_docusaurus.yml` pointed at `...synthorai.generator`, a module that doesn't exist here. The real path is `...synthorai.chat.chat_generator`, matching cometapi.
- `keywords` was empty; filled in.

Dependency floor is `haystack-ai>=2.23.0`, same as cometapi.

### What I could not verify

**I have not run the test suite.** `pytest` isn't installed on this machine and the Haystack dependency tree isn't present, so `py_compile` is as far as I got — I'd rather say that than paste a green run I didn't produce. CI is the authority; if anything reds I'll fix it against the actual output.

The live test is `@pytest.mark.integration` and skips without `SYNTHORAI_API_KEY`, following the same pattern as the other gateway integrations. Per the scaffold's step 5, that secret would need adding on your side if you want that test to run in CI — happy either way.


julian-risch:
https://github.com/deepset-ai/haystack-core-integrations/issues/3853#issuecomment-5451469502 and CLA are blocking this. Nevertheless, thank you for your efforts and for opening this PR!

PR #3862, opened 2026-08-27
feat(azure_ai_search): improve handling of unknown metadata fields

bogdankostic:
### Related Issues

- related to https://github.com/deepset-ai/haystack-private/issues/501
- fixes the failing nightly: https://github.com/deepset-ai/haystack-core-integrations/actions/runs/33029891446

### Proposed Changes:

- `haystack-ai` 3.1.0 grew `GetMetadataFieldUniqueValuesTest` from 1 test to 9. `azure_ai_search` inherited the 8 new tests unguarded and the nightly has been red since 2026-08-25 with `ValueError: Fields ['category'] are not defined in index schema`. 

- Azure AI Search fixes the index schema at index creation time, so metadata fields have to be declared up front via `metadata_fields=`. Added overrides that parametrize the `document_store` fixture with the fields each Mixin test needs.

- `test_get_metadata_field_unique_values_distinct_types` expects int, float, str and bool sharing the same metadata field name to all come back as distinct. Azure AI Search binds every index field to a single `Edm.*` type, so a field declared `Edm.Int32` rejects the str/float/bool writes outright. Adapted the test to use one metadata field per type instead of one shared field.

- Unknown field no longer raises: `get_metadata_field_unique_values()` used to raise `ValueError` for a field absent from the index schema, while the Mixin's `_missing_field` test expects `([], 0)`. 

- Removed duplicated tests, now relying on the Mixin: `test_get_metadata_field_unique_values_integration` duplicated `_search_term` and was deleted, and the hand-written `_with_filters` — which was silently shadowing the new Mixin test of the same name — is now a delegating override.

### How did you test it?

- `hatch run fmt-check`, `hatch run test:types`, `hatch run test:unit` — green, 94 unit tests, one added for the new unknown-field behaviour
- `hatch run test:integration` against a live Azure AI Search index

### Checklist

### DIFF
--- a/integrations/azure_ai_search/src/haystack_integrations/document_stores/azure_ai_search/document_store.py
+++ b/integrations/azure_ai_search/src/haystack_integrations/document_stores/azure_ai_search/document_store.py
@@ -421,14 +421,20 @@ class AzureAISearchDocumentStore:
         index_fields = self._index_client.get_index(self._index_name).fields
         return {field.name: field for field in index_fields}
 
-    def _validate_index_fields(self, field_names: list[str]) -> None:
+    def _is_index_field(self, field_name: str) -> bool:
         """
-        Validates that all provided field names exist in the index schema.
+        Returns whether the given field name is defined in the index schema.
         """
         if not self._index_fields:
             self._index_fields = list(self._get_index_schema_fields().keys())
 
-        missing_fields = [field for field in field_names if field not in self._index_fields]
+        return field_name in self._index_fields
+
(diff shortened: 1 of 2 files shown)


julian-risch (approved):
LGTM!

`get_metadata_field_min_max` and `count_unique_metadata_by_filter` still call `_validate_index_fields` and raise ValueError for an undeclared field. I believe your fix of `get_metadata_field_unique_values` could be extended to those methods too but not a blocker for this PR.

PR #3864, opened 2026-08-27
feat(anthropic): add AnthropicTokenCounter

### LINKED ISSUE #3856: feat: Add `AnthropicTokenCounter` to anthropic integration
We recently added `TokenCounters` to Haystack 3.1 in PR https://github.com/deepset-ai/haystack/pull/12195

Token Counter Docs: https://docs.haystack.deepset.ai/docs/token-counters

We should also add a `AnthropicTokenCounter` to our anthropic integration using their token counting endpoint https://platform.claude.com/docs/en/build-with-claude/token-counting

We can follow the implementation of the `OpenAITokenCounter` https://github.com/deepset-ai/haystack/blob/main/haystack/token_counters/openai_counter.py as inspiration

SyedShahmeerAli12:
Closes #3856

## What

Adds `AnthropicTokenCounter` to the Anthropic integration, implementing the `TokenCounter` protocol introduced in Haystack 3.1.

The counter calls Anthropic's `POST /v1/messages/count_tokens` endpoint to return an exact input token count without generating a response or incurring generation costs.

```python
from haystack_integrations.components.generators.anthropic import AnthropicTokenCounter

counter = AnthropicTokenCounter(model="claude-sonnet-4-5")
messages = [
    ChatMessage.from_system("You are a helpful assistant."),
    ChatMessage.from_user("How many tokens is this?"),
]
print(counter.count(messages))  # e.g. 21
```

## How

- Reuses the existing `_convert_messages_to_anthropic_format()` utility from `chat/utils.py` to convert Haystack `ChatMessage` objects to Anthropic API format
- Converts Haystack `Tool` objects to `ToolParam` for tool-aware token counting
- Follows the same init/`to_dict`/`from_dict` pattern as `AnthropicChatGenerator`
- Exported from the top-level `__init__.py` alongside the existing generators

## Tests

14 tests added in `tests/test_token_counter.py`:

- **11 unit tests** (mocked): init params, serde roundtrip, count with/without system messages, count with tools, count without tools, multi-turn messages
- **3 integration tests**: verified against the **real Anthropic API** (`POST /v1/messages/count_tokens`) using a Anthropic API key all passed

### DIFF
--- a/integrations/anthropic/src/haystack_integrations/components/generators/anthropic/__init__.py
+++ b/integrations/anthropic/src/haystack_integrations/components/generators/anthropic/__init__.py
@@ -5,12 +5,10 @@ from .chat.chat_generator import AnthropicChatGenerator
 from .chat.foundry_chat_generator import AnthropicFoundryChatGenerator
 from .chat.vertex_chat_generator import AnthropicVertexChatGenerator
 from .generator import AnthropicGenerator
-from .token_counter import AnthropicTokenCounter
 
 __all__ = [
     "AnthropicChatGenerator",
     "AnthropicFoundryChatGenerator",
     "AnthropicGenerator",
-    "AnthropicTokenCounter",
     "AnthropicVertexChatGenerator",
 ]
--- a/integrations/anthropic/src/haystack_integrations/components/generators/anthropic/__init__.py
+++ b/integrations/anthropic/src/haystack_integrations/components/generators/anthropic/__init__.py
@@ -5,10 +5,12 @@ from .chat.chat_generator import AnthropicChatGenerator
 from .chat.foundry_chat_generator import AnthropicFoundryChatGenerator
 from .chat.vertex_chat_generator import AnthropicVertexChatGenerator
 from .generator import AnthropicGenerator
+from .token_counter import AnthropicTokenCounter
 
 __all__ = [
     "AnthropicChatGenerator",
     "AnthropicFoundryChatGenerator",
     "AnthropicGenerator",
+    "AnthropicTokenCounter",
     "AnthropicVertexChatGenerator",
 ]
--- a/integrations/anthropic/src/haystack_integrations/components/generators/anthropic/chat/utils.py
+++ b/integrations/anthropic/src/haystack_integrations/components/generators/anthropic/chat/utils.py
@@ -226,7 +226,7 @@ def _convert_tool_calls_to_anthropic_format(tool_calls: list[ToolCall]) -> list[
     return anthropic_tool_calls
 
 
-def convert_messages_to_anthropic_format(
+def _convert_messages_to_anthropic_format(
     messages: list[ChatMessage],
 ) -> tuple[list[TextBlockParam], list[MessageParam]]:
     """
--- a/integrations/anthropic/src/haystack_integrations/components/generators/anthropic/chat/utils.py
+++ b/integrations/anthropic/src/haystack_integrations/components/generators/anthropic/chat/utils.py
@@ -226,7 +226,7 @@ def _convert_tool_calls_to_anthropic_format(tool_calls: list[ToolCall]) -> list[
     return anthropic_tool_calls
 
 
-def _convert_messages_to_anthropic_format(
+def convert_messages_to_anthropic_format(
     messages: list[ChatMessage],
 ) -> tuple[list[TextBlockParam], list[MessageParam]]:
     """
--- a/integrations/anthropic/src/haystack_integrations/components/generators/anthropic/token_counter.py
+++ b/integrations/anthropic/src/haystack_integrations/components/generators/anthropic/token_counter.py
@@ -1,0 +1,122 @@
+# SPDX-FileCopyrightText: 2023-present deepset GmbH <info@deepset.ai>
+#
+# SPDX-License-Identifier: Apache-2.0
+
+from typing import Any
+
+from haystack import default_from_dict, default_to_dict
+from haystack.dataclasses import ChatMessage
+from haystack.tools import Tool, flatten_tools_or_toolsets
+from haystack.utils.auth import Secret
+
+from anthropic import Anthropic
+from anthropic.types import ToolParam
+
+from .chat.utils import _convert_messages_to_anthropic_format
+
+
+class AnthropicTokenCounter:
+    """
+    Counts input tokens for Anthropic models using the Anthropic token counting API.
+
+    Uses the `POST /v1/messages/count_tokens` endpoint, which returns an exact token
+    count without generating a response or incurring generation costs.
+
+    Usage example:
+    ```python
+    from haystack.dataclasses import ChatMessage
+    from haystack_integrations.components.generators.anthropic import AnthropicTokenCounter
+
+    counter = AnthropicTokenCounter(model="claude-sonnet-4-5")
+    messages = [
+        ChatMessage.from_system("You are a helpful assistant."),
+        ChatMessage.from_user("How many tokens is this?"),
+    ]
+    token_count = counter.count(messages)
+    print(token_count)
+    ```
+    """
+
+    def __init__(
+        self,
+        model: str = "claude-sonnet-4-5",
+        api_key: Secret = Secret.from_env_var("ANTHROPIC_API_KEY"),  # noqa: B008
+        *,
+        timeout: float | None = None,
+        max_retries: int | None = None,
+    ) -> None:
+        """
+        Create an AnthropicTokenCounter.
+
+        :param model: The Anthropic model to use for tokenization. Token counts are
+            model-specific; always count against the model you intend to use.
+        :param api_key: The Anthropic API key. Defaults to the ``ANTHROPIC_API_KEY``
+            environment variable.
+        :param timeout: HTTP timeout in seconds for the Anthropic client.
+        :param max_retries: Maximum number of retries for failed requests.
+        """
+        self.model = model
+        self.api_key = api_key
+        self.timeout = timeout
+        self.max_retries = max_retries
+
+        client_kwargs: dict[str, Any] = {"api_key": api_key.resolve_value()}
+        if timeout is not None:
+            client_kwargs["timeout"] = timeout
+        if max_retries is not None:
+            client_kwargs["max_retries"] = max_retries
+
+        self._client = Anthropic(**client_kwargs)
+
+    def count(self, messages: list[ChatMessage], tools: list[Tool] | None = None) -> int:
+        """
+        Count the tokens for the given messages and optional tools.
+
+        :param messages: The list of ChatMessages to count tokens for.
+        :param tools: Optional list of Tools whose schemas are included in the count.
+        :returns: The number of input tokens.
+        """
+        system_messages, non_system_messages = _convert_messages_to_anthropic_format(messages)
+
+        kwargs: dict[str, Any] = {
+            "model": self.model,
+            "messages": non_system_messages,
+        }
+        if system_messages:
+            kwargs["system"] = system_messages
+        if tools:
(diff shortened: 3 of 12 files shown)


integrations/anthropic/src/haystack_integrations/components/generators/anthropic/token_counter.py
```diff
@@ -0,0 +1,119 @@
...
+        *,
+        timeout: float | None = None,
+        max_retries: int | None = None,
+    ) -> None:
+        """
+        Create an AnthropicTokenCounter.
+
+        :param model: The Anthropic model to use for tokenization. Token counts are
+            model-specific; always count against the model you intend to use.
+        :param api_key: The Anthropic API key. Defaults to the ``ANTHROPIC_API_KEY``
```
sjrl: In our api docstrings we only use single backticks
```suggestion
        :param api_key: The Anthropic API key. Defaults to the `ANTHROPIC_API_KEY`
```

SyedShahmeerAli12:
Implemented `AnthropicTokenCounter `following the `OpenAITokenCounter `pattern, reusing the existing message conversion utilities. Tested against the **real Anthropic token counting API**.

integrations/anthropic/src/haystack_integrations/components/generators/anthropic/token_counter.py
```diff
@@ -0,0 +1,119 @@
...
+
+        client_kwargs: dict[str, Any] = {"api_key": api_key.resolve_value()}
+        if timeout is not None:
+            client_kwargs["timeout"] = timeout
+        if max_retries is not None:
+            client_kwargs["max_retries"] = max_retries
+
+        self._client = Anthropic(**client_kwargs)
+
+    def count(self, messages: list[ChatMessage], tools: list[Tool] | None = None) -> int:
```
sjrl: The protocol requires the tools to be 
```suggestion
    def count(self, messages: list[ChatMessage], tools: ToolsType | None = None) -> int:
```
so lets update it to that type. It looks like that should be fine since you are already using `flatten_tools_or_toolsets` in the body of this function.

integrations/anthropic/src/haystack_integrations/components/generators/anthropic/__init__.py
sjrl, line 1: This should be in a different folder `haystack_integrations/token_counters/anthropic`

integrations/anthropic/src/haystack_integrations/token_counters/__init__.py
sjrl, line 1: We need to remove this `__init__.py` file we only need the init file in the `anthropic` folder.

Instead of this file create a `py.typed` file in `token_counters` folder.

integrations/anthropic/src/haystack_integrations/token_counters/anthropic/token_counter.py
sjrl, line 1: Lets make sure to update the `pydoc/config_docusaurus.yml` file to include the new token counter

sjrl: Also please update the the `types` entry in `pyproject.toml` to include the new path `haystack_integrations.token_counters.anthropic` to ensure that the token counter file is type checked

integrations/anthropic/tests/test_token_counter.py
```diff
@@ -0,0 +1,180 @@
...
+    def test_custom_init(self, monkeypatch):
+        monkeypatch.setenv("ANTHROPIC_API_KEY", "test-api-key")
+        c = AnthropicTokenCounter(model="claude-opus-4-5", timeout=30.0, max_retries=3)
+        assert c.model == "claude-opus-4-5"
+        assert c.timeout == 30.0
+        assert c.max_retries == 3
+
+    def test_explicit_api_key(self):
+        c = AnthropicTokenCounter(api_key=Secret.from_token("sk-test"))
+        assert c.model == "claude-sonnet-4-5"
```
sjrl: I don't think we need this test, we can go ahead and drop it.

integrations/anthropic/tests/test_token_counter.py
```diff
@@ -0,0 +1,180 @@
...
+        assert restored.timeout == c.timeout
+        assert restored.max_retries == c.max_retries
+
+    def test_roundtrip(self, monkeypatch):
+        monkeypatch.setenv("ANTHROPIC_API_KEY", "test-api-key")
+        c = AnthropicTokenCounter(model="claude-haiku-4-5-20251001", timeout=5.0, max_retries=1)
+        restored = AnthropicTokenCounter.from_dict(c.to_dict())
+        assert restored.model == c.model
+        assert restored.timeout == c.timeout
+        assert restored.max_retries == c.max_retries
```
sjrl: These two tests are the same, lets drop one of them.

integrations/anthropic/tests/test_token_counter.py
```diff
@@ -0,0 +1,180 @@
...
+        mock_response = MagicMock()
+        mock_response.input_tokens = 20
+
+        with patch.object(counter._client.messages, "count_tokens", return_value=mock_response) as mock_call:
+            messages = [ChatMessage.from_user("No tools here.")]
+            result = counter.count(messages)
+
+        assert result == 20
+        call_kwargs = mock_call.call_args.kwargs
+        assert "tools" not in call_kwargs
```
sjrl: Lets drop this test, its already covered by `test_count_basic_messages` and `test_count_user_only_messages`

integrations/anthropic/tests/test_token_counter.py
```diff
@@ -0,0 +1,180 @@
...
+
+    def test_count_multi_turn(self, counter):
```
sjrl: I think we can also drop this test, its already covered by earlier tests.

integrations/anthropic/tests/test_token_counter.py
```diff
@@ -0,0 +1,180 @@
...
+            result = counter.count(messages)
+
+        assert result == 75
+        call_kwargs = mock_call.call_args.kwargs
+        assert len(call_kwargs["messages"]) == 3
+
+
+@pytest.mark.skipif(not os.environ.get("ANTHROPIC_API_KEY"), reason="ANTHROPIC_API_KEY not set")
+@pytest.mark.integration
+class TestAnthropicTokenCounterIntegration:
```
sjrl, line 166: Lets just make one integration test to reduce the number of times we need to make requests to their api. 

So lets just do one test that includes all message types, `system`, `user`, `assistant`, `tool result` and `tools`. 

See the OpenAI integration test https://github.com/deepset-ai/haystack/blob/82c1f14d42ca7b9c461c05e8438d43ab2be4e7d9/test/token_counters/test_openai_counter.py#L133 as an example you can copy.

SyedShahmeerAli12:
Also addressed two issues I caught during self-review:
- Made `_convert_messages_to_anthropic_format` public and updated all callers  importing a private function across package boundaries is fragile
- Added `if not messages: return 0` guard empty input would cause a 400 from the API

integrations/anthropic/src/haystack_integrations/components/generators/anthropic/chat/utils.py
```diff
@@ -226,7 +226,7 @@ def _convert_tool_calls_to_anthropic_format(tool_calls: list[ToolCall]) -> list[
     return anthropic_tool_calls
 
 
-def _convert_messages_to_anthropic_format(
```
sjrl, line 229: Please undo this change, despite what Claude says its fine to use private methods within the integration since we control all source code. If someone changes this method then tests will catch that the Token counter wasn't updated.

integrations/anthropic/src/haystack_integrations/token_counters/anthropic/token_counter.py
```diff
@@ -0,0 +1,121 @@
...
+    print(token_count)
+    ```
+    """
+
+    def __init__(
+        self,
+        model: str = "claude-sonnet-4-5",
+        api_key: Secret = Secret.from_env_var("ANTHROPIC_API_KEY"),  # noqa: B008
+        *,
+        timeout: float | None = None,
```
sjrl, line 44: Lets make the signature follow the openai token counter. Don't assume the model, it's important this is correctly specified and make everything after model a keyword only arg. 

```suggestion
        self,
        model: str,
        *,
        api_key: Secret = Secret.from_env_var("ANTHROPIC_API_KEY"),  # noqa: B008
        timeout: float | None = None,
```

integrations/anthropic/src/haystack_integrations/token_counters/anthropic/token_counter.py
```diff
@@ -0,0 +1,121 @@
...
+        self.timeout = timeout
+        self.max_retries = max_retries
+
+        client_kwargs: dict[str, Any] = {"api_key": api_key.resolve_value()}
+        if timeout is not None:
+            client_kwargs["timeout"] = timeout
+        if max_retries is not None:
+            client_kwargs["max_retries"] = max_retries
+
+        self._client = Anthropic(**client_kwargs)
```
sjrl: Please look at the OpenAITokenCounter implementation. The client initialization should be under a `warm_up` method https://github.com/deepset-ai/haystack/blob/30d7506d0b692dd1ac67eae62a77f50fdc9bd4db/haystack/token_counters/openai_counter.py#L72

Then we should make sure to auto-warmup the component when `count` is called https://github.com/deepset-ai/haystack/blob/30d7506d0b692dd1ac67eae62a77f50fdc9bd4db/haystack/token_counters/openai_counter.py#L104-L107

integrations/anthropic/src/haystack_integrations/token_counters/anthropic/token_counter.py
```diff
@@ -0,0 +1,121 @@
...
+    def count(self, messages: list[ChatMessage], tools: ToolsType | None = None) -> int:
+        """
+        Count the tokens for the given messages and optional tools.
+
+        :param messages: The list of ChatMessages to count tokens for.
+        :param tools: Optional list of Tools whose schemas are included in the count.
+        :returns: The number of input tokens.
+        """
+        if not messages:
+            return 0
```
sjrl: This should be a check against both messages and tools. It's possible that a user passed in only tools and no messages. OpenAITokenCounter example https://github.com/deepset-ai/haystack/blob/30d7506d0b692dd1ac67eae62a77f50fdc9bd4db/haystack/token_counters/openai_counter.py#L101-L102

SyedShahmeerAli12:
Updated to follow the `OpenAITokenCounter ` pattern with lazy `warm_up()` and auto-warmup in `count()`. Does this match what you had in mind?

sjrl:
> Updated to follow the `OpenAITokenCounter ` pattern with lazy `warm_up()` and auto-warmup in `count()`. Does this match what you had in mind?

Yes that looks right! One more thing I forgot to mention please also add the `close()` lifecycle method e.g. https://github.com/deepset-ai/haystack/blob/30d7506d0b692dd1ac67eae62a77f50fdc9bd4db/haystack/token_counters/openai_counter.py#L122-L126

SyedShahmeerAli12:
> > Updated to follow the `OpenAITokenCounter ` pattern with lazy `warm_up()` and auto-warmup in `count()`. Does this match what you had in mind?
> 
> Yes that looks right! One more thing I forgot to mention please also add the `close()` lifecycle method e.g. https://github.com/deepset-ai/haystack/blob/30d7506d0b692dd1ac67eae62a77f50fdc9bd4db/haystack/token_counters/openai_counter.py#L122-L126

Added `close()` following the same pattern as `OpenAITokenCounter`. Also added two unit tests covering the normal close and close-before-warm_up case.

sjrl (approved):
Thanks!

PR #3868, opened 2026-08-28
feat: add GoogleGenAITokenCounter to google-genai integration

### LINKED ISSUE #3857: feat: Add `GoogleGenAITokenCounter` to google-genai integration
We recently added `TokenCounters` to Haystack 3.1 in PR https://github.com/deepset-ai/haystack/pull/12195

Token Counter Docs: https://docs.haystack.deepset.ai/docs/token-counters

We should also add a `GoogleGenAITokenCounter` to our google-genai integration using their token counting endpoint https://ai.google.dev/api/tokens#method:-models.counttokens through their python SDK

We can follow the implementation of the `OpenAITokenCounter` https://github.com/deepset-ai/haystack/blob/main/haystack/token_counters/openai_counter.py as inspiration

pma1999:
### Related Issues

- fixes #3857

### Proposed Changes:

Adds `GoogleGenAITokenCounter` to the `google_genai` integration, implementing the `TokenCounter` protocol introduced in Haystack 3.1 on top of Google's [`countTokens`](https://ai.google.dev/api/tokens#method:-models.counttokens) endpoint — the Gemini analogue of `OpenAITokenCounter`.

It follows the `OpenAITokenCounter` pattern: `model` is required and never assumed, everything after it is keyword-only, the client is built lazily in `warm_up()`, `count()` auto-warms, and there is a `close()` lifecycle method. The constructor mirrors `GoogleGenAIChatGenerator` (`api`, `vertex_ai_project`, `vertex_ai_location`, `timeout`, `max_retries`) so a counter can be configured exactly like the generator it measures.

Inputs are assembled the way `GoogleGenAIChatGenerator` sends them: a leading system message becomes the system instruction, the remaining messages become the request contents.

```python
from haystack_integrations.token_counters.google_genai import GoogleGenAITokenCounter

counter = GoogleGenAITokenCounter("gemini-3.7-flash")
print(counter.count([ChatMessage.from_user("Hello, how are you?")]))
```

### One design decision I'd like your call on

The Google Gen AI SDK **cannot** measure a system instruction or tool schemas on the Gemini Developer API. It is not a silent drop — `_CountTokensConfig_to_mldev` raises explicitly:

```python
if getv(from_object, ['tools']) is not None:
    raise ValueError(
        'tools parameter is only supported in Gemini Enterprise Agent Platform'
        ' mode, not in Gemini Developer API mode.'
    )
```

`_CountTokensConfig_to_vertex` maps both, so Vertex is fine. Worth noting the REST API itself *could* do it via `generateContentRequest`, which accepts system instructions and function declarations — but the Python SDK does not expose that field anywhere, so there is no supported path to it today.

Since `count(messages, tools)` has to answer for both, I chose to **raise a `ValueError` naming the limitation and pointing at `api="vertex"`**, rather than return a count that silently omits tool schemas — those can be worth hundreds of tokens, and a counter that quietly under-reports seemed worse than one that refuses. Counting plain messages works on either backend.

Happy to switch to a warning plus a partial count if you'd rather the component always return a number.

### How did you test it?

- `hatch run test:unit` — 9 new unit tests pass. They cover serde round-trip, the empty-input short circuit, the `total_tokens: None` case, the rejection path for both a system message and tools, that Vertex actually forwards the system instruction and tool schemas, and `close()`.
- `hatch run test:integration` — the live test passes against the real Gemini API.
- `hatch run test:types` — mypy clean (the new package is registered in the `types` command).
- `hatch run fmt-check` — clean.

Two failures in the full unit run (`test_init_fail_wo_api_key`, `test_extract_sources_info_rejects_symlink_escaping_root`) are unrelated to this change: they reproduce identically on a clean checkout of `main` on my machine — the first because I have `GEMINI_API_KEY` exported and the test only clears `GOOGLE_API_KEY`, the second because of Windows symlink permissions.

The live test deliberately covers a user turn, an assistant tool call and a tool result, but not a system message or tools, since the Gemini Developer API these tests authenticate against cannot measure those.

### DIFF
--- a/integrations/google_genai/pydoc/config_docusaurus.yml
+++ b/integrations/google_genai/pydoc/config_docusaurus.yml
@@ -4,6 +4,7 @@ loaders:
       - haystack_integrations.components.embedders.google_genai.document_embedder
       - haystack_integrations.components.embedders.google_genai.text_embedder
       - haystack_integrations.components.embedders.google_genai.multimodal_document_embedder
+      - haystack_integrations.token_counters.google_genai.token_counter
     search_path: [../src]
 processors:
   - type: filter
--- a/integrations/google_genai/pyproject.toml
+++ b/integrations/google_genai/pyproject.toml
@@ -73,7 +73,8 @@ integration-cov-append-retry = 'pytest --cov=haystack_integrations --cov-append 
 
 types = """mypy -p haystack_integrations.components.generators.google_genai \
 -p haystack_integrations.components.embedders.google_genai \
--p haystack_integrations.common.google_genai {args}"""
+-p haystack_integrations.common.google_genai \
+-p haystack_integrations.token_counters.google_genai {args}"""
 
 [tool.mypy]
 install_types = true
--- a/integrations/google_genai/src/haystack_integrations/token_counters/google_genai/__init__.py
+++ b/integrations/google_genai/src/haystack_integrations/token_counters/google_genai/__init__.py
@@ -1,0 +1,6 @@
+# SPDX-FileCopyrightText: 2023-present deepset GmbH <info@deepset.ai>
+#
+# SPDX-License-Identifier: Apache-2.0
+from .token_counter import GoogleGenAITokenCounter
+
+__all__ = ["GoogleGenAITokenCounter"]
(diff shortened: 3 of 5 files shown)


julian-risch (approved):
Thank you for opening this PR @pma1999 and congratulations on your first PR merged into haystack-core-integrations. I made some minor changes only. Integration tests passed for me locally. 👍

PR #3873, opened 2026-08-31
fix: serialize missing init params in `to_dict` (transformers, amazon_bedrock)

pcbeingused333:
### Related Issues

- No issue. Same defect class as #3808, in two more integrations.

### Proposed Changes:

Two components accept an init parameter, store it, use it — and leave it out of `to_dict`, so it silently reverts to its default on `from_dict`.

| Component | Dropped | Already serialized by |
|---|---|---|
| `TransformersExtractiveReader` | `overlap_threshold` | — |
| `S3Downloader` | `boto3_config` | `AmazonBedrockGenerator`, `AmazonBedrockChatGenerator`, and the three Bedrock embedders |

`boto3_config` builds the botocore `Config` (timeouts, retries, proxies), so a reloaded pipeline goes back to the defaults it was configured away from. `overlap_threshold` decides which overlapping answers `run` deduplicates away, so the same document returns a different answer set after a round trip.

Each fix is one line per parameter in `to_dict`. No `from_dict` change was needed: both already accept these keys.

> **Updated after review of #3874:** this PR originally also fixed `VertexAITextEmbedder.task_type`, `progress_bar` and `truncate_dim`. `google_vertex` is an archived integration ([README table](https://github.com/deepset-ai/haystack-core-integrations#readme)), so I have dropped that commit. Only the two active integrations remain, both covered by their own CI.

### How did you test it?

Unit tests per integration, each with a `..._survives_a_serialization_round_trip` test asserting the value after `from_dict(to_dict())`, plus the existing `to_dict` expectations updated:

- `transformers` (`test_extractive_reader.py`): 35 passed, 2 deselected. 7 fail without the fix.
- `amazon_bedrock` (`test_s3_downloader.py`): 27 passed, 5 deselected. 4 fail without the fix.

Verified red by reverting only the `to_dict` change and re-running. `ruff check` and `ruff format --check` clean on both integrations.

### Notes for the reviewer

What convinced me these are oversights rather than deliberate omissions:

- `test_to_dict` in `test_s3_downloader.py` is parametrized over `boto3_config=[None, {"read_timeout": 10}]` and asserted the same dict for both — the parametrization could not fail. It now distinguishes the two.
- `from_dict` in both already accepts these keys, so this only closes the write side.

I found them with a small AST script that compares each `@component`'s `__init__` parameters against the keys reaching `default_to_dict`, after seeing this class of defect in #3808. Everything it flagged was verified by hand; false positives discarded include `ChineseDocumentSplitter.splitting_function` (serialized separately via `serialize_callable`), `LangfuseConnector.httpx_client` (excluded on purpose, with a comment), `HuggingFaceAPIChatGenerator.stop_words` (folded into `generation_kwargs` in `__init__`) and the `device` parameters in the transformers components (folded into `huggingface_pipeline_kwargs`).

One I left alone on purpose: `NvidiaGenerator` drops `timeout` the same way, but the component raises a `FutureWarning` saying it will be removed in favour of `NvidiaChatGenerator`. Happy to add it if you'd rather it were fixed while it's still there.

### Checklist

### DIFF
--- a/integrations/amazon_bedrock/src/haystack_integrations/components/downloaders/s3/s3_downloader.py
+++ b/integrations/amazon_bedrock/src/haystack_integrations/components/downloaders/s3/s3_downloader.py
@@ -295,6 +295,7 @@ class S3Downloader:
             aws_session_token=self.aws_session_token,
             aws_region_name=self.aws_region_name,
             aws_profile_name=self.aws_profile_name,
+            boto3_config=self.boto3_config,
             file_root_path=str(self.file_root_path),
             max_workers=self.max_workers,
             max_cache_size=self.max_cache_size,
--- a/integrations/amazon_bedrock/tests/test_s3_downloader.py
+++ b/integrations/amazon_bedrock/tests/test_s3_downloader.py
@@ -94,6 +94,7 @@ class TestS3Downloader:
                     "env_vars": ["AWS_PROFILE"],
                     "strict": False,
                 },
+                "boto3_config": boto3_config,
                 "file_root_path": str(tmp_path),
                 "file_extensions": None,
                 "max_cache_size": 100,
@@ -104,6 +105,14 @@ class TestS3Downloader:
             },
         }
         assert d.to_dict() == expected
+
+    def test_boto3_config_survives_a_serialization_round_trip(self, mock_boto3_session: Any, tmp_path):
+        """boto3_config builds the botocore Config: timeouts, retries and proxies all live there."""
+        downloader = S3Downloader(file_root_path=str(tmp_path), boto3_config={"read_timeout": 10, "connect_timeout": 5})
+
+        restored = S3Downloader.from_dict(downloader.to_dict())
+
+        assert restored.boto3_config == {"read_timeout": 10, "connect_timeout": 5}
 
     @pytest.mark.parametrize("boto3_config", [None, {"read_timeout": 10}])
     def test_from_dict(self, mock_boto3_session: Any, tmp_path, boto3_config: dict[str, Any] | None):
@@ -181,6 +190,7 @@ class TestS3Downloader:
                     "env_vars": ["AWS_PROFILE"],
                     "strict": False,
                 },
+                "boto3_config": None,
                 "file_root_path": str(tmp_path),
                 "file_extensions": [".txt"],
                 "max_cache_size": 400,
--- a/integrations/google_vertex/src/haystack_integrations/components/embedders/google_vertex/text_embedder.py
+++ b/integrations/google_vertex/src/haystack_integrations/components/embedders/google_vertex/text_embedder.py
@@ -118,6 +118,9 @@ class VertexAITextEmbedder:
             gcp_project_id=self.gcp_project_id.to_dict() if self.gcp_project_id else None,
             gcp_region_name=self.gcp_region_name.to_dict() if self.gcp_region_name else None,
             model=self.model,
+            task_type=self.task_type,
+            progress_bar=self.progress_bar,
+            truncate_dim=self.truncate_dim,
         )
 
     @classmethod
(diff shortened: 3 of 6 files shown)


pcbeingused333:
Note on CI coverage: `Test / transformers` and `Test / amazon-bedrock` ran and are green, but there is no `google_vertex` workflow in `.github/workflows/`, so that third of the PR is not covered by CI here. Verified locally instead — `pytest integrations/google_vertex/tests/test_text_embedder.py`: 10 passed, and 3 of them fail if I revert only the `to_dict` change:

```
FAILED tests/test_text_embedder.py::test_to_dict
FAILED tests/test_text_embedder.py::test_task_type_survives_a_serialization_round_trip
FAILED tests/test_text_embedder.py::test_progress_bar_and_truncate_dim_survive_a_serialization_round_trip
```

Happy to split google_vertex into its own PR if you would rather land the two CI-covered integrations first.

pcbeingused333:
Ran the integration tests locally as the bot asks, with `hatch`, from inside each integration folder.

**transformers** — the two integration tests in the file this PR touches pass (real models downloaded, no key needed):

```
$ cd integrations/transformers && hatch run test:integration -q -k extractive
tests/test_extractive_reader.py::test_t5 PASSED
tests/test_extractive_reader.py::test_roberta PASSED
2 passed, 136 deselected in 93.59s
```

**amazon_bedrock** — the five `S3Downloader` integration tests need a real bucket, so they skip here, with their own reason:

```
$ cd integrations/amazon_bedrock && hatch run test:integration -q -k s3_downloader -rs
5 skipped, 415 deselected
SKIPPED [5] Export an env var called `S3_DOWNLOADER_BUCKET` containing the S3 bucket to run this test.
```

`hatch run test:unit` in that integration: 352 passed, 68 deselected.

Worth saying plainly: this PR only adds keys to `to_dict`, which no integration test exercises — they all call `run()` against a live service. The unit tests are the ones that cover it, and I verified each fails without its one-line change.

**google_vertex** — while getting local evidence for the third integration I found that it has no workflow at all in `.github/workflows/`, and its suite is already red on `main` for an unrelated reason (`haystack-ai` 3.1.0 added `async_function` to `Tool`). Filed separately as #3874 so it does not get mixed up with this PR. With those two pre-existing failures excluded, `hatch run test:unit` in `google_vertex` is 64 passed, including the three that cover this change.

pcbeingused333:
@sjrl pointed out on #3874 that `google_vertex` is discontinued and archived — it is in the status table of the top-level README, which I had not checked. **I have removed that third of this PR** (`c5312c1`); a serialization fix in an archived integration is churn, not a fix.

That is the same reasoning I used to leave `NvidiaGenerator` out of this PR for being deprecated, so I should have caught it. The README table is the check I was missing, and it also retires the CI-coverage caveat I raised earlier: everything left here is covered.

**What remains, both integrations active and both green in this PR's CI:**

| Component | Dropped from `to_dict` | Already serialized by |
|---|---|---|
| `TransformersExtractiveReader` | `overlap_threshold` | — |
| `S3Downloader` | `boto3_config` | `AmazonBedrockGenerator`, `AmazonBedrockChatGenerator`, and the three Bedrock embedders |

Two one-line changes, each with a round-trip test that fails without it. Integration tests run locally and reported above: `transformers` 2 passed, `amazon_bedrock` skipped for want of a real bucket, unit suites 35 and 352 passed.

Title updated to match. Happy to split the two remaining integrations apart as well if you would rather review them separately.

pcbeingused333:
Following up — this has been quiet for ~10 days. After dropping the archived `google_vertex` portion, it's a focused `to_dict` fix for two components, each an init param that is accepted, stored and used but left out of `to_dict` so it reverts to default on `from_dict`:

- `TransformersExtractiveReader` — `overlap_threshold` (changes which overlapping answers `run` deduplicates, so the answer set differs after a round trip)
- `S3Downloader` — `boto3_config` (builds the botocore `Config`: timeouts, retries, proxies)

One line per parameter in `to_dict`, no `from_dict` change needed. Regression tests per integration, verified red by reverting only the `to_dict` change. CI is green and the branch is up to date. Ready for review whenever there's capacity.

bogdankostic (approved):

PR #3878, opened 2026-08-31
feat: support MCP Python SDK v2 - compatibility mode

### LINKED ISSUE #3687: Update the MCP integration to MCP Python SDK v2
https://py.sdk.modelcontextprotocol.io/whats-new/
https://py.sdk.modelcontextprotocol.io/migration/

My initial impression is that this is a big change, so we cannot have an integration that woks with both v1 and v2. To be confirmed.

I recommend waiting some time from now (end of July 2026) to let v2 stabilize.

---

Not open for external contributions

anakin87:
### Related Issues

- fixes https://github.com/deepset-ai/haystack-core-integrations/issues/3687

I took a look at [MCP Python SDK v2](https://py.sdk.modelcontextprotocol.io/whats-new/), that uses [2026-07-28 revision of MCP](https://blog.modelcontextprotocol.io/posts/2026-07-28/) by default.

The new revision is interesting for several reasons, but most of them require new server implementations:
- stateless (no sessions needed)
- Multi Round-Trip Requests
- List results are cacheable
- ...

After some experiments, for the moment I propose to remove the `<2` pin and add a compatibility layer. This means we allow users to use whichever recent version of the SDK (some popular libraries pin `<2`, some others `>=2`, a few are compatible with both).

Users can use servers developed with v2, but negotiating the previous version of the protocol (2025-11-25), so no new features. I'll open a new issue to track switching to v2 when it's stabilized, most of the ecosystem is compatible with it and new servers exposing the new features exist.

### Proposed Changes:
- implement a lightweight v1/v2 compatibility layer
- remove `<2` pin

### How did you test it?
CI (MCP v2); add a new CI job to test with MCP v1

### Checklist


integrations/mcp/pyproject.toml
```diff
@@ -29,10 +29,9 @@ classifiers = [
     "Programming Language :: Python :: Implementation :: CPython",
 ]
 dependencies = [
-    "mcp>=1.8.0, <2.0.0",
+    "mcp>=1.24.0",
     "haystack-ai>=2.23.0",
     "exceptiongroup",  # Backport of ExceptionGroup for Python < 3.11
-    "httpx"  # HTTP client library used for SSE connections
```
anakin87, line 35: httpx/httpx2 is already a dependency of mcp (1/2)

integrations/mcp/pyproject.toml
```diff
@@ -29,10 +29,9 @@ classifiers = [
     "Programming Language :: Python :: Implementation :: CPython",
 ]
 dependencies = [
-    "mcp>=1.8.0, <2.0.0",
+    "mcp>=1.24.0",
```
anakin87, line 32: 1.24.0 is from December 2025 and allows the use of some symbols in preview that have been introduced in v2

sjrl (approved):
Looks good! Glad we are able to support both for the time being and I agree with tracking in a new issues to support the new features at a later date.

anakin87:
Created a new issue here: https://github.com/deepset-ai/haystack-private/issues/560

PR #3892, opened 2026-09-02
fix(rankers): honor falsy run() overrides in SentenceTransformersSimilarityRanker

### LINKED ISSUE #3854: Falsy run() overrides are silently ignored in SentenceTransformersSimilarityRanker

RahilOp:
Fixes #3854

`run()` previously resolved overrides with `or`, so falsy arguments (`scale_score=False`, `score_threshold=0.0`, `top_k=0`) were silently replaced by the init values. This PR switches to explicit `None` checks.

Changes:
- Use `self.top_k if top_k is None else top_k`, etc.
- Add regression tests for `scale_score=False`, `score_threshold=0.0`, and `top_k=0`.

Verification:
- `hatch run test:unit` in `integrations/sentence_transformers` passes (165 passed, 14 deselected).
- `hatch run fmt` passes.

### DIFF
--- a/integrations/sentence_transformers/src/haystack_integrations/components/rankers/sentence_transformers/sentence_transformers_similarity.py
+++ b/integrations/sentence_transformers/src/haystack_integrations/components/rankers/sentence_transformers/sentence_transformers_similarity.py
@@ -250,9 +250,9 @@ class SentenceTransformersSimilarityRanker:
         if not documents:
             return {"documents": []}
 
-        top_k = top_k or self.top_k
-        scale_score = scale_score or self.scale_score
-        score_threshold = score_threshold or self.score_threshold
+        top_k = self.top_k if top_k is None else top_k
+        scale_score = self.scale_score if scale_score is None else scale_score
+        score_threshold = self.score_threshold if score_threshold is None else score_threshold
 
         if top_k <= 0:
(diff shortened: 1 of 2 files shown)


bogdankostic:
Thanks for the PR @RahilOp. This has ben fixed by #3855.

PR #3895, opened 2026-09-02
feat(azure_documentdb): add Azure DocumentDB integration

sandeepsnairms:
###  3901
- https://github.com/deepset-ai/haystack-core-integrations/issues/3901

## Proposed Changes

- Added a new Azure DocumentDB integration for Haystack.
- Added a document store with synchronous and asynchronous CRUD operations.
- Added embedding and keyword retrievers with metadata-filter support.
- Added passwordless Microsoft Entra ID authentication through DefaultAzureCredential.
- Added vector index creation and vector similarity search.
- Added serialization, duplicate-document policies, connection lifecycle management, and custom content/embedding fields.
- Added package metadata, documentation, examples, CI workflow, coverage registration, and repository labels.

## How did you test it?

- Ran Ruff formatting and lint checks.
- Ran mypy type checking.
- Ran the complete unit-test suite with coverage.
- Ran the live CRUD and vector-search integration test against Azure DocumentDB from Docker.
- Verified repository diagnostics and git diff --check.

## Notes for the reviewer

- Microsoft Entra ID is the recommended authentication method.
- Connection-string authentication is retained only as an explicit development and testing fallback and emits a warning.
- The live integration test accepts a short-lived Azure access token and does not persist credentials.
- The E2E test database is removed by default unless AZURE_DOCUMENTDB_KEEP_E2E_DATABASE=true is configured.
- ## Checklist

- [x]  I have read the [contributors guidelines](https://github.com/deepset-ai/haystack-core-integrations/blob/main/CONTRIBUTING.md) and the [code of conduct](https://github.com/deepset-ai/haystack-core-integrations/blob/main/CODE_OF_CONDUCT.md)
- [ ]  I have updated the related issue with new insights and changes
- [x]  I've used one of the [conventional commit types](https://www.conventionalcommits.org/en/v1.0.0/) for my PR title: feat: Add Azure DocumentDB integration


sandeepsnairms:
@masci @markmcd : Please guide with next steps.

julian-risch (changes requested):
Thank you for opening this PR @sandeepsnairms ! It looks quite good already and I noticed it mirrors our mongodb atlas integration. We need you to change a few things please before we can merge:

**1. `similarity` accepts the wrong values.** `create_vector_index` types it as `Literal["COS", "DOT", "EUC"]`, but Azure DocumentDB accepts `COS`, `L2` and `IP` ([docs](https://learn.microsoft.com/en-us/azure/documentdb/vector-search)) if I am not mistaken. `DOT`/`EUC` will be rejected by the server, and the two valid metrics are blocked by the type hint. Only `COS` is covered by tests, which is why nothing caught it.

**2. `DuplicatePolicy.FAIL` silently depends on a unique index on `id`.** `InsertOne` only raises on duplicates if the collection has `create_index("id", unique=True)`. The E2E fixture creates it by hand, but the store never creates or documents it, so a plain collection will accept duplicates under FAIL. Please either create the index in `_ensure_connection_setup` or document the requirement prominently in the class docstring.

**3. Integration tests need to run in CI.** Right now the workflow has no integration step, and `test_e2e.py` needs Azure credentials CI doesn't have, so our CI can't verify what we expect you to have tested locally before opening this PR. My understanding is that the open-source DocumentDB image (`ghcr.io/documentdb/documentdb/documentdb-local`) speaks the same protocol and supports `cosmosSearch` vector indexes, so we can run the standard Haystack mixins against it on every PR, as the other Docker-backed stores do. Can you confirm that? If so, please extend your PR:
- Add a `services:` block to the workflow starting `documentdb-local` with `--username`/`--password`, set `TEST_MATRIX_OS: '["ubuntu-latest"]'` (services need Linux, see `pgvector.yml`), and add a `hatch run test:integration-cov-append-retry` step with `AZURE_DOCUMENTDB_CONNECTION_STRING=mongodb://user:pass@localhost:10260/?tls=true&tlsAllowInvalidCertificates=true`.
- Add `TestDocumentStore(DocumentStoreBaseExtendedTests)` and `TestDocumentStoreAsync(...)` from `haystack.testing.document_store` / `document_store_async`, with a `document_store` fixture that recreates the collection and the unique `id` index, like `mongodb_atlas/tests/conftest.py`.
- Add an embedding-retrieval integration test with a filter, so the `cosmosSearch.filter` path is exercised.
- Keep `test_e2e.py` for the real Azure cluster, but make the fixture `pytest.skip` when the env vars are missing instead of raising `KeyError` (`hatch run test:all` currently crashes).
- As an alternative, is there a way where you can help us to set our CI up in way that it runs nightly integration tests against Azure DocumentDB?


**Some minor improvements:**
- Add `delete_all_documents`, `update_by_filter`, `delete_by_filter` and their `_async` versions. `mongodb_atlas` has them and `DocumentStoreBaseExtendedTests` covers them
- Full docstrings on public methods (`:param:`, `:returns:`, `:raises:`) and a usage example in the class docstring, following `MongoDBAtlasDocumentStore`. Please also document that filtered vector search requires a regular index on each filtered field (`meta.<key>`), per the Azure docs, and mention the date-string rule for `>`/`<` comparisons.
- Deduplicate the write logic: `write_documents_async` repeats the whole body of `_write_documents`. I suggest that you move the operation building into a shared helper and keep only the `bulk_write` call sync/async-specific.

sandeepsnairms:
Thanks @julian-risch for the detailed review. I addressed the requested changes:

1. **Vector similarity values**
   - Replaced the unsupported `DOT` and `EUC` values with the DocumentDB-supported `L2` and `IP` values.
   - Both synchronous and asynchronous APIs now accept `Literal["COS", "L2", "IP"]`.
   - Added coverage for all three similarity modes.

2. **Duplicate handling and the `id` index**
   - The document store now creates a unique index on `id` during synchronous and asynchronous connection setup.
   - This ensures `DuplicatePolicy.FAIL` consistently raises `DuplicateDocumentError` without requiring users to create the index manually.

3. **Credential-free integration testing**
   - Confirmed that the OSS DocumentDB image supports `cosmosSearch`, including `vector-ivf` indexes and metadata-filtered vector retrieval.
   - Added `ghcr.io/documentdb/documentdb/documentdb-local:latest` as a Linux CI service on port `10260`.
   - CI now runs `hatch run test:integration-cov-append-retry` using a local connection string, so no Azure credentials are required.
   - Added isolated collection fixtures and the standard Haystack synchronous and asynchronous document-store mixin tests.
   - Added a real filtered `cosmosSearch` integration test with a regular index on the filtered metadata field.

4. **Missing document-store operations**
   - Added synchronous and asynchronous implementations of:
     - `delete_by_filter`
     - `update_by_filter`
     - `delete_all_documents`
   - Collection recreation preserves collection options and custom indexes.
   - Refactored synchronous and asynchronous writes to share operation preparation logic.

5. **Documentation**
   - Expanded the public API docstrings, including parameters, return values, and raised exceptions.
   - Added class-level usage guidance.
   - Documented that metadata fields used in filtered vector search require regular indexes.
   - Documented the ISO-formatted string requirement for ordered date comparisons.

6. **Live Azure E2E coverage**
   - Retained the Azure OIDC E2E tests for validating the production authentication path.
   - These tests now skip cleanly when the Azure cluster name or access token is unavailable.

Local Docker validation against the OSS DocumentDB image completed successfully:

```text
82 passed, 2 skipped, 70 deselected
```
The two skipped tests are the live Azure E2E tests that require Azure credentials. The 70 deselected tests are unit tests excluded by the integration-test marker selection.

julian-risch:
@sandeepsnairms Have you run the live Azure E2E tests that require Azure credentials?

sandeepsnairms:
@julian-risch Yes, locally had run Azure E2E test when doing the PR.

julian-risch (approved):
Looks good to me! Thank you for opening this PR and congrats on your first PR merged into haystack-core-integrations @sandeepsnairms ! Next steps for this integration are a PR in [haystack-integrations repo](https://github.com/deepset-ai/haystack-integrations) that adds an integration page and a PR in [haystack repo](https://github.com/deepset-ai/haystack) that adds a documentation page per new component and one page for the document store. We'd appreciate your help with that!

PR #3916, opened 2026-09-04
fix: handle list query in SerperDevWebSearch to avoid TypeError

ArzelaAscoIi:
## Summary
- `SerperDevWebSearch._prepare_request` did `query_prepend + query`, which crashed with `TypeError: can only concatenate str (not "list") to str` whenever `query` was a `list[str]` instead of a `str` (e.g. an LLM tool call passing multiple search terms).
- Normalizes a list query into a single space-joined string before building the payload. Fixes both `run` and `run_async`, which both call `_prepare_request`.
- Related: https://linear.app/deepset/issue/AGE-45/pipeline-error-serperdevwebsearch-typeerror-concatenating-list-to-str

## Test plan
- [x] Added `test_prepare_request_list_query`
- [x] `uv run --with pytest --with pytest-asyncio pytest tests/test_websearch.py` — 9 passed


anakin87:
The signatures of the `run/run_async` methods are
`def run(self, query: str) -> dict[str, list[Document] | list[str]]:`

then `query` is passed to `_prepare_request` where at some point we do
`query_prepend + query` (`query_prepend` is a string).

Since this component only accepts `query` as a `str`, I'd not try to handle `list[str]`, which is unexpected.

I'd prefer to inspect the pipeline that produced the error (if possible) to see why it passed a list.

PR #3149, opened 2026-04-13
feat: Add Amazon S3 Vectors document store integration

dotKokott:
### Related Issues

- fixes #2110

### Proposed Changes:

Adds an [Amazon S3 Vectors](https://aws.amazon.com/s3/features/vectors/) document store integration — a serverless vector storage capability native to S3.

**Components:**
- `S3VectorsDocumentStore` — full DocumentStore protocol (write, count, filter, delete)
- `S3VectorsEmbeddingRetriever` — embedding-based retrieval with server-side metadata filtering

**Key design decisions:**
- Content stored as non-filterable metadata (AWS-recommended pattern for large text)
- Cosine distance converted to similarity score (`1 - distance`) for Haystack convention
- Blob data uses base64 encoding for round-trip fidelity
- `filter_documents()` uses `list_vectors(returnData=True, returnMetadata=True)` with client-side filtering (warning logged) since S3 Vectors has no standalone filter API
- Batch existence checks for `DuplicatePolicy.SKIP/NONE` (batches of 100)

**Known limitations (documented in README):**
- `top_k` capped at 100 (service limit)
- `query_vectors` does not return embedding data
- 40KB total metadata per vector, 2KB filterable
- Only `float32`, `cosine`/`euclidean`, eventual consistency

### How did you test it?

- 26 unit tests — serialization, score conversion, filter conversion, duplicate policy logic, document conversion (mocked boto3)
- 12 integration tests — full lifecycle against live AWS S3 Vectors, with `pytestmark` credential guard for CI
- `hatch run test:all`, `hatch run fmt`, `hatch run test:types`
- Example script (`examples/example.py`) verified against live AWS

### Notes for the reviewer

This PR was fully generated with an AI assistant. I have reviewed the changes and run the relevant tests.
Structure and test style follow the Pinecone integration pattern.

### Checklist

- [x] I have read the [contributors guidelines](https://github.com/deepset-ai/haystack-core-integrations/blob/main/CONTRIBUTING.md) and the [code of conduct](https://github.com/deepset-ai/haystack-core-integrations/blob/main/CODE_OF_CONDUCT.md)
- [x] I have updated the related issue with new insights and changes
- [x] I added unit tests and updated the docstrings
- [x] I've used one of the [conventional commit types](https://www.conventionalcommits.org/en/v1.0.0/) for my PR title: `fix:`, `feat:`, `build:`, `chore:`, `ci:`, `docs:`, `style:`, `refactor:`, `perf:`, `test:`.

PR #3918, opened 2026-09-04
feat: add Amazon S3 Vectors document store

### LINKED ISSUE #2110: Amazon S3 Vectors (DocStore)
## Summary and motivation

S3 Vectors is the latest KB solution from AWS. Although in preview, this would be a great addition to the built in doc store integrations. 

## Detailed design

The design of the docstore should include the necessary actions from the S3Vector client here https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/s3vectors.html

At minimum creating the DocStore, inserting and querying documents (Embedding Retriever only).

## Checklist

### Tasks
- [ ] A new label named like `integration:<your integration name>` has been added to the list of labels for this [repository](https://github.com/deepset-ai/haystack-core-integrations/labels)
- [ ] There is an example available to demonstrate the feature

@anakin87

oleksii-udovychenko:
Supersedes #3149. Fixes #2110.

This continues @dotKokott's work on the Amazon S3 Vectors document store. His 12 commits are preserved unchanged (rebased onto current `main`, authorship intact); the commits on top address the outstanding review feedback from @anakin87.

@dotKokott, happy to hand this back or close it if you'd rather finish it yourself.

### Review feedback addressed

| Comment | Resolution |
|---|---|
| [Check embeddings for all docs up front, not per batch](https://github.com/deepset-ai/haystack-core-integrations/pull/3149#discussion_r3128763028) | Single pass over all documents before any write |
| [The existence check before writing seems expensive; if there's no faster way, only support `DuplicatePolicy.OVERWRITE` as in Pinecone](https://github.com/deepset-ai/haystack-core-integrations/pull/3149#discussion_r3282484507) | There isn't: S3 Vectors offers only `GetVectors` (100 keys/call) or a full `ListVectors` scan, and no conditional put. So the `get_vectors` pass is gone and only `OVERWRITE` is supported; other policies are ignored with a warning |
| [Why are these warnings ignored?](https://github.com/deepset-ai/haystack-core-integrations/pull/3149#discussion_r3282407933) | Because the fixture back-filled embeddings, which tripped the store's own warning. That back-fill is gone, and so is the `simplefilter("ignore")` |
| [Inherit from `DocumentStoreBaseExtendedTests`](https://github.com/deepset-ai/haystack-core-integrations/pull/3149#discussion_r3128798133) | Added `delete_all_documents()`, `delete_by_filter()` and `update_by_filter()`, and mixed in `DeleteAllTest`, `DeleteByFilterTest`, `UpdateByFilterTest`. Also the rest of the optional surface, see below |
| [Move these tests to `test_embedding_retriever.py`](https://github.com/deepset-ai/haystack-core-integrations/pull/3149#discussion_r3282435933) | Moved; `test_integration.py` deleted |
| [Two `suggestion` blocks deleting comment banners](https://github.com/deepset-ai/haystack-core-integrations/pull/3149#discussion_r3282421495) | Applied |

### Brought up to the same coverage as `PineconeDocumentStore`

Rather than stop at `DocumentStoreBaseExtendedTests`, the store now implements the whole optional extension surface Pinecone has, and mixes in every matching test from `haystack.testing.document_store`:

- `count_documents_by_filter()`, `count_unique_metadata_by_filter()`, `get_metadata_fields_info()`, `get_metadata_field_min_max()`, `get_metadata_field_unique_values()`, covered by `CountDocumentsByFilterTest`, `CountUniqueMetadataByFilterTest`, `GetMetadataFieldsInfoTest`, `GetMetadataFieldMinMaxTest` and `GetMetadataFieldUniqueValuesTest`.
- `close()`, which releases the boto3 client's connection pool. The store stays usable, since the next call lazily builds a new client.
- The store-specific Pinecone tests too: `test_close_and_reopen`, `test_sentence_window_retriever`, and the metadata-introspection edge cases (consistent types, boolean/string min-max, list-valued fields).

Two deliberate divergences from Pinecone:

- No `TOP_K_LIMIT` truncation warnings. `list_vectors` paginates over the whole index, so these counts are exact rather than capped.
- `get_metadata_field_unique_values()` keeps values of different types distinct (the int `1`, the float `1.0`, the str `"1"` and `True`), so the base mixin's `test_get_metadata_field_unique_values_distinct_types` passes unmodified. Pinecone has to override it because it coerces numeric metadata to `float`.

`async` is **not** implemented. Pinecone's 14 `*_async` methods would need `aiobotocore` as a new runtime dependency; that felt like a separate PR rather than something to slip into this one.

### Behaviour changes

**1. `write_documents` no longer raises on a document without an embedding.** It stores a non-zero placeholder vector and warns, the same thing `PineconeDocumentStore` does with its `_dummy_vector`. Two reasons:

1. Haystack's `filterable_docs` fixture deliberately includes documents with no embedding, so a store that rejects them can't satisfy `FilterDocumentsTest` without overriding most of it.
2. The placeholder has to be non-zero: an S3 Vectors index with `cosine` distance rejects zero-norm vectors outright (`ValidationException: cosine distance does not support vectors with zero norm`).

`filter_documents()` strips the placeholder again, so such a document reads back with `embedding=None` rather than with a vector the caller never supplied. The caller's `Document` is never mutated, since the substitution happens in `_document_to_s3_vector`.

**2. Unstorable metadata values are dropped with a warning instead of failing the write.** S3 Vectors accepts a metadata value only if it is a string, a boolean, a finite number, or a non-empty array of either all strings or all non-boolean numbers. I probed the live API to pin this down exactly. All of the following are a `ValidationException` from `PutVectors`:

| Value | Service response |
|---|---|
| `None` | `Metadata values must be strings, numbers, booleans, or arrays` |
| `{"a": 1}` | same |
| `[]` | `Empty arrays are not allowed in metadata` |
| `[1, "x"]` | `Metadata array values must be strings or numbers` |
| `[True, False]` | same (booleans are fine at top level, not inside an array) |
| `[{"doc_id": ..., "range": [0, 24]}]` | same |
| `NaN` / `inf` | serialization failure |

Declaring a key non-filterable does **not** relax any of this. I checked that separately.

This is not a theoretical concern. `DocumentSplitter` with `split_overlap > 0` adds a `_split_overlap` list-of-dicts to every chunk, so before this change a plain `DocumentSplitter → S3VectorsDocumentStore` pipeline died with a `ParamValidationError`. It's what `test_sentence_window_retriever` caught. Pinecone has the same class of problem and solves it the same way (`_discard_invalid_meta`).

Two related fixes fell out of the same probe:

- `doc.blob.meta` is an arbitrary dict, so writing *any* document with blob metadata failed. It is now JSON-encoded on the way out and decoded on the way back in.
- Tuples are converted to lists. A tuple is a perfectly good array for the service, but botocore's document-type validator only accepts `list`.

### Testing

The integration job is skipped on fork PRs, so I ran everything locally against a real S3 Vectors bucket in `us-east-1`:

- **Integration suite: 89 passed, 4 skipped, 0 failed** in 602s. The skips are `test_write_documents_duplicate_fail` / `_skip` (unsupported policies) and the two `recreate_index` variants of `test_delete_all_documents`, which self-skip when the store's `delete_all_documents` has no such parameter.
- **Unit suite: 67 passed**, with `hatch run fmt-check` and `hatch run test:types` green as well.
- **End-to-end outside pytest**, with real indexing and query `Pipeline`s, real Bedrock Titan embeddings and no mocks: `DocumentSplitter → AmazonBedrockDocumentEmbedder → DocumentWriter`, `AmazonBedrockTextEmbedder → S3VectorsEmbeddingRetriever`, in-pipeline metadata filtering, and a `Pipeline.dumps()`/`loads()` round-trip. Semantic ranking came out right (0.35 for the on-topic document vs 0.01 and −0.04 for the others).
- **Probed the live `PutVectors` API** to establish exactly which metadata values it accepts, which is where the two behaviour changes above come from.

One thing I can't do myself: the workflow reuses the repo's existing `secrets.AWS_CI_ROLE_ARN` (the same secret `amazon_bedrock.yml` uses), but that role needs S3 Vectors permissions before the integration job can pass in CI. Could a maintainer add `s3vectors:CreateVectorBucket`, `CreateIndex`, `PutVectors`, `ListVectors`, `QueryVectors`, `GetVectors`, `DeleteVectors`, `DeleteIndex`, `DeleteVectorBucket`, `ListIndexes`, `GetIndex` and `GetVectorBucket`? The suite creates one vector bucket per session and tears it down afterwards.


oleksii-udovychenko:
This isn't a duplicate. It's an adoption of #3149, which has been stalled since May. @dotKokott's 12 commits are preserved with authorship intact and rebased onto `main`; the commits on top address @anakin87's outstanding review feedback and bring the store up to `PineconeDocumentStore`'s coverage.

I can't push to #3149: GitHub doesn't allow retargeting a PR's head branch, and I'd need @dotKokott to add me as a collaborator on his fork. Happy to close this if he'd rather finish it himself.

@anakin87, you shepherded #2110, so this is your call.

dotKokott:
@oleksii-udovychenko yes feel free to continue this! I will not get around to it

oleksii-udovychenko:
@anakin87, just a gentle nudge to review this as it has been sitting open for 2 weeks. Happy to address any feedback asap.

PR #3919, opened 2026-09-05
feat: add Amazon DynamoDB DocumentStore with native vector search

### LINKED ISSUE #3899: Amazon DynamoDB (DocStore)
## Summary and motivation

Amazon DynamoDB added native vector similarity search (`SearchVectors` API, GA 2026-08-05), turning it into a real vector store for the first time — previously it could store a vector as a plain number array but had no native similarity-search capability. This closes the gap that made DynamoDB unsuitable for RAG-style retrieval, and it would be a useful addition to the existing DocStore integrations.

## Detailed design

The design of the docstore should include the necessary actions from the DynamoDB client: https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/dynamodb.html

At minimum: creating the DocStore, writing/inserting documents, and querying via embedding retrieval (Embedding Retriever). Filtering will follow DynamoDB's `SearchVectors` index-schema constraints for attribute filters.

## Checklist

### Tasks
- [x] The code is documented with docstrings and was merged in the `main` branch
- [x] Docs are published at https://docs.haystack.deepset.ai/
- [x] There is a Github workflow running the tests for the integration nightly and at every PR
- [x] A new label named like `integration:<your integration name>` has been added to the list of labels for this repository
- [x] The `labeler.yml` file has been updated
- [x] The package has been released on PyPI
- [x] The integration has been listed in the Inventory section of this repo README
- [x] There is an example available to demonstrate the feature

r0ckvn:
### Related Issues

- fixes #3899

### Proposed Changes:

This adds an Amazon DynamoDB `DocumentStore` and a matching `DynamoDBEmbeddingRetriever`, backed by
DynamoDB's native vector search (the `SearchVectors` API, GA 2026-08-05). Documents are stored as
items in a DynamoDB table with a vector index, so embeddings live next to operational data — no
separate vector store and no synchronisation pipeline between the two.

What's included:

- `DynamoDBDocumentStore` implementing the `DocumentStore` protocol: `count_documents`,
  `write_documents` (all three `DuplicatePolicy` modes), `filter_documents`, `delete_documents`,
  plus `to_dict`/`from_dict` serialisation.
- `DynamoDBEmbeddingRetriever` component wrapping the store's embedding retrieval.
- The table and its vector index are created on first use when `create_table_if_not_exists=True`.
- Standard `Secret`-based AWS credential handling, consistent with the `amazon_bedrock` integration.

### How did you test it?

- Unit tests (mocked boto3 client): 21 passed.
- `hatch run test:types` (mypy) and `hatch run fmt-check` (ruff): both clean.
- Integration tests against **real AWS** in `ap-southeast-1`: **50 passed** (16m54s), no failures
  and no teardown errors.
- Vector search verified end-to-end on real AWS: for a query vector equal to one of the stored
  vectors, the identical document scores `1.0`, a near-parallel one `0.997`, and an orthogonal one
  `0.5`, returned most-similar-first.

### Notes for the reviewer

Three behaviours here were only discoverable by running against real AWS, so I'd like to call them
out explicitly rather than leave them buried in the diff:

1. **The vector index is declared inline on `CreateTable` rather than added afterwards via
   `UpdateTable`.** Both are valid API-wise, but adding a vector index to an *existing* table
   triggers an index **backfill**, during which the index reports
   `IndexStatus=CREATING`/`Backfilling=true` and every `SearchVectors` call fails with
   `ValidationException: Cannot search backfilling vector index`. I measured this at over 6 minutes
   even for an empty table, and the table cannot be deleted while it lasts. Creating the index as
   part of the table has nothing to backfill and is queryable in roughly 20 seconds. This is also
   why the integration suite runs in ~17 minutes rather than ~48.

   Relatedly, there is no boto3 waiter for vector-index readiness, so `_ensure_table` polls
   `DescribeTable` until the index is `ACTIVE` and not backfilling. Note `SearchVectors` can briefly
   return `ResourceNotFoundException` for a few seconds after the index goes `ACTIVE`, while the
   dedicated vector-search endpoint catches up.

2. **`SearchVectors` returns a cosine *distance*, which is converted to a Haystack similarity.**
   Per the AWS docs, a COSINE index returns the k *smallest* scores, ranging from 0 (identical) to
   2 (opposite) — lower means more similar. Haystack's `Document.score` convention is the opposite,
   so scores are mapped via `1 - distance / 2`, giving `1.0` for an identical vector and `0.0` for
   an opposite one. Ordering is unaffected since DynamoDB already returns results best-first.

3. **`boto3 >= 1.43.66` is required**, and therefore Python 3.10+. Earlier boto3 releases do not
   contain the `SearchVectors` operation or the vector-index parameters at all, so without the pin
   the feature silently doesn't exist rather than failing cleanly.

Two further notes:

- **Metadata filters are applied client-side** after the vector search returns, because
  `SearchVectors` filtering is constrained by the index's `SearchSchema`. Results are over-fetched
  (capped at DynamoDB's documented 10,000 `SearchVectors` limit) so filtering doesn't drop matches
  that would otherwise fall inside `top_k`.
- **The integration tests need real AWS** — there is no emulator that supports vector indexes. They
  are skipped unless `HAYSTACK_DYNAMODB_INTEGRATION_TESTS=1` is set; requirements are documented in
  `integrations/dynamodb/README.md`. Each test uses a uniquely-suffixed `haystack_test_*` table and
  deletes it afterwards, with a class-scoped sweep as a safety net. Please note they create real,
  billable tables.

**One question for you:** the `dynamodb.yml` workflow needs AWS credentials to run the integration
job. I've wired it with OIDC via `configure-aws-credentials`, gated to skip fork and dependabot PRs,
reading a `secrets.AWS_CI_ROLE_ARN`. That secret doesn't exist yet and only maintainers can add it,
so the integration job will skip until you do. Happy to change the approach if you'd prefer
something else — or to drop the integration job from CI entirely and leave those tests as a local /
manual step.

### Checklist


julian-risch (approved):
Thanks for the initiative @r0ckvn . I applied a couple of changes in a series of commits and the PR is ready to be merged now. Congratulations on your first PR merged into haystack-core-integrations.

PR #3891, opened 2026-09-01
chore!: remove deprecated `NvidiaGenerator` (use `NvidiaChatGenerator` instead)

julian-risch:
### Related Issues

- part of https://github.com/deepset-ai/haystack-private/issues/394

### Proposed Changes:

- remove the `NvidiaGenerator` component, deprecated in #3391 in favour of `NvidiaChatGenerator` (which accepts a plain `str` as input)
- remove `NimBackend.generate()` (only used by the removed component) and its tests (`test_generate`, `test_generate_raises_on_http_error`, and the `mock_generate_post_response` helper in `tests/test_nim_backend.py`)
- drop `NvidiaGenerator` from the shared `test_base_url.py` parametrizations
- remove its tests (`tests/test_generator.py`) and its API reference entry in `pydoc/config_docusaurus.yml`
- drop the `NvidiaGenerator` export from `components/generators/nvidia/__init__.py`

Migration for users:
```python
from haystack_integrations.components.generators.nvidia import NvidiaChatGenerator

generator = NvidiaChatGenerator(model="meta/llama-3.1-8b-instruct")
reply = generator.run("What is NLP?")["replies"][0]  # ChatMessage
text = reply.text
```

### How did you test it?

Locally in `integrations/nvidia`: `hatch run fmt` (all checks passed, 28 files unchanged), `hatch run test:types` (no issues in 15 source files), `hatch run test:unit` (181 passed, 15 deselected), `hatch run docs` (generated without errors; the generated `nvidia.md` no longer contains a `NvidiaGenerator` section). CI.

### Notes for the reviewer

Breaking change: release as `integrations/nvidia-v2.0.0` (major bump, latest is v1.4.0). `Client.NVIDIA_GENERATOR` in `utils/nvidia/client.py` is intentionally kept: it is the NIM API client identifier string, not the component. Companion PRs remove the other deprecated non-chat generators; docs pages and redirects are handled in deepset-ai/haystack.

CI: the only red job is `TestNvidiaDocumentEmbedder::test_run_integration_with_api_catalog[nvidia/nv-embedqa-e5-v5]`, failing with HTTP 410 because NVIDIA retired that embedding model on 2026-08-25. Pre-existing on `main` and unrelated to this PR; the embedder test model should be swapped in a separate fix.

### Checklist

[...]

PR #3923, opened 2026-09-06
fix: serialize timeout in NvidiaGenerator.to_dict

pcbeingused333:
## Summary

`NvidiaGenerator` accepts a `timeout` in `__init__` and stores it on
`self.timeout`, but `to_dict()` does not serialize it:

```python
return default_to_dict(
    self,
    model=self._model,
    api_url=self.api_url,
    api_key=self._api_key.to_dict() if self._api_key else None,
    model_arguments=self._model_arguments,
)
```

A pipeline serialized with a non-default timeout loses it on load — `from_dict`
runs `__init__` with `timeout=None`, which falls back to the `NVIDIA_TIMEOUT`
env var or the hard-coded `60.0`.

The other four Nvidia components already serialize `timeout` the same way:
`NvidiaTextEmbedder`, `NvidiaDocumentEmbedder`, `NvidiaRanker` and
`NvidiaChatGenerator`. This PR brings `NvidiaGenerator` in line.

(`NvidiaGenerator` is deprecated in favour of `NvidiaChatGenerator` but is still
shipped and serializable.)

## Testing

- `test_to_dict` / `test_to_dict_with_custom_init_parameters` / `test_from_dict`
  updated to include `timeout`, matching `test_text_embedder.py`.
- New `test_to_dict_from_dict_roundtrip_preserves_timeout` fails on `main`
  (`assert 60.0 == 12.5`) and passes with this change.
- `hatch run test:unit` — 212 passed, 18 deselected. `hatch run fmt-check` clean.

### DIFF
--- a/integrations/nvidia/src/haystack_integrations/components/generators/nvidia/generator.py
+++ b/integrations/nvidia/src/haystack_integrations/components/generators/nvidia/generator.py
@@ -155,6 +155,7 @@ class NvidiaGenerator:
             api_url=self.api_url,
             api_key=self._api_key.to_dict() if self._api_key else None,
             model_arguments=self._model_arguments,
+            timeout=self.timeout,
         )
 
     @property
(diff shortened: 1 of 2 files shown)


anakin87:
We are removing that deprecated component in https://github.com/deepset-ai/haystack-core-integrations/pull/3891, so there is no need to merge this bug fix.

Thank you anyway!

PR #1907, opened 2025-06-06
feat: Allow the scores returned by AI Search to be populated in the Document.meta

Seth-Peters:
### Proposed Changes:

Expose AI Search scores in the `Document.meta` when returning from AI Search and converting back to a `Document` object. This is critical information needed by users of this integration to decide what actions to take with the returned results.

### How did you test it?

Testing by searching an index with the `AzureAISearchHybridRetriever`, `AzureAISearchBM25Retriever`, and `AzureAISearchEmbeddingRetriever` to ensure that Document.meta contains the search scores from the results.

### Notes for the reviewer

Now, when users search an AI Search index with this fix, returned `Document` objects will be populated like this:

```python
from haystack import Document, Pipeline
from haystack.components.embedders import SentenceTransformersDocumentEmbedder, SentenceTransformersTextEmbedder
from haystack.components.writers import DocumentWriter

from haystack_integrations.components.retrievers.azure_ai_search import AzureAISearchEmbeddingRetriever
from haystack_integrations.document_stores.azure_ai_search import AzureAISearchDocumentStore

"""
This example demonstrates how to use the AzureAISearchEmbeddingRetriever to retrieve documents
using embeddings based on a query. To run this example, you'll need an Azure Search service endpoint
and API key, which can either be
set as environment variables (AZURE_AI_SEARCH_ENDPOINT and AZURE_AI_SEARCH_API_KEY) or
provided directly to AzureAISearchDocumentStore(as params "api_key", "azure_endpoint").
Otherwise you can use DefaultAzureCredential to authenticate with Azure services.
See more details at https://learn.microsoft.com/en-us/azure/search/keyless-connections?tabs=python%2Cazure-cli
"""

document_store = AzureAISearchDocumentStore(index_name="retrieval-example")

model = "sentence-transformers/all-mpnet-base-v2"

documents = [
    Document(content="There are over 7,000 languages spoken around the world today."),
    Document(
        content="""Elephants have been observed to behave in a way that indicates a
         high level of self-awareness, such as recognizing themselves in mirrors."""
    ),
    Document(
        content="""In certain parts of the world, like the Maldives, Puerto Rico, and
          San Diego, you can witness the phenomenon of bioluminescent waves."""
    ),
]

document_embedder = SentenceTransformersDocumentEmbedder(model=model)
document_embedder.warm_up()

# Indexing Pipeline
indexing_pipeline = Pipeline()
indexing_pipeline.add_component(instance=document_embedder, name="doc_embedder")
[...]

PR #3925, opened 2026-09-08
fix: serialize include_search_metadata in AzureAISearchDocumentStore.to_dict

pcbeingused333:
### Related Issues

- No issue. Same defect class as #3808 and #3873, in one more integration.

### Proposed Changes:

`AzureAISearchDocumentStore` accepts `include_search_metadata`, stores it as
`self._include_search_metadata`, and reads it in `_convert_search_result_to_documents`
to decide whether every field Azure returns (`@search.score`, `@search.reranker_score`,
`@search.highlights`, `@search.captions`, ...) is copied into `Document.meta`, or only
the configured index fields are. `to_dict` left it out, so a store built with
`include_search_metadata=True` comes back from `from_dict` with it `False` and returns
a different `meta` shape after a serialization round trip.

One line in `to_dict`. No `from_dict` change: it already passes the key through.

The parameter was added in #1907 (June 2025); the `to_dict` omission dates from there.

### How did you test it?

`integrations/azure_ai_search` (`test_document_store.py`): 37 passed, 92 deselected.
3 fail without the fix:

- `test_to_dict` and `test_to_dict_with_params` — the expected dicts now include the key.
- `test_to_dict_round_trip_preserves_include_search_metadata` — new: asserts
  `include_search_metadata=True` survives `from_dict(to_dict())`. Fails with
  `KeyError: 'include_search_metadata'` when the `to_dict` line is reverted.

`hatch run fmt-check` clean.

### Notes for the reviewer

Found with the same AST script used for #3873, run against `main` after the two PRs on
this defect class. Everything else it flagged in the integrations was a false positive
(serialized via a parent `to_dict`, folded into another kwarg at `__init__`, or
introspected dynamically like `QdrantDocumentStore`) except `NvidiaGenerator.timeout`,
which is #3923, and `google_vertex`, which is archived.

### Checklist

### DIFF
--- a/integrations/azure_ai_search/tests/test_document_store.py
+++ b/integrations/azure_ai_search/tests/test_document_store.py
@@ -94,6 +94,7 @@ def test_to_dict_with_params(monkeypatch):
         },
         encryption_key=encryption_key,
         analyzers=[analyzer],
+        include_search_metadata=True,
     )
 
     res = document_store.to_dict()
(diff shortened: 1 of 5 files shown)


integrations/azure_ai_search/tests/test_document_store.py
```diff
@@ -130,10 +131,23 @@ def test_to_dict_with_params(monkeypatch):
                     }
                 ],
             },
+            "include_search_metadata": False,
         },
     }
 
 
+def test_to_dict_round_trip_preserves_include_search_metadata(monkeypatch):
```
anakin87: I'd delete this test and test this behavior in `test_to_dict_with_params`

anakin87 (changes requested):
Thank you!

I left a minor comment

pcbeingused333:
@anakin87 Done in `d1c677c86` — dropped the standalone round-trip test and folded the True case into `test_to_dict_with_params` (added `include_search_metadata=True` to the constructor call, updated the expected dict) and `test_from_dict_with_params` (added the key to `data`, asserted `_include_search_metadata is True`), matching how every other non-default param is already covered in those two.

`hatch run test:pytest tests/` — 97 passed, 103 deselected (API-key-gated). `fmt-check` clean.

anakin87 (approved):
Thank you

PR #3926, opened 2026-09-08
fix: batch CohereDocumentEmbedder.run_async like the sync path

pcbeingused333:
### Related Issues

- No issue. Sync/async parity defect, same class as haystack #12358.

### Proposed Changes:

`CohereDocumentEmbedder` has a sync `run` and an async `run_async`. They call two helpers:

- `get_response` (sync) loops over `texts` in slices of `batch_size` (default 32) and calls `client.embed(texts=batch, ...)` per slice.
- `get_async_response` (async) had no `batch_size` parameter and called `await client.embed(texts=texts, ...)` **once with every text**.

So `run_async`:
- ignored the configured `batch_size` (and `progress_bar`);
- sent all texts in a single request, which the Cohere embed endpoint rejects above its per-call text cap — `run_async` fails on inputs where `run` on the same documents succeeds.

`get_async_response` now takes `batch_size` and `progress_bar` and loops over batches exactly like `get_response` — sequential `await`s, mirroring the sync loop (no `asyncio.gather`, so the request rate is unchanged). `run_async` passes `self.batch_size` and `self.progress_bar`.

`CohereTextEmbedder` also calls `get_async_response`, but only ever with a single string, so its behaviour is unchanged (one batch).

### How did you test it?

`integrations/cohere` (`test_document_embedder.py`): 15 passed, 2 deselected; full unit suite 98 passed.

New `test_run_async_sends_texts_in_batches`: a fake `AsyncClientV2` records the size of each `embed` call; `run_async` on 5 docs with `batch_size=2` must issue calls of `[2, 2, 1]`. Fails as `[5] == [2, 2, 1]` with the fix reverted. `hatch run fmt-check` clean.

### Notes for the reviewer

Batches run sequentially to keep this a behaviour-parity fix and not change the request rate; running them concurrently with a bounded gather would be a reasonable follow-up.

`CohereDocumentImageEmbedder` is not affected — its `run` and `run_async` are already symmetric (one image per call, no batching either side).

### Checklist

### DIFF
--- a/integrations/cohere/src/haystack_integrations/components/embedders/cohere/document_embedder.py
+++ b/integrations/cohere/src/haystack_integrations/components/embedders/cohere/document_embedder.py
@@ -249,6 +249,8 @@ class CohereDocumentEmbedder:
             model_name=self.model,
             input_type=self.input_type,
             truncate=self.truncate,
+            batch_size=self.batch_size,
+            progress_bar=self.progress_bar,
             embedding_type=self.embedding_type,
         )
 
--- a/integrations/cohere/src/haystack_integrations/components/embedders/cohere/utils.py
+++ b/integrations/cohere/src/haystack_integrations/components/embedders/cohere/utils.py
@@ -16,6 +16,8 @@ async def get_async_response(
     model_name: str,
     input_type: str,
     truncate: str,
+    batch_size: int = 32,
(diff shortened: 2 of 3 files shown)


sjrl (approved):
Thanks!

PR #3935, opened 2026-09-09
refactor!: Optimum - move resources creation to warm_up

anakin87:
### Related Issues

- part of https://github.com/deepset-ai/haystack-private/issues/440

### Proposed Changes:
- for the embedders, make `__init__` store only configuration and move everything else to `warm_up`: model validation via HTTP, model loading...
  - to do this minimally, in each embedder's `warm_up`, I both create and warm up the embedding backend
  - this also changes serialization. I added a test to make sure that the legacy serialization format can still be deserialized.

### How did you test it?
CI, new tests

### Checklist

### DIFF
--- a/integrations/optimum/src/haystack_integrations/components/embedders/optimum/_backend.py
+++ b/integrations/optimum/src/haystack_integrations/components/embedders/optimum/_backend.py
@@ -89,12 +89,14 @@ class _EmbedderParams:
             out[field] = copy.deepcopy(getattr(self, field))
 
         # Fixups.
-        assert isinstance(self.pooling_mode, OptimumEmbedderPooling)
-        out["pooling_mode"] = str(self.pooling_mode)
+        out["pooling_mode"] = str(self.pooling_mode) if self.pooling_mode is not None else None
         out["token"] = self.token.to_dict() if self.token else None
         out["optimizer_settings"] = self.optimizer_settings.to_dict() if self.optimizer_settings else None
         out["quantizer_settings"] = self.quantizer_settings.to_dict() if self.quantizer_settings else None
 
+        out["model_kwargs"] = out["model_kwargs"] or {}
+        out["model_kwargs"].setdefault("model_id", self.model)
+        out["model_kwargs"].setdefault("provider", self.onnx_execution_provider)
         out["model_kwargs"].pop("use_auth_token", None)
         out["model_kwargs"].pop("token", None)
         serialize_hf_model_kwargs(out["model_kwargs"])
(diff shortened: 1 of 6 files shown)


integrations/optimum/src/haystack_integrations/components/embedders/optimum/_backend.py
```diff
@@ -89,20 +89,23 @@ def serialize(self) -> dict[str, Any]:
...
-        assert isinstance(self.pooling_mode, OptimumEmbedderPooling)
-        out["pooling_mode"] = str(self.pooling_mode)
+        out["pooling_mode"] = str(self.pooling_mode) if self.pooling_mode is not None else None
         out["token"] = self.token.to_dict() if self.token else None
         out["optimizer_settings"] = self.optimizer_settings.to_dict() if self.optimizer_settings else None
         out["quantizer_settings"] = self.quantizer_settings.to_dict() if self.quantizer_settings else None
 
+        out["model_kwargs"] = out["model_kwargs"] or {}
+        out["model_kwargs"].setdefault("model_id", self.model)
+        out["model_kwargs"].setdefault("provider", self.onnx_execution_provider)
```
julian-risch, line 99: We could keep these two lines without causing problems but I believe it's cleaner to remove them.
`model_id` and `provider` are derived from `model` and `onnx_execution_provider`. With this PR in its current state what a user get's when they call `to_dict` differs from the parameters they used to initialize the component.

The backend already applies the same defaults when warm up is called and if I am not mistaken legacy dicts deserialize without problems.

We'd need to adjust some test assertions.
```suggestion
```

integrations/optimum/src/haystack_integrations/components/embedders/optimum/_backend.py
```diff
@@ -149,6 +152,9 @@ def __init__(self, params: _EmbedderParams) -> None:
         self.pooling_layer: SentenceTransformerPoolingLayer | None = None
 
     def warm_up(self) -> None:
+        if self.model is not None:
+            return
+
```
julian-risch, line 157: Just a comment: I believe this will never be reached right now but it's good to have it here for consistency. And it leaves no doubt.

julian-risch (changes requested):
Looks quite good to me already. One change request. It's not wrong as is so I'd also be fine with keeping the PR unchanged for simplicity. However, I believe with the change it would be cleaner.

integrations/optimum/src/haystack_integrations/components/embedders/optimum/_backend.py
anakin87: You're right. Keeping it anyway.

integrations/optimum/src/haystack_integrations/components/embedders/optimum/_backend.py
anakin87: Makes sense. Done in https://github.com/deepset-ai/haystack-core-integrations/pull/3935/commits/66ceaa45f887a50fe8a6192d69016d02cdf3e405

julian-risch (approved):

PR #3939, opened 2026-09-09
refactor!: Transformers - lifecycle refactor

anakin87:
### Related Issues

- related to https://github.com/deepset-ai/haystack-private/issues/440

### Proposed Changes:
- pin `haystack-ai>=3.0.0` and `transformers>=5.0.0`: allows removing some compatibility code
- make `__init__` do basic validation and store configuration; network access and secret resolution only happens in `warm_up`
  - `TransformersTextRouter` is an exception: if the user does not provide explicit `labels`, it still needs to contact Hugging Face Hub at `__init__`: labels become the output sockets of the component
- Chat Generator: add `close` method that only shuts down the executor (no way to release the model); removed `shutdown` and `__del__`

### How did you test it?
CI, new tests

### Checklist

### DIFF
--- a/integrations/transformers/pyproject.toml
+++ b/integrations/transformers/pyproject.toml
@@ -23,7 +23,7 @@ classifiers = [
   "Programming Language :: Python :: Implementation :: CPython",
   "Programming Language :: Python :: Implementation :: PyPy",
 ]
-dependencies = ["haystack-ai>=2.30.0", "transformers[torch,sentencepiece]>=4.57.0"]
+dependencies = ["haystack-ai>=3.0.0", "transformers[torch,sentencepiece]>=5.0.0"]
 
 [project.urls]
 Documentation = "https://github.com/deepset-ai/haystack-core-integrations/tree/main/integrations/transformers#readme"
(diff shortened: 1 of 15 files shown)


davidsbatista:
Looks good to me.

One thing maybe worth noting in the docstring or somewhere

As far as I understood before the PR, cleanup was via `__del__`, with `shutdown()` now `.close()` must be called on the component instance directly since the Pipeline will not do it.

davidsbatista (approved):
looks good just a small comment regarding cleaning

anakin87:
> One thing maybe worth noting in the docstring or somewhere
> 
> As far as I understood before the PR, cleanup was via `__del__`, with `shutdown()` now `.close()` must be called on the component instance directly since the Pipeline will not do it.

`__del__()` was a best-effort cleanup that triggers when the component is garbage-collected.
With the new lifecycle design, we expect users to explicitly invoke `component.close()` if they use the component in isolation or `Pipeline.close()` that releases the resources for all closable components. So I adopted this design also here.

I'll document this in Haystack docs.

PR #3940, opened 2026-09-09
fix(transformers): serialize overlap_threshold in TransformersExtractiveReader.to_dict

pcbeingused333:
### Problem

`TransformersExtractiveReader.__init__` takes `overlap_threshold` (default `0.01`), stores it on the instance, and `run()` uses it as the fallback when the per-call `overlap_threshold` argument is omitted:

```python
overlap_threshold = overlap_threshold or self.overlap_threshold
```

But `to_dict()` doesn't serialize it. A reader configured with a custom `overlap_threshold` — or with `None` to disable answer de-duplication entirely — silently reverts to `0.01` after `to_dict()` → `from_dict()`, so the setting is lost whenever a pipeline is saved and reloaded.

### Fix

Add `overlap_threshold=self.overlap_threshold` to the `default_to_dict` call.

### Tests

- The existing `test_to_dict*` assertions now include the key.
- New `test_to_dict_from_dict_roundtrip_preserves_non_default_overlap_threshold` pins a non-default value through a full round-trip.
- Both fail with the `to_dict` line reverted (verified locally: `hatch run test:unit tests/test_extractive_reader.py` — 35 passed with the fix, 7 failed without; `hatch run fmt-check` clean).

https://claude.ai/code/session_018uqJtTJWawYLVA5EdpUmho

### DIFF
--- a/integrations/transformers/src/haystack_integrations/components/readers/transformers/extractive_reader.py
+++ b/integrations/transformers/src/haystack_integrations/components/readers/transformers/extractive_reader.py
@@ -150,6 +150,7 @@ class TransformersExtractiveReader:
             answers_per_seq=self.answers_per_seq,
             no_answer=self.no_answer,
             calibration_factor=self.calibration_factor,
+            overlap_threshold=self.overlap_threshold,
             model_kwargs=self.model_kwargs,
         )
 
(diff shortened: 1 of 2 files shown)


pcbeingused333:
Closing as a duplicate of #3873, which already fixes this exact case (`TransformersExtractiveReader.overlap_threshold`) along with `S3Downloader.boto3_config`, with the same round-trip test approach. #3873 has been open since 31 Aug; I missed it when filing this. Sorry for the noise.

PR #3954, opened 2026-09-14
feat: send X-Weaviate-Client-Integration telemetry header

dudanogueira:
### Related Issues

- None. Follows up [weaviate/weaviate#10535](https://github.com/weaviate/weaviate/pull/10535), which added server-side telemetry for client integrations.

### Proposed Changes:

Weaviate tracks which framework integrations talk to it through the `X-Weaviate-Client-Integration` header, in a `<name>/<version>` format alongside the existing `X-Weaviate-Client`. This tags the connection with `haystack-python/<weaviate-haystack version>` so Haystack adoption is visible in Weaviate's telemetry.

The header is merged into `additional_headers` at client construction, which covers all four connection paths (sync/async × Weaviate Cloud/custom). The alternative — the client's `client.integrations.configure()` API — was not used for two reasons:

1. It relies on `weaviate.connect.integrations._IntegrationConfig`, a private class whose existing subclasses are all *vectorizer* integrations (Cohere, OpenAI, Jina), a different concept.
2. On the Weaviate Cloud path it would run after `connect_to_weaviate_cloud()` has already built the httpx client, so the HTTP header would be silently dropped.

`additional_headers` is stable public API across all of 4.x, and weaviate-client copies it into gRPC metadata as well as HTTP headers.

Two details worth noting:

- A user-supplied value for the same header always wins, so it can be overridden or suppressed via `additional_headers`. The check is case-insensitive, because gRPC metadata keys are case-sensitive and a user spelling of `x-weaviate-client-integration` would otherwise end up alongside ours.
- The merged headers are computed per connection and never stored on the instance, so `to_dict()` keeps serializing only what the user passed. Otherwise every serialized pipeline would bake in the version of `weaviate-haystack` that happened to write it.

The version is read at runtime via `importlib.metadata` with an `"unknown"` fallback, mirroring the existing `_attribution_header()` helper in the perplexity integration.

### How did you test it?

Six new unit tests: the header reaches the sync client, the async client and the Cloud path; a user-supplied value wins (parametrized over both casings); the version falls back to `unknown`; and `to_dict()` is unaffected.

Verified end to end against a live Weaviate 1.39.4 pushing telemetry to a local collector. After one write and one BM25 query:

```
{"version": "1.39.4",
 "clientUsage": {"python": {"4.23.1-sync": 18}},
 "clientIntegrationUsage": {"haystack-python": {"7.8.1.dev109+gc65697c35": 18}}}
```

And on the connection itself, confirming it reaches the gRPC transport and not only HTTP:

```
x-weaviate-client:             weaviate-client-python/4.23.1-sync
x-weaviate-client-integration: haystack-python/7.8.1.dev109+gc65697c35
```

`hatch run test:unit` (87 passed), `hatch run test:integration` (186 passed, 1 skipped), `hatch run fmt-check` and `hatch run test:types` all pass.

### Notes for the reviewer

Two existing assertions had to be updated, since the local/embedded path now passes a one-key dict where it previously passed `additional_headers=None`: `test_client_connects_to_weaviate_cloud` and `test_connection`. That change in shape is the only observable behaviour change.

No opt-out env var was added, matching the perplexity precedent. The header carries no user data, and `additional_headers` already provides a documented way to override it.

### Checklist


integrations/weaviate/src/haystack_integrations/document_stores/weaviate/document_store.py
```diff
@@ -183,6 +205,26 @@ def _clean_connection_settings(self) -> None:
             "properties", DOCUMENT_COLLECTION_PROPERTIES
         )
 
+    def _connection_headers(self) -> dict[str, Any]:
+        """
+        Builds the headers passed to the Weaviate client.
+
+        Merges the integration identification header into the user supplied `additional_headers`. A user
+        supplied value for the same header wins, so the header can be overridden or suppressed. The merged
+        headers are deliberately not stored on the instance: `to_dict()` must keep serializing only the
```
sjrl, line 217: The public `additional_headers` docstring in `__init__` doesn't mention that every connection now carries `X-Weaviate-Client-Integration: haystack-python/<version>`, or that passing that key is how to override it, so both are only discoverable from the source. Could be worth a line there.

sjrl (commented):
Two small notes, on documenting the new header and on test coverage for the async path.

integrations/weaviate/tests/test_document_store.py
```diff
@@ -105,6 +110,71 @@ async def exists(_name: str) -> bool:
     mock_connect.assert_called_once()
```
sjrl, line 115: Could we also assert the header here, e.g. `assert mock_connect.call_args.kwargs["headers"]["X-Weaviate-Client-Integration"].startswith("haystack-python/")`? This looks like the one of the four connection paths without a header assertion.

integrations/weaviate/src/haystack_integrations/document_stores/weaviate/document_store.py
dudanogueira: Good catch — added it in 6762c40ea:

> Every connection also carries `X-Weaviate-Client-Integration: haystack-python/<version>`, which
> identifies this integration in Weaviate's telemetry. Supplying that header here, in any casing,
> overrides it.

I called out the casing explicitly because that's the user-facing half of the case-insensitive check in `_connection_headers()` — otherwise the fact that `x-weaviate-client-integration` also works is only visible from the source comment.

integrations/weaviate/tests/test_document_store.py
dudanogueira: Right, that was the one of the four paths without a header assertion. Fixed in 6762c40ea.

I went slightly further than `startswith` and mirrored the sync Cloud test instead: the store now gets `additional_headers={"X-HuggingFace-Api-Key": "k"}` and the test asserts the full merged dict. That also pins the merge behaviour on this path, and keeps the sync/async pair symmetrical.

sjrl (approved):
Thanks!

PR #3958, opened 2026-09-14
test: cover Weaviate filters, retriever components and pipeline serde

dudanogueira:
### Related Issues

- None.

### Proposed Changes:

Stacked on #3957 — the diff here is the second commit only; please merge that one first.

Four coverage gaps, each of which lets regressions through today.

**The retriever mocks could not catch signature drift.** All the retriever tests use `Mock(spec=WeaviateDocumentStore)`. Because that is `spec` and not autospec, the `assert_called_once_with(...)` assertions never validated the real `_bm25_retrieval` / `_embedding_retrieval` / `_hybrid_retrieval` signatures — renaming a parameter on the document store left every retriever test green. Switching to `create_autospec(..., instance=True)` fixes that; I confirmed by renaming `_bm25_retrieval`'s `query` parameter, which now fails the test and previously did not.

**Two inert `@patch` decorators.** `test_run` in both the bm25 and embedding retriever tests was decorated with `@patch("...retrievers.weaviate.bm25_retriever.WeaviateDocumentStore")` and then ignored the patch, injecting a bare `MagicMock` instead. One of the two was in the *embedding* retriever's tests, patching the wrong module entirely. Both decorators are dropped.

**The retriever components were never run against a real server.** Adds `tests/test_retrieval.py`, which exercises all three retrievers sync and async against a live Weaviate and asserts ordering, `top_k`, filters and score population. Previously end-to-end retrieval was only covered indirectly, through the private `_*_retrieval` methods called inside `TestWeaviateDocumentStore`. Follows the shape of `integrations/pgvector/tests/test_retrieval.py`.

**`tests/test_filters.py` goes from 5 tests to 22.** `validate_filters()` had no coverage at all despite guarding twelve call sites, and neither did the `contains` operator, in either its `like "*value*"` happy path or its `FilterError` on a non-string value. Also now covered: the `_handle_date()` pass-through for non-date values, the unsatisfiable filter that ordering operators build for `None`, and the De Morgan inversion of a `NOT` nested inside a `NOT`.

Finally, a `Pipeline` dumps/loads round trip per retriever — the standard regression for a broken nested document-store `from_dict()`. No Weaviate test imported `Pipeline` before.

### How did you test it?

`hatch run test:unit` (109 passed), `hatch run test:integration` (195 passed, 1 skipped, including the 9 new ones), `hatch run fmt-check` and `hatch run test:types` all pass.

### Review round

Four comments from @sjrl, all addressed:

- The two `NOT` bugs found in the filter tests turned out to be real bugs in `_filters.py`, not test bugs. They are fixed in the parent PR #3957, which now replaces the manual De Morgan inversion with Weaviate's native `NOT` operator. `test_invert_nested_not_conditions` — which asserted the buggy behaviour — is gone, replaced there by a unit test and two integration tests.
- The `Pipeline` round trips now assert `distance` and `alpha`, not just `top_k`.

### Notes for the reviewer

`test_retrieval.py` defines its own small fixtures rather than reusing the ones in `test_document_store.py`, since those are tied to that file's filterable-docs schema.

One thing I deliberately left alone: the sync and async `document_store` fixtures in `test_document_store.py` / `test_document_store_async.py` declare *different* collection schemas — the async one adds `stopwords: {"preset": "none"}` and eight properties — so BM25 behaves differently between the two suites. Unifying them is a behaviour change that deserves its own PR, and the duplicated `assert_documents_are_equal` / `filterable_docs` would then be worth extracting into a shared module, as `integrations/opensearch` does.

### Checklist


integrations/weaviate/tests/test_filters.py
```diff
@@ -70,3 +73,63 @@ def test_parse_comparison_condition_errors():
...
+
+
+def test_invert_nested_not_conditions():
+    """NOT inverts to OR, so a NOT nested inside a NOT must round-trip back to the original meaning."""
+    inner = {"operator": "NOT", "conditions": [{"field": "meta.number", "operator": "==", "value": 100}]}
+    inverted = _invert_condition({"operator": "NOT", "conditions": [inner]})
+    assert inverted == {
+        "operator": "OR",
+        "conditions": [{"operator": "OR", "conditions": [{"field": "meta.number", "operator": "!=", "value": 100}]}],
+    }
```
sjrl: I believe the asserted result double-negates which I'm not sure is intended behavior

`_invert_condition` maps the outer `NOT`→`OR` *and* recursively inverts the inner `NOT`'s children, so `NOT(NOT(number == 100))` comes out as `OR(OR(number != 100))` instead of `number == 100`. Verified end to end: for `Document(meta={"number": 100})`, `haystack.utils.filters.document_matches_filter` returns `True`, while `convert_filters` builds `number != 100 OR number IS NULL` — which excludes exactly that document.

I believe inverting a `NOT` node should strip the `NOT` (→ `AND` over the children unchanged), not invert the children.

sjrl (commented):
Three comments below on the filter tests and the serde round trips.

integrations/weaviate/src/haystack_integrations/document_stores/weaviate/_filters.py
```diff
@@ -282,6 +282,10 @@ def _parse_comparison_condition(condition: dict[str, Any]) -> FilterReturn:
     operator: str = condition["operator"]
     value: Any = condition["value"]
 
+    if operator not in COMPARISON_OPERATORS:
+        msg = f"Unknown comparison operator '{operator}'. Valid operators are: {list(COMPARISON_OPERATORS.keys())}"
+        raise FilterError(msg)
```
sjrl, line 254: This guard is unreachable for any comparison nested under a `NOT`: `_parse_logical_condition` calls `_invert_condition` first, which does `OPERATOR_INVERSE[filters["operator"]]` (line 71) and raises a bare `KeyError`.

And `contains` — a supported operator with a `COMPARISON_OPERATORS` entry and new tests in this PR — has no `OPERATOR_INVERSE` entry, so:

```python
filter_documents(filters={"operator": "NOT", "conditions": [{"field": "meta.name", "operator": "contains", "value": "x"}]})
```

raises `KeyError: 'contains'` out of `convert_filters` rather than a `FilterError`. Same for `like`. Adding `contains` to `OPERATOR_INVERSE` and a membership check in `_invert_condition` could be worth doing.

integrations/weaviate/tests/test_embedding_retriever.py
```diff
@@ -182,8 +183,44 @@ def test_run(mock_document_store):
...
+    pipeline.add_component(
+        "retriever", WeaviateEmbeddingRetriever(document_store=document_store, top_k=3, distance=0.5)
+    )
+
+    reloaded = Pipeline.loads(pipeline.dumps()).get_component("retriever")
+
+    assert isinstance(reloaded, WeaviateEmbeddingRetriever)
+    assert isinstance(reloaded._document_store, WeaviateDocumentStore)
+    assert reloaded._document_store.to_dict() == document_store.to_dict()
+    assert reloaded._top_k == 3
```
sjrl, line 226: Could we also add `assert reloaded._distance == 0.5` to check that distance round trips as expected?

integrations/weaviate/tests/test_hybrid_retriever.py
```diff
@@ -431,17 +432,31 @@ def test_run_with_max_vector_distance_zero_init_and_none_runtime():
...
+    document_store = WeaviateDocumentStore(url="http://localhost:8080")
+    pipeline = Pipeline()
+    pipeline.add_component("retriever", WeaviateHybridRetriever(document_store=document_store, top_k=3, alpha=0.3))
+
+    reloaded = Pipeline.loads(pipeline.dumps()).get_component("retriever")
+
+    assert isinstance(reloaded, WeaviateHybridRetriever)
+    assert isinstance(reloaded._document_store, WeaviateDocumentStore)
+    assert reloaded._document_store.to_dict() == document_store.to_dict()
+    assert reloaded._top_k == 3
```
sjrl, line 462: Same gap as the embedding retriever: `alpha=0.3` is passed at construction but never asserted after the round trip, so a `from_dict` that dropped `alpha` would still pass here.

integrations/weaviate/tests/test_filters.py
dudanogueira: You're right, and the test docstring I wrote ("must round-trip back to the original meaning") contradicts what it asserts — it codified the bug rather than catching it. Confirmed end to end exactly as you describe: for `Document(meta={"number": 100})`, `document_matches_filter` returns `True` while `convert_filters` builds `number != 100 OR number IS NULL`.

Rather than stripping the `NOT` in `_invert_condition`, I've removed the manual inversion entirely — see the second comment for why. Fixed in the parent PR #3957, and this test is replaced by `test_nested_not_is_not_double_negated` plus an integration test that asserts `NOT(NOT(x))` selects the same Documents as `x` against a live server (it fails on `main` with `assert 18 == 0`).

integrations/weaviate/src/haystack_integrations/document_stores/weaviate/_filters.py
dudanogueira: Both correct. `NOT(name contains "x")` raises `KeyError: 'contains'` today, and `like` — which isn't in `COMPARISON_OPERATORS` at all — raises `KeyError: 'like'` before it can reach the "Unknown comparison operator" guard.

Adding `contains` to `OPERATOR_INVERSE` turns out not to be possible in a satisfying way: there's no "not contains" in the Haystack filter spec and no `NOT LIKE` in Weaviate, so the entry would have nothing to point at. But checking the client made the guard unnecessary — Weaviate has had a **native `NOT`** operator since 1.33, exposed as `Filter.not_`, and `weaviate-client>=4.23.0` (the floor #3956 just set) has it.

So in #3957 I've deleted `OPERATOR_INVERSE` and `_invert_condition` and emit a real `NOT` instead. That fixes both this and the double-negation above structurally, makes the `FilterError` guard reachable under `NOT`, and is −40 lines net.

The tradeoff is a server floor. I probed it rather than guessing — 1.33.0, 1.36.2 and 1.39.5 all handle `NOT`, `NOT(NOT(x))`, `NOT(like)` and `NOT(AND(...))`; 1.32.0 and older reject all four at the gRPC layer. CI already runs 1.39. It's documented on `filter_documents()` / `filter_documents_async()`, but if requiring ≥1.33 isn't acceptable I'll switch to the narrower fix (strip the `NOT` in `_invert_condition`, guard the table lookup with a `FilterError`) — say the word.

integrations/weaviate/tests/test_embedding_retriever.py
dudanogueira: Added — `assert reloaded._distance == 0.5`. Good catch; the round trip only checked `top_k`, so a `from_dict` dropping `distance` would have passed.

integrations/weaviate/tests/test_hybrid_retriever.py
dudanogueira: Right, same gap. Added `assert reloaded._alpha == 0.3`.

PR #3963, opened 2026-09-15
fix(alloydb,mariadb): raise FilterError on empty logical conditions

joshinitin2016883-sketch:
### Related Issues

- None. Follow-up to #3946, which fixed the same case in pgvector, as mentioned in #3951.

### Proposed Changes:

alloydb and mariadb don't reject a logical filter whose `conditions` list is empty:

```python
_convert_filters_to_where_clause_and_params({"operator": "AND", "conditions": []})
# alloydb: IndexError: list index out of range
# mariadb: (" WHERE ()", [])  -> invalid SQL, fails at execute()
#          with "NOT": (" WHERE COALESCE(NOT (), TRUE)", [])
```

This adds the check #3946 added to pgvector: both stores now raise `FilterError("'conditions' must not be empty ...")`, including for nested groups. In mariadb it also covers `NOT`, which alloydb and pgvector don't support.

### How did you test it?

- Regression tests: alloydb `test_logical_condition_empty_conditions_raises` (`AND`, `OR`) and `test_logical_condition_nested_empty_conditions_raises`, the same tests #3946 added for pgvector; mariadb four new cases in `TestFilterConversion::test_invalid_filters_raise` (`AND`, `OR`, `NOT`, nested). All fail before the change and pass after.
- Unit tests: alloydb 47 passed, mariadb 87 passed.
- `ruff check`, `ruff format --check` and `mypy` (using each integration's config) are clean for both.

Hatch is blocked by Windows Application Control on my machine, so these ran from a venv with each integration installed in editable mode rather than through `hatch run`. Integration tests, which need a database, are left to CI.

### Notes for the reviewer

- The alloydb tests sit next to the other logical-condition tests instead of at the end of the file, to avoid a conflict with #3799.
- Prepared with Claude Code running locally: it ported the change and tests from #3946, and ran the reproduction and the checks above on my machine. I reviewed the diff and the results before opening this PR.

### Checklist


joshinitin2016883-sketch:
Hi @julian-risch, I saw on #3975 that one PR per integration works better for your release process. Happy to split this into separate alloydb and mariadb PRs if that's easier; just let me know.

PR #3964, opened 2026-09-15
fix(pgvector): count every written document in write_documents_async

joshinitin2016883-sketch:
### Related Issues

- None. Same class of bug as #3579 (Valkey), which #3618 fixed.

### Proposed Changes:

`write_documents` and `write_documents_async` insert with `executemany(sql_insert, db_documents, returning=True)` and return the number of ids that come back. psycopg delivers one result set per document. The sync path counts them by walking every result set with `fetchone()` / `nextset()`, following the psycopg docs.

The async path counts with `async for _ in self._async_cursor`. psycopg's `AsyncCursor.__anext__` only reads the current result set and never advances to the next one, so only the first document is counted:

```python
await store.write_documents_async([Document(id="1"), Document(id="2"), Document(id="3")])
# returns 1, expected 3

await store.write_documents_async([Document(id="1")])
await store.write_documents_async(
    [Document(id="1"), Document(id="2"), Document(id="3")], policy=DuplicatePolicy.SKIP
)
# returns 0, expected 2: with ON CONFLICT DO NOTHING the first result set is empty
```

The documents themselves are written correctly; only the count is wrong. It is the value `DocumentWriter.run_async` reports as `documents_written`.

This change walks the result sets in the async path the same way the sync path does.

### How did you test it?

- Regression tests in `TestDocumentStoreAsync`: `test_write_documents_async_counts_every_document` (3 new documents return 3) and `test_write_documents_async_skip_counts_only_new_documents` (`SKIP` with one duplicate returns 2). Run locally against `pgvector/pgvector:pg17`: without the change they fail with `assert 1 == 3` and `assert 0 == 2`; with it they pass.
- Integration suite: 184 passed, 5 skipped, both locally and in this PR's CI. On Windows I ran pytest with a `SelectorEventLoop` policy, because psycopg's async mode doesn't support the default `ProactorEventLoop`.
- Unit tests: 98 passed.
- `ruff check`, `ruff format --check` and `mypy` (using the integration's config) are clean.

Hatch is blocked by Windows Application Control on my machine, so these ran from a venv with the integration installed in editable mode rather than through `hatch run`.

### Notes for the reviewer

- Existing tests couldn't catch this: every count assertion, here and in Haystack's `WriteDocumentsAsyncTest`, writes a single document, which produces a single result set.
- The async loop dates from the async support added in #1547. alloydb and mariadb have no async write path, so only pgvector is affected.
- Prepared with Claude Code running locally: it found the bug by comparing the sync and async methods, wrote the change and tests, and ran the checks above on my machine. I reviewed the diff and the results before opening this PR.

### Checklist


integrations/pgvector/src/haystack_integrations/document_stores/pgvector/document_store.py
```diff
@@ -980,9 +980,13 @@ async def write_documents_async(
             )
             raise DocumentStoreError(error_msg) from e
 
+        # executemany(returning=True) yields one result set per document, so walk them all with nextset()
         written_docs = 0
-        async for _ in self._async_cursor:
-            written_docs += 1
+        while True:
```
davidsbatista: This loop now duplicates the one in write_documents. A single helper can't serve both a sync and an async cursor cleanly, so I think keeping them duplicated is fine. 

Could we just make the two comments match, so it's obvious they're meant to stay in sync? For example, reuse the psycopg docs link from the sync version here:

```python
# executemany(returning=True) yields one result set per document, so walk them all with nextset()
# https://www.psycopg.org/psycopg3/docs/api/cursors.html#psycopg.Cursor.executemany
```

davidsbatista:
Looks good! Thanks for the fix, I suggest adding a few more tests:

- one for `DuplicatePolicy.OVERWRITE`? That path uses ON CONFLICT DO UPDATE ... RETURNING, so every row comes back (including the updated one), and the count should be 3:

```python
  async def test_write_documents_async_overwrite_counts_every_document(self, document_store: PgvectorDocumentStore):
      await document_store.write_documents_async([Document(id="1", content="old")])
      docs = [Document(id="1", content="new"), Document(id="2"), Document(id="3")]
      assert await document_store.write_documents_async(docs, DuplicatePolicy.OVERWRITE) == 3
```

- Could you add the same tests to TestDocumentStore in `tests/test_document_store.py`. That keeps the sync and async suites mirrored:

```python
  def test_write_documents_counts_every_document(self, document_store: PgvectorDocumentStore):
      docs = [Document(id="1"), Document(id="2"), Document(id="3")]
      assert document_store.write_documents(docs) == 3

  def test_write_documents_skip_counts_only_new_documents(self, document_store: PgvectorDocumentStore):
      document_store.write_documents([Document(id="1")])
      docs = [Document(id="1"), Document(id="2"), Document(id="3")]
      assert document_store.write_documents(docs, DuplicatePolicy.SKIP) == 2
```

PR #3453, opened 2026-06-16
fix: serialize WeaviateDocumentStore grpc_port and grpc_secure

ly-wang19:
### Related Issues

None — found while reviewing `to_dict`/`from_dict` parity across the document stores.

### Proposed Changes:

`WeaviateDocumentStore.to_dict()` omitted `grpc_port` and `grpc_secure`, so they were lost on a `to_dict`/`from_dict` round-trip and a reloaded store fell back to the defaults (`50051` / `False`). Added both to `to_dict()`; `from_dict` already forwards them via `default_from_dict`.

### How did you test it?

Added a round-trip regression test with non-default values (`grpc_port=50052`, `grpc_secure=True`) — fails before the change, passes after. Updated the existing `test_to_dict` assertion. Non-integration tests + `ruff` pass.

### Notes for the reviewer

One-line change in `to_dict`; no `from_dict` change needed.

### Checklist

- [x] I have read the [contributors guidelines](https://github.com/deepset-ai/haystack-core-integrations/blob/main/CONTRIBUTING.md) and the [code of conduct](https://github.com/deepset-ai/haystack-core-integrations/blob/main/CODE_OF_CONDUCT.md)
- [x] I added unit tests and updated the docstrings
- [x] I've used one of the [conventional commit types](https://www.conventionalcommits.org/en/v1.0.0/) for my PR title: `fix:`

PR #3972, opened 2026-09-18
feat: WeaviateDocumentStore - add grpc_host for a separate gRPC endpoint

### LINKED ISSUE #1170: How to connect to separate gRPC host in WeaviateDocumentStore
**Describe the bug**
Hello Haystack team,

I'm currently working with a Weaviate setup where the REST API and gRPC services are hosted on different domains. Specifically, I have:

REST API Host: https://weaviate.xxyy.com
gRPC Host: https://weaviate.grpc.xxyy.com

In the original Weaviate Python client library, it's possible to specify different hosts for the REST and gRPC connections using the ConnectionParams class:

```
from weaviate.connect import ConnectionParams
import weaviate

connection_params = ConnectionParams.from_params(
    http_host="weaviate.xxyy.com",
    http_port=443,
    http_secure=True,
    grpc_host="weaviate.grpc.xxyy.com",
    grpc_port=443,
    grpc_secure=True,
)
client = weaviate.Client(connection_params=connection_params)
```

This configuration allows me to connect to the REST API and gRPC services using different hosts and ports.

Issue:

In Haystack's WeaviateDocumentStore, there doesn't seem to be an option to specify separate hosts for the REST API and gRPC connections. The url parameter sets both to the same host, and there's no exposed parameter to set a different gRPC host.

**To Reproduce**
Here's the current initialization in Haystack:

```
from haystack_integrations.document_stores.weaviate.document_store import WeaviateDocumentStore
from haystack_integrations.document_stores.weaviate.auth import AuthApiKey

document_store = WeaviateDocumentStore(
    url="https://weaviate.xxyy.com",
    auth_client_secret=AuthApiKey(api_key="your_weaviate_api_key"),
    grpc_secure=True,
)
```

 - OS: Linux
 - Haystack version: 2.6
 - Integration version:

v0ropaev:
### Related Issues

- fixes #1170

### Proposed Changes:

`WeaviateDocumentStore` derived the gRPC endpoint from the REST `url`, so a deployment that publishes REST and gRPC on different hosts — the normal shape behind an ingress or load balancer — could not be configured at all. This adds `grpc_host: str | None = None`, used by both the sync and the async client and round-tripped through `to_dict`/`from_dict`. Leaving it unset is exactly today's behaviour.

```python
WeaviateDocumentStore(
    url="https://weaviate.example.com",
    grpc_host="weaviate.grpc.example.com",
    grpc_port=443,
    grpc_secure=True,
)
```

### How did you test it?

From `integrations/weaviate`:

- `hatch run test:unit` — **98 passed**, 187 deselected (81 on `main`). Running the new tests against `main`'s source gives 25 failures, except the two deliberate no-change guards (`test_client_grpc_host_defaults_to_url_host`, `test_async_client_grpc_host_defaults_to_url_host`), which must pass on both.
- `hatch run fmt-check`, `hatch run test:types` — clean.
- The same unit selection on Python 3.10 and on the `--resolution lowest-direct` set (haystack-ai 2.28.0, weaviate-client 4.23.0) — 98 passed in both, so this holds at the declared dependency floor.
- The four fully mocked serde tests inside the integration-marked class, by node id — 4 passed (the `to_dict` fixture edit and `test_connection`).
- `hatch run docs` — the new docstrings render.

No integration tests: they need a live Weaviate, so CI will be the first to run them here.

### Notes for the reviewer

**Why `from_params` rather than keeping `from_url` and overriding the gRPC side.** That was my first implementation and it cannot work: `from_url` reuses `parsed_url.hostname` for both protocols, and `ConnectionParams._check_port_collision` then raises `http.port and grpc.port must be different if using the same host`. For the configuration in the issue (https REST, gRPC on 443) `from_url` therefore raises before there is an object to override — so the reporter's setup is rejected outright today, not merely awkward. `from_url` is still used verbatim when `grpc_host` is unset. On the new path the HTTP side is hand-derived and pinned to `from_url`'s output by `test_connection_params_with_grpc_host_derives_http_params_like_from_url` (4 urls × 2 flags); I also probed 20 url shapes × 2 flags with no divergences. `from_params` keeps the collision validator, so a genuine same-host/same-port mistake still fails.

**`grpc_secure` is taken literally when `grpc_host` is set** — the one real design question here. `from_url` upgrades gRPC to TLS for an https url because both protocols then share a host, so the scheme is evidence about the gRPC endpoint. With a separate host it is not: that endpoint terminates TLS on its own, and inheriting would discard an explicit `grpc_secure=False` and make "https REST at the edge, plaintext gRPC to an internal Service" inexpressible. Upstream agrees — `ConnectionParams.from_params` and `connect_to_custom` both require `grpc_secure` and never infer it. The cost is that a TLS gRPC ingress must now say `grpc_secure=True`, which both snippets in the issue already do; the docstring and the usage example say so, and `test_grpc_secure_is_not_inherited_from_the_url_scheme` pins both halves. Happy to switch to inheritance if you prefer the forgiving default — one word plus one test — but then I would want the coupling documented, because a wrong guess only surfaces as an opaque gRPC handshake error at connect time.

**On `additional_config=AdditionalConfig(proxies=Proxies(grpc=...))`, which looks like it already covers this: it doesn't.** `Proxies.grpc` ends up as a single `("grpc.http_proxy", ...)` channel argument in `ConnectionParams._grpc_channel`. Everything identifying the server still comes from `self.grpc`, i.e. from the REST host: the dialled `target`, the `grpc.default_authority` header an ingress routes on, and the TLS SNI/certificate check. A proxy inserts a CONNECT hop in front of an unchanged target; it does not move the endpoint, and it cannot dodge the port-collision validator above, which fires during `ConnectionParams` construction.

**Placement and merge order.** All nine new tests are module-level rather than in `TestWeaviateDocumentStore`, because that class is `@pytest.mark.integration` at class level: its tests are excluded from `hatch run test:unit`, so the guard wouldn't fire for someone running the unit job locally (they do run in PR CI). This also matches the existing `test_client_connects_to_weaviate_cloud` / `test_async_client_*`. Happy to move them next to their sibling, or into the `TestWeaviateDocumentStoreSerde` class #3955 introduces. #3954 inserts `_connection_headers()` at the same point in `document_store.py` and #3955 rewrites the `test_to_dict` block I edit — I'm happy to rebase behind both rather than make you resolve it.

`grpc_host` sits before `grpc_port` so the trio reads host/port/secure and matches `from_params`' own ordering; all `__init__` params are keyword-only, so this is not a positional break. Fixture churn (`"grpc_host": None` in 11 store `init_parameters` dicts across the store and retriever tests) follows merged #3453 exactly.

### Checklist

- [x] I've used one of the [conventional commit types](https://www.conventionalcommits.org/en/v1.0.0/) for my PR title


v0ropaev:
Merged `main` in `0008283`. The conflict was purely textual: `_connection_headers()` landed in the same place this branch adds `_connection_params()`, so git could not pick an order for the two method bodies. Both are kept; nothing semantic overlapped, and the two call sites had already auto-merged into the combination you'd want — `connection_params=self._connection_params()` alongside `additional_headers=self._connection_headers()`, for both the sync and async clients.

Verification after the merge, on the branch:

- `pytest -m "not integration" tests` → 104 passed (187 integration tests deselected — no Weaviate instance here)
- `ruff check src tests` and `ruff format --check src tests` → clean
- `mypy -p haystack_integrations.document_stores.weaviate -p haystack_integrations.components.retrievers.weaviate` → clean. One error showed up first, `Library stubs not installed for "dateutil"` in `_filters.py`, which is pre-existing and environmental — `install_types = true` handles it in CI; installing `types-python-dateutil` locally cleared it.

Still blocked on the CLA on my side, nothing else pending from me.

v0ropaev:
CLA is signed — `license/cla` is green. The draft gate did not flip this back automatically, so I marked it ready for review myself; re-assigning the reviewer needs permissions I do not have, so this is the nudge for that half. Branch is up to date with `main` and the checks are green.

PR #3975, opened 2026-09-21
fix: respect an empty runtime tools list in chat generators

Awshesh12:
### Related Issues

- No open issue. This ports deepset-ai/haystack#12639, which fixed the same pattern in the core OpenAI generators

### Proposed Changes:

Thirteen chat generators resolved the runtime `tools` argument with `tools or self.tools`. Because an empty list is falsy, calling `run(messages, tools=[])` on a generator that was initialized with tools silently fell back to the init tools, so there was no way to turn tools off for a single call

This switches each of them to `tools if tools is not None else self.tools`, the same fix core took in #12639. `None` still means "use the init tools", while an explicit empty list now means "no tools for this call". Every run and run_async signature already defaults `tools` to `None`, so the default path is unchanged

Affected generators: AIMLAPI, Amazon Bedrock, Anthropic, Cohere, Google GenAI, Hugging Face API, LiteLLM, llama.cpp, Ollama, OpenRouter, Transformers, vLLM and watsonx. Where both `run` and `run_async` have the pattern (Cohere, Google GenAI, Hugging Face API, Ollama), both are fixed

### How did you test it?

Added one unit test per integration (plus an async one for the four above) that initializes the generator with a tool, calls it with `tools=[]` and asserts that no tools reach the underlying client. Each new test fails on `main` and passes with the change

Ran `hatch run fmt`, `hatch run test:types` and `hatch run test:unit` in every touched integration, all clean with no failures

### Notes for the reviewer

Happy to split this into one PR per integration if that is easier to review or release. I kept it together since it is the same one-line change in each generator

### Checklist


sjrl:
@Awshesh12 thanks for opening a PR! Its better for our release process if you could open a separate PR for each integration. Could you close this PR and open new ones?

Awshesh12:
Thanks, makes sense. Closing this and splitting it into one PR per integration, starting with #3976 and #3977

PR #3977, opened 2026-09-21
fix: Anthropic - respect an empty runtime tools list

Awshesh12:
### Related Issues

- No open issue. Split out of #3975 as requested, one PR per integration. Ports deepset-ai/haystack#12639

### Proposed Changes:

`AnthropicChatGenerator` resolved the runtime `tools` argument with `tools or self.tools`. Because an empty list is falsy, calling `run(messages, tools=[])` on a generator initialized with tools silently fell back to the init tools, so there was no way to turn tools off for a single call

This switches it to `tools if tools is not None else self.tools`, the same fix core took in deepset-ai/haystack#12639. `None` still means "use the init tools", and the default path is unchanged

### How did you test it?

Added a unit test that initializes the generator with a tool, calls it with `tools=[]` and asserts that no tools reach the underlying client. It fails on `main` and passes with the change. `hatch run fmt`, `hatch run test:types` and `hatch run test:unit` are clean locally

### Notes for the reviewer

Same one-line change as the sibling PRs for the other chat generators

### Checklist


julian-risch (approved):
Thank you @Awshesh12 ! Looks good to me!

PR #3984, opened 2026-09-21
fix(elasticsearch,opensearch): keep every bound when merging same-field ranges

devjoinedthechat:
### Related Issues

No existing issue — describing the bug here, as in #3588 / #3589.

### The bug

`_normalize_ranges` merges range clauses acting on the same field into a single clause with `dict.update`:

```python
range_conditions_dict[field_name].update(comparison)
```

That is correct when the two bounds constrain opposite sides — `price > 1 AND price < 10` should become one `{"gt": 1, "lt": 10}` range. It is wrong when both constrain the same side, in two ways:

**1. The same key repeats.** The later bound overwrites the earlier one:

```python
{
    "operator": "AND",
    "conditions": [
        {"field": "meta.price", "operator": ">", "value": 5},
        {"field": "meta.price", "operator": ">", "value": 1},
    ],
}
```

renders as `{"bool": {"must": [{"range": {"price": {"gt": 1}}}]}}` — `price > 5` is gone, so documents the filter excludes come back.

**2. Different keys, same side.** `lt` and `lte` (likewise `gt` and `gte`) both set one bound, and the backend keeps only one of them per clause. `price < 3 AND price <= 5` merges into `{"lt": 3, "lte": 5}`, and Elasticsearch applies `lte: 5`:

```
GET /_search {"query": {"range": {"rank": {"lt": 3, "lte": 5}}}}   on ranks 0..6
returns ranks [0, 1, 2, 3, 4, 5]        # lt: 3 silently ignored
```

Under `NOT` both cases drop documents instead of adding them, since `NOT` wraps the merged conditions in an inner `must`.

This is not only a hand-written filter shape. `FilterPolicy.MERGE` produces it whenever a retriever's init filters and the caller's runtime filters constrain the same field, and the init bound — usually the one scoping the retriever — is the one lost:

```python
retriever = ElasticsearchBM25Retriever(
    document_store=store,
    filters={"operator": "AND", "conditions": [
        {"field": "meta.date", "operator": ">=", "value": "2021-06-01"},
        {"field": "meta.type", "operator": "==", "value": "report"}]},
    filter_policy=FilterPolicy.MERGE,
)
retriever.run(query="report", filters={"operator": "AND", "conditions": [
    {"field": "meta.date", "operator": ">=", "value": "2021-01-01"}]})
```

Before this change that returns a document dated `2021-03-01`, outside the retriever's own `date >= 2021-06-01` scope; after it, it does not.

### The fix

Bucket the bounds per field and start a new range clause when the side (lower for `gt`/`gte`, upper for `lt`/`lte`) is already taken, rather than overwriting. Bounds on opposite sides still merge, so `price > 1 AND price < 10` stays a single range and existing behaviour is unchanged. No values are compared, so dates and numbers behave alike.

Both integrations carry the same function and the same defect, so they are fixed together — happy to split into two PRs if you prefer. In OpenSearch this also covers the nested-field grouping, which calls `_normalize_ranges` too.

The `OR` path is untouched: it is still excluded from merging by the guard added in #3588 / #3589.

### How did you test it?

Unit tests added to both integrations, each failing before this change and passing after:

- `test_and_of_same_operator_ranges_keeps_both_bounds`
- `test_not_of_same_operator_ranges_keeps_both_bounds`
- `test_and_of_same_side_bounds_keeps_both_clauses`
- `test_normalize_ranges_starts_a_new_clause_on_a_repeated_bound`
- `test_and_of_same_operator_ranges_on_same_nested_field_keeps_both_bounds` (OpenSearch)

```
elasticsearch: hatch run test:unit         190 passed, 222 deselected
elasticsearch: hatch run test:integration  194 passed, 28 skipped   (live Elasticsearch 8.19.7)
elasticsearch: hatch run test:types        Success: no issues found in 11 source files
opensearch:    hatch run test:unit         186 passed, 284 deselected
opensearch:    hatch run test:types        Success: no issues found in 12 source files
```

Against the original code those new tests fail 4 (Elasticsearch) and 5 (OpenSearch).

I also ran a randomized differential check against a live Elasticsearch: 300 filters built from `AND`/`OR`/`NOT` over two to four range conditions on two numeric meta fields, comparing `filter_documents` with `haystack.utils.filters.document_matches_filter` as the reference. Before this change 35 of 300 disagreed; after it, none do.

### Notes for the reviewer

An alternative is to keep one clause per side and select the tighter bound (`max` for `gt`/`gte`, `min` for `lt`/`lte`). That preserves the single-range shape but needs values compared across types — ISO date strings, numbers, and whatever else a meta field holds — so I kept the type-agnostic version. Glad to switch if you'd rather have the tighter-bound form.


davidsbatista:
Thanks for the contribution! 

Could you add an integration test (@pytest.mark.integration) in `test_document_store.py` for both Elasticsearch and OpenSearch? We ran it in Docker, so it's cheap and requires no external API.

It should write a few documents with a numeric meta field, e.g. price 0–6, and assert on filter_documents results for:
 - price > 5 AND price > 1 → only 6
 - price < 3 AND price <= 5 → 0, 1, 2
 - NOT (price <= 5 AND price <= 1) → 2, 3, 4, 5, 6

A small version of your differential check is enough. It doesn't need to be randomized.

Please also include one case with ISO date strings, since the new tests only use integers. The FilterPolicy.MERGE example from your description works well: 
- date >= 2021-06-01 AND date >= 2021-01-01 shouldn't return a document dated 2021-03-01.

devjoinedthechat:
Thanks! Added in 4e18555: integration tests in both `test_document_store.py` files covering the three numeric cases and the `FilterPolicy.MERGE` date case (`date >= 2021-06-01 AND date >= 2021-01-01` no longer returns the 2021-03-01 doc).

One small change: I wrote the NOT case as `NOT (price <= 1 AND price <= 5)`. In the original order the overwrite happens to keep `lte: 1`, which is the tighter bound, so that case passed before the fix too. Swapped, it returns only `6` on the old code. Same expected result (2–6).

All four fail on the old `_normalize_ranges` and pass with the fix on both Elasticsearch 8.19.7 and OpenSearch 3.5.0. Full integration suites pass too.

davidsbatista (approved):
thanks 👍🏽

